Issue #23, opened 2021-08-12
Set up logging via easylogging


Mstrutov:
Replace `cout` usages with `LOG(ERROR)`, `LOG(INFO)`, etc. and set up `logging.conf` file such that only the most high-level logs would be visible to a user.

PR #83, opened 2022-02-02
Fix Pyro datarace errors

polyntsov:
One datarace was in SearchSpace::discover() with std::cout usage from different threads. Replaced std::cout with thread-safe LOG(). We really need to replace all output via std::cout with easylogging lib, at least to avoid such problems (#23). Registering mined uccs and fds from different threads also lead to the datarace. Synchronized registration to fix it.


Mstrutov:
Wow, gonna look out for those `std::cout`s!

Nice

PR #111, opened 2022-05-24
Upgrade README with image preview

Mstrutov:


Mstrutov:
Also: introduce `./pull_datasets.sh` to the README readers

PR #154, opened 2022-11-29
GFD validation

AntonChern:
Algorithm for checking if a graph functional dependency satisfies a given graph.

### DIFF
--- a/src/core/algorithms/gfd/balancer.cpp
+++ b/src/core/algorithms/gfd/balancer.cpp
@@ -1,0 +1,172 @@
+#include "balancer.h"
+
+#include <algorithm>
+#include <map>
+#include <numeric>
+#include <vector>
+
+std::vector<std::vector<int>> Balancer::Balance(std::vector<int> const& weights,
+                                                int const& processors_num) {
+    m_ = std::min(processors_num, static_cast<int>(weights.size()));
+    result_ = {};
+    if (weights.empty()) {
+        result_.resize(processors_num);
+        return result_;
+    }
+    for (std::size_t i = 0; i < m_; ++i) {
+        // the first value is index
+        std::vector<int> temp = {static_cast<int>(i)};
+        result_.push_back(temp);
+    }
+    // fill processors initially
+    // count optimal
+    optimal_ = 0;
+    std::size_t i = 0;
+    for (int const& weight : weights) {
+        result_.at(i++).push_back(weight);
+        i = i == m_ ? 0 : i;
+        optimal_ += weight;
+    }
+    optimal_ /= m_;
+    // sort processors (for convenience)
+    for (std::vector<int>& processor : result_) {
+        std::sort(processor.begin() + 1, processor.end());
+    }
+    // ALGORITHM
+    DeleteLarge();
+    Prepare();
+    DeleteFirstSmall();
+    DeleteSecondSmall();
+    FullLarge();
+    FullSmall();
+    // delete indices
+    for (std::vector<int>& processor : result_) {
+        processor.erase(processor.begin());
+    }
+    for (std::size_t i = 0; i < processors_num - m_; ++i) {
+        std::vector<int> empty = {};
+        result_.push_back(empty);
+    }
+    return result_;
+}
+
+void Balancer::DeleteLarge() {
+    deleted_large_ = {};
+    for (std::vector<int>& processor : result_) {
+        auto border = processor.end();
+        for (auto it = --processor.end(); it != processor.begin() + 1; --it) {
+            if (*(it - 1) > optimal_ / 2) {
+                deleted_large_.push_back(*it);
+                border = it;
+            } else {
+                break;
+            }
+        }
+        processor.erase(border, processor.end());
+    }
+}
+
+void Balancer::Prepare() {
+    for (std::size_t i = 0; i < m_; ++i) {
+        quality_.emplace(i, std::tuple<int, int, int>(0, 0, 0));
+    }
+    for (std::vector<int> const& processor : result_) {
+        auto last_small = processor.end();
+        auto last = processor.end();
+        if (*(--processor.end()) > optimal_ / 2) {
+            --last_small;
+        }
+        if (processor.begin() + 1 == last_small) {
+            continue;
+        }
+        int a = 0;
+        int b = 0;
+        float sum_small = std::accumulate(processor.begin() + 1, last_small, 0, std::plus<int>());
+        float sum = std::accumulate(processor.begin() + 1, last, 0, std::plus<int>());
+        while (sum_small > optimal_ / 2) {
+            ++a;
+            --last_small;
+            sum_small -= *last_small;
+        }
+        while (sum > optimal_) {
+            ++b;
+            --last;
+            sum -= *last;
+        }
+        std::get<0>(quality_.at(processor.at(0))) = a;
+        std::get<1>(quality_.at(processor.at(0))) = b;
+        std::get<2>(quality_.at(processor.at(0))) = a - b;
+    }
+}
+
+void Balancer::DeleteFirstSmall() {
+    // sort for convenience
+    deleted_small_ = {};
+    std::vector<std::vector<int>> small_processors = {};
+    std::vector<std::vector<int>> large_processors = {};
+    for (std::vector<int> const& processor : result_) {
+        if (*(--processor.end()) > optimal_ / 2) {
+            large_processors.push_back(processor);
+        } else {
+            small_processors.push_back(processor);
+        }
+    }
+    auto cGreater = [this](std::vector<int> const& a, std::vector<int> const& b) {
+        return std::get<2>(quality_.at(a.at(0))) > std::get<2>(quality_.at(b.at(0)));
+    };
+    sort(small_processors.begin(), small_processors.end(), cGreater);
+    sort(large_processors.begin(), large_processors.end(), cGreater);
+    result_.clear();
+    result_.insert(result_.end(), small_processors.begin(), small_processors.end());
+    result_.insert(result_.end(), large_processors.begin(), large_processors.end());
+    large_procs_num_ = large_processors.size();
+    std::size_t larges_num = large_processors.size() + deleted_large_.size();
+    // work
+    border_ = larges_num < m_ ? result_.end() - larges_num : result_.begin();
+    for (auto it = border_; it != result_.end(); ++it) {
+        auto last = it->end();
+        if (*(last - 1) > optimal_ / 2) {
+            --last;
+        }
+        for (auto cur = last - std::get<0>(quality_.at(*it->begin())); cur != last; ++cur) {
+            deleted_small_.push_back(*cur);
+        }
+        it->erase(last - std::get<0>(quality_.at(*it->begin())), last);
+    }
+}
+
+void Balancer::DeleteSecondSmall() {
+    for (auto it = result_.begin(); it != border_; ++it) {
+        auto last = it->end();
+        for (auto cur = last - std::get<1>(quality_.at(*it->begin())); cur != last; ++cur) {
+            deleted_small_.push_back(*cur);
+        }
+        it->erase(last - std::get<1>(quality_.at(*it->begin())), last);
+    }
+}
+
+void Balancer::PutWeight(int const& weight) {
+    sort(result_.begin(), result_.end(), [](std::vector<int> const& a, std::vector<int> const& b) {
+        return std::accumulate(a.begin(), a.end(), 0, std::plus<int>()) <
+               std::accumulate(b.begin(), b.end(), 0, std::plus<int>());
+    });
+    result_.begin()->push_back(weight);
+}
+
+void Balancer::FullLarge() {
+    std::size_t i = 0;
+    for (int const& weight : deleted_large_) {
+        if (i < m_ - large_procs_num_) {
+            (result_.begin() + i)->push_back(weight);
+        } else {
+            PutWeight(weight);
+        }
+        ++i;
+    }
+}
+
+void Balancer::FullSmall() {
+    for (int const& weight : deleted_small_) {
+        PutWeight(weight);
+    }
+}
--- a/src/core/algorithms/gfd/balancer.h
+++ b/src/core/algorithms/gfd/balancer.h
@@ -1,0 +1,27 @@
+#pragma once
+#include <map>
+#include <vector>
+
+class Balancer {
+private:
+    std::size_t m_;
+    double optimal_;
+    std::vector<std::vector<int>>::iterator border_;
+    std::vector<std::vector<int>> result_;
+    std::vector<int> deleted_large_ = {};
+    std::vector<int> deleted_small_ = {};
+    std::size_t large_procs_num_;
+    std::map<int, std::tuple<int, int, int>> quality_;
+
+    void DeleteLarge();
+    void Prepare();
+    void DeleteFirstSmall();
+    void DeleteSecondSmall();
+    void PutWeight(int const& weight);
+    void FullLarge();
+    void FullSmall();
+
+public:
+    std::vector<std::vector<int>> Balance(std::vector<int> const& weights,
+                                          int const& processors_num);
+};
--- a/src/core/algorithms/gfd/egfd_validation.cpp
+++ b/src/core/algorithms/gfd/egfd_validation.cpp
@@ -1,0 +1,1043 @@
+#include "egfd_validation.h"
+
+#include <iostream>
+
+#include <boost/graph/vf2_sub_graph_iso.hpp>
+
+#include "config/equal_nulls/option.h"
+#include "config/names_and_descriptions.h"
+#include "config/option_using.h"
+#include "config/tabular_data/input_table/option.h"
+
+namespace {
+
+using namespace algos;
+using Match = std::vector<std::pair<std::set<vertex_t>::iterator, std::set<vertex_t>::iterator>>;
+
+void FstStepForest(graph_t const& graph, std::map<vertex_t, std::set<vertex_t>>& rooted_subtree,
+                   std::map<vertex_t, vertex_t>& children_amount) {
+    typename boost::graph_traits<graph_t>::vertex_iterator it, end;
+    for (boost::tie(it, end) = vertices(graph); it != end; ++it) {
+        if (boost::degree(*it, graph) != 1) {
+            continue;
+        }
+        typename boost::graph_traits<graph_t>::adjacency_iterator adjacency_it =
+                boost::adjacent_vertices(*it, graph).first;
+
+        if (rooted_subtree.find(*adjacency_it) != rooted_subtree.end()) {
+            rooted_subtree.at(*adjacency_it).insert(*it);
+            children_amount[*adjacency_it]++;
+        } else {
+            std::set<vertex_t> value = {*it};
+            rooted_subtree.emplace(*adjacency_it, value);
+            children_amount.emplace(*adjacency_it, 1);
+        }
+    }
+}
+
+void BuildForest(graph_t const& graph, std::map<vertex_t, std::set<vertex_t>>& rooted_subtree,
+                 std::map<vertex_t, vertex_t>& children_amount) {
+    bool changed = true;
+    while (changed) {
+        changed = false;
+        std::map<vertex_t, std::set<vertex_t>> temp;
+        std::map<vertex_t, vertex_t> children_temp;
+        for (auto const& kv : rooted_subtree) {
+            auto desc = kv.first;
+            auto children = kv.second;
+
+            if (boost::degree(desc, graph) == (children_amount.at(desc) + 1)) {
+                changed = true;
+                typename boost::graph_traits<graph_t>::adjacency_iterator adjacency_it,
+                        adjacency_end;
+                boost::tie(adjacency_it, adjacency_end) = boost::adjacent_vertices(desc, graph);
+                for (; adjacency_it != adjacency_end; ++adjacency_it) {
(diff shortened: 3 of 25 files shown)


Mstrutov:
весь ПР ещё не смотрел + возможно это WIP, но большие датасеты (>100k строк) надо убирать из `input_data`. например, это увеличивает объём репо. вместо этого нужно использовать отдельный репозиторий https://github.com/Mstrutov/Desbordante-Data, откуда `./pull_datasets.sh` берёт датасеты

polyntsov:
> весь ПР ещё не смотрел + возможно это WIP, но большие датасеты (>100k строк) надо убирать из input_data. например, это увеличивает объём репо. вместо этого нужно использовать отдельный репозиторий https://github.com/Mstrutov/Desbordante-Data, откуда ./pull_datasets.sh берёт датасеты

И видимо класть их `git lfs` всё ещё

src/core/algorithms/gfd/parser.cpp
```diff
@@ -0,0 +1,287 @@
...
+
+// #include "/home/anton/boost_1_72_0/libs/graph/src/read_graphviz_new.cpp"
+// #include "/home/anton/boost_1_72_0/libs/graph/src/read_graphviz_new.cpp"
+// #include "/usr/local/include/boost/graph/detail/read_graphviz_new.cpp"
+
+#include "parser.h"
+
+namespace GFDtools {
+
+typedef boost::property_map<graph_t, std::map<std::string, std::string> vertex::*>::type AMap;
```
Mstrutov, line 19: prefer `using` to `typedef` in type alias declarations

src/core/algorithms/gfd/parser.cpp
```diff
@@ -0,0 +1,287 @@
...
+typedef boost::property_map<graph_t, std::string edge::*>::type RMap;
+
+namespace {
+template <typename AMap>
```
Mstrutov, line 23: It is only used with `AMap==AMap` --- does it really have to be a template?

src/core/algorithms/gfd/parser.cpp
```diff
@@ -0,0 +1,287 @@
...
+struct NewAttr {
+    typedef boost::shared_ptr<boost::dynamic_property_map> Ptr;
+
+private:
+    template <typename PMap>
+    static Ptr make_dyn(PMap m) {
```
Mstrutov, line 29: We use `PascalCase` for method names --- make sure to rename all the other methods in `gfd`

src/core/algorithms/gfd/balancer.cpp
```diff
@@ -0,0 +1,157 @@
...
+        i = i == m ? 0 : i;
+        optimal += weight;
+    }
+    optimal /= m;
+    // sort processors (for convenience)
+    for (std::vector<int>& processor : result) {
+        std::sort(processor.begin() + 1, processor.end());
+    }
+    // ALGORITHM
+    // 1st step
```
Mstrutov, line 39: Try to break down this huge method in a bunch of smaller ones. Each code block denoted by a comment looks like a good candidate for extraction to a separate method.

src/core/algorithms/gfd/graph_descriptor.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
+
+#include <map>
+#include <string>
+
+#include <boost/graph/adjacency_list.hpp>
+
+struct vertex {
```
Mstrutov, line 8: for struct and class names, we also use `PascalCase` (https://google.github.io/styleguide/cppguide.html#Type_Names)

Mstrutov (changes requested):

Mstrutov:
Also, you will have to rework the commits themselves. E.g., group "refactor" commits, remove commits on eigenvalues, squash development steps, etc. Finally, write descriptive commit messages. You can do this before the review-based refactor or after --- which would be more convenient for you. 

This is necessary to keep the commit history readable and informative. On the contrary, it is a bad practice to just squash everything into a single 2k-lines-of-code commit, as it would be hard to review and make sense of the reasons which lead to a particular change.

Personally, I have the following criteria for the perfect commit list:
- All the code made it into the PR final version, i.e. there are no intermediate commits. E.g., in your case, there are several commits on eigenvalues, which were removed at some point, therefore there were no eigenvalues before the PR, and there will be no eigenvalues after the PR. Say, the PR as a whole changes the code state from A to D. We are usually interested only in the final result, the whole A->[B->C->B->B'->C']->D chain can and should be omitted. 
- Each commit targets one conceptual change, e.g. "Add GFD algorithm", "Add efficient GFD algortihm", "Add tests".
- Each commit has a descriptive message.

This is a pretty subjective topic, but I hope my comment will point you in a right direction.
See examples of what I view as a good commit list: https://github.com/Mstrutov/Desbordante/pull/215/commits, https://github.com/Mstrutov/Desbordante/pull/139/commits

src/core/config/names.h
```diff
@@ -41,4 +41,6 @@ constexpr auto kCfdColumnsNumber = "columns_number";
 constexpr auto kCfdTuplesNumber = "tuples_number";
 constexpr auto kCfdMaximumLhs = "cfd_max_lhs";
 constexpr auto kCfdSubstrategy = "cfd_substrategy";
+// constexpr auto kGraphPath = "graph_path";
+constexpr auto kGFDData = "gfd";
```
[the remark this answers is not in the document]
AntonChern: Yes, there is `./Desbordante_run --algo=gfdvalid --csv_path="input_data/graph_data/quadrangle.dot" --gfd="input_data/graph_data/quadrangle_gfd.dot"`
I had to interpret `csv_path` as the path to the graph, since this parameter is required for all algorithms

src/core/algorithms/gfd/balancer.cpp
```diff
@@ -0,0 +1,175 @@
...
+                                                const int& processors_num) {
+    m = std::min(processors_num, (int)weights.size());
+    this->result = {};
+    if (weights.begin() == weights.end()) {
+        for (int i = 0; i < processors_num; ++i) {
+            std::vector<int> temp = {};
+            result.push_back(temp);
+        }
+        return result;
+    }
```
vs9h, line 18:
```suggestion
    if (weights.empty()) {
        result.resize(processors_num);
        return result;
    }

```

src/core/algorithms/gfd/balancer.cpp
```diff
@@ -0,0 +1,175 @@
+#include "balancer.h"
+
+#include <algorithm>
+#include <map>
+#include <numeric>
+#include <vector>
+
+std::vector<std::vector<int>> Balancer::Balance(const std::vector<int>& weights,
+                                                const int& processors_num) {
+    m = std::min(processors_num, (int)weights.size());
```
vs9h, line 10: don't use c-style cast

src/core/algorithms/gfd/balancer.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+#include <algorithm>
+#include <map>
+#include <numeric>
+#include <vector>
+
+std::vector<std::vector<int>> Balancer::Balance(const std::vector<int>& weights,
+                                                const int& processors_num) {
+    m = std::min(processors_num, (int)weights.size());
+    this->result = {};
```
vs9h, line 11: 1) the naming of the variable needs to be corrected (this applies not only to this variable): there are many fields in the class where the underscore is missing at the end or a non-snake case is used
2) in the code we do not use an explicit indication of the `this` keyword unless it is explicitly required (this applies to the entire pull request)

src/core/algorithms/gfd/balancer.cpp
```diff
@@ -0,0 +1,175 @@
...
+    }
+}
+
+void Balancer::FullSmall() {
+    for (const int& weight : deleted_small) {
+        sort(result.begin(), result.end(), [](std::vector<int> a, std::vector<int> b) {
+            return std::accumulate(a.begin(), a.end(), 0, std::plus<int>()) <
+                   std::accumulate(b.begin(), b.end(), 0, std::plus<int>());
+        });
+        result.begin()->push_back(weight);
```
vs9h, line 173: it's better to avoid code duplication (lines 157-161)

src/core/algorithms/gfd/egfd_validation.cpp
```diff
@@ -0,0 +1,915 @@
...
+                                 std::vector<std::set<vertex_t>>& forest) {
+    if (boost::num_vertices(graph) == (boost::num_edges(graph) + 1)) {
+        typename boost::graph_traits<graph_t>::vertex_iterator it, end;
+        for (boost::tie(it, end) = vertices(graph); it != end; ++it) {
+            core.insert(*it);
+        }
+        return;
+    }
+
+    typename boost::graph_traits<graph_t>::vertex_iterator it, end;
```
Mstrutov, line 24: This class (as well as other algorithm classes) contains a lot of huge methods. Please try to decompose them as much as possible. I'll leave some thoughts ahead.

Mstrutov: e.g. this `for` block looks extractable to a separate method.

Tip: use anonymous namespace to avoid bloating the class definition with utility methods, see HyFD/HyUCC for examples.

src/core/algorithms/gfd/egfd_validation.cpp
```diff
@@ -0,0 +1,915 @@
...
+            auto desc = kv.first;
+            auto children = kv.second;
+
+            if (boost::degree(desc, graph) == (children_amount.at(desc) + 1)) {
+                changed = true;
+                typename boost::graph_traits<graph_t>::adjacency_iterator adjacency_it,
+                        adjacency_end;
+                boost::tie(adjacency_it, adjacency_end) = boost::adjacent_vertices(desc, graph);
+                for (; adjacency_it != adjacency_end; ++adjacency_it) {
+                    if (children.find(*adjacency_it) == children.end()) {
```
Mstrutov, line 58: use `continue` to avoid high nestedness

```
if (children.find(*adjacency_it) != children.end()) {
    continue;
}

// code from body
```

src/core/algorithms/gfd/egfd_validation.cpp
```diff
@@ -0,0 +1,915 @@
...
+        if (label_classes.find(graph[*it].attributes.at("label")) != label_classes.end()) {
+            label_classes[graph[*it].attributes.at("label")].insert(*it);
+        } else {
+            std::set<vertex_t> value = {*it};
+            label_classes.emplace(graph[*it].attributes.at("label"), value);
+        }
+    }
+    std::vector<vertex_t> order(core.begin(), core.end());
+
+    auto cmp1 = [&graph, &query, &label_classes](const vertex_t& a, const vertex_t& b) {
```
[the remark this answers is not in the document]
Mstrutov: same for `cmp2`

src/core/algorithms/gfd/egfd_validation.cpp
```diff
@@ -0,0 +1,915 @@
...
+                bn++;
+            }
+        }
+        return an / a_degree < bn / b_degree;
+    };
+    std::sort(order.begin(), std::next(order.begin(), top), cmp2);
+    return *order.begin();
+}
+
+void EGfdValidation::BfsTree(const graph_t& query, const vertex_t& root,
```
Mstrutov, line 206: looks like this method is composed of two parts --- extract them

src/core/algorithms/gfd/egfd_validation.cpp
```diff
@@ -0,0 +1,915 @@
...
+            candidates.at(root).insert(*it);
+        }
+    }
+    std::set<vertex_t> visited = {root};
+    std::map<vertex_t, std::set<vertex_t>> visited_neighbours;
+    std::map<vertex_t, std::set<vertex_t>> unvisited_neighbours;
+    std::map<vertex_t, int> cnts;
+
+    std::vector<std::set<vertex_t>>::const_iterator i = std::next(levels.cbegin());
+    for (; i != levels.cend(); ++i) {
```
Mstrutov, line 278: this huge loop should be decomposed, you can:
- split body into separate parts
- extract whole body into separate method

src/core/algorithms/gfd/egfd_validation.cpp
```diff
@@ -0,0 +1,915 @@
...
+}
+
+void EGfdValidation::BottomUpRefinement(CPI& cpi, const graph_t& graph, const graph_t& query,
+                                        const std::vector<std::set<vertex_t>>& levels,
+                                        const std::map<vertex_t, vertex_t>& parent,
+                                        std::map<vertex_t, std::set<vertex_t>>& candidates) {
+    std::map<vertex_t, int> cnts;
+
+    std::vector<std::set<vertex_t>>::const_iterator lev_it;
+    for (lev_it = --levels.cend(); lev_it != std::next(levels.begin(), -1); --lev_it) {
```
Mstrutov, line 411: decompose

src/core/algorithms/gfd/egfd_validation.cpp
```diff
@@ -0,0 +1,915 @@
...
+        if (nta == 0) {
+            return false;
+        }
+        if (ntb == 0) {
+            return true;
+        }
+        return NumOfEmbeddings(cpi, a, *a.begin()) / nta <
+               NumOfEmbeddings(cpi, b, *b.begin()) / ntb;
+    };
+    std::vector<vertex_t> pi = *std::min_element(paths.begin(), paths.end(), cmp);
```
Mstrutov, line 541: extract this minimum-seeking part

src/core/algorithms/gfd/egfd_validation.cpp
```diff
@@ -0,0 +1,915 @@
...
+        for (; *pi_it == *path_it; ++pi_it, ++path_it) {
+        }
+        origins.push_back(*(--pi_it));
+    }
+    paths.erase(std::remove(paths.begin(), paths.end(), pi), paths.end());
+
+    std::vector<vertex_t> seq;
+    std::copy(pi.begin(), pi.end(), std::back_inserter(seq));
+    while (!paths.empty()) {
+        auto cmp = [this, &cpi, &paths_, &origins](std::vector<vertex_t> const& a,
```
Mstrutov, line 559: extract

src/core/algorithms/gfd/egfd_validation.cpp
```diff
@@ -0,0 +1,915 @@
...
+        if ((seq.at(i) != parent.at(u)) && boost::edge(seq.at(i), u, query).second) {
+            if (!boost::edge(*match.at(i).first, v, graph).second) {
+                return false;
+            }
+        }
+    }
+    return true;
+}
+
+bool EGfdValidation::Check(CPI& cpi, const graph_t& graph, const Gfd& gfd,
```
Mstrutov, line 592: decompose

Mstrutov (commented):
Good job!

There is one pending comment from my previous review, @vs9h comments, and there is some readability refactoring in `egfd_validation` to be done. Other than that, it seems, I have nothing to add.

src/core/algorithms/gfd/parser.h
```diff
@@ -0,0 +1,39 @@
...
+    static void ReadGraphTxt(std::string file_name, graph_t& result);
+
+    static void WriteGraph(std::ostream& stream, graph_t& result);
+    static void WriteGraph(std::string file_name, graph_t& result);
+
+    static void ReadGfd(std::istream& stream, Gfd& result);
+    static void ReadGfd(std::string file_name, Gfd& result);
+
+    static void WriteGfd(std::ostream& stream, Gfd& result);
+    static void WriteGfd(std::string file_name, Gfd& result);
```
vs9h, line 36: 1) Prefer using `std::filesystem::path const&` for `file_name`
2) Rename `file_name` to `path` (for consistency with `CSVParser`, for example)
3) Also copy-paste of `ReadGraph...(std::string filename, graph_t& result);` methods looks pretty bad. Maybe it's worth making one function for them?
4) Can we just return the graph instead of accepting it by reference? (https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs)
5) We use this class just to group `static` methods. According to google C++ style guide we should avoid this. (https://google.github.io/styleguide/cppguide.html#Nonmember,_Static_Member,_and_Global_Functions)
6) Why for `static void ReadGraph(std::string edges_name, std::string features_name, graph_t& result);` we use `std::string`? In other functions we use `std::istream`.

src/core/algorithms/gfd/gfd_handler.cpp
```diff
@@ -0,0 +1,62 @@
...
+    MakeOptionsAvailable({config::names::kGfdData, config::names::kCsvPath});
+};
+
+void GfdHandler::RegisterOptions() {
+    using namespace config::names;
+    using namespace config::descriptions;
+    DESBORDANTE_OPTION_USING;
+
+    RegisterOption(config::Option{&gfd_paths_, kGfdData, kDGfdData});
+    RegisterOption(config::Option{&graph_path_, kCsvPath, kDCsvPath});
```
vs9h, line 32: The use of this option should definitely be removed. I saw your comment regarding this: 

> I had to interpret csv_path as the path to the graph, since this parameter is required for all algorithms

But this is incorrect, since for example now we are adding algorithms that expect several tables (#296) as input - these algorithms accept a vector from tables. And in general, as the desbordante develops, various new formats will be added. GFD is just this case.

src/core/algorithms/gfd/gfd_handler.h
```diff
@@ -0,0 +1,42 @@
...
+#include "config/names_and_descriptions.h"
+#include "gfd.h"
+#include "parser.h"
+
+namespace algos {
+
+class GfdHandler : public Algorithm {
+protected:
+    std::filesystem::path graph_path_;
+    std::vector<std::string> gfd_paths_;
```
vs9h, line 14: why we use `std::string` here?

vs9h (changes requested):

src/core/algorithms/gfd/gfd_handler.h
vs9h, line 14: Why was it decided to directly receive paths from options? In the case when we work with CSV tables, the logic for creating tables is moved outside the algorithm into special options. (`config::InputTable`)

AntonChern: I decided not to clutter the code with new objects, since these fields are very simple. I think it's pretty natural to supply the paths directly

vs9h: Thanks for the answer. The motivation is clear, but it seems to me that code consistency is more important. But I would like to hear an opinion from @polyntsov.

src/core/algorithms/gfd/balancer.cpp
```diff
@@ -0,0 +1,175 @@
...
+    }
+}
+
+void Balancer::FullLarge() {
+    std::size_t i = 0;
+    for (const int& weight : deleted_large) {
+        if (i < m - numOfLargeProcs) {
+            (result.begin() + i)->push_back(weight);
+        } else {
+            sort(result.begin(), result.end(), [](std::vector<int> a, std::vector<int> b) {
```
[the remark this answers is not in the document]
Mstrutov: `auto cGreater` at 114 still has this problem

Mstrutov (approved):

src/core/parser/graph_parser/graph_parser.h
```diff
@@ -0,0 +1,32 @@
...
+#include <vector>
+
+#include "algorithms/gfd/gfd.h"
+#include "algorithms/gfd/graph_descriptor.h"
+
+namespace parser {
+
+std::vector<std::string> Split(std::string str, std::string sep);
+std::vector<Literal> ParseLiterals(std::istream& stream);
+void WriteLiterals(std::ostream& stream, std::vector<Literal> const& literals);
```
vs9h, line 14: These functions are needed only for implementation, so it is better not to declare them in the header. 
Don't forget to wrap these function to anonymous namespace in the source file.

src/core/algorithms/gfd/gfd_handler.h
polyntsov: The reason we have a separator option `Table` for CSV tables is because the table can be specified for the algorithm not only as path, but also as pandas dataframe. So what we need to think about here is how do we want to specify input graphs from python, if using path to the file, then storing only path is ok. 
Since in this PR we're not touching python bindings at all, there is little reason to create a separate option, and we can just store paths for simplicity. So IMO the separate option, if required, should be added in the future along with python bindings.

vs9h (approved):

src/core/algorithms/gfd/balancer.h
```diff
@@ -0,0 +1,30 @@
...
+    void Prepare();
+    void DeleteFirstSmall();
+    void DeleteSecondSmall();
+    void PutWeight(int const& weight);
+    void FullLarge();
+    void FullSmall();
+
+public:
+    Balancer() = default;
+    ~Balancer() = default;
```
polyntsov, line 26: not needed

src/core/algorithms/gfd/gfd_validation.cpp
```diff
@@ -0,0 +1,383 @@
...
+
+    std::vector<int> weights = {};
+    for (auto& kv : all_weighted_messages) {
+        weights.push_back(kv.first);
+    }
+
+    // balance
+    Balancer* balancer = new Balancer();
+    std::vector<std::vector<int>> balanced_weights = balancer->Balance(weights, threads_num_);
+    delete balancer;
```
polyntsov, line 335:
```suggestion
    Balancer balancer;
    std::vector<std::vector<int>> balanced_weights = balancer.Balance(weights, threads_num_);

```

PR #174, opened 2023-02-07
Reformat file structure

Mstrutov:
This PR aims at improving file tree structure and readability.

The main ways of refactoring are:
- deletion of unused code
- grouping algorithms by primitive
- moving algorithm-specific facilities to that algorithm directory
- refining directory responsibilities (e.g., extracting `config` out of `util` directory)

For a more detailed description of the changes, check out the commit messages.


Mstrutov:
Does `ucc` belong to `functional` group? Conceptually, it is a special case of a functional dependency, but at the same time it is categorized as a distinct primitive. Practically, it is very close to a `hyfd` algorithm and shares many components with it, but it doesn't reuse other functional dependency facilities.

The same question applies for`cfd`.

@polyntsov, what do you think?

python_bindings/create_dataframe_reader.h
```diff
@@ -2,9 +2,9 @@
 
 #include <pybind11/pybind11.h>
```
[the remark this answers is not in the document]
Mstrutov: Looks like python bindings are not built in CI after https://github.com/Mstrutov/Desbordante/pull/166, but they somehow get into clang-tidy compile commands.

I think this is a minor issues and it should be analyzed separately. For now, this warning serves as a reminder to compile python bindings manually. But in the future, we should find a way to compile them automatically.

src/structures/position_list_index.h
polyntsov, line 1: Personally, I don't really like the name "structures" for this because it's too generic. Also I'm not sure if we should have a separate directory in top-level `src` for position list indexes, because PLI is pretty close to be a "table representation" so it might even belong to `src/model` and at the same time it doesn't seem significant enough to have a separate module.

polyntsov (changes requested):
Firstly, I guess `src/caching` should be moved to at least `src/algorithms`.
Regarding the question about UCCs and CFDs, IMO they're distinct primitives and therefore should be placed in distinct directories with the following structure: `src/algorithms/ucc`, `src/algorithms/fd/`, `src/algorithms/cfd/`. The implementation similarity between `HyUCC` and `HyFD` seems to be exceptional due to the specific approach to discovery, I think it would be fine to put `hycommon` in `src/algorithms/fd/hyfd` for example and reuse it in `HyUCC` from there, regardless of the location of the `HyUCC` implementation. However, let's ask for another opinions.

Also, I think we should explicitly define the purposes of directories and individual modules to simplify reasoning about all this. Something like this:
- `src/algorithms/` -- primitive mining algorithms
- `src/algorithms/fd` -- FD mining algorithms
- `src/algorithms/...` -- .. mining algorithms
- `src/config/` -- configuration for the algorithms. Maybe it should be left in `utils/`, like `progress.h` which is pretty similar in nature
- `src/model` -- algorithm data representation things? Maybe we should use subdirectories for different data types like `src/model/table`, `src/model/graph` and so on? Also do classes like `FD` (implements FD definition) belong to "algorithm data representation"?
- `src/parser` -- csv parser
- `src/util`  -- generally useful utility? Not sure what we should keep here, `parallel_for.h` seems very general and may be used in any code, whereas `progress.h` and `primitive_collection.h` are supposed to be used in primitive mining algorithm classes like `FDAlgorithm`.
- ...?

Also, what do you think about placing `python_bindings` and `tests` to `src`, so that we have:
`src/python_bindings`, `src/tests`, `src/core` (and `src/cli` in the future)? The rationale for this is to avoid multiple directories with the source code in the root directory.

src/structures/position_list_index.h
BUYT-1: "structures" -> "data_structures"?

polyntsov: It doesn't actually make the name more specific

Mstrutov: I think we can move all the `structures` content to the `model/table`, as all the classes I ended up with (`AgreeSetFactory`, `IdentifierSet`, `PositionListIndex`, `VerticalMap`) are based on table representation and could be useful for any code employing it.

Mstrutov:
I mostly agree with invaluable @polyntsov comments, so the second iteration for the most part implements the proposed changes. However, there are some differences:
- Moved `src/caching` and `src/custom` to `src/util` as those two directories classes might be used by any code interested in caching and using user-provided hashes/random generators.
- Considering `src/util`, the only class that does not serve a general purpose is `progress.h`, therefore maybe it should be moved to `src/algorithms`, near the `algorithm.h`. Also, `primitive_collection.h` is really just an all-purpose thread-safe list with an `asList` feature, thus it may be renamed and used as such. Other that that, the number of classes under this directory is huge enough to keep it.

Additional thoughts:
- I think, all the code under `src/core/` more or less serves the algorithms, therefore the `algorithms` directory should only contain primitive directories, the abstract algorithm and algorithm instantiation classes. Thus, e.g., instead of creating `model/` under `algorithms/`, we place at under `core/`.

CMakeLists.txt
polyntsov, line 1: What do you think about moving `lib` under `src` and create `src/CMakeLists.txt` with parts of this file which are configuring compilation of the source code?

src/core/model/idataset_stream.h
polyntsov, line 1: This file has an interface describing operations on a parsed table so it should be moved to `model/table` I think. It's used by the classes implementing transactional data representation, but that's the consequence of the fact that transactional data is actually a table with special semantics for column and values

src/tests/input_data/BernoulliRelation.csv
polyntsov, line 1: I think we should move `input_data` out of the `src` to the root, let's try to keep only source code under `src`. Applying the same logic, examples of python bindings usage can also be moved out of `src`, however this is not necessary as the examples are technically code. And maybe rename `input_data` to `tests_input_data` or place it under `tests/` (in the root) dir with `datasets`?

polyntsov (changes requested):
I think we should also delete `run_test.sh`, it seems obsolete.
I'm still not sure about storing classes implementing primitives (`FD`) under `algorithms` and not `model`, but it's rather minor and opinion-based, in this state would probably be fine too.
Everything else seems nice.

src/python_bindings/examples/mine_typos.py
```diff
@@ -1,11 +1,10 @@
...
 
-from itertools import groupby, islice
-
-from colorama import Style, Fore
-from jellyfish import levenshtein_distance
 import desbordante as desb
 import pandas as pd
+from colorama import Style, Fore
+from itertools import groupby, islice
+from jellyfish import levenshtein_distance
```
BUYT-1, line 7: Why? The import order was correct, see [PEP-8](https://peps.python.org/pep-0008/#imports)

src/python_bindings/examples/dedupe.py
```diff
@@ -1,7 +1,6 @@
-from collections import defaultdict, deque
-
 import desbordante as desb
 import pandas as pd
+from collections import defaultdict, deque
```
BUYT-1, line 3: Why? The import order was [correct](https://peps.python.org/pep-0008/#imports)

BUYT-1 (dismissed):

CMakeLists.txt
```diff
@@ -43,23 +43,22 @@ message(${Boost_INCLUDE_DIRS})
...
+        "src/algorithms/functional/"
+        "src/algorithms/functional/aidfd"
+        "src/algorithms/functional/depminer"
+        "src/algorithms/functional/dfd"
+        "src/algorithms/statistics"
+        "src/caching"
+        "src/model"
+        "src/model/types"
+        "src/parser"
+        "src/parser/json"
```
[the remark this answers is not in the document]
polyntsov: IMHO `parser` name is better than `file_parsing`, it's precise and less verbose and to me `parser` sounds better for a module name than `file_parsing`.
I agree on creating `csv_parser` subdirectory

CMakeLists.txt
Mstrutov: I think we should totally encapsulate dependency management inside CMake using `FetchContent` module (e.g. [here](https://github.com/Mstrutov/Desbordante/blob/df5e0f516bf2fa33d2117cf5382a28215bec299d/external/CMakeLists.txt)) and this CMakeLists refinement would better be a part of a separate PR.

src/python_bindings/examples/mine_typos.py
Mstrutov: Oh, CLion uses a different format, reverted.

src/tests/input_data/BernoulliRelation.csv
Mstrutov: Went with the first option, because test input data and datasets relationship is not clear. Both file group are clearly sets of csv files, but the first one is used heavely by the tests, and the second one is intended for performance testing and showcase. It is also used in heavy datasets testsuite. Maybe we should have the following structure:
```
datasets/
  tests_input_data/
  datasets.zip
  CMakeLists.txt
```
But then I'm not sure, how both dataset and test CMakeLists should share responsibility for moving all the csv's to the binary directory. Maybe test CMakeLists should only check their presence and turn off the heavy testsuite if the are absent.

CMakeLists.txt
polyntsov: Ok, let's do it this way.

Mstrutov:
> I think we should also delete `run_test.sh`, it seems obsolete. I'm still not sure about storing classes implementing primitives (`FD`) under `algorithms` and not `model`, but it's rather minor and opinion-based, in this state would probably be fine too. Also, please try to squash commits of the form "Do x", "Revert x" to the single commit if possible and not too difficult with conflicts. Everything else seems nice.

I thought about getting rid of intermediate commits that were reworked later, but the commits are now too intertwined, there are too many related code changes, to reorder and squash them easily. I think it is easier to redo all the work, having the final image in mind, compiling after each commit so there is no need for fix-after commits. 

But IMO the final commit list is OK enough, surely better that a squashed single commit. A proper rework won't pay off.

src/tests/input_data/BernoulliRelation.csv
polyntsov: Well, we indeed don't use heavy datasets for unit tests on practice, but they are actually used in code and, moreover, right now testing on them is turned on by default. Also, not all the datasets in `datasets.zip` are considered heavy, as far as I remember at least `breast_cancer.csv` is used in FD tests on light datasets?
And even without this consideration, performance testing is technically testing, so IMHO it's ok to keep `datasets.zip` under tests dir. 

Despite all this, I think the directory structure you provided is also a viable solution (but maybe some renaming is required, like `datasets.zip` -> `heavy_datasets.zip`).

For both solutions I think we should have a single CMakeLists.txt if possible. Probably it will be ok to leave targets as they are, `input_data` is copied every time `tests` target is built (however, I'm not sure if it's possible to do this technically, since `tests` target is defined in `src/tests/CMakeLists.txt`) and `datasets.zip` have a separate target for unpacking.

polyntsov:
> But IMO the final commit list is OK enough

Okay, agree

polyntsov (approved):

PR #220, opened 2023-05-24
Implement parallel version of HyUCC

polyntsov:
Implements multithreaded version of HyUCC with `Validator` parallelized and with initialization step of `Sampler` parallelized. Also adds tests for multithreaded UCC algorithms. 
`Sampler` is the same for HyUCC and HyFD, but before using it in HyFD we should check how parallelized `Sampler` version performs for HyFD.

### DIFF
--- a/src/algorithms/ucc/hyucc/validator.cpp
+++ b/src/algorithms/ucc/hyucc/validator.cpp
@@ -1,5 +1,10 @@
 #include "validator.h"
 
+#include <future>
+
+#include <boost/asio/post.hpp>
+#include <boost/asio/thread_pool.hpp>
+
 #include "hycommon/efficiency_threshold.h"
 #include "hycommon/validator_helpers.h"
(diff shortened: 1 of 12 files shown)


src/algorithms/ucc/hyucc/validator.cpp
```diff
@@ -110,6 +115,41 @@ Validator::UCCValidations Validator::ValidateAndExtendSeq(
     return result;
 }
 
+Validator::UCCValidations Validator::ValidateAndExtendParallel(
+        std::vector<LhsPair> const& current_level) {
+    UCCValidations result;
+    boost::asio::thread_pool pool(threads_num_);
+    std::vector<std::future<UCCValidations>> validation_futures;
+    // split current_level on 4 spans
```
Mstrutov, line 123: `... on 'threads_num_' spans`?

Mstrutov (approved):
All good, just one minor comment --- you can ignore it and merge if time is running out.

PR #238, opened 2023-08-01
Implemented conversion mechanism from model::INumericType

Popov-Dmitriy-Ivanovich:
Each inheritor of model::INumericType now can be casted to any other INumericType inheritor and to any builtin type. That will help to improve working experience with custom type system, and allows to implement data storing inside model::NumericType class in future


polyntsov:
The commit name is still incorrect (commit message should have imperative mood and be no longer than 72 characters) and merge commits are prohibited, fix this, please.

PR #243, opened 2023-08-11
Minor bug fixes

vs9h:
Minor bug fixes:
* fix option names in `build.sh` script output
* remove duplication 'ac' in the description of the algo option
* fix 'toJSON' method for FD model (object keys must be in double quotes)

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -2,7 +2,8 @@
 
 function print_help() {
 cat << EOF
-Usage: ./build.sh [-h|--help] [-p|--pybind] [-t|--tests] [-u|--unpack] [-jN|--jobs=N]
+Usage: ./build.sh [-h|--help] [-p|--pybind] [-n|--no-test]
+                  [-u|--no-unpack] [-jN|--jobs=N] [-d|--debug]
 
   -h,         --help                  Display help
   -p,         --pybind                Compile python bindings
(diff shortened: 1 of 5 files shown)


tests/datasets.h
```diff
@@ -50,12 +50,12 @@ class LightDatasets {
...
-             {"EpicMeds.csv", 19617, '|', true},
-             {"EpicVitals.csv", 24082, '|', true},
-             {"iowa1kk.csv", 55285, ',', true},
-             {"LegacyPayors.csv", 34874, '|', true}}};
+            {{"adult.csv", 23075, ';', false},
+             {"CIPublicHighway.csv", 32696 /* TODO: update hash */, ',', true},
+             {"EpicMeds.csv", 50218, '|', true},
+             {"EpicVitals.csv", 24082 /* TODO: update hash */, '|', true},
+             {"iowa1kk.csv", 55285 /* TODO: update hash */, ',', true},
+             {"LegacyPayors.csv", 34874 /* TODO: update hash */, '|', true}}};
```
polyntsov, line 58: Hashes are mismatched due to the additional quotes? Are we calculating hashes of a JSON string directly?

vs9h: Yes, `fletcher` invokes `GetFDsJSON`

polyntsov: Hm, this doesn't seem right. Let's leave it as it is for now, however, I think we should calculate hash directly from left and right hand sides of an fd, not from any string representation of an fd (to avoid problems like this, we changed the representation and hashes are changed accordingly). @Mstrutov, what do you think? 

Anyways, could you fix this TODOs right away, Pyro/HyFD built in release mode will calculate it fast.

vs9h: TODOs fixed

polyntsov (approved):

PR #244, opened 2023-08-12
Rename Tane::RegisterFd to Tane::RegisterAndCountFd

polyntsov:
To avoid hiding FDAlgorithm::RegisterFd virtual function, this fixes a "virtual function hiding" warning on a newer versions of GCC (and in particular fixes the debug build on arch linux)


BUYT-1 (approved):
👍

PR #252, opened 2023-08-21
Date type implementation and tests

egshnov:
Added class DateType that inherits class Type for working with custom type-system. DateType operates with objects of boost::gregorian::date that supports converting date to/from string, subtracting dates, adding dates, adding date_duration (represents days) to a date.
Added tests.


polyntsov:
Please, don't change the formatting of the old code, it's a conceptually different change and should be done in a separate PR.

src/core/model/types/builtin.h
```diff
@@ -47,12 +48,14 @@ BETTER_ENUM(TypeId, char,
                  * (arbitrary-precision integer value) */
     kString,    /* Except for nulls and empties column contains only strings
                  * (string value, sequence of characters) */
+    kDate,      /* Except for nulls (not_a_date_time) and infinities(pos_infin, neg_infin) column contains only dates
```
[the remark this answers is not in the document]
polyntsov: This enum describes both types of values which can be stored in the table and types of the table columns.
This PR adds date as a type for a single value, but lacks support for `Date` table columns. The code which creates `TypedColumnData` knows nothing about newly added `Date` and this is to be fixed. Please, look at the `TypedColumnDataFactory::CreateFrom` and implement the corresponding functionality. I suppose adding the appropriate values to `TypedColumnDataFactory::type_id_to_regex_` should be enough

polyntsov (changes requested):
Please, fix the commit messages so that they satisfy the required style (line lengths in the first commit are wrong).
Also, "add missing CreateType statement fix indents" part of the second commit shouldn't present in the history, squash these changes into the first commit.

src/core/model/types/builtin.h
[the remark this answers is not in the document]
egshnov: It seems to me that checking whether string value represents a date via regular expression is not a good option because there are some instances that will satisfy regular expression but are not a valid date (e.g. «2023-02-29») in such cases `Date` is constructed as `not_a_date_time` so i think it would be better if we just leave it as a string value. `boost::gregorian` doesn’t provide any option to check whether string value represents a valid date without constructing a `Date` object so i don’t think this is a good option either. Is there any way to check it via regular expression more precisely (i use `^([0-9]{4})[-./](1[0-2]|0[1-9])[-./](3[0-1]|0[1-9]|[1-2][0-9])$` expression from "Regular Expressions Cookbook»)  or should i just leave it as it is?

src/tests/test_types.cpp
```diff
@@ -264,4 +264,72 @@ INSTANTIATE_TEST_SUITE_P(TestStringSuite, TestString,
...
+    mo::Date d1(boost::gregorian::from_string(l_val));
+    mo::Date d2(boost::gregorian::from_string(r_val));
+    model::CompareResult cr = mo::DateType::Compare(d1, d2);
+
+    using Owner = std::unique_ptr<std::byte[]>;
+    Owner l_owner(date_type->MakeValue(d1));
+    Owner r_owner(date_type->MakeValue(d2));
+    std::byte* l = l_owner.get();
+    std::byte* r = r_owner.get();
+    EXPECT_EQ(date_type->Compare(l, r), cr);
```
polyntsov, line 290: Given that `date_type->Compare` uses `mo::DateType::Compare` in its implementation, this check seems pretty useless because it will almost never fail

src/tests/test_types.cpp
```diff
@@ -264,4 +264,72 @@ INSTANTIATE_TEST_SUITE_P(TestStringSuite, TestString,
...
+    TestDateTypeParam(std::string l, std::string r) noexcept : l(std::move(l)), r(std::move(r)) {}
+};
+
+class TestDateType : public ::testing::TestWithParam<TestDateTypeParam> {};
+
+TEST_P(TestDateType, ComparisonTest) {
+    TestDateTypeParam p = GetParam();
+    auto const& [l_val, r_val] = GetParam();
+    std::unique_ptr<mo::Type> type(mo::CreateType(mo::TypeId::kDate, true));
+    auto* date_type = static_cast<mo::DateType*>(type.get());
```
polyntsov, line 280: Why not `std::unique_ptr<mo::Type> date_type(mo::CreateSpecificType<mo::DateType>(mo::TypeId::kDate, true));`?

src/tests/test_types.cpp
```diff
@@ -264,4 +264,72 @@ INSTANTIATE_TEST_SUITE_P(TestStringSuite, TestString,
...
+    model::CompareResult cr = mo::DateType::Compare(d1, d2);
+
+    using Owner = std::unique_ptr<std::byte[]>;
+    Owner l_owner(date_type->MakeValue(d1));
+    Owner r_owner(date_type->MakeValue(d2));
+    std::byte* l = l_owner.get();
+    std::byte* r = r_owner.get();
+    EXPECT_EQ(date_type->Compare(l, r), cr);
+
+    auto* val = new std::byte[date_type->GetSize()];
```
polyntsov, line 292: Don't use raw pointers, which own memory they point to. Don't allocate memory for the objects of Desbordante types by hand, use `Allocate` or `MakeValue` methods instead.

src/tests/test_types.cpp
```diff
@@ -264,4 +264,72 @@ INSTANTIATE_TEST_SUITE_P(TestStringSuite, TestString,
...
+    auto* date_type = static_cast<mo::DateType*>(type.get());
+    mo::Date d1(boost::gregorian::from_string(l_val));
+    mo::Date d2(boost::gregorian::from_string(r_val));
+    model::CompareResult cr = mo::DateType::Compare(d1, d2);
+
+    using Owner = std::unique_ptr<std::byte[]>;
+    Owner l_owner(date_type->MakeValue(d1));
+    Owner r_owner(date_type->MakeValue(d2));
+    std::byte* l = l_owner.get();
+    std::byte* r = r_owner.get();
```
polyntsov, line 289: I don't think it's worth creating these variables to only use them once, use unique pointers right away instead

src/tests/test_types.cpp
polyntsov: The same problem in the test below

src/tests/test_types.cpp
polyntsov: The same problem in the test below

src/tests/test_types.cpp
```diff
@@ -264,4 +264,72 @@ INSTANTIATE_TEST_SUITE_P(TestStringSuite, TestString,
...
+    mo::Date d2(boost::gregorian::from_string(r_val));
+
+    using Owner = std::unique_ptr<std::byte[]>;
+    Owner l_owner(date_type->MakeValue(d1));
+    Owner r_owner(date_type->MakeValue(d2));
+    std::byte* l = l_owner.get();
+    std::byte* r = r_owner.get();
+
+    boost::gregorian::date_duration delta = d1 - d2;
+    EXPECT_EQ(delta, d1 - d2);
```
polyntsov, line 313: strange check

src/core/model/table/typed_column_data.cpp
```diff
@@ -24,6 +24,9 @@ TypedColumnDataFactory::TypeMap TypedColumnDataFactory::CreateTypeMap() const {
         bool matched = false;
         for (auto const& [type_id, regex] : type_id_to_regex_) {
             if (std::regex_match(val, regex)) {
+                if (type_id == +TypeId::kDate &&
+                    boost::gregorian::from_simple_string(val).is_not_a_date())
+                    break;
```
polyntsov, line 29: I think adding braces will simplify the visual perception of the code

src/core/model/table/typed_column_data.h
```diff
@@ -165,16 +170,20 @@ class TypedColumnDataFactory {
...
         {TypeId::kDouble, std::regex(R"(^(\+|-)?\d+\.\d*$)")},
+        {TypeId::kDate,std::regex(R"(^([0-9]{4})[-.\/]?(1[0-2]|0[1-9])[-.\/]?(3[0-1]|0[1-9]|[1-2][0-9])$)")},
         {TypeId::kNull, std::regex(Null::kValue.data())},
         {TypeId::kEmpty, std::regex(R"(^$)")}};
 
     size_t CalculateMixedBufSize(std::vector<TypeId> const& types_layout,
                                  TypeIdToType const& type_id_to_type) const noexcept;
+    // clang-format on
+    size_t CalculateMixedBufSize(MixedType const* mixed, TypeIdToType const& tid_to_map,
+                                 TypeMap const& type_map) const noexcept;
```
polyntsov, line 186: Redundant

test_input_data/iowa1kk_cropped.csv
polyntsov, line 1: I don't think it's the best idea to add a cropped version of an existing dataset to use it only in one place. When dates support will be added to AC we will need to test it on some data anyway and cropped version doesn't look like the best candidate. I think a lot better solution is to recreate a dataset from the original paper describing AC: https://dl.acm.org/doi/10.5555/1315451.1315509, because besides testing purposes we could use it as a good demonstration example.

polyntsov (changes requested):
Please, fix the commit message of the first commit, it should be less or equal to 72 symbols

src/tests/test_types.cpp
```diff
@@ -264,4 +264,109 @@ INSTANTIATE_TEST_SUITE_P(TestStringSuite, TestString,
...
+    test("1941-06-22", "1945-05-09", 1417);
+    test("2023-10-28", "2297-08-12", 100000);
+}
+struct TestDateArithmeticsParam {
+    std::string const date1;
+    std::string const date2;
+    long const dist;
+    TestDateArithmeticsParam(std::string l, std::string r, long dist)
+        : date1(std::move(l)), date2(std::move(r)), dist(dist){};
+};
```
polyntsov, line 312: anonymous namespace

test_input_data/ACShippingDates.csv
polyntsov, line 1: It is better to add it in a separate commit

src/tests/test_types.cpp
```diff
@@ -264,4 +264,109 @@ INSTANTIATE_TEST_SUITE_P(TestStringSuite, TestString,
...
+            date_type->GetDeleater()};
+}
+
+TEST(TestDateSuite, Compare) {
+    std::unique_ptr<mo::DateType> date_type(
+            mo::CreateSpecificType<mo::DateType>(mo::TypeId::kDate, true));
+    auto test = [date_type = std::move(date_type)](
+                        std::string const& date1, std::string const& date2, mo::CompareResult res) {
+        auto ptr1 = DatePtrFromString(date1, date_type);
+        auto ptr2 = DatePtrFromString(date2, date_type);
```
polyntsov, line 280: Type is not obvious from the right side, so use it explicitly instead of relying on the type deduction

src/tests/test_types.cpp
```diff
@@ -264,4 +264,109 @@ INSTANTIATE_TEST_SUITE_P(TestStringSuite, TestString,
...
+    std::unique_ptr<mo::DateType> date_type(
+            mo::CreateSpecificType<mo::DateType>(mo::TypeId::kDate, true));
+
+    auto const& [date1, date2, delta_] = GetParam();
+    mo::DateType::Delta delta(delta_);
+
+    auto initial_date = DatePtrFromString(date1, date_type);
+    auto estimated_result = DatePtrFromString(date2, date_type);
+    auto actual_result =
+            DatePtr(date_type->SubDelta(initial_date.get(), delta), date_type->GetDeleater());
```
polyntsov, line 325:
```suggestion
    auto expected = DatePtrFromString(date2, date_type);
    auto actual =
            DatePtr(date_type->SubDelta(initial_date.get(), delta), date_type->GetDeleater());
```

src/core/model/table/typed_column_data.h
```diff
@@ -167,11 +172,13 @@ class TypedColumnDataFactory {
...
-        {TypeId::kDouble, std::regex(R"(^(\+|-)?\d+\.\d*$)")},
-        {TypeId::kNull, std::regex(Null::kValue.data())},
-        {TypeId::kEmpty, std::regex(R"(^$)")}};
+            {TypeId::kInt, std::regex(R"(^(\+|-)?\d{1,19}$)")},
+            {TypeId::kBigInt, std::regex(R"(^(\+|-)?\d{20,}$)")},
+            {TypeId::kDouble, std::regex(R"(^(\+|-)?\d+\.\d*$)")},
+            {TypeId::kDate,
+             std::regex(R"(^([0-9]{4})[-.\/]?(1[0-2]|0[1-9])[-.\/]?(3[0-1]|0[1-9]|[1-2][0-9])$)")},
+            {TypeId::kNull, std::regex(Null::kValue.data())},
+            {TypeId::kEmpty, std::regex(R"(^$)")}};
```
polyntsov, line 181: Formatting of the old code should be ideally done in a separate commit

src/core/model/table/typed_column_data.h
```diff
@@ -53,9 +53,14 @@ class TypedColumnData : public model::AbstractColumnData {
...
-            if (GetValueTypeId(i) == +TypeId::kString || GetValueTypeId(i) == +TypeId::kBigInt) {
+            if (GetValueTypeId(i) == +TypeId::kString || GetValueTypeId(i) == +TypeId::kBigInt ||
+                GetValueTypeId(i) == +TypeId::kDate) {
                 std::byte const* value = (mixed) ? mixed->RetrieveValue(data_[i]) : data_[i];
-                StringType::Destruct(value);
+                if (GetValueTypeId(i) == +TypeId::kDate) {
+                    DateType::Destruct(value);
+                } else {
+                    StringType::Destruct(value);
+                }
```
polyntsov, line 63: Actually, this case-handling doesn't look good, seems like we need a generic method `Type::Destruct<T>` to destruct values which will dispatch to `Destruct` method specific for passed type. 
However, I think, it should be implemented in a separate PR. You may implement it if you want.

src/tests/test_types.cpp
```diff
@@ -264,4 +264,109 @@ INSTANTIATE_TEST_SUITE_P(TestStringSuite, TestString,
...
+
+    EXPECT_EQ(date_type->Compare(estimated_result.get(), actual_result.get()),
+              mo::CompareResult::kEqual);
+}
+
+TEST_P(TestDateArithmetics, AddDelta) {
```
polyntsov, line 331: This test case has almost the same code as the test above. Common logic should be moved to a separate function which expects tested method, dates and probably an initial date as params.

src/tests/test_types.cpp
```diff
@@ -264,4 +264,109 @@ INSTANTIATE_TEST_SUITE_P(TestStringSuite, TestString,
...
+}
+
+TEST(TestDateSuite, Dist) {
+    std::unique_ptr<mo::DateType> date_type(
+            mo::CreateSpecificType<mo::DateType>(mo::TypeId::kDate, true));
+    auto test = [date_type = std::move(date_type)](std::string const& date1,
+                                                   std::string const& date2, long res) {
+        auto left_val = DatePtrFromString(date1, date_type);
+        auto right_val = DatePtrFromString(date2, date_type);
+        EXPECT_EQ(long(date_type->Dist(left_val.get(), right_val.get())), res);
```
polyntsov, line 298: Avoid `Type(object)` syntax for casting, it is really confusing because such expression could mean a function call, a constructor call and a variable declaration depending on the context. 
Moreover, it's a C-style cast and Google Style Guide [forbids](https://google.github.io/styleguide/cppguide.html#Casting) them.

polyntsov (changes requested):
Several rather minor concerns, otherwise looks good to me

src/core/model/table/typed_column_data.h
```diff
@@ -198,4 +205,4 @@ class TypedColumnDataFactory {
 std::vector<TypedColumnData> CreateTypedColumnData(IDatasetStream& dataset_stream,
                                                    bool is_null_equal_null);
 
-}  // namespace model
+}  // namespace model
```
polyntsov, line 208: End of line symbol :)

src/tests/test_types.cpp
```diff
@@ -263,5 +263,105 @@ INSTANTIATE_TEST_SUITE_P(TestStringSuite, TestString,
...
+struct TestDateArithmeticsParam {
+    std::string const date1;
+    std::string const date2;
+    long const dist;
+    TestDateArithmeticsParam(std::string l, std::string r, long dist)
+        : date1(std::move(l)), date2(std::move(r)), dist(dist){};
+};
+
+void TestBinop(std::string const& initial_date_, std::string const& expected_, long const delta_,
+               std::byte* (mo::DateType::*Binop)(std::byte const* date, mo::DateType::Delta const&)
```
polyntsov, line 318: Create an alias for this, it will look a lot simpler

src/tests/test_types.cpp
```diff
@@ -263,5 +263,105 @@ INSTANTIATE_TEST_SUITE_P(TestStringSuite, TestString,
...
+                       const) {
+    std::unique_ptr<mo::DateType> date_type(
+            mo::CreateSpecificType<mo::DateType>(mo::TypeId::kDate, true));
+    mo::DateType::Delta delta(delta_);
+
+    DatePtr initial_date = DatePtrFromString(initial_date_, date_type);
+    DatePtr expected = DatePtrFromString(expected_, date_type);
+    auto actual =
+            DatePtr((date_type.get()->*Binop)(initial_date.get(), delta), date_type->GetDeleater());
```
polyntsov, line 327:
```suggestion
            DatePtr(std::invoke(Binop, date_type.get(), initial_date.get(), delta), date_type->GetDeleater());
```
Method pointer call syntax looks atrocious, let's use `std::invoke`

src/tests/test_types.cpp
```diff
@@ -263,5 +263,105 @@ INSTANTIATE_TEST_SUITE_P(TestStringSuite, TestString,
...
+namespace {
+struct TestDateArithmeticsParam {
+    std::string const date1;
+    std::string const date2;
+    long const dist;
+    TestDateArithmeticsParam(std::string l, std::string r, long dist)
+        : date1(std::move(l)), date2(std::move(r)), dist(dist){};
+};
+
+void TestBinop(std::string const& initial_date_, std::string const& expected_, long const delta_,
```
polyntsov, line 317: Local variables shouldn't end with the underscore

polyntsov (approved):

PR #256, opened 2023-08-24
Reorganize python examples

BUYT-1:
Makes the code there less confusing to a new reader.


src/python_bindings/examples/mine_typos.py
```diff
@@ -4,23 +4,41 @@
...
+EXACT_ALGORITHM_TYPE = desbordante.HyFD
+EXACT_ALGO_CONFIG = {}
+
+# Algorithm that finds approximate FDs and its config.
+APPROXIMATE_ALGORITHM_TYPE = desbordante.Pyro
+ERROR = 0.005  # Highest error for almost holding FDs.
+APPROXIMATE_ALGO_CONFIG = {'error': ERROR}
+
+# Parameters for pandas.read_csv(...).
+DATASET_PATH = '../../../test_input_data/Workshop.csv'
```
polyntsov, line 24: I don't think it's a good idea to hardcode relative paths (or at least paths with `..`) in the code

polyntsov (changes requested):
I'd rename two last commits moving their body to the commit message, because "Rewrite mine_typos.py" sounds too general. "Enhance naming and add comments to typo mining example" would be better IMO and no additional information in the commit message body would be needed

src/python_bindings/examples/datasets/Workshop.csv
```diff
@@ -0,0 +1 @@
+../../../../test_input_data/Workshop.csv
```
polyntsov, line 1: How does it work? Pandas parses it?..

BUYT-1: It's a symlink

polyntsov: Looks weird. Why don't we just copy `test_input_data/Workshop.csv` here during build process as it's done for other test data?

polyntsov (approved):

PR #258, opened 2023-08-27
Fix float division by zero when constructing ranges in AC

polyntsov:
`weight` equal to 1 leads to the float division by zero when constructing ranges, avoid it by adding a corresponding check. This doesn't change the observable behavior of the code.
Also, use option member `weight_` when reconstructing ranges so that the value validation check is performed for weight for which ranges are reconstructed.

### DIFF
--- a/src/core/algorithms/algebraic_constraints/ac_algorithm.cpp
+++ b/src/core/algorithms/algebraic_constraints/ac_algorithm.cpp
@@ -263,25 +263,34 @@ std::vector<std::byte const*> ACAlgorithm::ConstructDisjunctiveRanges(ACPairs co
         ranges = std::vector<std::byte const*>();
         return ranges;
     }
+
     ACPair const* l_border = ac_pairs.front().get();
     ACPair const* r_border = nullptr;
-    double delta = num_type_->Dist(ac_pairs.front()->GetRes(), ac_pairs.back()->GetRes()) *
-                   (weight / (1 - weight));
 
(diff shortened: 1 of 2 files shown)


src/core/algorithms/algebraic_constraints/ac_algorithm.cpp
```diff
@@ -256,40 +256,50 @@ void ACAlgorithm::PrintRanges(std::vector<model::TypedColumnData> const& data) c
...
     return ranges;
 }
 
 RangesCollection ACAlgorithm::ReconstructRangesByColumns(size_t lhs_i, size_t rhs_i,
-                                                         double weight) const {
+                                                         double weight) {
     ACPairsCollection const& constraints_collection = GetACPairsByColumns(lhs_i, rhs_i);
     ACPairs const& ac_pairs = constraints_collection.ac_pairs;
-    std::vector<std::byte const*> ranges = ConstructDisjunctiveRanges(ac_pairs, weight);
+    SetOption(config::names::kWeight, weight);
```
BUYT-1, line 301: nit: I think this line should be at the beginning of the method so it fails earlier if a wrong value is passed and so that whoever reads the code sees that this sets an option.

polyntsov: Agree, good point

BUYT-1 (approved):
LGTM

polyntsov:
Waiting for approve from @Sched71 before merging

Sched71:
All good. Thank you :+1:

PR #259, opened 2023-08-27
Fix minor issues in DataStats found by UB sanitizer

polyntsov:
Explicitly make integer of 0 in data statistics functions instead of relying on the fact that `Type::Allocate` zeroes a buffer it's allocated. Integer and float point numbers represent a zero as all zero bits, but this is rather a coincidence on which we shouldn't rely.

Also introduce a check to the `Execute` that table is not empty to avoid float division by zero.


Mstrutov (approved):
LGTM

PR #265, opened 2023-09-07
Configuration exceptions

BUYT-1:
Distinguishes configuration exceptions from standard library exceptions. Python support included.

### DIFF
--- a/src/core/algorithms/algorithm.cpp
+++ b/src/core/algorithms/algorithm.cpp
@@ -2,6 +2,8 @@
 
 #include <cassert>
 
+#include "config/exceptions.h"
+
 namespace algos {
 
 bool Algorithm::SetExternalOption([[maybe_unused]] std::string_view option_name,
@@ -90,13 +92,13 @@ void Algorithm::SetOption(std::string_view option_name, boost::any const& value)
     auto it = possible_options_.find(option_name);
     if (it == possible_options_.end()) {
         if (ext_opt_set) return;
-        throw std::invalid_argument("Unknown option \"" + std::string{option_name} + '"');
+        throw config::ConfigurationError("Unknown option \"" + std::string{option_name} + '"');
     }
     std::string_view name = it->first;
     config::IOption& option = *it->second;
     if (available_options_.find(name) == available_options_.end()) {
         if (ext_opt_set) return;
-        throw std::invalid_argument("Invalid option \"" + std::string{name} + '"');
+        throw config::ConfigurationError("Invalid option \"" + std::string{name} + '"');
     }
 
     if (option.IsSet()) {
--- a/src/core/algorithms/cfd/fd_first_algorithm.cpp
+++ b/src/core/algorithms/cfd/fd_first_algorithm.cpp
@@ -10,6 +10,7 @@
 #include "algorithms/cfd/util/set_util.h"
 #include "algorithms/cfd/util/tidlist_util.h"
 #include "config/equal_nulls/option.h"
+#include "config/exceptions.h"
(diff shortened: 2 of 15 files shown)


src/core/config/exceptions.h
```diff
@@ -0,0 +1,43 @@
...
+    explicit ConfigurationError(std::string message) : message_(std::move(message)) {}
+
+    const char* what() const noexcept override {
+        return message_.data();
+    }
+};
+
+class UnknownOption : public ConfigurationError {
+    using ConfigurationError::ConfigurationError;
+};
```
polyntsov, line 20: I think part of an exception message could be derived from the exception type and we should use this.
For example, almost any exception of type `UnknownOption` can start its message with "Unknown option: " with option name concatenated after. Or, as it is implemented right now, `"Option named " option name here " does not exist"`.
So I propose to add a constructor(s) from params from which the error message will be constructed using predefined message specific for exception type. This will unify exception messages and reduce amount of boilerplate (look, for example, at `InvalidChoice` usage in `src/python_bindings/py_to_any.cpp`)

src/core/algorithms/algorithm.cpp
```diff
@@ -90,13 +92,13 @@ void Algorithm::SetOption(std::string_view option_name, boost::any const& value)
...
         if (ext_opt_set) return;
-        throw std::invalid_argument("Unknown option \"" + std::string{option_name} + '"');
+        throw config::UnknownOption("Unknown option \"" + std::string{option_name} + '"');
     }
     std::string_view name = it->first;
     config::IOption& option = *it->second;
     if (available_options_.find(name) == available_options_.end()) {
         if (ext_opt_set) return;
-        throw std::invalid_argument("Invalid option \"" + std::string{name} + '"');
+        throw config::UnknownOption("Invalid option \"" + std::string{name} + '"');
```
polyntsov, line 101: There is the same exception type but different error messages because the errors are slightly logically different. Maybe we should create a separate exception type for existing but not available options?

polyntsov (changes requested):

src/core/algorithms/algorithm.cpp
BUYT-1: Now that I think of it, is there any situation where one would need to catch a particular type of configuration exception? Perhaps there should only be one `ConfigurationError`. What are your thoughts?

polyntsov: Well, I'm not sure about a possible usecase where we want to catch only a specific type of configuration related exceptions, but this doesn't seem unrealistic.. Also, separate exception types reduce amount of boilerplate code which generates exception messages.
However, all this is rather minor and right now there is no need to handle different configuration errors differently. I think it's acceptable to use only `ConfigurationError` and get rid of all other now redundant types. Other exceptions can be added when they're needed.

src/core/algorithms/cfd/fd_first_algorithm.cpp
```diff
@@ -54,36 +55,38 @@ unsigned long long FDFirstAlgorithm::ExecuteInternal() {
 }
 
 void FDFirstAlgorithm::CheckForIncorrectInput() const {
+    // TODO: should be checked by Option
```
polyntsov, line 58: Could you create an issue for this?

polyntsov (approved):

src/core/algorithms/cfd/fd_first_algorithm.cpp
BUYT-1: done

Issue #150, opened 2022-11-26
Fix ValueFromStr for DoubleType


Sched71:
`expected_ranges` contains string representations of double values. `num_type` has `std::unique_ptr<model::INumericType>` type. `byte_ranges.ranges` has `std::vector<const std::byte *>` type.
For `"7.4"` string in `expected_ranges`, when compairing `expected` to the value from `byte_ranges.ranges`, which string representation returned by `byte_ranges.num_type->ValueToString` equals `"7.4"`, the `EXPECT_EQ` fails:
``` 
    auto expected = std::unique_ptr<std::byte[]>(byte_ranges.num_type->Allocate());
    for (size_t i = 0; i < expected_ranges.size(); ++i) {
        byte_ranges.num_type->ValueFromStr(expected.get(), expected_ranges[i]);
        EXPECT_EQ(byte_ranges.num_type->Compare(expected.get(), byte_ranges.ranges[i]),
                  model::CompareResult::kEqual);
    }
```
But string representation of `expected` is `"7.4"`, so the `EXPECT_EQ` in the code below succeeds:
``` 
    auto expected = std::unique_ptr<std::byte[]>(byte_ranges.num_type->Allocate());
    for (size_t i = 0; i < expected_ranges.size(); ++i) {
        byte_ranges.num_type->ValueFromStr(expected.get(), expected_ranges[i]);
        EXPECT_EQ(byte_ranges.num_type->ValueToString(expected.get()),
                  byte_ranges.num_type->ValueToString(byte_ranges.ranges[i])); 
    }
```
Also I want to note, that values in `byte_ranges.ranges` were obtained by `NumericType::Add`

Also see https://github.com/Mstrutov/Desbordante/pull/124#discussion_r1026187015

polyntsov:
https://github.com/Mstrutov/Desbordante/pull/124#discussion_r1026187015

PR #267, opened 2023-09-09
Change model::DoubleType::Compare

Popov-Dmitriy-Ivanovich:
model::DoubleType::Compare returns kEqual if Double values are close enough.

### DIFF
--- a/src/core/model/types/double_type.h
+++ b/src/core/model/types/double_type.h
@@ -2,6 +2,9 @@
 
 #include <cassert>
 
+#include <boost/math/special_functions/next.hpp>
+#include <boost/math/special_functions/relative_difference.hpp>
+
 #include "numeric_type.h"
 
 namespace model {
@@ -10,11 +13,19 @@ class DoubleType final : public NumericType<Double> {
 public:
     DoubleType() noexcept : NumericType<Double>(TypeId::kDouble) {}
 
+    static const unsigned int kDefaultEpsCount = 5;
+
     CompareResult Compare(std::byte const* l, std::byte const* r) const final {
+        return CompareEPS(l, r, kDefaultEpsCount);
+    }
+
+    static CompareResult CompareEPS(std::byte const* l, std::byte const* r,
+                                    unsigned int eps_count) {
         Double l_val = GetValue(l);
         Double r_val = GetValue(r);
 
-        if (std::abs(l_val - r_val) < std::numeric_limits<Double>::epsilon()) {
+        if (boost::math::relative_difference(l_val, r_val) <
+            eps_count * std::numeric_limits<model::Double>::epsilon()) {
             return CompareResult::kEqual;
         }
 
--- a/src/tests/test_double_compare.cpp
+++ b/src/tests/test_double_compare.cpp
@@ -19,19 +19,19 @@ protected:
     std::byte* double_max;
     std::byte* result;
 
-    static void add(std::byte* a, std::byte* b, std::byte* res) {
+    static void Add(std::byte* a, std::byte* b, std::byte* res) {
         model::DoubleType double_type;
         double_type.Add(a, b, res);
     }
-    static void sub(std::byte* a, std::byte* b, std::byte* res) {
+    static void Sub(std::byte* a, std::byte* b, std::byte* res) {
         model::DoubleType double_type;
         double_type.Sub(a, b, res);
     }
-    static void mul(std::byte* a, std::byte* b, std::byte* res) {
+    static void Mul(std::byte* a, std::byte* b, std::byte* res) {
         model::DoubleType double_type;
         double_type.Mul(a, b, res);
     }
-    static void div(std::byte* a, std::byte* b, std::byte* res) {
+    static void Div(std::byte* a, std::byte* b, std::byte* res) {
         model::DoubleType double_type;
         double_type.Div(a, b, res);
     }
@@ -115,23 +115,23 @@ TEST_F(DoubleCompare, ComprasionFromAcAlgorithm) {
 }
 
 TEST_F(DoubleCompare, AddNumbers) {
-    TestArithmetic(7.7, 8.8, 7.7 + 8.8, add);
-    TestArithmetic(777.777, 888.888, 777.777 + 888.888, add);
+    TestArithmetic(7.7, 8.8, 7.7 + 8.8, Add);
+    TestArithmetic(777.777, 888.888, 777.777 + 888.888, Add);
 }
 
 TEST_F(DoubleCompare, SubNumbers) {
-    TestArithmetic(7.7, 8.8, 7.7 - 8.8, sub);
-    TestArithmetic(777.777, 888.888, 777.777 - 888.888, sub);
+    TestArithmetic(7.7, 8.8, 7.7 - 8.8, Sub);
+    TestArithmetic(777.777, 888.888, 777.777 - 888.888, Sub);
 }
 
 TEST_F(DoubleCompare, MulNumbers) {
-    TestArithmetic(7.7, 8.8, 7.7 * 8.8, mul);
-    TestArithmetic(777.777, 888.888, 777.777 * 888.888, mul);
+    TestArithmetic(7.7, 8.8, 7.7 * 8.8, Mul);
+    TestArithmetic(777.777, 888.888, 777.777 * 888.888, Mul);
 }
 
 TEST_F(DoubleCompare, DivNumbers) {
-    TestArithmetic(7.7, 8.8, 7.7 / 8.8, div);
-    TestArithmetic(777.777, 888.888, 777.777 / 888.888, div);
+    TestArithmetic(7.7, 8.8, 7.7 / 8.8, Div);
+    TestArithmetic(777.777, 888.888, 777.777 / 888.888, Div);
 }
 
 TEST_F(DoubleCompare, TwoSmallNumbers) {
@@ -153,11 +153,11 @@ TEST_F(DoubleCompare, EpsilonMin) {
 }
 TEST_F(DoubleCompare, LowFractionalNumber) {
     std::unique_ptr<std::byte[]> ten_power_minus_2(double_type.MakeValue(1e-2));
-    std::unique_ptr<std::byte[]> ten_power_minus_2_plus_eps_div_ten(
+    std::unique_ptr<std::byte[]> ten_power_minus_2_plus_eps_Div_ten(
             double_type.MakeValue(1e-2 + 2.5e-17));
 
     ASSERT_EQ(double_type_ref.Compare(ten_power_minus_2.get(),
-                                      ten_power_minus_2_plus_eps_div_ten.get()),
+                                      ten_power_minus_2_plus_eps_Div_ten.get()),
               model::CompareResult::kLess);
 }
 TEST_F(DoubleCompare, BigFractionalNumber) {
--- a/src/tests/test_double_compare.cpp
+++ b/src/tests/test_double_compare.cpp
@@ -1,0 +1,209 @@
+#include <boost/math/special_functions/next.hpp>
+#include <gtest/gtest.h>
+
+#include "types.h"
+
+namespace tests {
+
+class DoubleCompare : public ::testing::Test {
+private:
+    std::unique_ptr<std::byte[]> double_min_ptr;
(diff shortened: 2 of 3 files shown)


src/core/model/types/double_type.h
```diff
@@ -12,7 +13,7 @@ class DoubleType final : public NumericType<Double> {
         Double l_val = GetValue(l);
         Double r_val = GetValue(r);
 
-        if (std::abs(l_val - r_val) < std::numeric_limits<Double>::epsilon()) {
+        if (util::AlmostEqualUlps(l_val, r_val, 10)) {
```
polyntsov, line 16: Maybe 10 as a number of ulps works fine in some cases, but I'm not sure that we should use it for every double comparison at least without alternative method which allows to specify the precision. The desired precision actually depends on the nature of doubles we're comparing and hence on the task we're solving, for some algorithm 10 might be either too large or too small.

polyntsov: Also, please, stop using [magic number anti-pattern](https://en.wikipedia.org/wiki/Magic_number_(programming)) and replace all non-obvious values (that is all numbers besides 0 and 1 in most cases) with named constants. This regards all the code, including tests.

polyntsov (changes requested):

src/core/model/types/double_type.h
```diff
@@ -1,6 +1,8 @@
 #pragma once
 
 #include "numeric_type.h"
+#include <boost/math/special_functions/relative_difference.hpp>
```
polyntsov, line 4: Style guide requires an empty line between different include groups (that is, between local project headers and between library headers)

src/core/model/types/double_type.h
```diff
@@ -9,10 +11,16 @@ class DoubleType final : public NumericType<Double> {
     DoubleType() noexcept : NumericType<Double>(TypeId::kDouble) {}
 
     CompareResult Compare(std::byte const* l, std::byte const* r) const final {
+        return CompareULPs(l, r, 5);  // eps = 3 is too low for comparing doubles calculated using
```
polyntsov, line 14: I'd make this `5` a static public const field with some meaningfull name

src/core/model/types/double_type.h
```diff
@@ -9,10 +11,16 @@ class DoubleType final : public NumericType<Double> {
...
+                                      // arithmetic operations
+    }
+
+    CompareResult CompareULPs(std::byte const* l, std::byte const* r, int eps) const {
```
polyntsov, line 18: Why is it named `eps`? It's not an epsilon in any way, because of this `eps * std::numeric_limits<model::Double>::epsilon()` looks like we're multiplying epsilon values which is really strange

polyntsov, line 18: This function doesn't compare ULPs anymore

src/tests/test_double_compare.cpp
```diff
@@ -0,0 +1,159 @@
...
+    std::byte* less = less_ptr.get();
+    double_type_ref.Mul(number_128, double_min, result);
+    double_type_ref.Mul(number_100, double_min, less);
+    for (int i = 0; i < 128; i++) {
+        double_type_ref.Add(buff, double_min, buff);
+    }
+    ASSERT_EQ(double_type_ref.Compare(buff, result), model::CompareResult::kEqual);
+    ASSERT_EQ(double_type_ref.Compare(less, buff), model::CompareResult::kLess);
+}
+
```
polyntsov, line 46: Please, implement a test that compares two denormalized values with each over and a test which compares denormalized value with normalized value

src/core/model/types/double_type.h
polyntsov: Also let's replace `int` with `unsigned int` since this value is not expected to be negative

src/core/model/types/double_type.h
polyntsov: And the order of includes is wrong also

polyntsov (changes requested):
Please, fix the commit message names and the way the patch is splitted into commits, the history looks messy which makes reviewing harder than it should be.
Also, please, make sure that CI passes (all checks including formattiing) before requesting a review

polyntsov (changes requested):
Please, fix the commit messages, they're still wrong. Some of them have wrong format, they shouldn't end with dots and body of a commit message shouldn't have spaces in the beggining. Also "Change thing X" is not informative at all, make sure you read and understand [the guidelines](https://cbea.ms/git-commit/) on how to write a good commit message.
Also try to not add separate commits which are not useful to keep in history, like "Improve codestyle" changing the code which is being added in the current patch. Squash it to the commit which adds the changes initially.

polyntsov:
Closes #150

src/tests/test_double_compare.cpp
```diff
@@ -0,0 +1,220 @@
...
+    std::unique_ptr<std::byte[]> seven_seven(double_type.MakeValue(7.7));
+    std::unique_ptr<std::byte[]> eight_eight(double_type.MakeValue(8.8));
+    std::unique_ptr<std::byte[]> expect(double_type.MakeValue(7.7 * 8.8));
+
+    double_type_ref.Mul(seven_seven.get(), eight_eight.get(), result);
+    ASSERT_EQ(double_type_ref.Compare(result, expect.get()), model::CompareResult::kEqual);
+
+    seven_seven.reset(double_type.MakeValue(777.777));
+    eight_eight.reset(double_type.MakeValue(888.888));
+    expect.reset(double_type.MakeValue(691356.641976));
```
polyntsov, line 130:
```suggestion
    expect.reset(double_type.MakeValue(777.777 * 888.888));
```

src/tests/test_double_compare.cpp
```diff
@@ -0,0 +1,220 @@
...
+    std::unique_ptr<std::byte[]> few_eight(double_type.MakeValue(8.8));
+    std::unique_ptr<std::byte[]> expect(double_type.MakeValue(8888.8888 / 8.8));
+
+    double_type_ref.Div(many_eight.get(), few_eight.get(), result);
+    ASSERT_EQ(double_type_ref.Compare(result, expect.get()), model::CompareResult::kEqual);
+
+    expect.reset(double_type.MakeValue(8.8 / 8888.8888));
+    double_type_ref.Div(few_eight.get(), many_eight.get(), result);
+    ASSERT_EQ(double_type_ref.Compare(result, expect.get()), model::CompareResult::kEqual);
+}
```
polyntsov, line 146: All these methods have the same logic, it should be moved out to a function and reused here

src/tests/test_double_compare.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+    ASSERT_EQ(double_type_ref.Compare(double_zero, double_min), model::CompareResult::kEqual);
+    double_type_ref.Div(double_min, double_max, result);
+    ASSERT_EQ(double_type_ref.Compare(result, double_zero), model::CompareResult::kEqual);
+}
+
+TEST_F(DoubleCompare, ComprasionFromAcAlgorithm) {
+    std::unique_ptr<std::byte[]> seven_point_seven(double_type.MakeValue(7.7));
+    std::unique_ptr<std::byte[]> six_point_nine(double_type.MakeValue(6.9));
+    std::unique_ptr<std::byte[]> expect(double_type.MakeValue(14.6));
```
polyntsov, line 83:
```suggestion
    std::unique_ptr<std::byte[]> expect(double_type.MakeValue(7.7 + 6.9);
```

src/tests/test_double_compare.cpp
```diff
@@ -0,0 +1,220 @@
...
+        double_type_ref.Add(buff, double_min, buff);
+    }
+
+    ASSERT_EQ(double_type_ref.Compare(buff, result), model::CompareResult::kEqual);
+    ASSERT_EQ(double_type_ref.Compare(less, buff), model::CompareResult::kLess);
+}
+
+TEST_F(DoubleCompare, DivMax) {
+    std::unique_ptr<std::byte[]> three_ptr(double_type.MakeValue(3.0));
+    std::unique_ptr<std::byte[]> three_power_20_ptr(double_type.MakeValue(3486784401.0));
```
polyntsov, line 56:
```suggestion
    std::unique_ptr<std::byte[]> three_power_20_ptr(double_type.MakeValue(std::pow(3., 20.)));
```
And twenty can be stored in a variable `power` since it's also used later in the loop condition

polyntsov:
First commit (“Rework DoubleType::Compare to work correctly with any floating point number”) shouldn't be kept in history, since it just incorrectly implements double comparison, squash it into the second one. Also, commit messages should be less than 72 symbols in length, this is the value many graphical git tools use when showing commit messages (look, for example, how GitHub formats this commit message)

src/tests/test_double_compare.cpp
```diff
@@ -0,0 +1,207 @@
...
+namespace tests {
+
+class DoubleCompare : public ::testing::Test {
+private:
+    std::unique_ptr<std::byte[]> double_min_ptr;
+    std::unique_ptr<std::byte[]> double_max_ptr;
+    std::unique_ptr<std::byte[]> result_ptr;
+
+protected:
+    void test_arithmetic(model::Double a, model::Double b, model::Double expected,
```
polyntsov, line 15: Google style guide names methods using `PascalCase`

polyntsov, line 15: I'd rename a and b to left and right, I think it would be more readable

src/tests/test_double_compare.cpp
```diff
@@ -0,0 +1,207 @@
...
+    auto sub_ = [](std::byte* a, std::byte* b, std::byte* res) {
+        model::DoubleType double_type;
+        double_type.Sub(a, b, res);
+    };
+    test_arithmetic(7.7, 8.8, 7.7 - 8.8, sub_);
+    test_arithmetic(777.777, 888.888, 777.777 - 888.888, sub_);
+}
+
+TEST_F(DoubleCompare, MulNumbers) {
+    auto mul_ = [](std::byte* a, std::byte* b, std::byte* res) {
```
polyntsov, line 118: local variables shouldn't have the underscore in the end of their names

src/tests/test_double_compare.cpp
polyntsov: Also move this method definition below fields definitions, don't mix them with each other

src/tests/test_double_compare.cpp
```diff
@@ -0,0 +1,208 @@
...
+        double_min_ptr.reset(double_type.MakeValue(std::numeric_limits<model::Double>().min()));
+        double_max_ptr.reset(double_type.MakeValue(std::numeric_limits<model::Double>().max()));
+        result_ptr.reset(double_type.MakeValue(0.0));
+        result = result_ptr.get();
+        double_min = double_min_ptr.get();
+        double_max = double_max_ptr.get();
+    }
+
+    void TestArithmetic(model::Double left, model::Double right, model::Double expected,
+                        std::function<void(std::byte*, std::byte*, std::byte*)> operation) {
```
polyntsov, line 32: nit: In this situation, template is a better approach than `std::function` ([rationale](https://stackoverflow.com/questions/14677997/stdfunction-vs-template)).

polyntsov (commented):
Everything looks good except for the last commit "[Improve DoubleCompare test readability](https://github.com/Mstrutov/Desbordante/pull/267/commits/e69fed5e76fe8ac8af758e87bc0ef526b524e759)". It shouldn't really be kept in history, instead it should be squashed to a commit which implements the tests for double comparison. The best approach here IMO is to split the first commit into two separate commits; first fixes `DoubleCompare`, second adds tests, then squash the third (currently) commit into the second one.

polyntsov (dismissed):

src/tests/test_double_compare.cpp
```diff
@@ -0,0 +1,209 @@
...
+        double_type.Sub(a, b, res);
+    }
+    static void mul(std::byte* a, std::byte* b, std::byte* res) {
+        model::DoubleType double_type;
+        double_type.Mul(a, b, res);
+    }
+    static void div(std::byte* a, std::byte* b, std::byte* res) {
+        model::DoubleType double_type;
+        double_type.Div(a, b, res);
+    }
```
polyntsov, line 37: Functions are named with the first letter capitalized

polyntsov (approved):

PR #269, opened 2023-09-12
Fix operator precedence error in dataframe reader

BUYT-1:


src/python_bindings/dataframe_reader.cpp
```diff
@@ -56,13 +56,13 @@ std::vector<std::string> StringDataframeReader::GetNextRow() {
 
 std::vector<std::string> ArbitraryDataframeReader::GetNextRow() {
     std::vector<std::string> strings{};
-    auto list = *df_iter_++.cast<py::list>();
-    for (py::handle el : list) {
+    auto tuple_row = py::reinterpret_borrow<py::object>(*df_iter_++);
+    for (py::handle el : tuple_row) {
```
polyntsov, line 60: I don't really like expressions like `*df_iter_++` (and `*df_iter_++.cast<py::list>();` was atrocious), IMO it's better to use separate variables to state the intention clearly and thereby to prevent the problems like the one that's being fixed.

polyntsov (changes requested):

polyntsov (approved):

PR #270, opened 2023-09-13
Column layout classes cleanup

BUYT-1:


src/core/model/table/column.h
```diff
@@ -16,10 +16,8 @@ class Column {
     RelationalSchema const* schema_;
 
 public:
-    Column(RelationalSchema const* schema, std::string name, int index) :
-        name_(std::move(name)),
-        index_(index),
-        schema_(schema) {}
+    Column(RelationalSchema const* schema, std::string name, size_t index)
+        : name_(std::move(name)), index_(index), schema_(schema) {}
```
polyntsov, line 20: If we're going down the path of replacing `int` for indices with `size_t`, we should also create an alias for this type  to make next such refactoring much easier (and the code more readable). 
Also, almost any algorithm works with columns and represents them using indices of some type, ideally these types should all be replaced with the introduced alias. However, it may be a lot of work to do, so I think it's acceptable to write a TODO comment/create an issue about this.

src/core/model/table/column_layout_relation_data.cpp
```diff
@@ -20,56 +20,45 @@ std::vector<int> ColumnLayoutRelationData::GetTuple(int tuple_index) const {
...
     while (data_stream.HasNextRow()) {
         row = data_stream.GetNextRow();
 
-        if (row.empty() && num_columns == 1) {
-            row.emplace_back("");
-        } else if ((int)row.size() != num_columns) {
-            LOG(WARNING) << "Skipping incomplete rows";
+        if (row.size() != num_columns) {
+            LOG(WARNING) << "Unexpected number of columns for a row, skipping (expected "
+                         << num_columns << ", got " << row.size() << ")";
```
polyntsov, line 37: I think the warning message should also contain the previous message, information that such rows are skipped is important

src/core/model/table/column_layout_typed_relation_data.cpp
```diff
@@ -23,32 +18,20 @@ std::unique_ptr<ColumnLayoutTypedRelationData> ColumnLayoutTypedRelationData::Cr
...
-                    break;
-                }
-            }
-        } else {
-            LOG(WARNING) << "Processed " << row_num << " rows and " << max_rows - row_num
-                         << " rows remain unprocessed due to `max_rows` parameter.";
-            break;
+        for (size_t index = 0; index < row.size(); ++index) {
+            std::string& field = row[index];
+            columns[index].push_back(std::move(field));
```
polyntsov, line 29: Why not range-based for?

polyntsov (changes requested):

src/core/model/table/column.h
polyntsov: BTW, there is already an alias for this in `src/core/algorithms/fd/hycommon/types.h`, probably we can extract it and reuse here

src/core/model/table/column_layout_typed_relation_data.cpp
BUYT-1: Forgot to change.

src/core/model/table/column_layout_relation_data.cpp
BUYT-1: It already says "skipping", is it not enough?

polyntsov: oops, didn't see it

src/core/model/table/column_layout_typed_relation_data.cpp
BUYT-1: No, hold on, `columns[index]` uses `index`, I can't use a range-based `for` here.

src/core/model/table/column.h
polyntsov: I meant that reusing the alias in all algorithms can be postponed, introducing the alias IMHO should be done in this PR along with changing the index type from `int` to `size_t`. Creating something like `Column::IndexType = size_t` and using it instead of `int` for `Column` index is simple enough, but still eases the next refactoring of rewriting all algorithms to use it

polyntsov (approved):

PR #272, opened 2023-09-18
Implement infrastructure for IND discovery and Faida algorithm

alexandrsmirn:
This pull request introduces Faida algorithm for inclusion dependency discovery.

It is easier to review this PR by individual commits:
* The first commit adds infrastructure for IND discovery algorithms. Also the commit adds support for multi-file datasets in Desbordante;
* The second commit adds the implementation of Faida IND discovery algorithm and some tests for the algorithm;
* Third commit integrates Faida to Desbordante codebase.

Node: this PR contains source code of third-party libraries: emhash, murmurhash3, atomicbitvector. The code is added in the second commit. This is a temporary solution, so later third-party code should be removed, and libraries should be downloaded at the build stage.


vs9h (changes requested):
One of these days I will try to review some points in more detail and add some comments.

src/core/algorithms/ind/faida/faida.h
```diff
@@ -0,0 +1,58 @@
...
+
+#include "algorithms/ind/faida/inclusion_testing/iinclusion_tester.h"
+#include "algorithms/ind/faida/preprocessing/preprocessor.h"
+#include "algorithms/ind/faida/util/simple_ind.h"
+#include "algorithms/ind/ind_algorithm.h"
+#include "config/thread_number/type.h"
+
+namespace algos {
+
+class Faida : public algos::INDAlgorithm {
```
vs9h, line 11:
```suggestion
class Faida : public INDAlgorithm {
```

src/core/algorithms/ind/faida/faida.h
```diff
@@ -0,0 +1,58 @@
...
+    std::vector<SimpleIND> TestCandidates(std::vector<SimpleIND> const& candidates);
+    void RegisterInds(std::vector<SimpleIND> const& inds);
+
+public:
+    Faida();
+
+    void LoadDataInternal() final;
+    void MakeExecuteOptsAvailable() final;
+    unsigned long long ExecuteInternal() final;
+    void ResetState() final;
```
vs9h, line 43: `LoadDataInternal`/`ExecuteInternal`/`ResetState` - private
`MakeOptionsAvailable` - protected

src/tests/test_faida.cpp
```diff
@@ -0,0 +1,227 @@
...
+    }
+    SUCCEED();
+}
+
+class INDAlgorithmTest : public ::testing::Test {
+protected:
+    static algos::StdParamsMap GetParamMap(std::filesystem::path const& path, char separator = ',',
+                                           bool hasHeader = true, int sample_size = 500,
+                                           double hll_accuracy = 0.001, bool find_nary = true,
+                                           unsigned short num_threads = 16) {
```
vs9h, line 75: I'm not sure if specifying 16 threads by default is a good idea --- that seems like a lot. In the UCC tests I saw a maximum of 4 threads.

src/tests/test_faida.cpp
```diff
@@ -0,0 +1,227 @@
...
+            FAIL() << "Expected table name '" << expected[table_id] << "', but received '"
+                   << actual[table_id].table_name << "'";
+        }
+    }
+    SUCCEED();
+}
+
+class INDAlgorithmTest : public ::testing::Test {
+protected:
+    static algos::StdParamsMap GetParamMap(std::filesystem::path const& path, char separator = ',',
```
vs9h, line 72:
```suggestion
    static algos::StdParamsMap GetParamMap(fs::path const& path, char separator = ',',
```

src/core/algorithms/algo_factory.h
```diff
@@ -77,6 +78,36 @@ void LoadAlgorithm(Algorithm& algorithm, OptionMap&& options) {
                     details::ExtractOptionValue<char>(options, kSeparator),
                     details::ExtractOptionValue<bool>(options, kHasHeader));
             return boost::any{parser};
+        } else if (option_name == kTableSet
+                   && options.find(std::string{kTableSet}) == options.end()) {
+            auto const data = details::ExtractOptionValue<std::filesystem::path>(options, kCsvPath);
```
vs9h, line 83: Is it typo or we want to use `kCsvPath`?

src/tests/test_faida.cpp
alexandrsmirn: Yes, I agree. Changed to 4.

src/core/algorithms/algo_factory.h
```diff
@@ -77,6 +78,36 @@ void LoadAlgorithm(Algorithm& algorithm, OptionMap&& options) {
                     details::ExtractOptionValue<char>(options, kSeparator),
                     details::ExtractOptionValue<bool>(options, kHasHeader));
             return boost::any{parser};
+        } else if (option_name == kTableSet
+                   && options.find(std::string{kTableSet}) == options.end()) {
```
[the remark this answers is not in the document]
vs9h: There is also an interesting question: should we introduce a new option `--csv_paths` for multiple files or reuse the existing `--csv_path` field.
It looks like we'll have to introduce a new option. Otherwise, it will be necessary to change the existing code (for example, in tests the `--csv_path` parameter is often set and a single path is passed, but it will be necessary to pass a vector of paths)
But on the other hand, it may not be obvious to the user which option from the console to use when he wants to run an algorithm on one dataset that expects several datasets.
It may be necessary to be able to allow the `TableSet` to take into account the values of both options.

src/core/config/tabular_data/input_tableset_type.h
```diff
@@ -0,0 +1,14 @@
...
+#include <memory>
+
+#include "model/table/idataset_stream.h"
+
+namespace config {
+
+struct InputTableSet {
+    std::string name;
+    std::vector<std::shared_ptr<model::IDatasetStream>> parsers;
+};
```
vs9h, line 12: I believe this type should be defined like this:
```suggestion
using InputTableSet = std::vector<InputTable>;
```

src/core/algorithms/ind/ind_algorithm.h
```diff
@@ -0,0 +1,35 @@
...
+        std::vector<std::string> header;
+    };
+    using DatasetsOrder = std::vector<DatasetInfo>;
+    using IND = model::IND;
+    using INDList = std::list<IND>;
+
+public:
+    explicit INDAlgorithm(std::vector<std::string_view> phase_names);
+
+    virtual DatasetsOrder const& GetDatasetsOrder() const = 0;
```
vs9h, line 29: I don’t really like the fact that every developer who will implement algorithms for finding inclusion dependencies will need to implement this function.
Firstly, the user himself knows the datasets in what order he specified (this will be true after we switch to using a boost multitoken and manually tasking the list of tables) --- therefore, he must also know the order in which the datasets are specified.
Secondly, if we still want to give this information, then it is better to do it in a uniform way for all algorithms (which will be inherited from this class).
For example, we could simply add the following method:
```cpp
config::InputTableSet const& GetInputTableSet() const {
    return input_table_set_;
}
```
This will work well as we can get any information we need (table name/header).

src/core/algorithms/ind/ind_algorithm.h
```diff
@@ -0,0 +1,35 @@
...
+    virtual INDList IndList() const = 0;
```
vs9h, line 30: It's better to use `INDList const&` return type in order to avoid unnecessary copying (in the case when the algorithm uses the same type and does not construct a new list of dependencies with each call)

src/core/algorithms/ind/ind_algorithm.h
vs9h: Also, in order to avoid the need to provide the user information about the order of the specified tables with the corresponding column names, we can add a new user-friendly `RawIND` class, which will contain information about the names and indices of the table and the names and indices of the columns themselves.
```cpp
namespase model {

struct RawIND {
   struct AttributeInfo {
       unsigned int attribute_id;
       string attribute_name;
       unsigned int column_id;
       string column_name;
   }; 
   using INDSide = std:vector<AttributeInfo>;
   
   INDSide lhs, rhs;
};

};
```
And then add to existing `INDAlgorithm` class methods to transform dependencies from `model::IND` to `model::RawIND`.
For example we can add public method that takes one dependency and gives a `RawIND`: `INDAlgorithm::ToRawIND(model::IND const& ind)`
And then add public method `INDAlgorithm::GetRawINDs()`.

But if we want to do this, then it is better to think about naming: I would prefer to rename the existing class `model::IND` to `model::INDCC`, and suggested `model::RawIND` to `model::IND`

src/core/config/tabular_data/input_tableset_type.h
vs9h: nit: But I would think about using the word `set` in the naming --- this may lead to associations with the fact that we store tables using `std::set`.
I would like to know the opinion of reviewers who will also review this pull request.

src/core/algorithms/ind/faida/candidate_generation/apriori_candidate_generator.cpp
```diff
@@ -0,0 +1,99 @@
...
+            if (!first_ind.StartsWith(second_ind)) {
+                break;
+            }
+
+            if (first_ind.left()->GetLastColumn() == second_ind.left()->GetLastColumn() ||
+                first_ind.right()->GetLastColumn() == second_ind.right()->GetLastColumn()) {
+                continue;
+            }
+
+            CreateCandidate(first_ind, second_ind, ccs_on_level, last_result_set, candidates);
```
vs9h, line 92: nit: you can use `std::reference_wrapper` to get rid of the need to write a transformation and a custom comparator for sorting. There is also no need to write a double dereference (**).
The only drawback I find is that it is not possible to fill `last_result_set` using the default constructor due to the fact that the types are different. I think this version of the function looks a little better:

But it seems to me that this is insignificant and the pros outweigh the cons.
```cpp
    std::vector<std::reference_wrapper<const SimpleIND>> last_result(inds.begin(), inds.end());
    std::sort(last_result.begin(), last_result.end(), std::less<const SimpleIND&>{});

    std::unordered_set<const SimpleIND*> last_result_set;
    for (const auto& ind_rw : last_result) last_result_set.insert(&ind_rw.get());

    std::unordered_set<std::shared_ptr<SimpleCC>> ccs_on_level;
    std::vector<SimpleIND> candidates;
    for (auto left_iter = last_result.begin(); left_iter != last_result.end(); ++left_iter) {
        for (auto right_iter = std::next(left_iter); right_iter != last_result.end();
             ++right_iter) {
            SimpleIND const& left_ind = left_iter->get();
            SimpleIND const& right_ind = right_iter->get();

            if (!left_ind.StartsWith(right_ind)) {
                break;
            }

            if (left_ind.left()->GetLastColumn() == right_ind.left()->GetLastColumn() ||
                left_ind.right()->GetLastColumn() == right_ind.right()->GetLastColumn()) {
                continue;
            }

            CreateCandidate(left_ind, right_ind, ccs_on_level, last_result_set, candidates);
```

src/core/config/descriptions.h
```diff
@@ -70,4 +70,10 @@ constexpr auto kDWeight =
 constexpr auto kDBumpsLimit = "max considered intervals amount. Pass 0 to remove limit";
 constexpr auto kDIterationsLimit = "limit for iterations of sampling";
 constexpr auto kDACSeed = "seed, needed for choosing a data sample";
+constexpr auto kDTableSet =
+        "A set of tables constituting some dataset, which will be "
+        "processed by an algorithm";
+constexpr auto kDHllAccuracy = "(Faida special) HyperLogLog approximation accuracy";
+constexpr auto kDSampleSize = "(Faida special) Size of a table sample";
+constexpr auto kDFindNary = "(Faida special) Detect n-ary inclusion dependencies [true|false]";
```
vs9h, line 78: In the case of functional dependencies in the desbordante, it is possible to explicitly set a limit on the number of attributes on the left side of the dependency (`--max_lhs`). Is there any point and opportunity to do something similar here? This will be more consistent (relative to existing code) and convenient (for the user)

src/core/config/tabular_data/input_tableset_type.h
vs9h: You can use `IDatasetStream::GetRelationName()` - it seems that this will be enough.
Also now the `name` field looks a little strange, in one case it is the name of the directory (when we pass the path to the directory with files), in the other case it is the file itself (when the path to the file is passed).

Suggested name `InputTableCollection` acceptable

src/core/config/descriptions.h
vs9h: Thanks for the answer, I like the suggested name for the option.
It would be good to open an issue if we don’t want to implement it as part of this pr

src/core/algorithms/ind/faida/faida.cpp
```diff
@@ -0,0 +1,232 @@
...
+    std::vector<SimpleIND> candidates;
+    candidates.reserve(combinations.size() * combinations.size());
+
+    for (auto left_it = combinations.begin(); left_it != combinations.end(); left_it++) {
+        for (auto right_it = combinations.begin(); right_it != combinations.end(); right_it++) {
+            if (left_it != right_it) {
+                candidates.emplace_back(*left_it, *right_it);
+            }
+        }
+    }
```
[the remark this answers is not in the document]
vs9h: Thanks for the good point, comparison by pointer is really something strange. But also, since the `SimpleIND` class has a comparison operator defined, we can use this operator. It also seems like this is exactly what we need semantically.
It will look like this:
```suggestion
    for (auto const& left : combinations) {
        for (auto const& right : combinations) {
            if (left != right) {
                candidates.emplace_back(left, right);
            }
        }
    }
```

src/core/algorithms/ind/faida/inclusion_testing/hyperloglog.h
```diff
@@ -0,0 +1,235 @@
...
+#endif
+
+}
+#define _GET_CLZ(x, b) _get_leading_zero_count(x, b)
+#endif /* defined(__GNUC__) */
+
+namespace hll {
+
+static const double pow_2_32 = 4294967296.0;       ///< 2^32
+static const double neg_pow_2_32 = -4294967296.0;  ///< -(2^32)
```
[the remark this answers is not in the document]
vs9h: Thanks for the good comment.
I don't know the desbordante's policy regarding what to do with code that was taken from other sources. It seems that such code is anyway modified to match the coding style in the desbordante (the same happened with conditional functional dependencies)
I'd like to hear opinion @polyntsov. It seems to me that it would be worth correcting the naming and, for example, correcting this remark.

src/core/algorithms/ind/faida/inclusion_testing/hyperloglog.h
```diff
@@ -0,0 +1,235 @@
...
+
+    int zcf(size_t hash, int prec) {
+        size_t mask = ~(((1 << prec) - 1) << (64 - prec));
+        return _lzcnt_u64(hash & mask) - (uint8_t) prec;
+    }
+
+    void add_hash(size_t value_hash) {
+        int index = (int) (value_hash >> (8 * sizeof(size_t) - b_));
+        int rank = _lzcnt_u64((value_hash << b_) | ((1 << (b_ - 1)) + 1)) + 1;
+        // int rank = zcf(value_hash, b_) + 1;
```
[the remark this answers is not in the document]
vs9h: Here I do not have enough competence to assess the correctness of this approach.
I would prefer to use a macro with a comment explaining its meaning.

src/tests/test_faida.cpp
```diff
@@ -0,0 +1,227 @@
...
+TEST_F(INDAlgorithmTest, TestWide2) {
+    IndsTest expected_inds{{{0, {2}}, {0, {0}}}, {{0, {3}}, {0, {1}}}, {{0, {2, 3}}, {0, {0, 1}}}};
+
+    auto file_name = "TestWide2.csv";
+    char separator = ',';
+    bool has_header = false;
+    int sample_size = 500;
+    double hll_accuracy = 0.001;
+    bool find_nary = true;
+    unsigned short num_threads = 16;
```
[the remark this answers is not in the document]
vs9h: With this motivation, you can completely remove the default parameters

src/core/config/tabular_data/input_tableset_type.h
alexandrsmirn: `name` here is not the name of a table, but the name of a _whole_ dataset (a collection of tables). In the current approach the user specifies folder name, where csv-s are locaded, and this folder name becomes `InputTableSet.name`.

src/tests/test_faida.cpp
```diff
@@ -0,0 +1,227 @@
...
+    for (unsigned table_id = 0; table_id != expected.size(); ++table_id) {
+        if (expected[table_id] != actual[table_id].table_name) {
+            FAIL() << "Expected table name '" << expected[table_id] << "', but received '"
+                   << actual[table_id].table_name << "'";
+        }
+    }
+    SUCCEED();
+}
+
+class INDAlgorithmTest : public ::testing::Test {
```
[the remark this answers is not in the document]
vs9h: It might make sense to note the motivation for creating a separate file (about the approximate algorithm) in this file, for example in the description of a class for a test. This will help people who will learn code in the future.

src/core/algorithms/ind/faida/faida.h
vs9h: Interestingly, in the base class `MakeOptionsAvailable` is protected, but it is only used in the base class (`Algorithm`). I think, that it would be more correct to use 'private'.

src/core/algorithms/ind/ind_algorithm.h
vs9h: It seems to me that we don’t need this `DatasetInfo` at all. We have `input_table_collection_,` from which we can get both the name of the relation and the names of the columns.

src/core/algorithms/ind/faida/faida.cpp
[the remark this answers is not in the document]
alexandrsmirn: `left` and `right` here are  instances of `SimpleCC` which is essentially a vector of column indices, and the comparison operator compares this vectors per-element, which is not so fast. The size of `combinations` here equals to the total count of columns in all tables, so the comparison operation is perfrormed square of number of columns times.

src/core/algorithms/ind/faida/inclusion_testing/hyperloglog.h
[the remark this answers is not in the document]
alexandrsmirn: I do not think that the marco is needed (and do not know how a macro can be used here). This is just a line I did not use but did not want to just delete. In fact, it is mostly for me, in case in the future a problem occurs.

src/core/algorithms/ind/faida/hashing/hashing.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
+
+#include <string>
+
+#include "algorithms/ind/faida/include_lib/murmurhash3/murmur_hash_3.h"
+#include "immintrin.h"
+
+namespace algos::faida::hashing {
+
+inline size_t __attribute__((always_inline)) rotl(size_t hash, int num_bits) {
```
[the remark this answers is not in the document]
vs9h: It is usually believed that the compiler is smart enough and can handle the inlining of the necessary functions itself. And therefore, declaring a function as inline often will not affect anything.

I would prefer to ask @polyntsov for his opinion on this matter

src/core/config/tabular_data/input_tableset_type.h
vs9h: In a personal dialogue we discussed it in more detail and decided that we would remove the `name` field

src/core/config/tabular_data/input_table_collection_type.h
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
+#include <memory>
+
+#include "input_table_type.h"
+#include "model/table/idataset_stream.h"
```
vs9h, line 6: Don't forget to fix includes:
```suggestion
#include <vector>

#include "input_table_type.h"
```

src/tests/test_faida.cpp
```diff
@@ -0,0 +1,231 @@
...
+                   << actual[table_id]->GetRelationName() << "'";
+        }
+    }
+    SUCCEED();
+}
+
+class FaidaINDAlgorithmTest : public ::testing::Test {
+protected:
+    static algos::StdParamsMap GetParamMap(fs::path const& path, char separator = ',',
+                                           bool hasHeader = true, int sample_size = 500,
```
vs9h, line 73: `hasHeader` -> `has_header`

src/core/algorithms/ind/ind.h
```diff
@@ -0,0 +1,99 @@
...
+    struct ColumnCombinationInfo {
+        size_t table_idx;
+        std::shared_ptr<std::string> table_name;
+        std::vector<ColumnInfo> columns;
+
+        std::string ToString() const {
+            std::stringstream ss;
+            ss << *table_name << '[';
+
+            int col_idx = 0;
```
vs9h, line 77: Better use `size_t` to avoid `static_cast` on 80 line

vs9h:
@alexandrsmirn, please take a look at the two pull requests (#295 and #296). We communicated in a personal dialogue and I hope that I was able to convince of the need to merge the correct option for working with tables.
Changes:
- added a `--csv_paths` option that expects a vector of paths (in tests it will now be necessary to explicitly pass vector of files)
- minor changes to the table collection option (the option and the corresponding files have been renamed)
- removed `PrettyIND` class
- added type aliases for table and column indexes (now both `unsigned int`)

Regarding the `PrettyIND` class: the task of transformation into a more user-friendly form is more complex than it might seem at first glance. We need to understand whether we want to do this in the c++ core or in python. Also what type should be used. 
And since there is now no ambiguity in the order of the tables, there is no need to give information about the order of the tables.

src/core/config/descriptions.h
alexandrsmirn: Created an issue: https://github.com/Mstrutov/Desbordante/issues/302

src/core/algorithms/ind/faida/inclusion_testing/sampled_inverted_index.h
```diff
@@ -0,0 +1,90 @@
...
+#include <boost/dynamic_bitset.hpp>
+
+#include "algorithms/ind/faida/util/simple_ind.h"
+#include "hll_data.h"
+
+namespace algos::faida {
+
+class SampledInvertedIndex {
+private:
+    struct SpinLock {
```
vs9h, line 19: Have you tried using a spinlock from `boost`? @polyntsov is sure that writing a custom spinlock is not worth it.
If this implementation of a spinlock is faster than a `boost`, then you need to attach the results of experiments and explicitly move this implementation into the `util` directory

src/core/config/all_options.cpp
vs9h, line 1: It is better to immediately remove changes from this file, because very soon we will completely remove parsing options code from the с++ (see #309)

vs9h (changes requested):
`Faida` is a false positive algorithm, so it should find all dependencies, plus it may find the wrong dependencies. So I decided to run the `Faida` for unary dependencies (that is, I specified the appropriate option `findNary: false`) on my tests to make sure that the `Faida` works correctly. I found some differences in the results:
- for `TestWide.csv` `Faida` doesn't find any dependencies (but expected `2 -> 3` and `3 -> 2`)
- for `neighbors10k.csv` and `neighbors100k.csv` `Faida` doesn’t find dependency `5 -> 6` (expected `3 -> 4`, `4 -> 3`, `5 -> 6`, but last dependency wasn’t found — column 5 have unique values {“2”}, column 6 have unique values {“1”, “2”})

I also have a question that arose when comparing the results of dependency mining on the dataset `CIPublicHighway700.csv` and `EpicMeds.csv` (these datasets are necessary for testing work with nulls). Therefore, I will not provide the expected results, but will propose to discuss the rules for inferring inds in the case of working with nulls. I would also like to hear the opinion from @chernishev. I have also already discussed these rules with @polyntsov.

For simplicity, we will talk about unary dependencies. We will use the following notation:
Suppose dependency `A -> B` satisfied, where `A` and `B` are columns without nulls. Also, the notation `A + nulls` means that we added nulls to column `A`. 
Then the following is true:
- `A + nulls -> B` not satisfied
- `A -> B + nulls` satisfied
- `A + nulls -> B + nulls` not satisfied (if flag `null_equal_nulls` is `false`)
- `A + nulls -> B + nulls` satisfied (if flag `null_equal_nulls` is `true`)

We also separately consider the case when one of the columns consists entirely of nulls (`nulls -> B` and `A -> nulls`) - such dependencies are not satisfied.

src/core/algorithms/ind/faida/hashing/hashing.h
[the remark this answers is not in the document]
polyntsov: Firstly, `inline` keyword in c++ has nothing to do with function inlining, however, since we want to define this function in the header, `inline` is perfectly fine.
I think that compiler will inline it when compiling with optimizations regardless of presence of `__attribute__((always_inline))` so removing it won't influence performance. At the same time, this is a compiler extension, and IMO it's better not to use compiler extensions when we can (and here we certainly can).
Actually, I think that the best solution is to wait until C++ version is bumped to c++20 and use `std::rotl` here

src/core/algorithms/ind/faida/inclusion_testing/hyperloglog.h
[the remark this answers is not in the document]
polyntsov: I wouldn't change the library code unless absolutely necessary. Or better yet, I wouldn't commit the third-party library code, instead adding it as a dependency.

src/core/algorithms/ind/faida/inclusion_testing/sampled_inverted_index.h
alexandrsmirn: No, I have not tried it. In this PR I will not perform any experiments, so if you think this piece of code somehow impacts the code quality, I will just remove it. However, I still do not see any problems with leaving it here, since this is just an experimental feature located in the local code and related only to Faida. If in the future any programmer will want to use  spinlock, it may use one from boost himself.

alexandrsmirn:
> `Faida` is a false positive algorithm, so it should find all dependencies, plus it may find the wrong dependencies. So I decided to run the `Faida` for unary dependencies (that is, I specified the appropriate option `findNary: false`) on my tests to make sure that the `Faida` works correctly. I found some differences in the results:
> 
>     * for `TestWide.csv` `Faida` doesn't find any dependencies (but expected `2 -> 3` and `3 -> 2`)
> 
>     * for `neighbors10k.csv` and `neighbors100k.csv` `Faida` doesn’t find dependency `5 -> 6` (expected `3 -> 4`, `4 -> 3`, `5 -> 6`, but last dependency wasn’t found — column 5 have unique values {“2”}, column 6 have unique values {“1”, “2”})
> 
> 
> I also have a question that arose when comparing the results of dependency mining on the dataset `CIPublicHighway700.csv` and `EpicMeds.csv` (these datasets are necessary for testing work with nulls). Therefore, I will not provide the expected results, but will propose to discuss the rules for inferring inds in the case of working with nulls. I would also like to hear the opinion from @chernishev. I have also already discussed these rules with @polyntsov.
> 
> For simplicity, we will talk about unary dependencies. We will use the following notation: Suppose dependency `A -> B` satisfied, where `A` and `B` are columns without nulls. Also, the notation `A + nulls` means that we added nulls to column `A`. Then the following is true:
> 
>     * `A + nulls -> B` not satisfied
> 
>     * `A -> B + nulls` satisfied
> 
>     * `A + nulls -> B + nulls` not satisfied (if flag `null_equal_nulls` is `false`)
> 
>     * `A + nulls -> B + nulls` satisfied (if flag `null_equal_nulls` is `true`)
> 
> 
> We also separately consider the case when one of the columns consists entirely of nulls (`nulls -> B` and `A -> nulls`) - such dependencies are not satisfied.

This was because my implementation of Faida used some optimization and ignored columns which are constant (the sutuation when all the values in a column are the same), and the columns which are null-columns (all values are null). Now I added two options to control this feature, and now it disabled by default, so the results should be the same. Also, the algorithm now supports the rules proposed by you (except that Faida does not support the semantics `null_equal_nulls == false`, only `null_equal_nulls == true`)

src/tests/test_faida.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+    ASSERT_THROW(CreateFaidaInstance(std::vector{path}, separator, has_header, sample_size,
+                                     hll_accuracy, find_nary, num_threads),
+                 std::runtime_error);
+}
+
+TEST_F(FaidaINDAlgorithmTest, TestEmptyFolder) {
+    IndsTest expected_inds{};
+
+    auto file_name = "empty_dataset";
```
vs9h, line 124: it is worth adding configs to the [all_tables_config.h](https://github.com/Mstrutov/Desbordante/blob/main/src/tests/all_tables_config.h) and using them in this file

src/core/config/descriptions.h
```diff
@@ -77,4 +77,13 @@ constexpr auto kDWeight =
...
 constexpr auto kDACSeed = "seed, needed for choosing a data sample";
+constexpr auto kDHllAccuracy = "(Faida special) HyperLogLog approximation accuracy";
+constexpr auto kDSampleSize = "(Faida special) Size of a table sample";
+constexpr auto kDFindNary = "(Faida special) Detect n-ary inclusion dependencies [true|false]";
+constexpr auto kDIgnoreNullCols =
+        "(Faida special) Ignore INDs which contain columns filled only with NULLs. May increase "
+        "performance but impacts the result. [true|false]";
+constexpr auto kDIgnoreConstantCols =
+        "(Faida special) Ignore INDs which contain columns filled with only one value. May "
+        "increase performance but impacts the result. [true|false]";
```
vs9h, line 88: You definitely need to remove these postscripts at the beginning. From python `cli` we can get a set of its options for each specific algorithm. And these inscriptions `(Faida special)` will look strange.

vs9h (approved):
I left a couple of small comments, but I approve.

polyntsov (approved):

PR #276, opened 2023-10-21
Move Python examples to repo root

BUYT-1:


polyntsov (approved):
I think we should create tests from these examples to ensure that examples are always working. However, it's of course a separate issue (#281)

PR #277, opened 2023-10-23
Rename clang-tidy failure step

BUYT-1:


.github/workflows/check-codestyle.yml
```diff
@@ -58,6 +58,6 @@ jobs:
 
     - uses: ZedThree/clang-tidy-review/upload@v0.13.4
 
-    - name: Fail on comments
+    - name: Fail check if clang-tidy comments are present
```
polyntsov, line 61: I think “Fail the check if clang-tidy reported issues” is a more precise name

polyntsov (approved):

PR #279, opened 2023-10-27
Add pybinds for statistic methods

xJoskiy:


src/python_bindings/bindings.cpp
```diff
@@ -156,7 +156,57 @@ PYBIND11_MODULE(desbordante, module) {
...
+    DEFINE_ALGORITHM(DataStats, Algorithm)
+            .def("get_result_string", &PyDataStats::GetResultString)
+            .def("get_number_of_values", &PyDataStats::GetNumberOfValues,
+                 "Get number of values in column.")
+            .def("get_number_of_columns", &PyDataStats::GetNumberOfColumns,
+                 "Get number of columns in table.")
+            .def("get_columns_with_null", &PyDataStats::GetColumnsWithNull,
+                 "Get number columns with null.")
+            .def("get_columns_with_unique_values", &PyDataStats::GetColumnsWithUniqueValues,
+                 "Get number of columns with unique values.")
```
BUYT-1, line 168:
```suggestion
            .def("get_columns_with_all_unique_values", &PyDataStats::GetColumnsWithUniqueValues,
                 "Get indices of columns where all values are distinct.")
```
There has to be a name for this type of column, but I don't know it.

src/python_bindings/bindings.cpp
```diff
@@ -156,7 +156,57 @@ PYBIND11_MODULE(desbordante, module) {
...
+            .def("get_number_of_distinct", &PyDataStats::Distinct,
+                 "Get number of unique values in the column.")
+            .def("is_categorical", &PyDataStats::IsCategorical,
+                 "Check if quantity <= count of unique values in the column.")
```
BUYT-1, line 172:
```cpp
bool DataStats::IsCategorical(size_t index, size_t quantity) {
    return this->Distinct(index) <= quantity;
}
```

```suggestion
                 "Check if quantity >= number of unique values in the column.")
```

src/python_bindings/bindings.cpp
```diff
@@ -156,7 +156,57 @@ PYBIND11_MODULE(desbordante, module) {
...
+            .def("show_sample", &PyDataStats::ShowSample, py::arg("start_row"), py::arg("end_row"),
+                 py::arg("start_col"), py::arg("end_col"), py::arg("str_len") = 10,
+                 py::arg("usigned_len") = 5, py::arg("double_len") = 10,
```
BUYT-1, line 175:
```suggestion
                 py::arg("unsigned_len") = 5, py::arg("double_len") = 10,
```

src/python_bindings/bindings.cpp
```diff
@@ -156,7 +156,57 @@ PYBIND11_MODULE(desbordante, module) {
...
+                 "Returns table slice from start_row to end_row and from start_col to end_col. "
+                 "Data values converted to string type.")
```
BUYT-1, line 177: Are you sure this method is supposed to be exposed to the user? That length specification makes me doubt it.

src/python_bindings/bindings.cpp
```diff
@@ -156,7 +156,57 @@ PYBIND11_MODULE(desbordante, module) {
...
+            .def("get_average", &PyDataStats::GetAverage,
+                 "Returns avarage value in the column if it's numeric.")
```
BUYT-1, line 179:
```suggestion
                 "Returns average value in the column if it's numeric.")
```

src/python_bindings/bindings.cpp
```diff
@@ -156,7 +156,57 @@ PYBIND11_MODULE(desbordante, module) {
...
+            .def("get_corrected_std", &PyDataStats::GetCorrectedSTD,
+                 "Returns corrected standard deviation of the column if it's numeric.")
+            .def("get_skewness", &PyDataStats::GetSkewness,
+                 "Returns skewness of the column if it's numeric.")
+            .def("get_kurtosis", &PyDataStats::GetKurtosis,
+                 "Returns kurtosis of the column if it's numeric.")
+            .def("get_min", &PyDataStats::GetMin,
+                 "Returns minimum (maximumin if order = mo::CompareResult::kGreater) value of the "
+                 "column.")
```
BUYT-1, line 188:
```suggestion
                 "Returns minimum value of the column.")
```
Irrelevant

src/python_bindings/bindings.cpp
```diff
@@ -156,7 +156,57 @@ PYBIND11_MODULE(desbordante, module) {
...
+            .def("get_max", &PyDataStats::GetMax, "Returns maximumin value of the column.")
+            .def("get_sum", &PyDataStats::GetSum,
+                 "Returns sum of the column's values if it's numeric.")
+            .def("get_quantile", &PyDataStats::GetQuantile,
+                 "Returns quantile of the column if it's type is comparable.")
```
BUYT-1, line 193:
```suggestion
                 "Returns quantile of the column if its type is comparable.")
```

src/python_bindings/py_data_stats.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+#include "statistics/data_stats.h"
+
+namespace python_bindings {
+
+namespace py = pybind11;
+namespace mo = model;
+
+size_t PyDataStats::GetNumberOfValues(size_t index) const {
+    return GetAlgorithm<algos::DataStats>(algorithm_).NumberOfValues(index);
```
BUYT-1, line 11: You should make a function template for this and use it everywhere applicable.

src/python_bindings/py_data_stats.cpp
```diff
@@ -0,0 +1,162 @@
...
+            return py::str(mo::Type::GetValue<mo::String>(data));
+        default:
+            assert(false);
+            __builtin_unreachable();
+    }
+}
+
+py::object PyDataStats::GetAverage(size_t index) const {
+    algos::Statistic stat = GetAlgorithm<algos::DataStats>(algorithm_).GetAvg(index);
+    return GetValue(stat);
```
BUYT-1, line 64: You should make a function template for this and use it everywhere applicable.

src/python_bindings/py_data_stats.h
```diff
@@ -1,18 +1,53 @@
 #pragma once
 
+#include "algorithms/algo_factory.h"
```
BUYT-1, line 3: not needed

BUYT-1 (changes requested):

src/python_bindings/bindings.cpp
xJoskiy: It's nothing else but giving a sample to user and length specification just tells how many signs will be used to display value in the cell for each value type. Like if there are long string in table we may shorten them with `str_len = 5`

src/python_bindings/bindings.cpp
xJoskiy: It does consider all possible types of columns

src/python_bindings/py_data_stats.cpp
xJoskiy: Don't quite understand what function templates I should implement

src/python_bindings/py_data_stats.cpp
BUYT-1: Nevermind, it doesn't look good in this case. I was thinking of something like this (doesn't work for all cases):
```cpp
template <typename T, typename... Args>
auto GetDirectStatValue(std::unique_ptr<algos::Algorithm> const& data_stats,
                        T(algos::DataStats::*method)(Args...) const, Args... args) {
    return (GetAlgorithm<algos::DataStats>(data_stats).*method)(args...);
}
```

src/python_bindings/py_data_stats.cpp
BUYT-1: I doubt this one would look good either, don't bother.

BUYT-1 (changes requested):

src/python_bindings/bindings.cpp
BUYT-1: You forgot to do it here

src/python_bindings/bindings.cpp
BUYT-1: I think `get_unique_columns` should work, referring to the `UNIQUE` constraint in databases. Probably not ideal, but whatever.

src/python_bindings/py_data_stats.cpp
```diff
@@ -0,0 +1,166 @@
...
+}
+
+py::object PyDataStats::GetMedianAD(size_t index) const {
+    algos::Statistic stat = GetAlgorithm<algos::DataStats>(algorithm_).GetMedianAD(index);
+    return GetValue(stat);
+}
+
+size_t PyDataStats::GetNumNulls(size_t index) const {
+    algos::Statistic stat = GetAlgorithm<algos::DataStats>(algorithm_).GetNumNulls(index);
+    return (size_t)mo::Type::GetValue<mo::Int>(stat.GetData());
```
BUYT-1, line 163:
```suggestion
    return static_cast<size_t>(mo::Type::GetValue<mo::Int>(stat.GetData()));
```

src/python_bindings/bindings.cpp
```diff
@@ -156,7 +156,57 @@ PYBIND11_MODULE(desbordante, module) {
             .def("get_num_error_rows", &PyFDVerifier::GetNumErrorRows)
             .def("get_highlights", &PyFDVerifier::GetHighlights);
 
-    DEFINE_ALGORITHM(DataStats, Algorithm).def("get_result_string", &PyDataStats::GetResultString);
+    DEFINE_ALGORITHM(DataStats, Algorithm)
+            .def("get_result_string", &PyDataStats::GetResultString)
+            .def("get_number_of_values", &PyDataStats::GetNumberOfValues,
+                 "Get number of values in column.")
```
BUYT-1, line 162:
```suggestion
                 "Get number of values in the column.")
```

src/python_bindings/bindings.cpp
```diff
@@ -156,7 +156,57 @@ PYBIND11_MODULE(desbordante, module) {
...
+            .def("get_number_of_columns", &PyDataStats::GetNumberOfColumns,
+                 "Get number of columns in table.")
```
BUYT-1, line 164:
```suggestion
                 "Get number of columns in the table.")
```

src/python_bindings/bindings.cpp
```diff
@@ -156,7 +156,57 @@ PYBIND11_MODULE(desbordante, module) {
...
+                 "Get indices of columns where all values are distinct.")
+            .def("get_number_of_distinct", &PyDataStats::Distinct,
+                 "Get number of unique values in the column.")
+            .def("is_categorical", &PyDataStats::IsCategorical,
+                 "Check if quantity => number of unique values in the column.")
+            .def("show_sample", &PyDataStats::ShowSample, py::arg("start_row"), py::arg("end_row"),
+                 py::arg("start_col"), py::arg("end_col"), py::arg("str_len") = 10,
+                 py::arg("unsigned_len") = 5, py::arg("double_len") = 10,
+                 "Returns table slice from start_row to end_row and from start_col to end_col. "
+                 "Data values converted to string type.")
```
BUYT-1, line 179:
```suggestion
                 "Returns a table slice containing values from rows in the range [start_row, end_row] and columns in the range [start_col, end_col]. Data values are converted to strings.")
```
Split into lines however is correct

src/python_bindings/bindings.cpp
```diff
@@ -156,7 +156,57 @@ PYBIND11_MODULE(desbordante, module) {
...
+            .def("get_number_of_zeros", &PyDataStats::GetNumberOfZeros,
+                 "Returns number of zeros in the column if it's numeric.")
+            .def("get_number_of_negatives", &PyDataStats::GetNumberOfNegatives,
+                 "Returns number of negative numbers in the column if it's numeric.")
+            .def("get_sum_of_squares", &PyDataStats::GetSumOfSquares,
+                 "Returns sum of numbers' squares in the column if it's numeric.")
+            .def("get_geometric_mean", &PyDataStats::GetGeometricMean,
+                 "Returns geometric mean of numbers in the column if it's numeric.")
+            .def("get_mean_ad", &PyDataStats::GetMeanAD,
+                 "Returns mean absolute deviation if it's numeric.")
```
BUYT-1, line 203:
```suggestion
                 "Returns mean absolute deviation of the column if it's numeric.")
```

src/python_bindings/bindings.cpp
```diff
@@ -156,7 +156,57 @@ PYBIND11_MODULE(desbordante, module) {
...
+            .def("get_median", &PyDataStats::GetMedian,
+                 "Returns median of the column if it's numeric.")
+            .def("get_median_ad", &PyDataStats::GetMedianAD,
+                 "Returns meadian absolute deviation in the column if it's numeric.")
```
BUYT-1, line 207:
```suggestion
                 "Returns meadian absolute deviation of the column if it's numeric.")
```

BUYT-1 (changes requested):

src/python_bindings/bindings.cpp
xJoskiy: Unique in this context refers more to the columns that are unique among all other columns in a table

BUYT-1 (changes requested):
change commit names to:
Add Python bindings for DataStats methods
Fix spelling in \`DataStats::GetQuantile\` comment

BUYT-1 (approved):
LGTM

src/python_bindings/py_data_stats.cpp
```diff
@@ -0,0 +1,166 @@
...
+std::vector<std::vector<std::string>> PyDataStats::ShowSample(size_t start_row, size_t end_row,
+                                                              size_t start_col, size_t end_col,
+                                                              size_t str_len, size_t unsigned_len,
+                                                              size_t double_len) const {
+    const algos::DataStats& stat = GetAlgorithm<algos::DataStats>(algorithm_);
+    return stat.ShowSample(start_row, end_row, start_col, end_col, str_len, unsigned_len,
+                           double_len);
+}
+
+py::object GetValue(algos::Statistic stat) {
```
polyntsov, line 47:
```suggestion
py::object GetValue(algos::Statistic const& stat) {
```
Copying `Statistic` is a rather heavy operation

polyntsov, line 47: Should be placed into anonymous namespace

src/python_bindings/py_data_stats.cpp
```diff
@@ -0,0 +1,166 @@
...
+std::vector<size_t> PyDataStats::GetNullColumns() const {
+    return GetAlgorithm<algos::DataStats>(algorithm_).GetNullColumns();
+}
+
+std::vector<size_t> PyDataStats::GetColumnsWithNull() const {
+    return GetAlgorithm<algos::DataStats>(algorithm_).GetColumnsWithNull();
+}
+
+std::vector<size_t> PyDataStats::GetColumnsWithUniqueValues() const {
+    return dynamic_cast<algos::DataStats&>(*algorithm_).GetColumnsWithUniqueValues();
```
polyntsov, line 27: All `dynamic_cast`s should be replaced with `GetAlgorithm`

src/python_bindings/py_data_stats.h
```diff
@@ -6,13 +6,48 @@
 
 namespace python_bindings {
 
+namespace py = pybind11;
+
 class PyDataStats : public PyAlgorithm<algos::DataStats, PyAlgorithmBase> {
     using PyAlgorithmBase::algorithm_;
 
 public:
     [[nodiscard]] std::string GetResultString() const {
```
polyntsov, line 15: I'd rename this to something like `GetAllStatisticsAsString`, `ResultString` sounds to vague

src/python_bindings/py_data_stats.h
```diff
@@ -6,13 +6,48 @@
 
 namespace python_bindings {
 
+namespace py = pybind11;
+
```
polyntsov, line 10: Don't introduce such “general” aliases to header files

src/python_bindings/py_data_stats.cpp
polyntsov: Actually, all these casts look a bit unreasonable since we know the specific algorithm type in this context. Seems like `algorithm_` should be moved to `PyAlgorithm`, however, this is for a different PR

polyntsov (changes requested):

src/python_bindings/py_data_stats.cpp
xJoskiy: Since these stat methods can't be called via a const object I can't use `GetAlgoritm` there

BUYT-1: Changing `T const&` to `T&` in `GetAlgorithm` should work.

src/python_bindings/py_data_stats.cpp
```diff
@@ -0,0 +1,168 @@
...
+bool PyDataStats::IsCategorical(size_t index, size_t quantity) {
+    return dynamic_cast<algos::DataStats&>(*algorithm_).IsCategorical(index, quantity);
+}
+
+std::vector<std::vector<std::string>> PyDataStats::ShowSample(size_t start_row, size_t end_row,
+                                                              size_t start_col, size_t end_col,
+                                                              size_t str_len, size_t unsigned_len,
+                                                              size_t double_len) const {
+    const algos::DataStats& stat = GetAlgorithm<algos::DataStats>(algorithm_);
+    return stat.ShowSample(start_row, end_row, start_col, end_col, str_len, unsigned_len,
```
BUYT-1, line 64:
```suggestion
    return GetAlgorithm<algos::DataStats>(algorithm_).ShowSample(
            start_row, end_row, start_col, end_col, str_len, unsigned_len,
```
consistency

BUYT-1 (changes requested):

src/python_bindings/py_data_stats.cpp
```diff
@@ -0,0 +1,167 @@
...
+}
+
+py::object PyDataStats::GetSum(size_t index) const {
+    algos::Statistic stat = GetAlgorithm<algos::DataStats>(algorithm_).GetSum(index);
+    return GetValue(stat);
+}
+
+py::object PyDataStats::GetQuantile(double part, size_t index, bool calc_all) {
+    algos::Statistic stat =
+            dynamic_cast<algos::DataStats&>(*algorithm_).GetQuantile(part, index, calc_all);
```
BUYT-1, line 123:
```suggestion
            GetAlgorithm<algos::DataStats>(algorithm_).GetQuantile(part, index, calc_all);
```

BUYT-1 (changes requested):
missed one

BUYT-1 (approved):

src/python_bindings/py_data_stats.cpp
```diff
@@ -0,0 +1,167 @@
...
+}
+
+py::object PyDataStats::GetMedianAD(size_t index) const {
+    algos::Statistic stat = GetAlgorithm<algos::DataStats>(algorithm_).GetMedianAD(index);
+    return GetValue(stat);
+}
+
+size_t PyDataStats::GetNumNulls(size_t index) const {
+    algos::Statistic stat = GetAlgorithm<algos::DataStats>(algorithm_).GetNumNulls(index);
+    return static_cast<size_t>(mo::Type::GetValue<mo::Int>(stat.GetData()));
```
polyntsov, line 164: Hm, I'm wondering why `GetNumNulls` returns a `Statistic` instead of a plain integer

src/python_bindings/bindings.cpp
```diff
@@ -160,7 +160,66 @@ PYBIND11_MODULE(desbordante, module) {
...
+            .def("get_null_columns", &PyDataStats::GetNullColumns,
+                 "Get indices of columns with only null values.")
+            .def("get_columns_with_null", &PyDataStats::GetColumnsWithNull,
+                 "Get indices of columns which contain null value.")
+            .def("get_columns_with_all_unique_values", &PyDataStats::GetColumnsWithUniqueValues,
+                 "Get indices of columns where all values are distinct.")
+            .def("get_number_of_distinct", &PyDataStats::Distinct,
+                 "Get number of unique values in the column.", py::arg("index"))
+            .def("is_categorical", &PyDataStats::IsCategorical,
+                 "Check if quantity => number of unique values in the column.", py::arg("index"),
```
polyntsov, line 178: what does "quantity => number" means? Looks like a typo

BUYT-1:
```suggestion
                 "Check if quantity >= number of unique values in the column.", py::arg("index"),
```

polyntsov: I think "greater than" is better than ">="

src/python_bindings/bindings.cpp
```diff
@@ -160,7 +160,66 @@ PYBIND11_MODULE(desbordante, module) {
...
+                 py::arg("quantity"))
+            .def("show_sample", &PyDataStats::ShowSample, py::arg("start_row"), py::arg("end_row"),
+                 py::arg("start_col"), py::arg("end_col"), py::arg("str_len") = 10,
+                 py::arg("unsigned_len") = 5, py::arg("double_len") = 10,
```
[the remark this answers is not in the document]
xJoskiy: So, it worth getting rid of the trimming?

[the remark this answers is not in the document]
polyntsov: We should fix the existing method

myeasylog.log
polyntsov, line 1: Should not be commited

polyntsov (changes requested):
Please, move adding example for statistics and tests into separate PR and I'll merge

polyntsov (approved):

PR #284, opened 2023-11-09
Fix max_lhs option

AndreyChernyshev555:
Declaration of the option was moved to FDAlgorithm class due to both tane and fastfds use it. The test was added.

### DIFF
--- a/src/core/algorithms/fd/fd_algorithm.cpp
+++ b/src/core/algorithms/fd/fd_algorithm.cpp
@@ -1,6 +1,8 @@
 #include "fd_algorithm.h"
 
+#include <map>
 #include <thread>
+#include <vector>
 
 #include "config/equal_nulls/option.h"
 #include "config/tabular_data/input_table/option.h"


src/tests/test_max_lhs_option.cpp
```diff
@@ -0,0 +1,60 @@
...
+                {kHasHeader, hasHeader}, {kTable, GetInputTable(path, separator, hasHeader)},
+                {kMaximumLhs, max_lhs},  {kEqualNulls, EqNullsType}};
+
+        return algos::CreateAndLoadAlgorithm<Algo>(params);
+    }
+};
+
+TEST_F(MaxLHSOptionTest, TaneMaxLHSOptionWork) {
+    unsigned int max_lhs = 3;
+    auto const path = test_data_dir / "TestFD.csv";
```
BUYT-1, line 42: HyFD reports these FDs on this dataset: `[( ) -> 0, ( 5 ) -> 2, ( 5 ) -> 4, ( 5 ) -> 1, ( 5 ) -> 3, ( 2 ) -> 1, ( 2 4 ) -> 5, ( 2 4 ) -> 3, ( 2 3 ) -> 5, ( 2 3 ) -> 4, ( 1 4 ) -> 3, ( 3 4 ) -> 1, ( 1 3 ) -> 4]`. None of them have an LHS longer than 3, so the option does nothing. Either lower the tested value or choose another dataset for these tests.
Also, the 2 tests for Tane and FastFDs look exactly the same. Try `TYPED_TEST_P`, like in `test_algorithm.h`

BUYT-1 (changes requested):

PR #285, opened 2023-11-11
Change destruction of typed_column_data and mixed

egshnov:
Added generic destruct method Type::Destruct<T>.
Added method Destruct(TypeId, std::byte* ) to typed_column_data and mixed_type
(maybe putting this method to a separate header file to avoid code duplication is a better idea).


src/core/model/table/typed_column_data.h
```diff
@@ -40,6 +40,15 @@ class TypedColumnData : public model::AbstractColumnData {
...
     friend class TypedColumnDataFactory;
 
+    void Destruct(TypeId type_id, std::byte const* value) {
+        if (type_id == +TypeId::kString || type_id == +TypeId::kBigInt) {
+            Type::Destruct<StringType>(value);
+        }
+        if (type_id == +TypeId::kDate) {
+            Type::Destruct<DateType>(value);
+        }
+    }
```
polyntsov, line 50: It's the same as in `MixedType`, why implementing it twice?
I think it should be made `static` and moved to `Type` instead of unneeded templated version of `Destruct` and used everywhere

egshnov: Seems like adding such method to 'Type' won't work as we can't use functionality of a child class in the base class. I'll try to move destruct to a separate header.

BUYT-1:
To reviewer: #330 also implements a destructor for types

src/core/model/table/typed_column_data.h
Mstrutov: Separate header seems OK for now

src/core/model/types/destruct.h
```diff
@@ -0,0 +1,15 @@
...
+namespace model {
+inline void Destruct(TypeId type_id, std::byte const* value) {
+    if (type_id == +TypeId::kString || type_id == +TypeId::kBigInt) {
+        StringType::Destruct(value);
+    }
+    if (type_id == +TypeId::kDate) {
+        DateType::Destruct(value);
+    }
+}
+}  // namespace model
```
polyntsov, line 15: New line symbol

PR #286, opened 2023-11-15
Remove `const` in GetAlgorithm return type

BUYT-1:


src/python_bindings/get_algorithm.h
```diff
@@ -7,8 +7,8 @@
 namespace python_bindings {
 
 template <typename T>
-[[nodiscard]] T const& GetAlgorithm(std::unique_ptr<algos::Algorithm> const& ptr) {
-    return dynamic_cast<T const&>(*ptr);
+[[nodiscard]] decltype(auto) GetAlgorithm(std::unique_ptr<algos::Algorithm> const& ptr) {
```
polyntsov, line 10: Why are we using `decltype(auto)` here? It seems like it doesn't make the code clearer/safe, but rather harder to read and to understand

BUYT-1: I don't think it makes a difference either way. I just didn't want to write `T&` again.

polyntsov (approved):

PR #290, opened 2023-11-16
Add UUCVerifier with python bindings 

aartdem:
Completed PR #216
Fix naming, adapt to the current version of the project, add new tests, add default implementation for `IndicesOption` in `UCCVerifier`

### DIFF
--- a/src/core/algorithms/ucc/ucc_verifier/ucc_verifier.cpp
+++ b/src/core/algorithms/ucc/ucc_verifier/ucc_verifier.cpp
@@ -1,0 +1,68 @@
+#include "ucc_verifier.h"
+
+#include <chrono>
+#include <stdexcept>
+
+#include "config/equal_nulls/option.h"
+#include "config/indices/option.h"
+#include "config/names_and_descriptions.h"
+#include "config/option_using.h"
+#include "config/tabular_data/input_table/option.h"
+
+namespace algos {
+
+UCCVerifier::UCCVerifier() : Algorithm({}) {
+    RegisterOptions();
+    MakeOptionsAvailable({config::TableOpt.GetName(), config::EqualNullsOpt.GetName()});
+}
+
+void UCCVerifier::RegisterOptions() {
+    DESBORDANTE_OPTION_USING;
+    auto get_schema_cols = [this]() { return relation_->GetSchema()->GetNumColumns(); };
+    RegisterOption(config::TableOpt(&input_table_));
+    RegisterOption(config::EqualNullsOpt(&is_null_equal_null_));
+    RegisterOption(
+            config::IndicesOption{kUCCIndices, kDUCCIndices}(&column_indices_, get_schema_cols));
+}
+
+void UCCVerifier::MakeExecuteOptsAvailable() {
+    using namespace config::names;
+    MakeOptionsAvailable({kUCCIndices});
+}
+
+void UCCVerifier::LoadDataInternal() {
+    relation_ = ColumnLayoutRelationData::CreateFrom(*input_table_, is_null_equal_null_);
+
+    if (relation_->GetColumnData().empty()) {
+        throw std::runtime_error("Got an empty dataset: UCC verifying is meaningless.");
+    }
+}
+
+unsigned long long UCCVerifier::ExecuteInternal() {
+    auto start_time = std::chrono::system_clock::now();
+    VerifyUCC();
+    auto elapsed_milliseconds = std::chrono::duration_cast<std::chrono::milliseconds>(
+            std::chrono::system_clock::now() - start_time);
+    return elapsed_milliseconds.count();
+}
+
+void UCCVerifier::VerifyUCC() {
+    std::shared_ptr<model::PLI const> pli =
+            relation_->GetColumnData(column_indices_[0]).GetPliOwnership();
+    for (size_t i = 1; i < column_indices_.size(); ++i) {
+        pli = pli->Intersect(relation_->GetColumnData(column_indices_[i]).GetPositionListIndex());
+        if (pli->GetNumCluster() == 0) {
+            return;
+        }
+    }
+    CalculateStatistics(pli->GetIndex());
+}
+
+void UCCVerifier::CalculateStatistics(std::deque<model::PLI::Cluster> const& clusters) {
+    for (auto const& cluster : clusters) {
+        num_rows_violating_ucc_ += cluster.size();
+        clusters_violating_ucc_.push_back(cluster);
+    }
+}
+
+}  // namespace algos
--- a/src/core/algorithms/ucc/ucc_verifier/ucc_verifier.cpp
+++ b/src/core/algorithms/ucc/ucc_verifier/ucc_verifier.cpp
@@ -1,6 +1,7 @@
 #include "ucc_verifier.h"
 
 #include <chrono>
+#include <numeric>
 #include <stdexcept>
 
 #include "config/equal_nulls/option.h"
@@ -19,10 +20,15 @@ UCCVerifier::UCCVerifier() : Algorithm({}) {
 void UCCVerifier::RegisterOptions() {
     DESBORDANTE_OPTION_USING;
     auto get_schema_cols = [this]() { return relation_->GetSchema()->GetNumColumns(); };
+    auto calculate_default = [get_schema_cols]() {
+        config::IndicesType indices(get_schema_cols());
+        std::iota(indices.begin(), indices.end(), 0);
+        return indices;
+    };
     RegisterOption(config::TableOpt(&input_table_));
     RegisterOption(config::EqualNullsOpt(&is_null_equal_null_));
-    RegisterOption(
-            config::IndicesOption{kUCCIndices, kDUCCIndices}(&column_indices_, get_schema_cols));
+    RegisterOption(config::IndicesOption{kUCCIndices, kDUCCIndices, std::move(calculate_default)}(
+            &column_indices_, std::move(get_schema_cols)));
 }
 
 void UCCVerifier::MakeExecuteOptsAvailable() {
--- a/src/core/algorithms/ucc/ucc_verifier/ucc_verifier.h
+++ b/src/core/algorithms/ucc/ucc_verifier/ucc_verifier.h
@@ -1,0 +1,65 @@
+#pragma once
+
+#include <vector>
+
+#include "algorithms/algorithm.h"
+#include "config/equal_nulls/type.h"
+#include "config/indices/type.h"
+#include "config/tabular_data/input_table_type.h"
+#include "model/table/column_layout_relation_data.h"
+
+namespace algos {
+
+/* Algorithm used to verify that a set of columns is UCC and retrieving useful information in
+ * case it is not */
+class UCCVerifier : public Algorithm {
+private:
+    /* input options */
+    config::IndicesType column_indices_;
+    config::EqNullsType is_null_equal_null_{};
+
+    config::InputTable input_table_;
+    std::unique_ptr<ColumnLayoutRelationData> relation_;
+
+    /* results of work */
+    size_t num_rows_violating_ucc_ = 0;
+    std::vector<model::PLI::Cluster> clusters_violating_ucc_;
+
+    void VerifyUCC();
+    void CalculateStatistics(std::deque<model::PLI::Cluster> const& clusters);
+    void RegisterOptions();
+    void LoadDataInternal() override;
+    void MakeExecuteOptsAvailable() override;
+    unsigned long long ExecuteInternal() override;
+    void ResetState() override {
+        clusters_violating_ucc_.clear();
+        num_rows_violating_ucc_ = 0;
+    }
+
+public:
+    /* Returns true if UCC holds and false otherwise */
+    bool UCCHolds() const {
+        return clusters_violating_ucc_.empty();
+    }
+
+    /* Returns the number of clusters where the UCC is violated, that is, the number of sets of rows
+     * where each set consists of rows equal to each other in the specified columns */
+    size_t GetNumClustersViolatingUCC() const {
+        return clusters_violating_ucc_.size();
+    }
+
+    /* Returns the total number of table rows that violate the UCC */
+    size_t GetNumRowsViolatingUCC() const {
+        return num_rows_violating_ucc_;
+    }
+
+    /* Returns clusters where the UCC is violated, that is, sets of rows where each set consists of
+     * rows equal to each other in the specified columns */
+    std::vector<model::PLI::Cluster> const& GetClustersViolatingUCC() const {
+        return clusters_violating_ucc_;
+    }
+
+    UCCVerifier();
+};
+
+}  // namespace algos
--- a/src/core/config/common_option.h
+++ b/src/core/config/common_option.h
@@ -2,6 +2,9 @@
 
 #include <optional>
 #include <string_view>
+#include <variant>
+
+#include <boost/functional/overloaded_function.hpp>
 
 #include "config/option.h"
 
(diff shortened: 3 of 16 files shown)


src/tests/test_ucc_verifier.cpp
```diff
@@ -0,0 +1,76 @@
...
+    const std::vector<model::PLI::Cluster>& GetExpectedClustersNotUCC() const {
+        return clusters_not_ucc_;
+    }
+};
+
+class TestUCCVerifying : public ::testing::TestWithParam<UCCVerifyingParams> {};
+
+TEST_P(TestUCCVerifying, DefaultTest) {
+    auto const& p(GetParam());
+    auto mp = algos::StdParamsMap(p.GetParams());
```
polyntsov, line 53: Strange cast, `p.GetParams()` already has `algos::StdParamsMap` type, there is no need to cast it. Moreover, this `mp` variable is not needed at all, since it is only used in the line below, `p.GetParams()` could be passed there right away

src/tests/test_ucc_verifier.cpp
```diff
@@ -0,0 +1,76 @@
...
+                          UCCVerifyingParams({0, 1}, 4, 12,
+                                             {{0, 1, 2}, {3, 4, 5}, {6, 7, 8}, {9, 10, 11}},
+                                             "TestFD.csv"),
+                          UCCVerifyingParams({0, 1, 2}, 4, 8, {{0, 1}, {3, 4}, {6, 7}, {9, 10}},
+                                             "TestFD.csv"),
+                          UCCVerifyingParams({0, 1, 2, 3, 4, 5}, 3, 6, {{3, 4}, {6, 7}, {9, 10}},
+                                             "TestFD.csv"),
+                          UCCVerifyingParams({0}, 0, 0, {}, "TestWide.csv"),
+                          UCCVerifyingParams({0, 1, 2, 3, 4}, 0, 0, {}, "TestWide.csv")));
+
```
polyntsov, line 75: It would be good to implement a test on big tables using HyUCC to find uccs

src/tests/test_ucc_verifier.cpp
```diff
@@ -0,0 +1,76 @@
...
+    size_t GetExpectedNumClustersNotUCC() const {
+        return num_clusters_not_ucc_;
+    }
+
+    size_t GetExpectedNumRowsNotUCC() const {
+        return num_rows_not_ucc_;
+    }
+
+    const std::vector<model::PLI::Cluster>& GetExpectedClustersNotUCC() const {
```
polyntsov, line 44:
```suggestion
    std::vector<model::PLI::Cluster> const& GetExpectedClustersNotUCC() const {
```

polyntsov (changes requested):

src/core/algorithms/ucc/ucc_verifier/ucc_verifier.h
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /* Returns the total number of table rows that violate the UCC */
+    size_t GetNumRowsNotUCC() const {
+        return num_rows_not_ucc_;
+    }
+
+    /* Returns clusters where UCC is violated, that is, in which rows equal in the
+     * specified columns */
+    std::vector<model::PLI::Cluster> const& GetClustersNotUCC() const {
```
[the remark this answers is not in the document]
aartdem: Question about method description: maybe "... in the specified columns" (instead of  "... in the UCC columns) is better because specified columns (i. e. input columns) does not hold UCC in case when returned vector is not empty. So ".. the UCC columns" is a little confusing

src/tests/test_ucc_verifier.cpp
aartdem: We need `mp` variable because `CreateAndLoadAlgorithm` accepts a rvalue-reference and `p.GetParams` returns const l-value reference

src/core/algorithms/ucc/ucc_verifier/ucc_verifier.h
```diff
@@ -0,0 +1,65 @@
...
+/* Algorithm used for verify that a set of columns is UCC and retrieving useful information in
+ * case it is not */
+class UCCVerifier : public Algorithm {
+private:
+    /* input options */
+    config::IndicesType column_indices_;
+    config::EqNullsType is_null_equal_null_{};
+
+    config::InputTable input_table_;
+    std::shared_ptr<ColumnLayoutRelationData> relation_;
```
polyntsov, line 22:
```suggestion
    std::unique_ptr<ColumnLayoutRelationData> relation_;
```
There is no reason to use `shared_ptr` here

src/core/algorithms/ucc/ucc_verifier/ucc_verifier.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+UCCVerifier::UCCVerifier() : Algorithm({}) {
+    RegisterOptions();
+    MakeOptionsAvailable({config::TableOpt.GetName(), config::EqualNullsOpt.GetName()});
+}
+
+void UCCVerifier::RegisterOptions() {
+    DESBORDANTE_OPTION_USING;
+    auto get_schema_cols = [this]() { return relation_->GetSchema()->GetNumColumns(); };
+    std::function<config::IndicesType()> calculate_default = [get_schema_cols]() {
```
polyntsov, line 23: nit
```suggestion
    auto calculate_default = [get_schema_cols]() {
```

src/core/config/common_option.h
```diff
@@ -17,22 +17,25 @@ class CommonOption {
...
     typename Option<T>::ValueCheckFunc const value_check_func_;
+    typename Option<T>::DefaultFunc const default_func_;
 
 public:
     CommonOption(std::string_view name, std::string_view description,
                  std::optional<T> default_value = std::nullopt,
                  typename Option<T>::NormalizeFunc normalize_func = nullptr,
-                 typename Option<T>::ValueCheckFunc value_check_func = nullptr)
+                 typename Option<T>::ValueCheckFunc value_check_func = nullptr,
+                 typename Option<T>::DefaultFunc default_func_ = nullptr)
```
polyntsov, line 27: I don't think it should be possible to pass to the constructor simultaneously default value and default function, because it's not obvious which takes preference when both are specified. Let's better add a separate constructor with `default_func` parameter, but without `default_value`, as in `Option`

src/core/config/descriptions.h
```diff
@@ -33,6 +33,7 @@ const auto kDMetric = _kDMetric.c_str();
 constexpr auto kDLhsIndices = "LHS column indices";
 constexpr auto kDRhsIndices = "RHS column indices";
 constexpr auto kDRhsIndex = "RHS column index";
+constexpr auto kDUCCIndices = "column indices for verify UCC";
```
polyntsov, line 36:
```suggestion
constexpr auto kDUCCIndices = "column indices to verify for UCC";
```
or
```suggestion
constexpr auto kDUCCIndices = "column indices for UCC verification";
```

src/tests/test_ucc_verifier.cpp
```diff
@@ -0,0 +1,153 @@
...
+          num_rows_violating_ucc_(num_rows_violating_ucc),
+          clusters_violating_ucc_(std::move(clusters_violating_ucc)) {
+        // if column_indices is empty then it is not inserted into params_map_ (and this option is
+        // not passed to the algorithm), in this case default implementation will be used
+        if (!column_indices.empty()) {
+            params_map_.insert({onam::kUCCIndices, std::move(column_indices)});
+        }
+    }
+
+    algos::StdParamsMap GetParamsMap() const {
```
polyntsov, line 39:
```suggestion
    algos::StdParamsMap const& GetParamsMap() const {
```

src/tests/test_ucc_verifier.cpp
```diff
@@ -0,0 +1,153 @@
...
+                                    {onam::kSeparator, separator},
+                                    {onam::kHasHeader, has_header},
+                                    {onam::kEqualNulls, true}}),
+          hyucc_params_map_({{onam::kThreads, static_cast<config::ThreadNumType>(1)},
+                             {onam::kCsvPath, test_data_dir / dataset},
+                             {onam::kSeparator, separator},
+                             {onam::kHasHeader, has_header},
+                             {onam::kEqualNulls, true}}) {}
+
+    algos::StdParamsMap GetUCCVerifierParamsMap() const {
```
polyntsov, line 76:
```suggestion
    algos::StdParamsMap const& GetUCCVerifierParamsMap() const {
```

src/tests/test_ucc_verifier.cpp
```diff
@@ -0,0 +1,153 @@
...
+        return ucc_verifier_params_map_;
+    }
+
+    algos::StdParamsMap GetHyUCCParamsMap() const {
```
polyntsov, line 80:
```suggestion
    algos::StdParamsMap const& GetHyUCCParamsMap() const {
```

src/tests/test_ucc_verifier.cpp
```diff
@@ -0,0 +1,153 @@
...
+                       {onam::kSeparator, separator},
+                       {onam::kHasHeader, has_header},
+                       {onam::kEqualNulls, true}}),
+          num_clusters_violating_ucc_(num_clusters_violating_ucc),
+          num_rows_violating_ucc_(num_rows_violating_ucc),
+          clusters_violating_ucc_(std::move(clusters_violating_ucc)) {
+        // if column_indices is empty then it is not inserted into params_map_ (and this option is
+        // not passed to the algorithm), in this case default implementation will be used
+        if (!column_indices.empty()) {
+            params_map_.insert({onam::kUCCIndices, std::move(column_indices)});
```
polyntsov, line 35:
```suggestion
            params_map_.emplace(onam::kUCCIndices, std::move(column_indices));
```

src/tests/test_ucc_verifier.cpp
```diff
@@ -0,0 +1,153 @@
...
+    UCCVerifierWithHyUCCParams const& p(GetParam());
+
+    auto hyucc = algos::CreateAndLoadAlgorithm<algos::HyUCC>(p.GetHyUCCParamsMap());
+    hyucc->Execute();
+    std::list<model::UCC> const& mined_uccs = hyucc->UCCList();
+
+    // run ucc_verifier on each UCC from mined_ucc (UCC must hold)
+    for (auto const& current_ucc : mined_uccs) {
+        algos::StdParamsMap ucc_verifier_params_map = p.GetUCCVerifierParamsMap();
+        ucc_verifier_params_map.insert({onam::kUCCIndices, current_ucc.GetColumnIndicesAsVector()});
```
polyntsov, line 124:
```suggestion
        ucc_verifier_params_map.emplace(onam::kUCCIndices, current_ucc.GetColumnIndicesAsVector());
```

src/tests/test_ucc_verifier.cpp
```diff
@@ -0,0 +1,153 @@
...
+    // Сases of prevent false negative triggering
+    // run ucc_verifier on each UCC from mined_ucc with one column index removed
+    // (UCC must not hold because HyUCC returns minimal UCCs)
+    for (auto const& current_ucc : mined_uccs) {
+        algos::StdParamsMap ucc_verifier_params_map = p.GetUCCVerifierParamsMap();
+        std::vector<unsigned int> current_ucc_vec = current_ucc.GetColumnIndicesAsVector();
+        if (current_ucc_vec.size() < 2) {
+            continue;
+        }
+        current_ucc_vec.erase(--current_ucc_vec.end());
```
polyntsov, line 140:
```suggestion
        current_ucc_vec.pop_back();
```

src/tests/test_ucc_verifier.cpp
```diff
@@ -0,0 +1,153 @@
...
+        ucc_verifier_params_map.insert({onam::kUCCIndices, std::move(current_ucc_vec)});
+        auto verifier = algos::CreateAndLoadAlgorithm<algos::UCCVerifier>(
+                std::move(ucc_verifier_params_map));
+        verifier->Execute();
```
polyntsov, line 144: These lines are the same as in the loop above, that's because they're doing logically the same thing: creating `verifier` object with specified ucc candidate and executing the algorithm. It should be moved out into a separate function and reused here

src/tests/test_ucc_verifier.cpp
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    std::vector<model::PLI::Cluster> const& GetExpectedClustersViolatingUCC() const {
+        return clusters_violating_ucc_;
+    }
+};
+
+class TestUCCVerifierSimple : public ::testing::TestWithParam<UCCVerifierSimpleParams> {};
+
+class UCCVerifierWithHyUCCParams {
```
polyntsov, line 58: I'd move this class below the test for `TestUCCVerifierSimple` fixture, I think it will be easier to read that way, because fixture definition will be close to its usage

src/tests/test_ucc_verifier.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+#include "algorithms/algo_factory.h"
+#include "algorithms/ucc/ucc_verifier/ucc_verifier.h"
+#include "config/indices/type.h"
+#include "datasets.h"
+
+namespace tests {
+namespace onam = config::names;
+
+class UCCVerifierSimpleParams {
```
polyntsov, line 11: Move all classes definitions which are required only in this file to the anonymous namespace

src/tests/test_ucc_verifier.cpp
```diff
@@ -0,0 +1,153 @@
...
+                            bool const has_header = true)
+        : params_map_({{onam::kCsvPath, test_data_dir / dataset},
+                       {onam::kSeparator, separator},
+                       {onam::kHasHeader, has_header},
+                       {onam::kEqualNulls, true}}),
+          num_clusters_violating_ucc_(num_clusters_violating_ucc),
+          num_rows_violating_ucc_(num_rows_violating_ucc),
+          clusters_violating_ucc_(std::move(clusters_violating_ucc)) {
+        // if column_indices is empty then it is not inserted into params_map_ (and this option is
+        // not passed to the algorithm), in this case default implementation will be used
```
polyntsov, line 33:
```suggestion
        // if column_indices is empty then it is not inserted into params_map_, and `ucc_indices` option is
        // not passed to the algorithm. Instead we assume that the option will be initialized with the default value (set of all table columns)
```

src/tests/test_ucc_verifier.cpp
```diff
@@ -0,0 +1,153 @@
...
+        ucc_verifier_params_map.insert({onam::kUCCIndices, current_ucc.GetColumnIndicesAsVector()});
+        auto verifier = algos::CreateAndLoadAlgorithm<algos::UCCVerifier>(
+                std::move(ucc_verifier_params_map));
+        verifier->Execute();
+        EXPECT_TRUE(verifier->UCCHolds());
+    }
+
+    // Сases of prevent false negative triggering
+    // run ucc_verifier on each UCC from mined_ucc with one column index removed
+    // (UCC must not hold because HyUCC returns minimal UCCs)
```
polyntsov, line 133: It doesn't check for false negatives, but for false positives, because the code below this comment verifies that `verifier` never returns `true` for non-ucc
```suggestion
    // Сase to prevent false positives
    // run ucc_verifier on each UCC from mined_ucc with one column index removed
    // (UCC must not hold because HyUCC returns minimal UCCs)
```

polyntsov (changes requested):

src/tests/test_ucc_verifier.cpp
aartdem: If we do this we will not able to pass this method return value into `CreateAndLoadAlgorithm(OptionMap&& options)` (because this function need r-value reference) directly, we need to create an additional variable. In the first review, it was proposed to [avoid this](https://github.com/Mstrutov/Desbordante/pull/290#discussion_r1404596051). As a result, add the `const` modifier?

src/tests/test_ucc_verifier.cpp
aartdem: [same](https://github.com/Mstrutov/Desbordante/pull/290#discussion_r1416305731)

src/tests/test_ucc_verifier.cpp
polyntsov:
```cpp
template <typename T, typename OptionMap>
CreateAndLoadAlgorithm(OptionMap&& options)
```
`option` here is not passed by rvalue reference, but by ["universal" reference](https://isocpp.org/blog/2012/11/universal-references-in-c11-scott-meyers), so passing lvalues to it works perfectly fine.
Moreover, there is almost no difference from the caller point of view between `T const&` and `T` return types, in almost any context where `T` works `T const&` will also work

aartdem: Thanks for the article, the proposed version is implemented in the branch

src/core/config/indices/option.h
```diff
@@ -9,7 +9,8 @@ namespace config {
 
 // This class is meant for creating options that are collections of indices.
 struct IndicesOption {
-    IndicesOption(std::string_view name, std::string_view description);
+    IndicesOption(std::string_view name, std::string_view description,
+                  typename Option<config::IndicesType>::DefaultFunc calculate_default = nullptr);
```
polyntsov, line 13: By the way, why are we adding `calculate_default` to constructor and not to `operator()` as it is done for `value_check_func`? Using the same `IndicesOption` object to create options with different `calculate_default` functions seems like a useful thing to do

polyntsov (changes requested):

src/core/config/common_option.h
```diff
@@ -17,22 +17,35 @@ class CommonOption {
     std::optional<T> const default_value_;
     typename Option<T>::NormalizeFunc const normalize_func_;
     typename Option<T>::ValueCheckFunc const value_check_func_;
+    typename Option<T>::DefaultFunc const default_func_;
 
 public:
     CommonOption(std::string_view name, std::string_view description,
-                 std::optional<T> default_value = std::nullopt,
+                 std::optional<T> default_value,
```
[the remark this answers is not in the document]
aartdem: Another constructor will be used for this

src/core/config/indices/option.h
aartdem: We can add `calculate_default` to `operator()`, I will implement this way in separate commit. By the way, in this way we don't need two constructors

aartdem: advantages: we need only one constructor and don't need "strange cast"

src/core/config/thread_number/option.cpp
```diff
@@ -8,7 +8,7 @@
 namespace config {
 using names::kThreads, descriptions::kDThreads;
 extern const CommonOption<ThreadNumType> ThreadNumberOpt{
-        kThreads, kDThreads, 0, [](auto &value) {
+        kThreads, kDThreads, static_cast<int>(0), [](auto &value) {
```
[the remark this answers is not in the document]
polyntsov: Type of `0` is exactly `int`, so `static_cast<int>(0)` does absolutely nothing, I doubt that it can anyhow influence overload resolution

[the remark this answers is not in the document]
BUYT-1: `+0` should work

polyntsov (approved):

PR #295, opened 2023-11-22
Add option `Tables`

vs9h:
* Apply changes from `clang-format` in `src/test/` directory.
* Small code fixes (make non-static methods static, change naming, etc.)
* Add option `Tables` for multipath algorithms.

### DIFF
--- a/src/core/algorithms/algo_factory.h
+++ b/src/core/algorithms/algo_factory.h
@@ -11,6 +11,7 @@
 #include "algorithms/create_algorithm.h"
 #include "algorithms/pipelines/typo_miner/typo_miner.h"
 #include "config/names.h"
+#include "tabular_data/input_tables_type.h"
 
 namespace algos {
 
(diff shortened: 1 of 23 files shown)


src/core/config/all_options.cpp
```diff
@@ -42,6 +42,7 @@ boost::program_options::options_description GeneralOptions() {
     po::options_description general_options("General options");
     general_options.add_options()
             (kCsvPath, po::value<std::filesystem::path>()->required(), kDCsvPath)
+            (kCsvPaths, po::value<std::vector<std::filesystem::path>>()->required(), kDCsvPaths)
```
BUYT-1, line 45: `->required` should be removed

src/core/config/tabular_data/input_tables/option.h
```diff
@@ -0,0 +1,11 @@
+// This option is meant for the case where an algorithm accepts several
+// tables.
```
BUYT-1, line 2: I added the comment to input_table/option.h so it wouldn't accidentally be used if an algorithm uses 2 tables, for example. But I don't see how this option could be confusing, so this comment isn't needed.

BUYT-1 (changes requested):

src/core/config/all_options.cpp
vs9h: Also it seems, that we should remove `required()` for `kCsvPath`, isn't it?

vs9h: Also it seems, that we should remove `required()` for `kCsvPath`, isn't it?

src/core/config/tabular_data/input_tables/option.h
vs9h: Ack

src/core/config/all_options.cpp
BUYT-1: yes, that's why I highlighted both lines

BUYT-1 (approved):
Looks okay

src/core/algorithms/algo_factory.h
```diff
@@ -71,12 +72,28 @@ template <typename OptionMap>
...
         using namespace config::names;
+        namespace fs = std::filesystem;
         if (option_name == kTable && options.find(std::string{kTable}) == options.end()) {
             config::InputTable parser = std::make_shared<CSVParser>(
-                    details::ExtractOptionValue<std::filesystem::path>(options, kCsvPath),
+                    details::ExtractOptionValue<fs::path>(options, kCsvPath),
                     details::ExtractOptionValue<char>(options, kSeparator),
                     details::ExtractOptionValue<bool>(options, kHasHeader));
             return boost::any{parser};
+        } else if (option_name == kTables && options.find(std::string{kTables}) == options.end()) {
```
polyntsov, line 82: This case handling looks bad, we certainly don't want here a big if-else spaghetti handling all cases where input data option may be constructed from several others. Let's leave it as it is for now, and I'll create an issue

polyntsov (approved):

PR #296, opened 2023-11-23
Add INDAlgorithm infrastructure

vs9h:
Add infrastructure for inclusion dependency mining algorithms.

Motivation: this commit was taken from #272 and modified in order to merge the common infrastructure for inclusion dependency mining algorithms separately from the `faida` algorithm (since this is also necessary for my algorithm).


src/core/algorithms/ind/ind_algorithm.h
```diff
@@ -0,0 +1,50 @@
...
+    virtual void RegisterIND(std::shared_ptr<IND::ColumnCombination> lhs,
+                             std::shared_ptr<IND::ColumnCombination> rhs) {
+        ind_collection_.Register(std::move(lhs), std::move(rhs));
+    }
+    virtual void RegisterIND(IND ind) {
+        ind_collection_.Register(std::move(ind));
+    }
+
+public:
+    std::list<IND> const& IndList() const noexcept {
```
polyntsov, line 45: `INDList` here or `Ind` in the names of all other methods

src/core/algorithms/ind/ind.cpp
```diff
@@ -0,0 +1,25 @@
+#include "ind.h"
+
+#include <sstream>
+
+namespace model {
+
+std::string CCToString(IND::ColumnCombination const& cc) {
```
polyntsov, line 7: Should be placed into anonymous namespace

src/core/algorithms/ind/ind.h
```diff
@@ -0,0 +1,46 @@
+#pragma once
+
+#include <memory>
+#include <string>
+#include <vector>
+
+namespace model {
+
+using AttributeIndex = unsigned int;
```
polyntsov, line 9: We have `Column::IndexType` for this (also see https://github.com/Mstrutov/Desbordante/issues/273)

polyntsov: Although maybe `Column::IndexType` should be moved out of `Column` class, since we don't really want to include `column.h` here.
Nevertheless, we certainly don't want another alias which represents the exact same thing -- type of index of the column in the table, so please fix this somehow :)

src/core/algorithms/ind/ind.h
```diff
@@ -0,0 +1,46 @@
...
+        size_t GetTableIndex() const {
+            return table_index_;
+        }
+        std::vector<AttributeIndex> const& GetColumnIndices() const {
+            return column_indices_;
+        }
+    };
+
+private:
+    std::shared_ptr<ColumnCombination> lhs, rhs;
```
polyntsov, line 31:
```suggestion
    std::shared_ptr<ColumnCombination> lhs_;
    std::shared_ptr<ColumnCombination> rhs_;
```

src/core/algorithms/ind/ind.h
```diff
@@ -0,0 +1,46 @@
...
+
+#include <memory>
+#include <string>
+#include <vector>
+
+namespace model {
+
+using AttributeIndex = unsigned int;
+
+class IND {
```
polyntsov, line 11: It would be great to have an explanatory comment on what this type represents

src/core/algorithms/ind/ind.h
polyntsov: We already have `config::IndexType` in `src/core/config/indices/type.h` and `TablePos` in `src/core/algorithms/fd/hycommon/types.h`...

src/core/algorithms/ind/ind.cpp
```diff
@@ -0,0 +1,25 @@
...
+    std::vector<AttributeIndex> const& col_ids = cc.GetColumnIndices();
+    std::stringstream ss;
+    for (size_t i = 0; i != col_ids.size(); ++i) {
```
polyntsov, line 10: nit: I'd rewrite this using iterators

src/core/algorithms/ind/ind.h
```diff
@@ -0,0 +1,46 @@
...
+public:
+    class ColumnCombination {
```
polyntsov, line 13: Not sure if it should be a nested class defined in `IND`, it seems like a class that may be useful in, for example, some FD mining algorithm, it doesn't look specific for inclusion dependencies.
We also have (unused) `src/core/model/table/column_combination.h`, that should be deleted. Maybe it's a good idea to move this definition there, but somehow indicate in its name that it uses vector to store set of attributes

src/core/algorithms/ind/ind.h
```diff
@@ -0,0 +1,46 @@
...
+    protected:
+        size_t table_index_;
```
polyntsov, line 15: It should be of type alias also, not `size_t` directly

polyntsov (changes requested):

polyntsov (approved):

PR #297, opened 2023-11-24
Fix max_lhs option

AndreyChernyshev555:
Declaration of the option was moved to FDAlgorithm class due to both tane and fastfds use it. The typed test for Tane and FastFDs algorithms was added. The value of the max_lhs option was set to 1 so it filter out some FDs.


src/core/algorithms/fd/fd_algorithm.cpp
```diff
@@ -51,7 +52,7 @@ std::vector<Column const*> FDAlgorithm::GetKeys() const {
 
     for (FD const& fd : FdList()) {
         Vertical const& lhs = fd.GetLhs();
-
+        if (lhs.GetArity() > max_lhs_) continue;
```
polyntsov, line 55: strange check, in the case when `max_lhs_` is treated correctly by the algorithm, `FdList()` couldn't possibly return `lhs` with arity greater than `max_lhs_`

src/tests/test_max_lhs_option.cpp
```diff
@@ -0,0 +1,57 @@
...
+    algorithm->Execute();
+    std::list<FD> result_fds_list = algorithm->FdList();
+    for (auto& fd : result_fds_list) {
+        ASSERT_TRUE(fd.GetLhs().GetArity() <= max_lhs);
+    }
+}
+
+REGISTER_TYPED_TEST_SUITE_P(MaxLHSOptionTest, MaxLHSOptionWork);
+
+using Algorithms = ::testing::Types<algos::Tane, algos::FastFDs>;
```
polyntsov, line 55: This test passes only because `Tane` and `FastFDs` treat `max_lhs` correctly internally during mining process, try `HyFD` which doesn't do anything like this and the test will fail. The goal is to make all algorithms working with `max_lhs` option even when they know nothing about it.
So this test should pass on _all_ fd mining algorithms. Also, there is no need to create a separate file with the test for this single option, add a test case to `src/tests/test_algorithm.cpp` instead (there all tests for fd mining algorithms are implemented)

src/tests/test_max_lhs_option.cpp
```diff
@@ -0,0 +1,57 @@
...
+InputTable GetInputTable(const std::filesystem::path& path, char separator = ',',
+                         bool has_header = true) {
+    return std::make_unique<CSVParser>(path, separator, has_header);
+}
+
+template <typename T>
+class MaxLHSOptionTest : public ::testing::Test {
+protected:
+    static std::unique_ptr<algos::FDAlgorithm> CreateAlgoInstance(unsigned int max_lhs,
+                                                                  bool EqNullsType,
```
polyntsov, line 25: naming, all variables are named using `snake_case`, check the style guide

src/core/algorithms/fd/fd_algorithm.cpp
```diff
@@ -1,21 +1,22 @@
...
+#include "config/max_lhs/option.h"
 #include "config/tabular_data/input_table/option.h"
 
 namespace algos {
 
 FDAlgorithm::FDAlgorithm(std::vector<std::string_view> phase_names)
     : Algorithm(std::move(phase_names)) {
     RegisterOptions();
-    MakeOptionsAvailable({config::TableOpt.GetName(), config::EqualNullsOpt.GetName()});
+    MakeOptionsAvailable({config::MaxLhsOpt.GetName(), config::TableOpt.GetName(),
```
polyntsov, line 14: `max_lhs` should be made available in the execute stage of the algorithm, not in the data load stage (in the current code it is made available in the `MakeExecuteOptsAvailable`, not when the algorithm object is constructed). Not sure if it's a desirable behaviour overall, but it is how it's implemented right now, so for this PR let's leave `max_lhs` as execute stage option

src/core/algorithms/fd/fd_algorithm.cpp
```diff
@@ -1,21 +1,22 @@
 #include "fd_algorithm.h"
 
-#include <map>
 #include <thread>
-#include <vector>
```
polyntsov, line 5: Why? This file uses classes defined in both of these headers

polyntsov (changes requested):

PR #303, opened 2023-11-27
Add datasets info for `tests`

vs9h:
In this pull request, a datasets info file has been added, which should contain information about all datasets in order to avoid code duplication in each test file.
It is important to note that this pull request does not aim to unify all test files, but is limited to only a part of the tests (namely `fd` and `ucc`).
In the future it can also be used in other tests.

Also fixed dataset info for `abalone.csv`: `has_header` `true` -> `false`


src/tests/datasets.h
```diff
@@ -4,74 +4,31 @@
...
 #include <vector>
 
-static const auto test_data_dir = std::filesystem::current_path() / "input_data";
+#include "config/tabular_data/input_table_type.h"
+#include "parser/csv_parser/csv_parser.h"
 
-struct Dataset {
-    std::string name;
-    size_t hash;
+static auto const test_data_dir = std::filesystem::current_path() / "input_data";
```
polyntsov, line 10: Why is it not in the namespace?

src/tests/datasets.h
```diff
@@ -4,74 +4,31 @@
...
+
+namespace tests {
+
+/// dataset configuration info to create an input table
+struct DatasetInfo {
```
polyntsov, line 15: `DatasetInfo` name seems too general, let's rename it to something that indicates that it actually describes csv datasets

polyntsov: Also, it's probably a good idea to rename this file accordingly

src/tests/test_fd_util.h
```diff
@@ -0,0 +1,59 @@
...
+
+#include <gtest/gtest.h>
+
+#include "algorithms/algo_factory.h"
+#include "algorithms/fd/fd_algorithm.h"
+#include "all_datasets_info.h"
+#include "config/error/type.h"
+#include "config/names.h"
+#include "datasets.h"
+
```
polyntsov, line 13: `namespace tests`

src/tests/datasets.h
```diff
@@ -4,74 +4,31 @@
...
-public:
-    static inline const std::array<Dataset, 6> datasets_ = {
-            {{"adult.csv", 23075, ';', false},
-             {"CIPublicHighway.csv", 13035, ',', true},
-             {"EpicMeds.csv", 50218, '|', true},
-             {"EpicVitals.csv", 2083, '|', true},
-             {"iowa1kk.csv", 28573, ',', true},
-             {"LegacyPayors.csv", 43612, '|', true}}};
+/// a pair consisting of a dataset and the expected hash
+using DatasetHashPair = std::pair<DatasetInfo, size_t>;
```
polyntsov, line 30: I think struct with two members with descriptive names is better than a pair with explanatory comment

src/tests/datasets.h
```diff
@@ -4,74 +4,31 @@
...
-struct Dataset {
-    std::string name;
-    size_t hash;
+static auto const test_data_dir = std::filesystem::current_path() / "input_data";
+
+namespace tests {
+
+/// dataset configuration info to create an input table
+struct DatasetInfo {
+    std::string_view name;
```
polyntsov, line 16: I think we should use `std::string` here, since this data member actually _owns_ the data it holds, so it shouldn't be `view` type. And by using `string_view` here, we're basically restricting ourselves to use only string literals for dataset names in the future, which doesn't seem like a right thing

src/tests/all_datasets_info.h
polyntsov, line 1: `all_datasets_info` seems to general, I'd indicate in the name that datasets there represent tables

polyntsov (changes requested):

polyntsov (approved):

PR #304, opened 2023-11-28
Add Spider algorithm

vs9h:
In this pull request was added:
- a spider algorithm - an algorithm for mining unary inclusion dependencies
- tests for inclusion algorithms

> Note: the first two commits related to the refactoring of hash-based tests are placed in a separate pull request (#303) and must be merged before this pull request.


src/core/model/table/block_data.h
```diff
@@ -0,0 +1,100 @@
...
+    using Value = ColumnData::Value;
+
+private:
+    std::vector<ColumnData> columns_;
+
+public:
+    BlockData(ColumnIndex cols_count) : columns_(cols_count) {}
+
+    /// insert row from data stream
+    void InsertRow(std::vector<std::string>& row);
```
vs9h, line 88:
```suggestion
    void InsertRow(std::vector<std::string> const& row);
```

src/core/model/table/block_dataset_stream.cpp
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * BlockDatasetStream class methods definition
+ */
+#include "block_dataset_stream.h"
+
+#include <easylogging++.h>
+
+namespace model {
+
+size_t BlockDatasetStream::GetRowSize(const Row& row) {
```
vs9h, line 12:
```suggestion
size_t BlockDatasetStream::GetRowSize(Row const& row) {
```

src/core/algorithms/ind/spider/spider.h
```diff
@@ -0,0 +1,68 @@
...
+    /// timing information for algorithm stages
+    struct StageTimings {
+        size_t load;    /**< time taken for the data loading */
+        size_t init;    /**< time taken for the attributes initialization */
+        size_t compute; /**< time taken for the inds computing */
+        size_t total;   /**< total time taken for all stages */
+    };
+
+private:
+    using Domain = model::ColumnDomain;
```
vs9h, line 34: remove this alias

src/core/algorithms/ind/spider/spider.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+    RegisterOption(config::ThreadNumberOpt(&threads_num_));
+    RegisterOption(config::MemLimitMBOpt(&mem_limit_mb_));
+    RegisterOption(config::ErrorOpt(&max_ind_error_));
+    MakeLoadOptsAvailable();
+}
+
+void Spider::MakeLoadOptsAvailable() {
+    MakeOptionsAvailable({config::ThreadNumberOpt.GetName(), config::MemLimitMBOpt.GetName(),
+                          config::ErrorOpt.GetName()});
```
vs9h, line 32: remove `ErrorOpt`

src/core/config/all_options.cpp
vs9h, line 1: remove these changes

src/core/algorithms/ind/spider/attribute.cpp
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * Attribute class methods definition
+ */
+#include "attribute.h"
+
+namespace algos::spider {
+
+std::vector<AttributeIndex> AINDAttribute::GetRefIds(config::ErrorType max_error) const {
+    std::vector<AttributeIndex> refs;
+    auto dep_count = static_cast<double>(occurrences_[id_]);
```
vs9h, line 12:
```suggestion
    auto dep_count = static_cast<config::ErrorType>(occurrences_[id_]);
```

src/core/algorithms/ind/spider/attribute.cpp
```diff
@@ -0,0 +1,38 @@
+/** \file
+ * \brief Spider attribute
+ *
+ * Attribute class methods definition
+ */
+#include "attribute.h"
+
+namespace algos::spider {
+
+std::vector<AttributeIndex> AINDAttribute::GetRefIds(config::ErrorType max_error) const {
```
slesarev-hub, line 10:
```suggestion
std::vector<AttributeIndex> AINDAttribute::GetRefIds(config::ErrorType const max_error) const {
```

src/core/algorithms/ind/spider/attribute.cpp
slesarev-hub, line 12:
```suggestion
    const auto dep_count = static_cast<double>(occurrences_[id_]);
```

src/core/algorithms/ind/spider/attribute.cpp
```diff
@@ -0,0 +1,38 @@
...
+    std::vector<AttributeIndex> refs;
+    auto dep_count = static_cast<double>(occurrences_[id_]);
+    for (size_t ref_id = 0; ref_id != occurrences_.size(); ++ref_id) {
+        if (id_ == ref_id) continue;
+        config::ErrorType error = 1 - occurrences_[ref_id] / dep_count;
```
slesarev-hub, line 15:
```suggestion
        const config::ErrorType error = 1 - occurrences_[ref_id] / dep_count;
```

src/core/algorithms/ind/spider/attribute.h
```diff
@@ -0,0 +1,145 @@
...
+    AttributeIndex id_;         /* attribute unique identificator */
+    AttributeIndex attr_count_; /* attribute unique identificator */
+    Iterator it_;               /* domain iterator */
+
+public:
+    Attribute(AttributeIndex attr_id, AttributeIndex attr_count, model::ColumnDomain const& domain)
+        : id_(attr_id), attr_count_(attr_count), it_(domain) {}
+
+    /// get unqiue attribute id
+    AttributeIndex GetId() const {
```
slesarev-hub, line 39:
```suggestion
    AttributeIndex GetId() const noexcept {
```

src/core/algorithms/ind/spider/attribute.h
```diff
@@ -0,0 +1,145 @@
...
+        return id_;
+    }
+
+    /// get attribute iterator
+    Iterator& GetIt() {
```
slesarev-hub, line 44:
```suggestion
    Iterator& GetIt() noexcept {
```

src/core/algorithms/ind/spider/attribute.h
```diff
@@ -0,0 +1,145 @@
...
+        return it_;
+    }
+
+    /// get attribute iterator
+    Iterator const& GetIt() const {
```
slesarev-hub, line 49:
```suggestion
    Iterator const& GetIt() const noexcept {
```

src/core/model/table/column_domain.cpp
```diff
@@ -0,0 +1,306 @@
...
+    } else {
+        return std::make_unique<MemoryBackedReader>(values_);
+    }
+}
+
+size_t DomainPartition::GetMemoryUsage() const {
+    /* in reality, partition storage takes on average
+     * 5 times more memory than the presented estimate
+     */
+    static constexpr double kAverageOverhead = 5.0;
```
slesarev-hub, line 102: Could you provide information about the reason to choose this value?

src/core/model/table/column_domain.cpp
```diff
@@ -0,0 +1,306 @@
...
+    size_t total = 0;
+
+    if (IsSwapped()) return total;
+
+    total += values_.size() * sizeof(Values::node_type);
+
+    for (std::string const& str : values_) {
+        if (str.size() >= str.capacity()) {
```
slesarev-hub, line 110: `str.size() <= str.capacity()`, isn't it?

src/core/model/table/column_domain.cpp
```diff
@@ -0,0 +1,306 @@
...
+        return std::make_unique<MemoryBackedReader>(values_);
+    }
+}
+
+size_t DomainPartition::GetMemoryUsage() const {
+    /* in reality, partition storage takes on average
+     * 5 times more memory than the presented estimate
+     */
+    static constexpr double kAverageOverhead = 5.0;
+    size_t total = 0;
```
slesarev-hub, line 103:
```suggestion
    if (IsSwapped()) return 0;
    const size_t data_size = std::accumulate(values_ | std::views::transform([](const std::string& str){ return str.capacity(); }), 0);
    return (data_size + values_.size() * sizeof(Values::node_type)) * kAverageOverhead;
```

src/core/algorithms/ind/spider/attribute.h
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    /// check whether the attribute has processed
+    virtual bool HasFinished() const {
+        return !GetIt().HasNext();
+    }
+
+    /// compare attributes first by their values and then by their ids
+    bool operator>(Attribute const& rhs) const {
+        int cmp = GetIt().GetValue().compare(rhs.GetIt().GetValue());
```
slesarev-hub, line 60: Expression `GetIt().GetValue()` is used several times in different places, so maybe we should add method `GetValue()` to `Attribute` class?

src/core/algorithms/ind/spider/spider.cpp
```diff
@@ -0,0 +1,144 @@
...
+                                              config::EqNullsType is_null_equal_null) {
+    using AttributeRW = std::reference_wrapper<Attribute>;
+    std::vector attrs = InitAttributes<Attribute>(domains);
+    AttributeIndex attr_count = attrs.size();
+    std::priority_queue<AttributeRW, std::vector<AttributeRW>, std::greater<Attribute>> attr_pq(
+            attrs.begin(), attrs.end());
+    while (!attr_pq.empty()) {
+        boost::dynamic_bitset<> ids_bitset(attr_count);
+        AttributeRW attr_rw = attr_pq.top();
+        std::string const& value = attr_rw.get().GetIt().GetValue();
```
slesarev-hub, line 77:
```suggestion
        Value const& value = attr_rw.get().GetIt().GetValue();
```
Or just `auto`?

src/core/algorithms/ind/spider/spider.cpp
```diff
@@ -0,0 +1,144 @@
...
+    std::vector attrs = GetProcessedAttributes<AINDAttribute>(domains_, is_null_equal_null_);
+    for (auto const& dep : attrs) {
+        for (AttributeIndex ref_id : dep.GetRefIds(max_ind_error_)) {
+            RegisterIND(dep.ToCC(), attrs[ref_id].ToCC());
+        }
+    }
+}
+
+unsigned long long Spider::ExecuteInternal() {
+    auto mining_func = max_ind_error_ == 0 ? &Spider::MineINDs : &Spider::MineAINDs;
```
slesarev-hub, line 132:
```suggestion
    auto mining_func = (max_ind_error_ == 0) ? &Spider::MineINDs : &Spider::MineAINDs;
```

src/core/util/timed_invoke.h
```diff
@@ -0,0 +1,16 @@
...
+
+#include <chrono>
+#include <functional>
+#include <utility>
+
+namespace util {
+/// invoke and return the time taken for execution
+template <typename... Ts>
+size_t TimedInvoke(Ts&&... invoke_args) {
+    auto start = std::chrono::system_clock::now();
```
slesarev-hub, line 11:
```suggestion
    const auto start = std::chrono::system_clock::now();
```

src/core/util/timed_invoke.h
```diff
@@ -0,0 +1,16 @@
...
+    std::invoke(std::forward<Ts>(invoke_args)...);
+    auto end = std::chrono::system_clock::now();
```
slesarev-hub, line 13:
```suggestion
    const auto end = std::chrono::system_clock::now();
```

src/core/algorithms/ind/spider/spider.cpp
```diff
@@ -0,0 +1,144 @@
...
+    timings_.compute = util::TimedInvoke(mining_func, this);
+    timings_.total = timings_.load + timings_.compute;
+    return timings_.total;
+}
+
+void Spider::ResetINDAlgorithmState() {
+    size_t load = timings_.load;
```
slesarev-hub, line 139: Why not this:
```suggestion
    timings_.compute = 0;
    timings_.total = 0;
```
?

src/tests/test_hash_util.cpp
```diff
@@ -0,0 +1,37 @@
...
+    return seed;
+}
+
+std::size_t Hash(std::vector<std::vector<unsigned>> const& vec) {
+    return HashVec(vec, Hash);
+}
+
+std::size_t Hash(std::pair<unsigned, std::vector<unsigned>> const& pair) {
+    auto const& [lhs, rhs] = pair;
+    std::size_t hash = 31 * lhs + Hash(rhs);
```
slesarev-hub, line 33:
```suggestion
    return 31 * lhs + Hash(rhs);
```

src/core/algorithms/ind/spider/attribute.cpp
vs9h: I believe that making double a constant parameter in the declaration is optional and not necessary. It is clear that in theory this can help avoid bugs when we accidentally change a variable that we did not want to change. But then it is necessary to change it everywhere, but I don’t see much point in this.

src/core/model/table/column_domain.cpp
vs9h: Estimating memory consumption is the most difficult part. In fact, we rely on a specific implementation, which we would like to avoid. Ideally, it seems to me that we should monitor memory usage in another way (maybe use `std::pmr::set` with custom allocator).
I also made some changes to the code that made the estimation more accurate.
Regarding `str.size() >= str.capacity()`, this is an error. There should be a `!=` check here. We assume that the compiler has SSO (small string optimization). And enough small strings should fit completely into the `std::string`. If the SSO works for string, then the size will not match the capacity. Otherwise, we take the capacity, since in this case memory allocation occurred on the heap.
So, `data_total_size` will contain memory usage in bytes for string's data, which was allocated in the heap.
And it also remains to take into account how much memory is spent on the nodes themselves. Here we multiply the number of nodes by the sum of the `sizeof(std::string)` and `sizeof(std::set<std::string>::node_type)`. And we also have to multiply this estimate by the coefficient that was derived experimentally.

I will also give an example of how this estimation works on TPCH dataset (with scale factor 1):
```
memory usage (estimate): 0 mb, current rss: 8.625 mb
memory usage (estimate): 93 mb, current rss: 101.098 mb
memory usage (estimate): 309 mb, current rss: 317.098 mb
memory usage (estimate): 827 mb, current rss: 846.598 mb
memory usage (estimate): 827 mb, current rss: 846.598 mb
memory usage (estimate): 1214 mb, current rss: 1224.1 mb
memory usage (estimate): 1277 mb, current rss: 1285.97 mb
memory usage (estimate): 1331 mb, current rss: 1339.6 mb
memory usage (estimate): 1520 mb, current rss: 1533.8 mb
memory usage (estimate): 1520 mb, current rss: 1533.8 mb
```
Left is estimation from `GetMemoryUsage()`, right is current RSS.
In this case we get a fairly good estimate.

src/core/algorithms/ind/spider/spider.cpp
vs9h: I believe that it is still better to leave `std::string`, since it should be obvious from the code what type a particular variable is. 

I also removed the `Attribute::GetIt()` method to prevent direct iterator using. So if we want to do this now, then we must add public alias in the `Attribute` class (`using Value = Iterator::Value`), and then use it here and write something like `Attribute::Value const& value = ...`. But to find out the `Value` type you (as a reader) will have to go to the `Attribute` class, then to the class `ColumnDomainIterator`, then to the class `DomainPartition` and only at that moment will you see the definition of this alias. I believe that it is better to leave this type alias in the model namespace and don't use it here.

src/core/algorithms/ind/spider/spider.cpp
```diff
@@ -0,0 +1,144 @@
...
+    for (AttributeIndex attr_id = 0; attr_id != attr_count; ++attr_id) {
+        attrs.emplace_back(attr_id, attr_count, domains[attr_id]);
+    }
+    return attrs;
+}
+
+template <typename Attribute>
+std::vector<Attribute> GetProcessedAttributes(std::vector<model::ColumnDomain> const& domains,
+                                              config::EqNullsType is_null_equal_null) {
+    using AttributeRW = std::reference_wrapper<Attribute>;
```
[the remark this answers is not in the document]
vs9h: I use a `std::reference_wrapper` because I want to create a "vector of references". This class stores a pointer and has a reference casting operator `operator T&()`, as well as a constructor from reference.

Next, I will talk about the advantages of using this class compared to a raw pointer.
We can use a constructor from two iterators (just pass `attrs.begin()`, `attrs.end()`). In the case of raw pointers, we call priority queue constructor, then write loop to push pointers. 
Also, if we used raw pointers, we would have to redefine the compare operator, since we would not be able to use just `std::greater<Attribute>`.

However, we had to write `attr_rw.get()` several times, it seems unavoidable. 

To summarize, I think that in general there are noticeably more advantages, since I managed to avoid writing some code.

I also like the semantic approach to using a `std::reference_wrapper`. When we see raw pointer, then we wonder if it could be a null? In the case of a `std::reference_wrapper`, we work with it as with a reference.

src/core/algorithms/ind/spider/attribute.cpp
vs9h: Yes, indeed, a const can be added in a function definition without adding a const in the declaration. But I’m rather saying that if we want to add a constant in one place, then we should also add it in similar places. Because the code must be consistent at least within one pull request. Ideally, the coding style should be consistent throughout the entire project. But I have not seen such a use of constants for parameters in function definitions in desbordante.

src/core/algorithms/ind/spider/spider.cpp
```diff
@@ -0,0 +1,143 @@
...
+            attr_rw = attr_pq.top();
+            if (value.empty() && !is_null_equal_null) break;
+        } while (attr_rw.get().GetCurrentValue() == value);
+
+        auto ids_vec = util::BitsetToIndices<AttributeIndex>(ids_bitset);
+        for (auto id : ids_vec) {
+            if constexpr (std::is_same_v<Attribute, spider::INDAttribute>) {
+                attrs[id].IntersectRefs(ids_bitset, attrs);
+            } else {
+                attrs[id].IntersectRefs(ids_bitset);
```
slesarev-hub, line 91: Could we pass `ids_vec` to `IntersectRefs` and not calculate it inside?

src/core/algorithms/ind/spider/spider.cpp
```diff
@@ -0,0 +1,143 @@
...
+template <typename Attribute>
+std::vector<Attribute> GetProcessedAttributes(std::vector<model::ColumnDomain> const& domains,
+                                              config::EqNullsType is_null_equal_null) {
+    using AttributeRW = std::reference_wrapper<Attribute>;
+    std::vector attrs = InitAttributes<Attribute>(domains);
+    AttributeIndex attr_count = attrs.size();
+    std::priority_queue<AttributeRW, std::vector<AttributeRW>, std::greater<Attribute>> attr_pq(
+            attrs.begin(), attrs.end());
+    while (!attr_pq.empty()) {
+        boost::dynamic_bitset<> ids_bitset(attr_count);
```
slesarev-hub, line 75: How about move `ids_bitset` out of loop and use `reset` at the end of iteration?

src/core/config/mem_limit/option.cpp
```diff
@@ -0,0 +1,15 @@
+#include "config/mem_limit/option.h"
+
+#include "config/names_and_descriptions.h"
+
+namespace config {
+using names::kMemLimitMB, descriptions::kDMemLimitMB;
+extern CommonOption<MemLimitMBType> const MemLimitMBOpt{
```
slesarev-hub, line 7: Aggregate initialization could be more readable here

src/core/model/table/block_data.cpp
```diff
@@ -0,0 +1,55 @@
...
+}
+
+ColumnIterator::ColumnIterator(std::vector<char> const& data, std::vector<Offset> const& offsets)
+    : buffer_(data), cur_(offsets.cbegin()), end_(offsets.cend()) {
+    assert(!offsets.empty());
+    ConstructView();
+}
+
+bool ColumnIterator::TryMoveToNext() {
+    bool can_move = std::next(cur_) != end_;
```
slesarev-hub, line 32: const

src/core/model/table/block_data.cpp
```diff
@@ -0,0 +1,55 @@
...
+    if (can_move) {
+        std::advance(cur_, 1);
+        ConstructView();
+    }
+    return can_move;
+}
+
+void ColumnData::Insert(std::string const& str) {
+    assert(raw_data_.size() < std::numeric_limits<Offset>::max());
+    Offset offset = raw_data_.size();
```
slesarev-hub, line 42: const

src/core/model/table/block_dataset_stream.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+size_t BlockDatasetStream::GetRowSize(Row const& row) {
+    return std::accumulate(row.begin(), row.end(), 0UL,
+                           [](size_t acc, std::string const& value) { return acc + value.size(); });
+}
+
+bool BlockDatasetStream::TryStoreNextRow() {
+    if (cur_row_.empty() && stream_->HasNextRow()) {
+        cur_row_ = stream_->GetNextRow();
+        size_t cols_num = stream_->GetNumberOfColumns();
```
slesarev-hub, line 22:
```suggestion
       const size_t cols_num = stream_->GetNumberOfColumns();
```

src/core/model/table/block_data.h
```diff
@@ -0,0 +1,105 @@
...
+    /// insert column value
+    void Insert(std::string const& str);
+
+    /// check if column empty
+    bool IsEmpty() const {
+        return raw_data_.empty();
+    }
+
+    /// get column data iterator
+    ColumnIterator GetIt() const {
```
slesarev-hub, line 71:
```suggestion
    ColumnIterator GetIt() const noexcept {
```

src/core/model/table/block_data.h
```diff
@@ -0,0 +1,105 @@
...
+    std::vector<Offset> offsets_; /* column values offsets in `raw_data_`  */
+
+public:
+    ColumnData() = default;
+
+    /// insert column value
+    void Insert(std::string const& str);
+
+    /// check if column empty
+    bool IsEmpty() const {
```
slesarev-hub, line 66:
```suggestion
    bool IsEmpty() const noexcept {
```

src/core/algorithms/ind/spider/spider.h
```diff
@@ -0,0 +1,69 @@
...
+    void MineAINDs();
+
+    unsigned long long ExecuteInternal() final;
+    void ResetINDAlgorithmState() final;
+
+public:
+    explicit Spider();
+
+    /// get information about stage timings
+    StageTimings const& GetStageTimings() const {
```
slesarev-hub, line 64:
```suggestion
    StageTimings const& GetStageTimings() const noexcept {
```

src/core/algorithms/ind/spider/attribute.h
```diff
@@ -0,0 +1,141 @@
...
+    ///
+    /// processing is completed if all values have been processed
+    /// or there are no more dependent and referenced candidates
+    ///
+    bool HasFinished() const final {
+        return !it_.HasNext() || (refs_.none() && deps_.none());
+    }
+
+    /// get referenced attributes indices
+    boost::dynamic_bitset<>& GetRefsBitset() {
```
slesarev-hub, line 136:
```suggestion
    boost::dynamic_bitset<>& GetRefsBitset() noexcept {
```

src/core/algorithms/ind/spider/attribute.h
```diff
@@ -0,0 +1,141 @@
...
+    Attribute(AttributeIndex attr_id, AttributeIndex attr_count, model::ColumnDomain const& domain)
+        : id_(attr_id), attr_count_(attr_count), it_(domain) {}
+
+    /// get unqiue attribute id
+    AttributeIndex GetId() const noexcept {
+        return id_;
+    }
+
+    /// check whether the attribute has processed
+    virtual bool HasFinished() const {
```
slesarev-hub, line 45:
```suggestion
    virtual bool HasFinished() const noexcept {
```

src/core/algorithms/ind/spider/attribute.h
```diff
@@ -0,0 +1,141 @@
...
+        return !it_.HasNext();
+    }
+
+    std::string const& GetCurrentValue() const {
```
slesarev-hub, line 49:
```suggestion
    std::string const& GetCurrentValue() const noexcept {
```

src/core/model/table/column_domain.cpp
```diff
@@ -0,0 +1,301 @@
...
+public:
+    explicit MemoryBackedReader(Values const& values) : cur_(values.begin()), end_(values.end()) {
+        assert(cur_ != end_);
+    }
+
+    Value const& GetValue() const final {
+        return *cur_;
+    }
+
+    bool HasNext() const final {
```
slesarev-hub, line 42:
```suggestion
    bool HasNext() const final noexcept {
```

src/core/model/table/column_domain.cpp
```diff
@@ -0,0 +1,301 @@
...
+    using It = Values::iterator;
+
+    It cur_, end_;
+
+public:
+    explicit MemoryBackedReader(Values const& values) : cur_(values.begin()), end_(values.end()) {
+        assert(cur_ != end_);
+    }
+
+    Value const& GetValue() const final {
```
slesarev-hub, line 38:
```suggestion
    Value const& GetValue() const final noexcept {
```

src/core/config/mem_limit/option.cpp
vs9h: I don’t fully understand, `CommonOption` is a class

src/core/model/table/column_domain.cpp
```diff
@@ -0,0 +1,302 @@
...
+            Domain& domain = domains_[swap_candidate_];
+            size_t const domain_mem_usage = domain.GetMemoryUsage();
+            domain.Swap();
+            mem_usage_ -= domain_mem_usage;
+            ++swap_candidate_;
+            return;
+        }
+        /* swap current table domains */
+        for (DomainRawData& raw_domain : raw_domains_) {
+            Partition& partition = raw_domain.back();
```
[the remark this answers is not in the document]
vs9h: we try to swap this partition after this line, so the partition isn't const

src/core/model/table/column_domain.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+std::vector<ColumnDomain> ColumnDomain::CreateFrom(
+        std::vector<std::shared_ptr<model::IDatasetStream>> const& streams,
+        config::MemLimitMBType mem_limit_mb, config::ThreadNumType threads_num) {
+    TableIndex const table_count = static_cast<TableIndex>(streams.size());
+    DomainManager manager{mem_limit_mb, threads_num};
+    for (TableIndex table_id = 0; table_id != table_count; ++table_id) {
+        manager.ProcessDatasetStream(table_id, streams[table_id]);
+    }
+    return std::move(manager).GetDomains();
```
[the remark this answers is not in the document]
vs9h: Yes, we can only call this method on rvalue object, since we do not make any guarantees about the state of the manager after calling this method. Otherwise, we need to add a method to reset the manager state. But this manager is not reusable and the use of this method clearly prohibits further use of this manager

src/core/model/table/column_domain.cpp
```diff
@@ -0,0 +1,302 @@
...
+        processed_block_count_ = 0;
+        size_t block_count = 0;
+        BlockDatasetStream block_stream{stream, block_capacity_};
+        do {
+            processed_block_count_ += block_count;
+            block_count = GetNumberOfBlocks();
+        } while (ProcessNext(block_stream, block_count));
+
+        for (DomainRawData& raw_domain : raw_domains_) {
+            if (!raw_domain.front().IsEmpty()) {
```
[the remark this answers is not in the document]
vs9h: 1) about empty domain: I added comments in the code and changed the function name from `IsEmpty` to `IsNull`. We do not work with columns that either contain no values at all or contain only nulls (i.e. empty values). For example, in this PR, the `IndTestNulls.csv` table was added. And when we mine inds on this table, the `null` and `null(2)` columns will be discarded at this stage and will not participate in further traversal of partitions. Also, all the logic for working with nulls for inclusion dependencies is described in the [pull request with faida](https://github.com/Mstrutov/Desbordante/pull/272#pullrequestreview-1766233923) (this is an algorithm for mining inclusion dependencies).
2) We assume that if a domain is created, it must contain at least one value. When creating iterators, we rely on the domain to contain at least one value. In theory, we can first add a domain and then remove the domain if it does not contain values. But I don’t think this solution is definitely better.

slesarev-hub (approved):

PR #315, opened 2023-12-09
Add tests and usage example for DataStats Python bindings

xJoskiy:


examples/data_stats.py
```diff
@@ -0,0 +1,54 @@
+import desbordante as db
+
+# Parameters for data_stats.load_data(...)
+DATASET_PATH = "TestDataStats.csv"
```
BUYT-1, line 4:
```suggestion
DATASET_PATH = "datasets/TestDataStats.csv"
```

BUYT-1 (changes requested):
Examples are expected to be copied to a specific directory along with their datasets

BUYT-1 (approved):

PR #316, opened 2023-12-10
Add ACAlgorithm python example

Sched71:


examples/algebraic_constraints.py
```diff
@@ -0,0 +1,44 @@
...
+BUMPS_LIMIT = 0
+WEIGHT = 0.1
+BIN_OPERATION = '-'
+AC_SEED = 11
+ITERATIONS_LIMIT = 4
+
+algo = desbordante.ACAlgorithm()
+
+df = pandas.read_csv(TABLE, sep=SEPARATOR, header=HEADER)
+df_without_id = df[['Delivery date', 'Dispatch date']]
```
BUYT-1, line 18:
```suggestion
df = pandas.read_csv(TABLE, sep=SEPARATOR, header=HEADER, index_col=['id'])
```

examples/algebraic_constraints.py
```diff
@@ -0,0 +1,44 @@
...
+
+print("Algorithm needs the following data format:", end=' ')
+print(algo.get_needed_options())
+algo.load_data(df=df_without_id)
+
+print("Algorithm needs the following options:", end=' ')
+print(algo.get_needed_options())
```
BUYT-1, line 25:
```suggestion
algo.load_data(df_without_id)
```
I don't think those are needed. Printing the options you are using, on the other hand, makes more sense.

examples/algebraic_constraints.py
```diff
@@ -0,0 +1,44 @@
...
+for ac_range in ac_ranges:
+    l_col = df.columns[ac_range.column_indices[0]]
+    r_col = df.columns[ac_range.column_indices[1]] 
+    print('Discovered ranges for', '(', l_col, BIN_OPERATION, r_col, ') are:')
+    print(ac_range.ranges)
+
+ac_exceptions = algo.get_ac_exceptions()
+for ac_exception in ac_exceptions:
+    print()
+    print('Row which is an exception, because it\'s difference is outside of discovered ranges:')
```
BUYT-1, line 39:
```suggestion
    print('Row which is an exception, because its difference is outside of discovered ranges:')
```
By the way, you should use double quotes when a single quote is inside the string literal and vice versa

examples/algebraic_constraints.py
```diff
@@ -0,0 +1,44 @@
...
+    print(f'id: {df.loc[ac_exception.row_index][0]}')
+    print(f'Dispatch date: {df.loc[ac_exception.row_index][2]}')
+    print(f'Delivery date: {df.loc[ac_exception.row_index][1]}')
+    print(f'Difference: {df.loc[ac_exception.row_index][1] - df.loc[ac_exception.row_index][2]}')
```
BUYT-1, line 43: `df.iloc[ac_exception.row_index]`
It should be in a separate variable. Use `iloc` instead ([doc](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.loc.html)):
> Allowed inputs are:
> * A single label, e.g. 5 or 'a', (note that 5 is interpreted as a *label* of the index, and **never** as an integer position along the index).

examples/algebraic_constraints.py
```diff
@@ -0,0 +1,44 @@
...
+print("Algorithm needs the following options:", end=' ')
+print(algo.get_needed_options())
+algo.execute(p_fuzz=P_FUZZ, fuzziness=FUZZINESS, bumps_limit=BUMPS_LIMIT, weight=WEIGHT,
+             bin_operation=BIN_OPERATION, ac_seed=AC_SEED, iterations_limit=ITERATIONS_LIMIT)
+
+ac_ranges = algo.get_ac_ranges()
+for ac_range in ac_ranges:
+    l_col = df.columns[ac_range.column_indices[0]]
+    r_col = df.columns[ac_range.column_indices[1]] 
+    print('Discovered ranges for', '(', l_col, BIN_OPERATION, r_col, ') are:')
```
BUYT-1, line 33:
```suggestion
    print(f'Discovered ranges for ({l_col} {BIN_OPERATION} {r_col}) are:')
```

examples/algebraic_constraints.py
BUYT-1: Should this text be printed on every iteration? Maybe it would be better to put this outside the loop. 

`print(f'Rows in which the result of the chosen operation ({BIN_OPERATION}) is outside of discovered ranges')`

BUYT-1 (changes requested):
Copy the example and its dataset using `examples/CMakeLists.txt`

examples/algebraic_constraints.py
Sched71: We left it because we wanted somewhere an example on how to get options for the algorithm

examples/algebraic_constraints.py
Sched71: I don't think I want it like that. Those are id's of cargos, it just happens so that they are identical to row indices

BUYT-1:
```suggestion
df = pandas.read_csv(TABLE, sep=SEPARATOR, header=HEADER, use_cols=['Delivery date', 'Dispatch date'])
```

Sched71: Very nice

Sched71: But I later use values from id column

BUYT-1: Then I guess that's fine

examples/algebraic_constraints.py
BUYT-1: No, I meant `row = df.iloc[ac_exception.row_index]`

BUYT-1: Or even `id_, delivery_date, dispatch_date = df.iloc[ac_exception.row_index]`, if that's possible

BUYT-1: `id_, delivery_date, dispatch_date = df.iloc[ac_exception.row_index].tolist()` if the above doesn't work

examples/algebraic_constraints.py
```diff
@@ -0,0 +1,43 @@
...
+ac_exceptions = algo.get_ac_exceptions()
+print()
+print(f'Rows in which the result of the chosen operation ({BIN_OPERATION}) is outside of discovered ranges:')
+for ac_exception in ac_exceptions:
+    id, delivery_date, dispatch_date = df.iloc[ac_exception.row_index]
+    print(f'id: {id}')
+    print(f'Dispatch date: {dispatch_date}')
+    print(f'Delivery date: {delivery_date}')
+    difference = delivery_date - dispatch_date
+    print(f'Difference: {difference}')
```
BUYT-1, line 41: This should probably depend on `BIN_OPERATION`.
```py
operations = {
    '+': (operator.add, 'Sum'),
    '-': (operator.sub, 'Difference'),
    ...
}
...
action, op_name = operations[BIN_OPERATION]
print(f'{op_name}: {action(delivery_date, dispatch_date)}')
```

BUYT-1 (approved):

PR #322, opened 2023-12-13
Fix bug with error option for afd_verification

DaniilGoncharov:


cli/cli.py
```diff
@@ -329,7 +329,7 @@ def check_mismatch(algo: str | None, task: str | None) -> None:
 
 
 def check_error_option_presence(task: str | None, error: str | None) -> None:
-    if task in (Task.afd, Task.afd_verification) and error is None:
+    if task in Task.afd and error is None:
```
BUYT-1, line 332:
```suggestion
    if task == Task.afd and error is None:
```

BUYT-1 (changes requested):

BUYT-1 (approved):

PR #323, opened 2023-12-14
Change afd_verification output

DaniilGoncharov:


cli/cli.py
```diff
@@ -370,7 +370,13 @@ def get_algo_result(algo: desbordante.Algorithm, algo_name: str) -> Any:
             case Algorithm.naive_fd_verifier:
                 result = algo.fd_holds()
             case Algorithm.naive_afd_verifier:
-                result = algo.get_error()
+                error = algo.get_error()
+                if error == 0.0:
+                    result = 'Functional dependency holds'
```
polyntsov, line 375:
```suggestion
                    result = 'Exact functional dependency holds'
```

cli/cli.py
```diff
@@ -370,7 +370,13 @@ def get_algo_result(algo: desbordante.Algorithm, algo_name: str) -> Any:
...
+                else:
+                    result = (f'Functional dependency does not hold, but '
```
polyntsov, line 377:
```suggestion
                    result = (f'Exact functional dependency does not hold, but '
```

polyntsov (changes requested):
Also, please make the commit message less vague

polyntsov (approved):

Issue #287, opened 2023-11-15
Ignore the last newline in CSVParser


BUYT-1:
For files that end with a newline, CSVParser::GetNextRow returns an extraneous empty line, which is not desirable

PR #326, opened 2023-12-17
Implement separator validation

MichaelS239:
This PR:
- Adds methods for separator validation in .csv tables
- Adds the last newline ignoring (issue #287)


src/core/algorithms/algo_factory.cpp
```diff
@@ -47,7 +48,9 @@ void LoadAlgorithm(Algorithm& algorithm, StdParamsMap const& options) {
     ConfigureFromFunction(algorithm, [&options](std::string_view option_name) {
         using namespace config::names;
         auto create_input_table = [](CSVConfig const& csv_config) -> config::InputTable {
-            return std::make_shared<CSVParser>(csv_config);
+            auto csv_parser = std::make_shared<CSVParser>(csv_config);
+            csv_parser->ValidateSeparator();
+            return csv_parser;
```
vs9h, line 53: Also, I'm not sure that we always want to validate the separator when creating a table. The main problem is that we have to read the table completely before running the algorithm itself. This causes the file system caches and disk cache to fill up. Therefore, if someone tries to conduct experiments and does not know that validation always takes place first, then the results of the experiments will be incorrect (since all caches must be reset before running the test).

vs9h (changes requested):

PR #330, opened 2023-12-19
Add GetDestructor method

BUYT-1:
Provides a more convenient way to destruct an object, used in #327, and perhaps it should be used in `TypedColumnData` too.

### DIFF
--- a/src/core/model/types/type.h
+++ b/src/core/model/types/type.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <functional>
 #include <memory>
 #include <sstream>
 
@@ -12,8 +13,9 @@ private:
     TypeId type_id_;
 
 public:
-    explicit Type(TypeId const type_id) noexcept : type_id_(type_id) {}
+    using Destructor = std::function<void(std::byte*)>;
 
+    explicit Type(TypeId const type_id) noexcept : type_id_(type_id) {}
     virtual ~Type() = default;
 
     /* Operations on the type itself */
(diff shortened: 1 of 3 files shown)


src/core/model/types/type.h
```diff
@@ -126,6 +128,10 @@ class Type {
         return !(type_id == +TypeId::kEmpty || type_id == +TypeId::kNull ||
                  type_id == +TypeId::kUndefined || type_id == +TypeId::kMixed);
     }
+
+    virtual Destructor GetDestructor() const {
+        return nullptr;
+    }
```
polyntsov, line 134: Why not make a `virtual Desturct` method with an empty body in the base class? Given that some types have non-trivial destruction logic and already have method `Destruct` adding `GetDestructor` seems a bit excessive. Also, the problem with `GetDestructor` is a default behavior in the base class, I don't think returning `nullptr` is a good idea, because call to `nullptr` will fail.

polyntsov (changes requested):
Also, please, make the commit message a bit more specific

src/core/model/types/type.h
BUYT-1: I thought it would make sense to check if the type was trivially destructible to avoid pointless calls. Checking for `nullptr` does just that

BUYT-1:
Nevermind, let's just use #285 
`~TypedColumnData` uses this type info, but a pointer needs to be created for virtual methods to work, which may throw

PR #333, opened 2023-12-21
Fix max_lhs option

AndreyChernyshev555:
Processing of the max_lhs option is now supported by all
fd algorithms. The typed test for it was added.

### DIFF
--- a/src/core/algorithms/fd/dfd/dfd.cpp
+++ b/src/core/algorithms/fd/dfd/dfd.cpp
@@ -3,6 +3,7 @@
 #include <boost/asio.hpp>
 #include <easylogging++.h>
 
+#include "config/max_lhs/option.h"
 #include "config/thread_number/option.h"
 #include "lattice_traversal/lattice_traversal.h"
 #include "model/table/column_layout_relation_data.h"
@@ -19,7 +20,7 @@ void DFD::RegisterOptions() {
     RegisterOption(config::kThreadNumberOpt(&number_of_threads_));
 }
 
-void DFD::MakeExecuteOptsAvailable() {
+void DFD::MakeExecuteOptsAvailableFDInternal() {
     MakeOptionsAvailable({config::kThreadNumberOpt.GetName()});
 }
 
--- a/src/core/algorithms/fd/fd_algorithm.h
+++ b/src/core/algorithms/fd/fd_algorithm.h
@@ -8,6 +8,7 @@
 
 #include "algorithms/algorithm.h"
 #include "config/equal_nulls/type.h"
+#include "config/max_lhs/type.h"
 #include "config/tabular_data/input_table_type.h"
 #include "fd.h"
 #include "model/table/column_layout_typed_relation_data.h"
@@ -29,10 +30,13 @@ private:
     void RegisterOptions();
 
     void ResetState() final;
+    virtual void MakeExecuteOptsAvailableFDInternal(){};
+    void MakeExecuteOptsAvailable() override;
     virtual void ResetStateFd() = 0;
 
 protected:
     config::InputTable input_table_;
+    config::MaxLhsType max_lhs_;
 
     /* Collection of all discovered FDs
      * Every FD mining algorithm should place discovered dependecies here. Don't add new FDs by
@@ -45,11 +49,12 @@ protected:
      * Should be overrided if custom behavior is needed
      */
     virtual void RegisterFd(Vertical lhs, Column rhs) {
-        fd_collection_.Register(std::move(lhs), std::move(rhs));
+        if (lhs.GetArity() <= max_lhs_) fd_collection_.Register(std::move(lhs), std::move(rhs));
     }
 
     virtual void RegisterFd(FD fd_to_register) {
-        fd_collection_.Register(std::move(fd_to_register));
+        if (fd_to_register.GetLhs().GetArity() <= max_lhs_)
+            fd_collection_.Register(std::move(fd_to_register));
     }
 
 public:
--- a/src/core/algorithms/fd/fdep/fd_tree_element.cpp
+++ b/src/core/algorithms/fd/fdep/fd_tree_element.cpp
@@ -243,15 +243,18 @@ void FDTreeElement::PrintDependencies(std::bitset<kMaxAttrNum>& active_path, std
     }
 }
 
-void FDTreeElement::FillFdCollection(RelationalSchema const& scheme,
-                                     std::list<FD>& fd_collection) const {
+void FDTreeElement::FillFdCollection(RelationalSchema const& scheme, std::list<FD>& fd_collection,
+                                     unsigned int max_lhs) const {
     std::bitset<kMaxAttrNum> active_path;
-    this->TransformTreeFdCollection(active_path, fd_collection, scheme);
+    this->TransformTreeFdCollection(active_path, fd_collection, scheme, max_lhs);
 }
 
 void FDTreeElement::TransformTreeFdCollection(std::bitset<kMaxAttrNum>& active_path,
                                               std::list<FD>& fd_collection,
-                                              RelationalSchema const& scheme) const {
+                                              RelationalSchema const& scheme,
+                                              unsigned int max_lhs) const {
+    if (active_path.count() > max_lhs) return;
+
     for (size_t attr = 1; attr <= this->max_attribute_number_; ++attr) {
         if (this->is_fd_[attr - 1]) {
             boost::dynamic_bitset<> lhs_bitset(this->max_attribute_number_);
@@ -268,8 +271,8 @@ void FDTreeElement::TransformTreeFdCollection(std::bitset<kMaxAttrNum>& active_p
     for (size_t attr = 1; attr <= this->max_attribute_number_; ++attr) {
         if (this->children_[attr - 1]) {
             active_path.set(attr);
-            this->children_[attr - 1]->TransformTreeFdCollection(active_path, fd_collection,
-                                                                 scheme);
+            this->children_[attr - 1]->TransformTreeFdCollection(active_path, fd_collection, scheme,
+                                                                 max_lhs);
             active_path.reset(attr);
         }
     }
--- a/src/tests/test_fd_algorithm.cpp
+++ b/src/tests/test_fd_algorithm.cpp
@@ -121,9 +121,40 @@ TYPED_TEST_P(AlgorithmTest, ConsistentRepeatedExecution) {
     }
 }
 
+namespace {
+void MaxLhsTestFun(CSVConfig config, std::list<FD> const& fds_list, config::MaxLhsType max_lhs) {
+    using namespace config::names;
+    algos::StdParamsMap verify_params = {
+            {kCsvConfig, config},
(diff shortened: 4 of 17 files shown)


src/core/algorithms/fd/dfd/dfd.cpp
```diff
@@ -20,7 +21,7 @@ void DFD::RegisterOptions() {
 }
 
 void DFD::MakeExecuteOptsAvailable() {
-    MakeOptionsAvailable({config::ThreadNumberOpt.GetName()});
+    MakeOptionsAvailable({config::MaxLhsOpt.GetName(), config::ThreadNumberOpt.GetName()});
```
polyntsov, line 24: This is a bad way to organize the code, which makes options available. `FDAlgorithm::MakeExecuteOptsAvailable` already makes `max_lhs` available, and this functionality should be reused in all other derived classes instead of overriding `MakeExecuteOptsAvailable` and once again making `max_lhs` available. 
The correct way to do this is to implement `FDAlgorithm::MakeExecuteOptsAvailable` similar to this:
```cpp
class FDAlgorithm {
//...

virtual void MakeExecuteOptsAvailableFDInternal() {};

void MakeExecuteOptsAvailable() final {
     MakeOptionsAvailable({config::MaxLhsOpt.GetName()});
     // This method can be overridden by any derived class if needed.
     // And only specific to the derived class options should be made available inside it
     MakeExecuteOptsAvailableFDInternal();
}
// ...
};
```

src/core/algorithms/fd/fd_algorithm.h
```diff
@@ -29,11 +30,12 @@ class FDAlgorithm : public Algorithm {
     void RegisterOptions();
 
     void ResetState() final;
+    virtual void MakeExecuteOptsAvailable();
     virtual void ResetStateFd() = 0;
 
 protected:
     config::InputTable input_table_;
-
+    config::MaxLhsType max_lhs_;
```
polyntsov, line 38: empty line below this

src/tests/test_fd_util.h
```diff
@@ -33,6 +33,17 @@ class AlgorithmTest : public ::testing::Test {
...
+    static algos::StdParamsMap MaxLHSParamMap(tests::TableConfig const& config,
+                                              unsigned int max_lhs_) {
+        using namespace config::names;
+        return {
+                {kTable, config.MakeInputTable()},
+                {kError, config::ErrorType{0.0}},
+                {kSeed, decltype(algos::pyro::Parameters::seed){0}},
+                {kMaximumLhs, max_lhs_},
+        };
+    }
```
polyntsov, line 45: I think we don't need a separate function for this, better add parameter with the default value to the existing `GetParamMap`

src/tests/test_fd_util.h
```diff
@@ -54,6 +65,11 @@ class AlgorithmTest : public ::testing::Test {
         return algos::CreateAndLoadAlgorithm<T>(GetParamMap(config));
     }
 
+    static std::unique_ptr<algos::FDAlgorithm> CreateMaxLHSAlgorithmInstance(
+            tests::TableConfig const& config, unsigned int max_lhs_) {
+        return algos::CreateAndLoadAlgorithm<T>(MaxLHSParamMap(config, max_lhs_));
+    }
```
polyntsov, line 71: Also seems redundant

src/tests/test_fd_algorithm.cpp
```diff
@@ -124,9 +124,21 @@ TYPED_TEST_P(AlgorithmTest, ConsistentRepeatedExecution) {
     }
 }
 
+TYPED_TEST_P(AlgorithmTest, MaxLHSOptionWork) {
+    using namespace config::names;
+    unsigned int max_lhs_ = 1;
```
polyntsov, line 129:
```suggestion
    MaxLhsType max_lhs = 1;
```

src/tests/test_fd_algorithm.cpp
```diff
@@ -124,9 +124,21 @@ TYPED_TEST_P(AlgorithmTest, ConsistentRepeatedExecution) {
...
+    auto algorithm = TestFixture::CreateMaxLHSAlgorithmInstance(kTestFD, max_lhs_);
+    algorithm->Execute();
+    std::list<FD> result_fds_list = algorithm->FdList();
+    for (auto& fd : result_fds_list) {
+        ASSERT_TRUE(fd.GetLhs().GetArity() <= max_lhs_);
+    }
```
polyntsov, line 135: This code verifies that all found fds has lhs size less than `max_lhs`, but this doesn't guarantee at all that fds are correct. Not sure, what is the best way to achieve this, but one approach is to verify all algorithm results via some fixed algorithm, for example, Pyro.
Also, I'd add more datasets, one small synthetic dataset doesn't look convincing for correctness verification.

polyntsov (changes requested):

src/tests/test_fd_algorithm.cpp
polyntsov: This is not yet fixed and is rather important

src/core/algorithms/fd/fdep/fd_tree_element.cpp
```diff
@@ -243,15 +243,16 @@ void FDTreeElement::PrintDependencies(std::bitset<kMaxAttrNum>& active_path, std
     }
 }
 
-void FDTreeElement::FillFdCollection(RelationalSchema const& scheme,
-                                     std::list<FD>& fd_collection) const {
+void FDTreeElement::FillFdCollection(RelationalSchema const& scheme, std::list<FD>& fd_collection,
+                                     unsigned int max_lhs_ = -1) const {
```
polyntsov, line 247:
```suggestion
                                     unsigned int max_lhs_ = std::numeric_limits<unsigned int>::max()) const {
```

polyntsov: And naming is incorrect btw, `_` postfix is redundant

polyntsov: Also, `max_lhs` parameter has default value here, but it's seen only in the definition of this class, which is rather useless, because this method is never called internally.
So either place the default parameter to the declaration or remove it :)

src/core/algorithms/fd/fdep/fd_tree_element.cpp
```diff
@@ -243,15 +243,16 @@ void FDTreeElement::PrintDependencies(std::bitset<kMaxAttrNum>& active_path, std
...
     std::bitset<kMaxAttrNum> active_path;
-    this->TransformTreeFdCollection(active_path, fd_collection, scheme);
+    this->TransformTreeFdCollection(active_path, fd_collection, scheme, max_lhs_);
 }
 
 void FDTreeElement::TransformTreeFdCollection(std::bitset<kMaxAttrNum>& active_path,
                                               std::list<FD>& fd_collection,
-                                              RelationalSchema const& scheme) const {
+                                              RelationalSchema const& scheme,
+                                              unsigned int max_lhs_ = -1) const {
```
polyntsov, line 255: also incorrect naming, also `std::numeric_limits::max()` instead of -1 with conversion (or remove the default value completely)

src/tests/test_fd_algorithm.cpp
```diff
@@ -120,9 +120,50 @@ TYPED_TEST_P(AlgorithmTest, ConsistentRepeatedExecution) {
...
+    auto verify_algo = algos::CreateAndLoadAlgorithm<algos::Pyro>(verify_params);
+    algorithm->Execute();
+    verify_algo->Execute();
+    std::list<FD> result_fds_list = algorithm->FdList();
+    auto verify_list = FDsToSet(verify_algo->FdList());
+    ASSERT_TRUE(CheckFdListEquality(verify_list, result_fds_list));
+    for (auto& fd : result_fds_list) {
+        ASSERT_TRUE(fd.GetLhs().GetArity() <= max_lhs_);
+    }
+}
```
polyntsov, line 161: This code is exactly the same as in the test above, it should be moved out to a separate function and reused. I'm not sure that we need two separate tests with the same logic, but on different datasets. IMO it's better to union it to a single test.  
Also, `kTestMetric` has FDs only with lhs of size <= 2, so this test is not very useful. Try to use `CIPublicHighway700.csv` for example

polyntsov (changes requested):

src/tests/test_fd_algorithm.cpp
```diff
@@ -121,9 +121,36 @@ TYPED_TEST_P(AlgorithmTest, ConsistentRepeatedExecution) {
     }
 }
 
+void MaxLhsTestFun(CSVConfig config, std::list<FD> const& fds_list, config::MaxLhsType max_lhs) {
```
polyntsov, line 124: Should be placed into anonymous namespace

src/tests/test_fd_algorithm.cpp
```diff
@@ -121,9 +121,36 @@ TYPED_TEST_P(AlgorithmTest, ConsistentRepeatedExecution) {
...
+}
+
+TYPED_TEST_P(AlgorithmTest, MaxLHSOptionWork) {
+    config::MaxLhsType max_lhs = 2;
+    auto algo = TestFixture::CreateAlgorithmInstance(kTestFD, max_lhs);
+    algo->Execute();
+    auto algo_large = TestFixture::CreateAlgorithmInstance(kCIPublicHighway700, max_lhs);
+    algo_large->Execute();
+    MaxLhsTestFun(kTestFD, algo->FdList(), max_lhs);
+    MaxLhsTestFun(kCIPublicHighway700, algo_large->FdList(), max_lhs);
```
polyntsov, line 147:
```suggestion
    config::MaxLhsType max_lhs = 2;
  
    auto algo = TestFixture::CreateAlgorithmInstance(kTestFD, max_lhs);
    algo->Execute();
    MaxLhsTestFun(kTestFD, algo->FdList(), max_lhs);
    
    auto algo_large = TestFixture::CreateAlgorithmInstance(kCIPublicHighway700, max_lhs);
    algo_large->Execute();
    MaxLhsTestFun(kCIPublicHighway700, algo_large->FdList(), max_lhs);
```

src/core/algorithms/fd/fdep/fd_tree_element.cpp
```diff
@@ -261,15 +262,15 @@ void FDTreeElement::TransformTreeFdCollection(std::bitset<kMaxAttrNum>& active_p
             }
             Vertical lhs(&scheme, lhs_bitset);
             Column rhs(&scheme, scheme.GetColumn(attr - 1)->GetName(), attr - 1);
-            fd_collection.emplace_back(FD{lhs, rhs});
+            if (lhs.GetArity() <= max_lhs) fd_collection.emplace_back(FD{lhs, rhs});
```
polyntsov, line 265: `lhs.GetArity()` is the same for all iterations of this loop and equals `active_path.count()`. There is no need to execute this loop and descend to children if `active_path.count() > max_lhs`. In such case we can return right away in the beginning of the function and by far shorten the traversal

polyntsov (approved):

polyntsov:
Squash commits into one, please

PR #334, opened 2023-12-22
Fix CSVParser issues

### LINKED ISSUE #287: Ignore the last newline in CSVParser

### LINKED ISSUE #288: CSVParser: "" is parsed incorrectly
`""` in a field should be parsed as `"`, but instead disappears.

MichaelS239:
- Add the last newline ignoring (closes #287)
- Fix CSVParser::Reset()
- Fix double quote parsing (closes #288)
- Add tests for CSVParser methods


src/core/parser/csv_parser/csv_parser.cpp
```diff
@@ -110,12 +123,26 @@ std::string CSVParser::GetUnparsedLine(unsigned long long const line_index) {
 }
 
 std::vector<std::string> CSVParser::ParseString(std::string const& s) const {
+    std::string t;
+    for (size_t index = 0; index < s.size(); index++) {
```
BUYT-1, line 127:
```suggestion
    for (std::size_t index = 0; index < s.size(); index++) {
```
nit: `std::size_t` is preferred, since `size_t` is not guaranteed to be defined by the standard

src/core/parser/csv_parser/csv_parser.cpp
```diff
@@ -110,12 +123,26 @@ std::string CSVParser::GetUnparsedLine(unsigned long long const line_index) {
...
+        if (s[index] == '\"' && index < s.size() - 1 && s[index + 1] == '\"' &&
+            (!t.size() || t[t.size() - 1] != '\\')) {
```
BUYT-1, line 129: Save `s.size()` to a variable

src/core/parser/csv_parser/csv_parser.cpp
```diff
@@ -110,12 +123,26 @@ std::string CSVParser::GetUnparsedLine(unsigned long long const line_index) {
 }
 
 std::vector<std::string> CSVParser::ParseString(std::string const& s) const {
+    std::string t;
+    for (size_t index = 0; index < s.size(); index++) {
+        if (s[index] == '\"' && index < s.size() - 1 && s[index + 1] == '\"' &&
```
BUYT-1, line 128: nit:
```suggestion
        if (s[index] == '"' && index < s.size() - 1 && s[index + 1] == '"' &&
```

src/core/parser/csv_parser/csv_parser.cpp
BUYT-1: `\` should be escaped too

src/core/parser/csv_parser/csv_parser.cpp
```diff
@@ -110,12 +123,26 @@ std::string CSVParser::GetUnparsedLine(unsigned long long const line_index) {
...
+
     std::vector<std::string> tokens;
     boost::escaped_list_separator<char> list_sep(escape_symbol_, separator_, quote_);
-    boost::tokenizer<boost::escaped_list_separator<char>> tokenizer(s, list_sep);
+    boost::tokenizer<boost::escaped_list_separator<char>> tokenizer(t, list_sep);
 
     for (auto& token : tokenizer) {
-        tokens.push_back(token);
+        if (token == "\"" && single_quote_is_empty_line) {
+            tokens.push_back("");
```
BUYT-1, line 142:
```suggestion
            tokens.emplace_back();
```
This doesn't create and move a `std::string`

BUYT-1 (changes requested):

src/core/parser/csv_parser/csv_parser.cpp
BUYT-1, line 129: What is this check needed for?

src/core/parser/csv_parser/csv_parser.cpp
BUYT-1: nit: add `#include <cstddef>`

src/core/parser/csv_parser/csv_parser.cpp
```diff
@@ -110,12 +129,30 @@ std::string CSVParser::GetUnparsedLine(unsigned long long const line_index) {
...
+
     std::vector<std::string> tokens;
     boost::escaped_list_separator<char> list_sep(escape_symbol_, separator_, quote_);
-    boost::tokenizer<boost::escaped_list_separator<char>> tokenizer(s, list_sep);
+    boost::tokenizer<boost::escaped_list_separator<char>> tokenizer(t, list_sep);
 
     for (auto& token : tokenizer) {
-        tokens.push_back(token);
+        if (token == "\"" && single_quote_is_empty_line) {
+            tokens.emplace_back("");
```
BUYT-1, line 152:
```suggestion
            tokens.emplace_back();
```

src/core/parser/csv_parser/csv_parser.h
```diff
@@ -20,6 +20,8 @@ class CSVParser : public model::IDatasetStream {
     char quote_ = '\"';
     bool has_header_;
     bool has_next_;
+    bool single_quote_is_empty_line =
+            false;  // states whether we should treat " in a field as an empty line
```
BUYT-1, line 24: Maybe this should always be true

BUYT-1 (changes requested):

src/tests/test_csv_parser.cpp
```diff
@@ -0,0 +1,95 @@
...
+    for (size_t index = 0; index < num_rows; index++) {
+        row = parser->GetNextRow();
+        actual.push_back(row);
+    }
+
+    ASSERT_THAT(actual, ContainerEq(expected)) << "Fail on " << path.c_str();
+}
+
+TEST(TestCSVParser, TestGetNextRow) {
+    CheckGetNextRow(test_data_dir / "NullEmpty.csv", ',', true, 4,
```
vs9h, line 29: it's better to use `TableConfig` from `all_tables_config.h`.
Add new tables if necessary.

src/core/parser/csv_parser/csv_parser.cpp
```diff
@@ -110,12 +129,42 @@ std::string CSVParser::GetUnparsedLine(unsigned long long const line_index) {
 }
 
 std::vector<std::string> CSVParser::ParseString(std::string const& s) const {
+    std::size_t const length = s.size();
+    std::string t;
+    for (std::size_t index = 0; index < length; ++index) {
+        if (s[index] == '\\') {  // preserve \ through boost parsing
+            t.append(2, '\\');
+        } else if (s[index] == '"') {  // preserve " through boost parsing
+            t.append("\\\"\"");
```
BUYT-1, line 138: And now quotes are useless, and we may as well use `boost::char_separator` instead of `boost::escaped_list_separator`.
`a,"a,a","a"` should be `a` `"a,a"` `a` but will be parsed as `a` `"a` `a"` `a`
I don't know. Maybe this is ok for now, but using a separate library for CSV parsing seems like the simplest option.

src/core/parser/csv_parser/csv_parser.cpp
```diff
@@ -110,12 +129,42 @@ std::string CSVParser::GetUnparsedLine(unsigned long long const line_index) {
...
+    }
+
     std::vector<std::string> tokens;
     boost::escaped_list_separator<char> list_sep(escape_symbol_, separator_, quote_);
-    boost::tokenizer<boost::escaped_list_separator<char>> tokenizer(s, list_sep);
+    boost::tokenizer<boost::escaped_list_separator<char>> tokenizer(t, list_sep);
 
     for (auto& token : tokenizer) {
-        tokens.push_back(token);
+        size_t token_length = token.size();
```
BUYT-1, line 149: nit:
```suggestion
        std::size_t const token_length = token.size();
```

src/core/parser/csv_parser/csv_parser.cpp
```diff
@@ -110,12 +129,42 @@ std::string CSVParser::GetUnparsedLine(unsigned long long const line_index) {
...
+        bool is_enclosed =
+                token_length >= 2 && token[0] == '"' &&
```
BUYT-1, line 151:
```suggestion
                token_length >= 2 && token.front() == '"' &&
```

src/core/parser/csv_parser/csv_parser.cpp
```diff
@@ -110,12 +129,42 @@ std::string CSVParser::GetUnparsedLine(unsigned long long const line_index) {
...
+                    token[index + 1] == '"') {  // transfer "" to " if the current field is enclosed
+                                                // in double-quotes
+                    new_token.push_back(token[index]);
+                    ++index;
+                }
+            } else {
+                new_token.push_back(token[index]);
+            }
+        }
+        tokens.emplace_back(new_token);
```
BUYT-1, line 167:
```suggestion
        tokens.emplace_back(std::move(new_token));
```

BUYT-1 (changes requested):

BUYT-1 (approved):

src/tests/test_csv_parser.cpp
```diff
@@ -0,0 +1,95 @@
...
+}
+
+TEST(TestCSVParser, TestReset) {
+    CheckReset(kACShippingDates, 6);
+    CheckReset(kadult, 32561);
+    CheckReset(kTestEmpty, 1);
+    CheckReset(kTest1, 20);
+}
+
+}  // namespace tests
```
vs9h, line 95: There must be line breaks at the end of all files (we also recently added #325 a line break at the end of files for csv tables, so it’s worth adding newline in `TestParse.csv`)

src/tests/test_csv_parser.cpp
```diff
@@ -0,0 +1,95 @@
+#include <cstddef>
```
polyntsov, line 1: Try to include all standard headers which define names you're using here. These are at least `vector`, `utility`, `string`

vs9h (approved):

src/tests/test_csv_parser.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+namespace tests {
+
+using ::testing::ContainerEq;
+
+class TestCSVParser : public ::testing::Test {};
+
+static void CheckGetNextRow(tests::TableConfig const& table, std::size_t num_rows,
+                            std::vector<std::vector<std::string>> const& expected) {
+    auto parser = table.MakeInputTable();
```
polyntsov, line 17: bad type deduction usage

src/tests/test_csv_parser.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include "all_tables_config.h"
+#include "parser/csv_parser/csv_parser.h"
+
+namespace tests {
+
+using ::testing::ContainerEq;
+
+class TestCSVParser : public ::testing::Test {};
+
+static void CheckGetNextRow(tests::TableConfig const& table, std::size_t num_rows,
```
polyntsov, line 15: Why do we need `num_rows` here? It seems like `expected.size()` already has this information?

src/core/parser/csv_parser/csv_parser.cpp
```diff
@@ -94,6 +103,16 @@ void CSVParser::GetLine(unsigned long long const line_index) {
...
     has_next_ = !source_.eof();
+
+    /* Check for the last newline */
+    if (has_next_) {
+        source_.get();
+        if (source_.eof()) {
+            has_next_ = false;
+        }
+        source_.unget();
+    }
```
polyntsov, line 114: A bit strange code given that right below there is an `if` with exactly the same condition, it's better to union them. Also, `get`, `eof`, `unget` sequence looks cumbersome and requires clearing of stream state in `PeekNext` which doesn't look right. Try something like this:
```cpp
    if (has_next_) {
        if (source_.peek() == std::ifstream::traits_type::eof()) {
            has_next_ = false;
            return;
        }
        GetNext();
    }
```

src/tests/test_csv_parser.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include <gtest/gtest.h>
+
+#include "all_tables_config.h"
+#include "parser/csv_parser/csv_parser.h"
+
+namespace tests {
+
+using ::testing::ContainerEq;
+
+class TestCSVParser : public ::testing::Test {};
```
polyntsov, line 13: nit: anonymous namespace

src/core/parser/csv_parser/csv_parser.cpp
```diff
@@ -94,6 +103,16 @@ void CSVParser::GetLine(unsigned long long const line_index) {
 
 void CSVParser::GetNextIfHas() {
     has_next_ = !source_.eof();
+
+    /* Check for the last newline */
```
polyntsov, line 107:
```suggestion
    // Check for the last newline
```
Be consistent in comment style

src/core/parser/csv_parser/csv_parser.cpp
```diff
@@ -77,6 +83,9 @@ void CSVParser::Reset() {
     if (has_header_) {
         SkipLine();
     }
+
+    /* For correctness of GetNextRow() after this method */
```
polyntsov, line 87:
```suggestion
    // For correctness of GetNextRow() after this method
```

polyntsov (changes requested):

src/core/parser/csv_parser/csv_parser.cpp
```diff
@@ -59,6 +60,11 @@ void CSVParser::GetNext() {
 void CSVParser::PeekNext() {
     int len = source_.tellg();
     GetNext();
+
+    if (source_.fail()) {
+        source_.clear();
+    }
```
polyntsov, line 66: This is not needed anymore?

polyntsov (approved):

PR #337, opened 2024-01-09
Add paper about MIND algorithm

Pechenux:


chernishev (changes requested):
The paper itself is fine, we looked though it a few days ago, together. However, can you please, squash commits?

chernishev (approved):

PR #338, opened 2024-01-09
Add HyMD review paper

BUYT-1:


chernishev (approved):
Its good, we have just finished looking it through together with Alexey.

PR #340, opened 2024-01-15
Update README.md with installation instructions

chernishev:


README.md
```diff
@@ -177,6 +177,17 @@ While the Python interface makes building interactive applications possible, Des
 
 You can try the deployed web version [here](https://desbordante.unidata-platform.ru/). You have to register in order to process your own datasets. Keep in mind that due to a large demand various time and memory limits are enforced: processing is aborted if they are exceeded. The source code of the web interface is kept in a separate [repo](https://github.com/vs9h/Desbordante).
 
+## Installation
+Desbordante is [available](https://pypi.org/project/desbordante/) at the Python Package Index (PyPI).
+
```
polyntsov, line 182: I think we should also add dependency list here (for now it's only Python >= 3.7)

polyntsov (approved):

PR #342, opened 2024-01-18
Add new column statistics

xJoskiy:
Add new string statistics and corresponding python bindigs.
Change the implemetation of some DataStats methods.


src/core/algorithms/statistics/data_stats.cpp
```diff
@@ -505,13 +487,119 @@ Statistic DataStats::GetMedianAD(size_t index) const {
...
+    if (all_stats_[index].vocab.HasValue()) return all_stats_[index].vocab;
+    mo::TypedColumnData const& col = col_data_[index];
+    if (col.GetTypeId() != +mo::TypeId::kString) return {};
+
+    mo::StringType string_type;
+    std::string string_data;
+    std::set<char> vocab;
+
+    for (size_t i = 0; i < col.GetNumRows(); i++) {
+        string_data = col.GetDataAsString(i);
```
BUYT-1, line 500: This will count the characters from the string representation of `NullType` (`"NULL"`) as characters in the column. There should be a check with `IsNullOrEmpty`, and then `TypedColumnData::GetValue` + `Type::GetValue<std::string>` can be used to avoid excessive copying

```suggestion
        if (col.IsNullOrEmpty(i)) continue;
        auto const& string_data = mo::Type::GetValue<std::string>(col.GetValue(i));
```

src/core/algorithms/statistics/data_stats.cpp
```diff
@@ -505,13 +487,119 @@ Statistic DataStats::GetMedianAD(size_t index) const {
...
-    size_t count = col.GetNumNulls();
+    if (col.GetTypeId() != +mo::TypeId::kString) return {};
+
+    size_t count = 0;
+    std::string string_data;
     mo::IntType int_type;
 
-    return Statistic(int_type.MakeValue(count), &int_type, false);
+    for (size_t i = 0; i < col.GetNumRows(); i++) {
+        string_data = col.GetDataAsString(i);
```
BUYT-1, line 555: Does not account for nulls (see above)

src/core/algorithms/statistics/data_stats.cpp
```diff
@@ -505,13 +487,119 @@ Statistic DataStats::GetMedianAD(size_t index) const {
...
+Statistic DataStats::GetNumberOfChars(size_t index) const {
+    mo::TypedColumnData const& col = col_data_[index];
+    if (col.GetTypeId() != +mo::TypeId::kString) return {};
+
+    size_t count = 0;
+    std::string string_data;
+    mo::IntType int_type;
+
+    for (size_t i = 0; i < col.GetNumRows(); i++) {
+        string_data = col.GetDataAsString(i);
```
BUYT-1, line 574: Does not account for nulls (see above)

src/core/algorithms/statistics/data_stats.cpp
```diff
@@ -505,13 +487,119 @@ Statistic DataStats::GetMedianAD(size_t index) const {
...
+
+    mo::TypedColumnData const& col = col_data_[index];
+    if (col.GetTypeId() != +mo::TypeId::kString) return {};
+
+    mo::DoubleType double_type;
+    mo::IntType int_type;
+    std::byte* res = double_type.Allocate();
+    Statistic num_stat = GetNumberOfChars(index);
+    std::byte const* num = num_stat.GetData();
+    std::byte const* count = double_type.MakeValueOfInt(col_data_.size());
```
BUYT-1, line 594:
```suggestion
    std::byte const* count = double_type.MakeValueOfInt(col.GetNumRows());
```
```suggestion
    std::byte const* count = double_type.MakeValueOfInt(col.GetNumRows() - col.GetNumNulls());
```
Not sure which one is better.

src/python_bindings/bindings.cpp
```diff
@@ -244,7 +243,21 @@ PYBIND11_MODULE(desbordante, module) {
...
-                 "Returns number of nulls in the column.", py::arg("index"));
+                 "Returns number of nulls in the column.", py::arg("index"))
+            .def("get_vocab", &PyDataStats::GetVocab,
+                 "Returns all the symbols of the columns as a sorted string.", py::arg("index"))
+            .def("get_number_of_non_letter_chars", &PyDataStats::GetNumberOfNonLetterChars,
+                 "Returns number of non-letter chars in a string column.", py::arg("index"))
+            .def("get_number_of_digit_chars", &PyDataStats::GetNumberOfDigitChars,
+                 "Returns number of digit chars in a string column.", py::arg("index"))
+            .def("get_number_of_lowercase_chars", &PyDataStats::GetNumberOfLowercaseChars,
+                 "Returns number of digit chars in a string column.", py::arg("index"))
```
BUYT-1, line 254:
```suggestion
                 "Returns number of lowercase chars in a string column.", py::arg("index"))
```

src/python_bindings/bindings.cpp
```diff
@@ -244,7 +243,21 @@ PYBIND11_MODULE(desbordante, module) {
...
+            .def("get_number_of_uppercase_chars", &PyDataStats::GetNumberOfUppercaseChars,
+                 "Returns number of digit chars in a string column.", py::arg("index"))
```
BUYT-1, line 256:
```suggestion
                 "Returns number of uppercase chars in a string column.", py::arg("index"))
```

BUYT-1 (changes requested):

src/core/algorithms/statistics/data_stats.h
```diff
@@ -107,7 +109,21 @@ class DataStats : public Algorithm {
     // Returns meadian absolute deviation in the column if it's numeric.
     Statistic GetMedianAD(size_t index) const;
     // Returns number of nulls in the column.
-    Statistic GetNumNulls(size_t index) const;
+    size_t GetNumNulls(size_t index) const;
+    // Returns all the symbols of the columns as a sorted string.
```
BUYT-1, line 113: better wording:
```suggestion
    // Returns all distinct symbols of the column as a sorted string.
```

BUYT-1 (changes requested):

src/core/algorithms/statistics/data_stats.cpp
```diff
@@ -505,13 +487,122 @@ Statistic DataStats::GetMedianAD(size_t index) const {
...
+    std::byte const* res = string_type.MakeValue(temp);
+
+    return Statistic(res, &string_type, false);
+}
+
+Statistic DataStats::GetNumberOfNonLetterChars(size_t index) const {
+    if (all_stats_[index].num_non_letter_chars.HasValue())
+        return all_stats_[index].num_non_letter_chars;
+
+    auto pred = [](int symbol) { return !std::isalpha(symbol); };
```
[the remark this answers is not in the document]
xJoskiy: You mean to put a predicate definition in a macro like this?
`#define predicate(function) [](unsigned char c) { function(c); }`
Or what

src/core/algorithms/statistics/data_stats.cpp
```diff
@@ -505,13 +487,122 @@ Statistic DataStats::GetMedianAD(size_t index) const {
...
+    for (size_t i = 0; i < col.GetNumRows(); i++) {
+        if (col.IsNullOrEmpty(i)) continue;
+        auto const& string_data = mo::Type::GetValue<std::string>(col.GetValue(i));
+        vocab.insert(string_data.begin(), string_data.end());
+    }
+    std::string temp(vocab.begin(), vocab.end());
+
+    std::byte const* res = string_type.MakeValue(temp);
+
+    return Statistic(res, &string_type, false);
```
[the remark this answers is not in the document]
xJoskiy: But we don't actually initialize members of Statistic object with Type pointer passed, the new object of the same Type we pass is being created whenever we call a ctor.
If we just return string, so we should return an empty one in case column isn't string and it is kinda weird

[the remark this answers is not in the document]
BUYT-1: You're right

BUYT-1 (approved):

PR #344, opened 2024-01-20
Reorganize Python bindings

BUYT-1:


src/core/algorithms/fd/mining_algorithms.h
```diff
@@ -0,0 +1,13 @@
...
+#include "algorithms/fd/depminer/depminer.h"
+#include "algorithms/fd/dfd/dfd.h"
+#include "algorithms/fd/fastfds/fastfds.h"
+#include "algorithms/fd/fd_mine/fd_mine.h"
+#include "algorithms/fd/fdep/fdep.h"
+#include "algorithms/fd/fun/fun.h"
+#include "algorithms/fd/hyfd/hyfd.h"
+#include "algorithms/fd/pyro/pyro.h"
+#include "algorithms/fd/tane/tane.h"
+#include "algorithms/statistics/data_stats.h"
```
vs9h, line 13: why we include `data_stats.h` here?

src/python_bindings/bind_fd.cpp
```diff
@@ -0,0 +1,98 @@
...
+    return fd.GetRhs().GetIndex();
+}
+
+std::string ToLongString(FD const& fd) {
+    std::stringstream ss;
+    ss << fd.GetLhs().ToString();
+    ss << " -> ";
+    ss << fd.GetRhs().GetName();
+    return ss.str();
+}
```
vs9h, line 30: Why in some cases do we write the implementation of some methods руку? I think it's better for consistency to add methods to the appropriate models and pass a pointer to the class method. 

The question applies only to those methods that do not involve bindings-related classes.

src/python_bindings/bind_fd.cpp
```diff
@@ -0,0 +1,98 @@
...
+    using namespace algos;
+
+    auto fd_module = main_module.def_submodule("fd");
+    py::class_<FD>(fd_module, "FD")
+            .def("__str__", ToLongString)
+            .def("to_long_string", ToLongString)
+            .def("to_short_string",
+                 [](FD const& fd) {
+                     std::stringstream stream;
+                     stream << "( ";
```
vs9h, line 56: Why do we use a parenthesis in this case? `ToLongString` method, for example, returns string with square brackets.

vs9h (changes requested):
I agree with the proposed changes and I left a few minor comments/questions.

And I also have a suggestion for organizing files inside the `src/python_bindings` directory. This directory now contains quite a lot of files, and the more new primitives there are, the larger and more inconvenient it will be to look at the files in this directory. Maybe it's worth grouping some files into subdirectories? For example, group by primitive (ac/ar/fd) or by their meaning (dependency/verification/..). 

I would also suggest adding subdirectories for the remaining files (dataframe related files and py-types related files), but this is less critical.

src/python_bindings/bind_fd.cpp
BUYT-1: No reason other than it being the old format.

vs9h (approved):

PR #346, opened 2024-01-30
Construct `RelationalSchema` with `IDatasetStream`

wildsor:


src/core/algorithms/fd/pyrocommon/model/pli_cache.cpp
```diff
@@ -29,17 +29,17 @@ PLICache::PLICache(ColumnLayoutRelationData* relation_data, CachingMethod cachin
...
+    for (auto& column : relation_data->GetSchema()->GetColumns()) {
+        index_->Put(static_cast<Vertical>(column),
+                    relation_data->GetColumnData(column.GetIndex()).GetPliOwnership());
     }
 }
 
 PLICache::~PLICache() {
-    for (auto& column_ptr : relation_data_->GetSchema()->GetColumns()) {
+    for (auto const& column : relation_data_->GetSchema()->GetColumns()) {
         // auto PLI =
```
Firsov62121, line 40:
```suggestion
```

src/core/algorithms/fd/pyrocommon/model/pli_cache.cpp
```diff
@@ -29,17 +29,17 @@ PLICache::PLICache(ColumnLayoutRelationData* relation_data, CachingMethod cachin
...
-        index_->Remove(static_cast<Vertical>(*column_ptr));
-        // relation_data_->GetColumnData(column_ptr->getIndex()).getPLI(std::move(PLI));
+        index_->Remove(static_cast<Vertical>(column));
+        // relation_data_->GetColumnData(column.getIndex()).getPLI(std::move(PLI));
```
Firsov62121, line 42:
```suggestion
```

src/core/model/table/relational_schema.cpp
```diff
@@ -3,49 +3,52 @@
...
+    column_names.reserve(number_of_columns);
+    for (model::ColumnIndex i = 0; i < number_of_columns; ++i) {
+        column_names.push_back(table.GetColumnName(i));
+    }
+    return std::make_unique<RelationalSchema>(table.GetRelationName(), std::move(column_names));
 }
 
-// TODO: В оригинале тут что-то непонятное + приходится пересоздавать empty_vertical_ -- тут
-// должен быть unique_ptr, тк создаём в остальных случаях новую вершину и выдаём наружу с овнершипом
+// TODO: В оригинале тут что-то непонятное
```
Firsov62121, line 35: Delete or insert a link to the original.

src/core/algorithms/cfd/model/cfd_relation_data.cpp
```diff
@@ -49,17 +49,23 @@ std::unique_ptr<CFDRelationData> CFDRelationData::CreateFrom(model::IDatasetStre
         return CFDRelationData::CreateFrom(parser, c_sample, r_sample);
     }
 
+    unsigned num_columns = parser.GetNumberOfColumns();
```
Firsov62121, line 52:
```suggestion
    size_t const num_columns = parser.GetNumberOfColumns();
```

Firsov62121: And change the signature of `AddNewItemsInFullTable`.

src/core/algorithms/cfd/cfd_discovery.h
```diff
@@ -26,8 +26,8 @@ class CFDDiscovery : public Algorithm {
 protected:
     config::InputTable input_table_;
 
-    unsigned columns_number_;
-    unsigned tuples_number_;
+    size_t columns_number_;
+    size_t tuples_number_;
```
vs9h, line 30: Why we use `size_t` for those fields? We have type aliases `model::ColumnIndex` and `model::TupleIndex` (they have type `unsigned int`)

BUYT-1: These are not indices, why not?

BUYT-1: I did this for easier overloading in `AddNewItemsInFullTable`. And these are not indices.

vs9h: It seems reasonable to me to use the same type for the number of columns as for the column index.
Because of this discrepancy in types, we have to write (line 55 in the `CreateFrom` function) a rather strange cast, which we probably would like to avoid:
```cpp
for (AttributeIndex i = 0; static_cast<size_t>(i) < num_columns; ++i) {
```
I'm sure it's better to change the type of the `num_columns` and `columns_number` to `AttributeIndex` to avoid this weird casting. I would prefer to fix this, but I don't insist.

BUYT-1: [May be related](https://github.com/Desbordante/desbordante-core/pull/346#discussion_r1564526352)
@Firsov62121

BUYT-1: I tried it, and it just leads to a bunch of other casts, not solving anything.

wildsor:
Superseded by #796

PR #349, opened 2024-02-12
Add bindings, cli support and a usage example for CFD discovery.

VanyaVolgushev:


src/python_bindings/cfd/bind_cfd.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+
+    py::class_<RawCFD>(cfd_module, "CFD")
+            .def("__str__", &RawCFD::ToString)
+            .def_property_readonly("lhs_items", &RawCFD::GetLhs)
+            .def_property_readonly("rhs_item", &RawCFD::GetRhs);
+
+
+    BindPrimitive<FDFirstAlgorithm>(cfd_module, py::overload_cast<>(&FDFirstAlgorithm::GetCfds,
+            py::const_), "CfdAlgorithm", "get_cfds", {"FDFirst"});
```
BUYT-1, line 30:
```suggestion
    BindPrimitive<FDFirstAlgorithm>(cfd_module, &CFDDiscovery::GetCfds, "CfdAlgorithm", "get_cfds", {"FDFirst"});
```

examples/mining_cfd.py
```diff
@@ -0,0 +1,131 @@
...
+            support -= 1
+            continue
+    return support
+
+
+def validate_cfd(lhs, rhs):
+    row_colors = ['red'] * len(table.index)
+    cfd_lhs_var_attributes = find_var_attributes(lhs)
+    num_rows_satisfy_cfd = 0
+    support = mark_unsupported(lhs, row_colors)
```
BUYT-1, line 66: Return the row indices

BUYT-1 (changes requested):
More to come

examples/mining_cfd.py
```diff
@@ -0,0 +1,131 @@
+import desbordante
+import pandas
+
+TABLEPATH = 'examples/datasets/university_fd.csv'
+MINIMUM_SUPPORT = 8
+MINIMUM_CONFIDENCE = 0.7
+MAXIMUM_LHS_COUNT = 3
+
+
+table = pandas.read_csv(TABLEPATH)
```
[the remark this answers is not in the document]
VanyaVolgushev: I changed `table` to be a local variable, totally right on that one, but I thought the capitalized globals would be like input parameters. The user could open the file and see right away at the top of the file that those are the CFD discovery parameters I picked as a good example but they can be changed. I guess I should add a comment stating that then. Or I could remove the capitalized globals and recieve them from the console with my values being the default. Just thought my version was more neat and more like the rest of the examples.

examples/mining_cfd.py
```diff
@@ -0,0 +1,131 @@
...
+        print(color_codes[row_colors[rowNum]], no_header_string, color_codes['default'])
+    print("lhs count: ", len(cfd.lhs_items))
+    print("support: ", str(support), "\033[1;42m \033[1;41m \033[1;49m")
+    print("confidence: \033[1;32m", num_rows_satisfy_cfd, "\033[1;37m/", support,
+          " = ", str(num_rows_satisfy_cfd/support)[0:6])
+
+
+if __name__ == '__main__':
+    algo = desbordante.cfd.algorithms.Default()
+    algo.load_data(table=(TABLEPATH, ',', True))
```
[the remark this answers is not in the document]
VanyaVolgushev: I'm not sure I get what you mean. `table` isn't explicitly used in main but it does get used in the other functions.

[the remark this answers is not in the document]
BUYT-1: You can pass the dataframe as the table (`algo.load_data(table=df)`). Since the dataframe is created anyway, we may as well use it.

examples/mining_cfd.py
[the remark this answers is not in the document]
BUYT-1: You may have forgotten to push

examples/mining_cfd.py
```diff
@@ -0,0 +1,127 @@
...
+#   key: a string representation of a dataframe with certain attribute values
+#   value: indices of all rows with those attribute values
+def make_lhs_to_row_nums(cfd_lhs_var_attributes, supported_rows, table):
+    lhs_to_row_nums = {}
+    for row_num in range(0, len(table.index)):
+        row_dataframe = table.iloc[[row_num], :]
+        if row_num not in supported_rows:
+            continue
+        if row_dataframe.iloc[0, cfd_lhs_var_attributes].to_string() in lhs_to_row_nums:
+            lhs_to_row_nums[row_dataframe.iloc[0, cfd_lhs_var_attributes].to_string()].append(row_num)
```
BUYT-1, line 45: Use tuples of values as keys in your PLI. tuple(table.iloc[row, lhs]). Same for rhs

examples/mining_cfd.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+def validate_cfd(lhs, rhs, table):
+    cfd_lhs_var_attributes = find_var_attributes(lhs)
+    supported_rows = get_supported(lhs, table)
+    rows_satisfying_cfd = []
+    lhs_to_row_nums = make_lhs_to_row_nums(cfd_lhs_var_attributes, supported_rows, table)
+
+    # mark rows with identical lhs with green color if they have the most frequent rhs
+    for string_lhs in lhs_to_row_nums:
```
BUYT-1, line 68: Use [dict.items](https://docs.python.org/3.8/library/stdtypes.html#dict.items), same for rhs

examples/mining_cfd.py
```diff
@@ -0,0 +1,127 @@
...
+            if row_dataframe.iloc[0, [rhs.attribute]].to_string() in rhs_to_row_nums:
+                rhs_to_row_nums[row_dataframe.iloc[0, [rhs.attribute]].to_string()].append(row_num)
+            else:
+                rhs_to_row_nums[row_dataframe.iloc[0, [rhs.attribute]].to_string()] = [row_num]
+
+        best_count = 0
+        most_frequent_rhs = None
+        for string_rhs in rhs_to_row_nums:
+            if len(rhs_to_row_nums[string_rhs]) > best_count:
+                best_count = len(rhs_to_row_nums[string_rhs])
```
BUYT-1, line 82: Store the cluster itself as well

examples/mining_cfd.py
```diff
@@ -0,0 +1,127 @@
...
+                most_frequent_rhs = string_rhs
+        for row_num in rhs_to_row_nums[most_frequent_rhs]:
+            rows_satisfying_cfd.append(row_num)
```
BUYT-1, line 85:
```suggestion
        rows_satisfying_cfd.update(most_frequent_cluster)
```

BUYT-1 (changes requested):
The bindings are okay.

examples/mining_cfd.py
BUYT-1: UPD: should be `tuple(table.iloc[row][lhs])`

examples/mining_cfd.py
```diff
@@ -0,0 +1,125 @@
...
+            true_array = np.full(
+                shape=table_height,
+                fill_value=True,
+                dtype=np.bool_
+            )
+            return pandas.Series(true_array)
+        columnName = table.columns[cfd_item.attribute]
+        return table[columnName].astype(str).str.lower() == cfd_item.value.lower()
+
+    return reduce(operator.and_, map(get_mask, pattern))
```
BUYT-1, line 48:
```suggestion
    table_height = table.shape[0]
    true_array = np.full(shape=table_height, fill_value=True, dtype=np.bool_)
    mask = pandas.Series(true_array)
    for cfd_item in pattern:
        item_value = cfd_item.value
        if item_value is None:
            continue
        mask &= table.columns[cfd_item.attribute].astype(str).str.lower() == item_value.lower()
    return mask
```
But if you really want to write it in a functional style, use [`filter`](https://docs.python.org/3/library/functions.html#filter) and an initializer for `reduce` for completeness.

examples/mining_cfd.py
```diff
@@ -0,0 +1,125 @@
...
+        for row_num in row_nums_with_lhs:
+            key = table.iloc[row_num, rhs.attribute]
+            rhs_to_row_nums[key].append(row_num)
+
+        best_count = 0
+        row_nums_with_most_frequent_rhs = None
+        for rhs_key, row_nums_with_rhs in rhs_to_row_nums.items():
+            if len(row_nums_with_rhs) > best_count:
+                best_count = len(row_nums_with_rhs)
+                row_nums_with_most_frequent_rhs = row_nums_with_rhs
```
BUYT-1, line 82:
```suggestion
        row_nums_with_most_frequent_rhs = max(rhs_to_row_nums.values(), key=len)
```

examples/mining_cfd.py
```diff
@@ -0,0 +1,125 @@
...
+            color_code = GREEN_BG_CODE
+        else:
+            color_code = RED_BG_CODE
+        print(color_code, row, DEFAULT_BG_CODE)
+
+    support = len(supported_rows)
+    num_rows_satisfy_cfd = len(rows_satisfying_cfd)
+    print("lhs count: ", len(cfd.lhs_items))
+    print("support: ", support, "\033[1;42m \033[1;41m \033[1;49m")
+    print("confidence: \033[1;32m", num_rows_satisfy_cfd, "\033[1;37m/", support,
```
BUYT-1, line 109: nit: hard-coded values (some of them have constants already), something like `f'{GREEN_BG_CODE} {RED_BG_CODE} {DEFAULT_BG_CODE}'` is preferred (though I'm not sure if this is supposed to be here, doesn't this just change the background where there is no text?).

examples/mining_cfd.py
```diff
@@ -0,0 +1,125 @@
...
+          " = ", format(num_rows_satisfy_cfd/support, '.4f'))
```
BUYT-1, line 110: nit:
```suggestion
          " = ", format(num_rows_satisfy_cfd / support, '.4f'))
```

examples/mining_cfd.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+if __name__ == '__main__':
+    tableDF = pandas.read_csv(TABLEPATH)
```
BUYT-1, line 114: nit:
```suggestion
    table_df = pandas.read_csv(TABLE_PATH)
```

examples/mining_cfd.py
```diff
@@ -0,0 +1,125 @@
...
+import numpy as np
+
+# CFD Discovery parameters:
+TABLEPATH = 'examples/datasets/play_tennis.csv'
+MINIMUM_SUPPORT = 8
+MINIMUM_CONFIDENCE = 0.7
+MAXIMUM_LHS_COUNT = 3
+
+
+GREEN_BG_CODE = '\033[1;42m'
```
BUYT-1, line 18: nit:
```suggestion
GREEN_BG_CODE = '\033[1;42m'
```

examples/mining_cfd.py
```diff
@@ -0,0 +1,125 @@
+"""CFD mining example with visualization using Desbordante algorithms."""
+
+from collections import defaultdict
+from functools import reduce
+import operator
+
+import desbordante
+import pandas
+import numpy as np
```
BUYT-1, line 9: nit (see PEP 8):
```suggestion
import numpy as np
import pandas
```

examples/mining_cfd.py
```diff
@@ -0,0 +1,125 @@
...
+GREEN_BG_CODE = '\033[1;42m'
+RED_BG_CODE = '\033[1;41m'
+DEFAULT_BG_CODE = '\033[1;49m'
+
+
+def items_to_indices(items):
+    return [item.attribute for item in items]
+
+
+def find_var_attributes(items):
```
BUYT-1, line 27: Unused

BUYT-1 (changes requested):

examples/mining_cfd.py
VanyaVolgushev: `mask &= table.columns[cfd_item.attribute].astype(str).str.lower() == item_value.lower()` doesn't work, I left it as it was for now, with "column_name" in my latest commit

examples/mining_cfd.py
VanyaVolgushev: purely visual thing to show that support is the amount of red and green lines, fixed to be global constants

examples/mining_cfd.py
BUYT-1: right, I meant `table[table.columns[cfd_item.attribute]]`, but splitting is better.

BUYT-1 (changes requested):
Small things

BUYT-1 (approved):

PR #369, opened 2024-03-08
Add CORDS

egshnov:
Implements the cords algorithm for mining of soft functional dependencies and correlations


chernishev (changes requested):
While I haven't reviewed the code, I've looked through the files and would like to ask some naming changes. In this PR you are adding the support of Soft Functional Dependency primitive. More precisely, you are adding mining algorithm which is called CORDS. This should be appropriately reflected in the naming scheme.

What I have noticed so far:
1) core/algorithms/ should contain not cords folder, but the sfd one
2) the contents of core/algorithms/cords should be moved into core/algorithms/sfd/cords
3) tests/test_cords.cpp should be renamed into test_sfd.cpp --- not sure, since we have test_faida and test_apriori

src/core/algorithms/sfd/cords/cords.h
```diff
@@ -0,0 +1,78 @@
...
+    long double minimum_cardinalty_;
+    long double max_diff_vals_proportion_;
+    long double min_sfd_strength_measure_;
+    long double min_skew_threshold_;
+    long double min_structural_zeroes_amount_;
+    long double max_false_positive_probability_;
+    long double delta_;
+    size_t max_amount_of_categories_;
+    std::vector<bool> is_skewed;
+    std::vector<long long> domains;
```
vs9h, line 31: underscore

src/core/algorithms/sfd/cords/cords_sample.cpp
```diff
@@ -0,0 +1,51 @@
...
+            map_lhs.emplace(data[l].GetDataAsString(row));
+        }
+        if (map_rhs.find(data[r].GetDataAsString(row)) == map_rhs.end()) {
+            map_rhs.emplace(data[r].GetDataAsString(row));
+        }
+
+        std::string tmp = data[l].GetDataAsString(row) + data[r].GetDataAsString(row);
+        if (map_cardinality.find(tmp) == map_cardinality.end()) {
+            map_cardinality.emplace(tmp);
+        }
```
vs9h, line 32:
```suggestion
        map_lhs.insert(data[l].GetDataAsString(row));
        map_rhs.insert(data[r].GetDataAsString(row));
        map_cardinality.insert(data[l].GetDataAsString(row) + data[r].GetDataAsString(row));
    }
```

vs9h (changes requested):
I left the first part of the review. I will most likely do the second part after the comments I left are fixed.

Also, please check how correctly the `include`'s are used (remove those that are not used and add those that are used). I added some comments, but you can probably find more.

Also, please fix the names of the commits (make them less abstract):
- Add string constants for CORDS (say about names and descriptions for CORDS options)
- Make CORDS visible (Integrate CORDS into Desbordante codebase)
- Add lineitem for tests (Add lineitem csv config test)
- Add tests (for ...) 
- Add bindings (for ...)

Also, regarding comments: if you corrected a comment and are confident in the correction, then just press resolved; if you're not sure, just leave the comment unresolved (i will click resolved); If you have an objection, write it down.
Also, after you fix the comments and the pull request is ready for a second review, click on the appropriate button to request it

src/core/config/names.h
```diff
@@ -54,4 +54,13 @@ constexpr auto kIgnoreConstantCols = "ignore_constant_cols";
...
 constexpr auto kGfdData = "gfd";
 constexpr auto kMemLimitMB = "mem_limit";
+constexpr auto kOnlySfd = "only_sfd";
+constexpr auto kMinCard = "min_cardinality";
+constexpr auto kMaxDiffValsProportion = "max_different_values_proportion";
+constexpr auto kMinSfdStrengthMeasure = "min_sfd_strength";
+constexpr auto kMinSkewThreshold = "min_skew_threshold";
+constexpr auto kMinStructuralZeroesAmount = "min_st_zeroes_amount";
+constexpr auto kMaxFalsePositiveProbability = "max_false_positive_probability";
+constexpr auto kCordsDelta = "cords_delta";
```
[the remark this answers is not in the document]
chernishev: > Why we use `cords` prefix for variable name and for string literal?

It may be a good idea, since there are like 5 parameters for this algorithm.

src/core/algorithms/sfd/cords/cords.h
```diff
@@ -0,0 +1,78 @@
...
+#include "algorithm.h"
+#include "config/tabular_data/input_table_type.h"
+#include "cords_sample.h"
+#include "frequency_handler.h"
+#include "model/table/column_layout_typed_relation_data.h"
+#include "model/types/types.h"
+
+namespace algos {
+
+class Cords : public Algorithm {
```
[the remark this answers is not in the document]
chernishev: > Why we don't introduce new class `SFDAlgorithm`?

While I'm not aware of any other algorithm for SDF discovery, it may be a good idea to introduce this additional class.

src/python_bindings/cords/bind_cords.cpp
```diff
@@ -0,0 +1,27 @@
...
+#include "py_util/bind_primitive.h"
+
+namespace {
+namespace py = pybind11;
+}  // namespace
+
+namespace python_bindings {
+void BindCords(py::module_& main_module) {
+    using namespace algos;
+    auto cords_module = main_module.def_submodule("cords");
```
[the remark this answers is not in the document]
chernishev: I concur, missed this one in my review.

src/tests/test_cords.cpp
```diff
@@ -0,0 +1,165 @@
...
+TEST_F(CordsAlgorithmTest, CIPublicHighway10k) {
+    auto a = CreateCordsInstance(kCIPublicHighway10k, true);
+    a->Execute();
+    AssertVectors(a->GetSoftKeys(), {0, 2});
+    AssertVectors(a->GetTrivialColumns(),
+                  {1,  7,  8,  9,  10, 14, 15, 16, 17, 7,  8,  9,  10, 14, 15, 16, 17, 7,
+                   8,  9,  10, 14, 15, 16, 17, 7,  8,  9,  10, 14, 15, 16, 17, 7,  8,  9,
+                   10, 14, 15, 16, 17, 14, 15, 16, 17, 14, 15, 16, 17, 14, 15, 16});
+    AssertVectors(a->GetSfd(), {{11, 3}, {11, 4}, {11, 5}, {11, 6}, {11, 12}});
+}
```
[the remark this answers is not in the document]
vs9h: I wrote, but forgot to attach a comment with motivation
1) I don't like using default parameters. When reading the tests themselves, which call functions that use these default parameters, it is awkward to go back to the function definition to find out the default values.
2) I don’t like such functions, since calling them turns into passing a set of numbers, the meaning of which is difficult to understand (how to map values to param names).
3) We have to duplicate parameters with their types in `GetParamMap` and `CreateCordsInstance`
4) At the moment, we need the following from these functions: we want to be able to set a config for several tests and use it. But also later we want to be able to add other tests that can use other configs.

Moreover, these problems may not be obvious when we have several parameters, but in this case there are too many of them.
Therefore, I would like to propose the following solution: let's separate the config into a structure (there is no table in the config, since we want to be able to reuse the same test config for different tables). And we also use Designated initializers when creating the config, i.e. We explicitly indicate the names of the fields.
Thanks to this, we solve the problems that I wrote about above.

src/core/config/names.h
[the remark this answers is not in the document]
vs9h: On the contrary, I am in favor of removing the prefix here too. We already have information about what options a particular algorithm uses, since we can call the corresponding method on the algorithm. Therefore, we usually omit the information that this is an option for a specific algorithm.

src/core/algorithms/sfd/cords/cords.cpp
```diff
@@ -0,0 +1,289 @@
...
+                continue;
+            }
+            if (handler.GetColumnCardinality(ind2) >= (1 - minimum_cardinalty_) * rows) {
+                soft_keys_.push_back(ind2);
+                is_soft_or_trivial[ind2] = true;
+                continue;
+            }
+            if (handler.GetColumnCardinality(ind2) == 1) {
+                trivial_columns_.push_back(ind2);
+                is_soft_or_trivial[ind1] = true;
```
vs9h, line 223: is it true, that `ind1` here is a typo?

src/core/algorithms/sfd/cords/cords.cpp
```diff
@@ -0,0 +1,289 @@
...
+                continue;
+            }
```
vs9h, line 225: try to avoid code duplication from 198 to 210 lines (extract to separate function/lambda)

src/core/config/names.h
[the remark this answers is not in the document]
chernishev: > On the contrary, I am in favor of removing the prefix here too. We already have information about what options a particular algorithm uses, since we can call the corresponding method on the algorithm. Therefore, we usually omit the information that this is an option for a specific algorithm.

Ok, seems reasonable. Lets do like you suggest.

src/core/algorithms/sfd/sfd_algorithm.cpp
```diff
@@ -0,0 +1,302 @@
...
+    if (handler.GetColumnCardinality(col_ind) >= (1 - minimum_cardinality_) * row_count) {
+        soft_keys_.push_back(col_ind);
+        is_soft_or_trivial[col_ind] = true;
+        return true;
+    }
+    if (handler.GetColumnCardinality(col_ind) == 1) {
+        trivial_columns_.push_back(col_ind);
+        is_soft_or_trivial[col_ind] = true;
+        return true;
+    }
```
vs9h, line 242: From this code it follows that the soft column cannot be trivial. Do we really want this?

src/core/algorithms/sfd/sfd_algorithm.cpp
```diff
@@ -0,0 +1,302 @@
...
+    Init(column_count);
+
+    auto start_time = std::chrono::high_resolution_clock::now();
+    FrequencyHandler handler(data, column_count, max_amount_of_categories_);
+
+    for (model::ColumnIndex ind1 = 0; ind1 < column_count - 1; ind1++) {
+        if (IsTrivialCase(ind1, is_soft_or_trivial, handler, row_count)) continue;
+
+        for (model::ColumnIndex ind2 = ind1 + 1; ind2 < column_count; ind2++) {
+            if (IsTrivialCase(ind2, is_soft_or_trivial, handler, row_count)) continue;
```
vs9h, line 261: I would like to suggest getting rid of the repeated calculation of `IsTrivialCase` for the case when the column is not trivial and not soft.
To do this, you can fill the vector `is_soft_or_trivial` before the loop.
```cpp
bool SFDAlgorithm::IsTrivialCase(model::ColumnIndex col_ind,
                                 FrequencyHandler const &handler, size_t row_count) {
    if (handler.GetColumnCardinality(col_ind) >= (1 - minimum_cardinality_) * row_count) {
        soft_keys_.push_back(col_ind);
        return true;
    }
    if (handler.GetColumnCardinality(col_ind) == 1) {
        trivial_columns_.push_back(col_ind);
        return true;
    }
    return false;
}
```

```cpp
unsigned long long SFDAlgorithm::ExecuteInternal() {
    ...
    std::vector<bool> is_soft_or_trivial;
    is_soft_or_trivial.reserve(column_count);
    for (model::ColumnIndex col_ind = 0; col_ind != column_count; ++col_ind)
        is_soft_or_trivial.push_back(IsTrivialCase(col_ind, handler, row_count));

    for (model::ColumnIndex ind1 = 0; ind1 < column_count - 1; ind1++) {
        if (is_soft_or_trivial[ind1]) continue;

        for (model::ColumnIndex ind2 = ind1 + 1; ind2 < column_count; ind2++) {
            if (is_soft_or_trivial[ind2]) continue;
    ...
```

src/core/config/descriptions.h
```diff
@@ -97,4 +97,31 @@ constexpr auto kDIgnoreConstantCols =
...
+constexpr auto kDMinSkewThreshold =
+        "(1-val) *|R| (where |R| is amount of rows in table R) is the minimum sum of the "
+        "frequencies of occurrence of the most frequent"
+        "values of the column in the table R, indicating that the distribution of values in the "
+        "column is skewed. Value lies in (0, 1).";
+constexpr auto kDMinStructuralZeroesAmount =
+        "val*d1*d2 is the minimum number of structural zeros in the contingency table indicating "
+        "that the columns are correlated. d1, d2 - the number of different values in columns C1, "
+        "C2, "
+        "respectively. Value lies in (0, 1).";
```
vs9h, line 119:
```suggestion
        "C2, respectively. Value lies in (0, 1).";
```

src/core/algorithms/sfd/sfd_algorithm.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+            auto [col_i, col_k] =
+                    SwitchIndicesIfNecessary(ind1, ind2, handler.GetColumnCardinality(ind1),
+                                             handler.GetColumnCardinality(ind2));
```
vs9h, line 265: let's also make code more readable (function name `SwitchIndicesIfNecessary` is not intuitive):
```cpp
    auto sort_indices_by_cardinality = [&handler](model::ColumnIndex ind1,
                                                  model::ColumnIndex ind2) -> ColumnIndexPair {
        if (handler.GetColumnCardinality(ind2) > handler.GetColumnCardinality(ind1)) {
            return {ind2, ind1};
        }
        return {ind1, ind2};
    };
    ...

    for (model::ColumnIndex ind1 = 0; ind1 < column_count - 1; ind1++) {
        ...

        for (model::ColumnIndex ind2 = ind1 + 1; ind2 < column_count; ind2++) {
            ...
            auto [col_i, col_k] = sort_indices_by_cardinality(ind1, ind2);
```
And we use `SwitchIndicesIfNecessary`  only here, so we can define this function as lambda right here.

src/core/algorithms/sfd/sfd_algorithm.h
```diff
@@ -0,0 +1,94 @@
...
+                                    std::vector<long double> const &n_i,
+                                    std::vector<long double> const &n_j, model::ColumnIndex col_i,
+                                    model::ColumnIndex col_k, long double sample_size) const;
+
+    void Init(model::ColumnIndex columns);
+
+    std::pair<model::ColumnIndex, model::ColumnIndex> SwitchIndicesIfNecessary(
+            model::ColumnIndex ind1, model::ColumnIndex ind2, size_t card1, size_t card2);
+
+    bool DetectAndEmplaceSfd(Sample const &smp);
```
vs9h, line 68: for consistency with other code
```suggestion
    bool DetectAndEmplaceSFD(Sample const &smp);
```

src/core/algorithms/sfd/frequency_handler.h
```diff
@@ -0,0 +1,30 @@
+#pragma once
```
vs9h, line 1: nit: put here and in other files empty line after `#pragma once`

vs9h (changes requested):

src/core/algorithms/sfd/sfd_algorithm.cpp
```diff
@@ -0,0 +1,302 @@
...
+    domains_.clear();
+    soft_keys_.clear();
+    trivial_columns_.clear();
+    correlations_.clear();
+    unrelated_.clear();
+    sfd_.clear();
+}
+
+void SFDAlgorithm::LoadDataInternal() {
+    typed_relation_ = model::ColumnLayoutTypedRelationData::CreateFrom(*input_table_, false);
```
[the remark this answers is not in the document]
egshnov: > Can you explain why we use `false` here?
Actually original paper doesn’t give any guidelines whether two distinct occurrences of a NULL should be considered the same value. In case sofd fd between columns is not present the next step will be test for correlation. Considering NULLS as distinct values will avoid skew of distribution caused by NULLS  and they will be filtered out before the correlation test if distribution is skewed, so as i was writing this piece of code it seemed reasonable to me to view NULLS as distinct. But now i am not sure about this because such approach affects detection of soft fds as it uses amount of distinct values in a column. Maybe it would be better to make new param for the algorithm `is_null_eq_null` to let the user decide whether NULLS are equal or not.

src/core/algorithms/sfd/sfd_algorithm.cpp
```diff
@@ -0,0 +1,302 @@
...
+        }
+    }
+}
+
+size_t SFDAlgorithm::Category(model::ColumnIndex col_ind, std::string const &val, size_t domain,
+                              bool skew, FrequencyHandler const &handler) const {
+    if (skew) {
+        return handler.GetValueOrdinalNumberAtColumn(val, col_ind);
+    }
+    return std::hash<std::string>{}(val) % domain;  //+1;
```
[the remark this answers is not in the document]
egshnov: debug leftovers

src/core/algorithms/sfd/sfd_algorithm.cpp
egshnov: According to the pseudo-code from the original paper soft column cannot be trivial.

src/core/algorithms/sfd/sfd_algorithm.h
```diff
@@ -0,0 +1,94 @@
...
+#include "config/tabular_data/input_table_type.h"
+#include "frequency_handler.h"
+#include "index_pair.h"
+#include "model/table/column_index.h"
+#include "model/table/column_layout_typed_relation_data.h"
+#include "sample.h"
+
+namespace algos {
+
+class SFDAlgorithm : public Algorithm {
```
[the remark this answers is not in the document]
egshnov: How should i name python bindings submodule for cords? Sfd or cords?

src/core/algorithms/fd/sfd/sample.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+namespace algos {
+Sample::Sample(unsigned long long sample_size, size_t rows, model::ColumnIndex l,
+               model::ColumnIndex r, std::vector<model::TypedColumnData> const &data,
+               RelationalSchema const *rel_schema_)
+    : lhs_col_(rel_schema_, rel_schema_->GetColumn(l)->GetName(), l),
+      rhs_col_(rel_schema_, rel_schema_->GetColumn(r)->GetName(), r) {
+    auto seed = std::chrono::high_resolution_clock::now().time_since_epoch().count();
+    std::mt19937 gen(seed);
+    std::uniform_int_distribution<size_t> distribution(0, rows - 1);
```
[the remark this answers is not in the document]
egshnov: If i'm not mistaken, `std::bernoulli_distribution` produces random boolean values. And here we are trying to get random row indices.

src/core/algorithms/fd/sfd/sample.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+    std::unordered_set<std::string> map_lhs;
+    std::unordered_set<std::string> map_rhs;
+    std::unordered_set<std::string> map_cardinality;
+
+    for (model::ColumnIndex i = 0; i < sample_size; i++) {
+        size_t row = distribution(gen);
+        row_indices_.push_back(row);
```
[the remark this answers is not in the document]
egshnov: Yes, paper assumes that sample is drawn with replacement

src/core/algorithms/fd/sfd/cords.h
```diff
@@ -0,0 +1,107 @@
...
+#include "correlation.h"
+#include "frequency_handler.h"
+#include "model/table/column.h"
+#include "model/table/column_index.h"
+#include "model/table/column_layout_typed_relation_data.h"
+#include "sample.h"
+
+namespace algos {
+/*  */
+class Cords : public FDAlgorithm {
```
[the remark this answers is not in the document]
egshnov: Yes, we discussed that Cords should be an inheritor of FDAlgorithm, since SFD is a relaxation of FD and there are no other (known to us) SFD mining algorithms so making some kind of intermediate class is redundant.

src/core/algorithms/fd/sfd/cords.cpp
```diff
@@ -0,0 +1,295 @@
...
+                   typed_relation_->GetSchema()->GetColumn(lhs_ind)->GetName(), lhs_ind);
+    Column rhs_col(typed_relation_->GetSchema(),
+                   typed_relation_->GetSchema()->GetColumn(rhs_ind)->GetName(), rhs_ind);
+    Correlation correlation_to_register(lhs_col, rhs_col);
+    correlations_collection_.Register(std::move(correlation_to_register));
+}
+
+void Cords::CheckAndRegisterCorrelation(model::ColumnIndex col_i, model::ColumnIndex col_k,
+                                        std::vector<model::TypedColumnData> const &data,
+                                        Sample &smp, FrequencyHandler const &handler) {
```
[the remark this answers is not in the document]
egshnov: It seems to me that sample should be left as it is because it is recreated on each iteration

src/core/algorithms/fd/sfd/frequency_handler.cpp
```diff
@@ -0,0 +1,49 @@
...
+                                            model::ColumnIndex columns,
+                                            size_t max_amount_of_categories) {
+    cardinality_.resize(columns, 0);
+    frequency_maps_.resize(columns);
+    freq_sums_.resize(columns, 0);
+    for (model::ColumnIndex col_ind = 0; col_ind < data.size(); col_ind++) {
+        std::unordered_map<std::string, size_t> value_frequency_counter;
+        auto const &col_data = data[col_ind];
+        for (model::TupleIndex row_ind = 0; row_ind < col_data.GetNumRows(); row_ind++) {
+            auto &freq = value_frequency_counter[col_data.GetDataAsString(row_ind)];
```
Sched71, line 24: Was it neccessary to view values as strings? Can't we save some memory and time by only looking at their byte representation and storing pointers to already existing values?

src/core/config/descriptions.h
```diff
@@ -109,4 +109,31 @@ constexpr auto kDInsertStatements = "Rows to be inserted into the table using th
 constexpr auto kDDeleteStatements = "Rows to be deleted from the table using the delete operation";
 constexpr auto kDUpdateStatements = "Rows to be replaced in the table using the update operation";
 constexpr auto kDNDWeight = "Weight of ND to verify (positive integer)";
+constexpr auto kDMinCard =
+        " (1-val) * |R| (where |R| is amount of rows in table R) denotes minimum "
+        "cardinality of a column to be considered a soft key. Value lies in (0, 1)";
```
Sched71, line 114: If it includes endpoints wouldn't it be better to write [0, 1]?

src/tests/test_sfd.cpp
```diff
@@ -0,0 +1,190 @@
...
+}
+
+TEST_F(CordsAlgorithmTest, CIPublicHighway10k) {
+    auto a = CreateCordsInstance(kCIPublicHighway10k, kTestConfigOnlySfdDefault);
+    a->Execute();
+    AssertVectors(a->GetSoftKeys(), {0, 2});
+    AssertVectors(a->GetTrivialColumns(), {1, 7, 8, 9, 10, 14, 15, 16, 17});
+    AssertSFDList(a->FdList(), {{11, 3}, {11, 4}, {11, 5}, {11, 6}, {11, 12}});
+}
+
```
Sched71, line 189: No correlation tests?

Sched71 (changes requested):

src/core/algorithms/fd/sfd/frequency_handler.cpp
egshnov: In order to use stl containers with byte representation properly we will have to do something like this  (at least that's the only thing I could think of.):
```cpp
#pragma once
#include <functional>

#include "model/types/types.h"

namespace algos {
struct TypedByte {
    std::byte const* data;
    model::Type const& type;
    TypedByte() = delete;
    TypedByte(std::byte const* data, model::Type const& type) : data(data), type(type) {}

    std::size_t GetHash() const {
        return this->type.GetHasher()(this->data);
    }

    TypedByte& operator=(TypedByte const& rhs) {
        if (this != &rhs) {
            data = rhs.data;
        }
        return *this;
    }

};
}  // namespace algos

template <>
struct std::hash<algos::TypedByte> {
    std::size_t operator()(algos::TypedByte const& tb) const noexcept {
        return tb.GetHash();
    }
};

template <>
struct std::equal_to<algos::TypedByte> {
    std::size_t operator()(algos::TypedByte const& lhs,
                           algos::TypedByte const& rhs) const noexcept {
        return lhs.type.Compare(lhs.data, rhs.data) == model::CompareResult::kEqual;
    }
};
```
and do something like this to all the used containers:
```cpp
std::unordered_map<TypedByte, size_t> value_frequency_counter;
```
It works fine with all the types introduced in Desbordante except for `EmptyType` which throws an error whenever hash or comparison function is called. So it seems to me that there is no way to avoid using strings here.

src/core/config/descriptions.h
egshnov: It doesn't include endpoints

src/core/algorithms/fd/sfd/frequency_handler.cpp
Sched71: Do we actually even need types?

egshnov: In order to implement hash and comparison functions on byte representation itself, without using typed implementations, we need to know the length of the byte sequence underneath `std::byte*`, which i don't know how to retrieve.

Sched71: I mostly talking about non-typed ColumnData, but ok

Sched71 (approved):

chernishev (approved):

vs9h (approved):

PR #371, opened 2024-03-11
Add CLI for INDs

Petua41:
Add command-line interface for Spider and Faida inclusion dependency detection algorithms

### DIFF
--- a/cli/cli.py
+++ b/cli/cli.py
@@ -5,6 +5,7 @@ from collections import namedtuple
 from enum import StrEnum, auto
 from time import process_time
 from typing import Any, Callable
+from os import scandir
 
 import click
 import desbordante
@@ -15,6 +16,7 @@ class Task(StrEnum):
     afd = auto()
     od = auto()
     pfd = auto()
+    ind = auto()
     fd_verification = auto()
     afd_verification = auto()
     mfd_verification = auto()
@@ -34,6 +36,8 @@ class Algorithm(StrEnum):
     aid = auto()
     fastod = auto()
     order = auto()
+    spider = auto()
+    faida = auto()
     naive_fd_verifier = auto()
     naive_afd_verifier = auto()
     icde09_mfd_verifier = auto()
@@ -47,6 +51,9 @@ FILENAME = 'filename'
 VERBOSE = 'verbose'
 ERROR = 'error'
 ERROR_MEASURE = 'error_measure'
+TABLES = 'tables'
+TABLES_LIST = 'tables_list'
+TABLES_DIRECTORY = 'tables_directory'
 
 PRIMARY_HELP = '''The Desbordante data profiler is designed to help users
 discover or verify various types of patterns in data. These patterns are
@@ -102,9 +109,10 @@ Currently, the console version of Desbordante supports:
 2) Discovery of approximate functional dependencies
 3) Discovery of probabilistic functional dependencies
 4) Discovery of exact order dependencies (set-based and list-based axiomatization)
-5) Verification of exact functional dependencies
-6) Verification of approximate functional dependencies
-7) Verification of metric dependencies
+5) Discovery of inclusion dependencies
+6) Verification of exact functional dependencies
+7) Verification of approximate functional dependencies
+8) Verification of metric dependencies
 
 If you need other types, you should look into the C++ code, the Python
 bindings or the Web version.
@@ -117,7 +125,9 @@ bindings or the Web version.
     specify the algorithm to run, e.g., PYRO
 
 --table=TABLE
-    specify the input file to be processed by the algorithm
+    specify the input file to be processed by the algorithm.
+    Algorithms for some tasks (currently, only IND) accept multiple
+    input files; see --task=TASK for more information
 
 --is_null_equal_null=BOOLEAN
     specify whether two NULLs should be considered equal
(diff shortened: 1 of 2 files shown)


cli/cli.py
```diff
@@ -482,11 +566,14 @@ def decorator(func: Callable) -> Callable:
                 in option_type_info.items():
             arg = f'--{opt_name}'
             if opt_main_type == list:
-                click.option(arg, multiple=True,
+                if opt_additional_types[0] == desbordante.data_types.Table:
+                    click.option(arg, type=(str, str, bool),
```
BUYT-1, line 570: [This](https://click.palletsprojects.com/en/8.1.x/parameters/#implementing-custom-types) should help avoid the workaround  (the elif will become redundant as well)

BUYT-1 (changes requested):
The CLI wasn't programmed with specifying the same option in multiple ways in mind.
I think making a dict that contains all option values out of kwargs instead of passing it around will be a better solution, though care must be taken with regards to the unused options set and perhaps something else.

cli/cli.py
```diff
@@ -585,11 +694,13 @@ def desbordante_cli(**kwargs: Any) -> None:
...
+    opts = process_tables_options(kwargs)
+
     start_point = process_time()
-    used_opts = set_algo_options(curr_algo, kwargs)
+    used_opts = set_algo_options(curr_algo, opts)
     curr_algo.load_data()
-    used_opts |= set_algo_options(curr_algo, kwargs)
-    provided_options = get_provided_options(kwargs)
+    used_opts |= set_algo_options(curr_algo, opts)
+    provided_options = get_provided_options(opts)
```
BUYT-1, line 703: If a user executes `./cli.py --tables-list ... --tables-directory ...`, but the algorithm doesn't use the "tables" option, it looks like this will output `Unused options: {"tables"}`, which is confusing.

cli/cli.py
```diff
@@ -393,6 +442,35 @@ def check_error_measure_option_presence(task: str | None, error_measure: str | N
         sys.exit(1)
 
 
+def parse_tables_list_file(file: click.File) \
+        -> list[tuple[str, str, bool]]:
+    try:
+        result = []
+        for line_num, line in enumerate(file.readlines()):
```
BUYT-1, line 449:
```suggestion
        for line_num, line in enumerate(file.readlines(), start=1):
```
Text editors normally count lines from 1

cli/cli.py
```diff
@@ -534,6 +614,29 @@ def get_option_type_info() -> dict[str, Any]:
...
+        if not is_omitted(opts[TABLES_DIRECTORY]):
+            new_tables.extend(parse_tables_directory(opts[TABLES_DIRECTORY]))
+        opts.pop(TABLES_DIRECTORY)
+
+    if new_tables:
+        if TABLES in opts:
+            if is_omitted(opts[TABLES]):
+                opts[TABLES] = new_tables
+            else:
+                opts[TABLES].extend(new_tables)
```
BUYT-1, line 633: `click` puts a tuple into the dict, but tuples are immutable, `.extend` won't work

cli/cli.py
```diff
@@ -534,6 +614,29 @@ def get_option_type_info() -> dict[str, Any]:
...
+        else:
+            opts.update(TABLES, new_tables)
```
BUYT-1, line 635: the branch on line 631 does exactly the same thing, change the `if`

By the way, correction. [Docs](https://docs.python.org/3.8/library/stdtypes.html#dict.update)
```suggestion
            opts.update(TABLES=new_tables)
```
(use line 631 instead)

cli/cli.py
```diff
@@ -534,6 +614,29 @@ def get_option_type_info() -> dict[str, Any]:
...
+def process_tables_options(opts: dict[str, Any]) -> dict[str, Any]:
+    new_tables = []
+    if TABLES_LIST in opts:
+        if not is_omitted(opts[TABLES_LIST]):
+            new_tables.extend(parse_tables_list_file(opts[TABLES_LIST]))
+        opts.pop(TABLES_LIST)
+    if TABLES_DIRECTORY in opts:
+        if not is_omitted(opts[TABLES_DIRECTORY]):
+            new_tables.extend(parse_tables_directory(opts[TABLES_DIRECTORY]))
+        opts.pop(TABLES_DIRECTORY)
```
BUYT-1, line 626:
```suggestion
    for option_name, parse_func in ((TABLES_LIST, parse_tables_list_file), (TABLES_DIRECTORY, parse_tables_directory)):
        value = opts.pop(option_name)
        if not is_omitted(value):
            new_tables.extend(parse_func(value))
```
I'm pretty sure those options are going to be in the dict

src/core/config/descriptions.h
```diff
@@ -86,8 +86,12 @@ constexpr auto kDBumpsLimit = "max considered intervals amount. Pass 0 to remove
 constexpr auto kDTimeLimitSeconds = "max running time of the algorithm. Pass 0 to remove limit";
 constexpr auto kDIterationsLimit = "limit for iterations of sampling";
 constexpr auto kDACSeed = "seed, needed for choosing a data sample";
-constexpr auto kDHllAccuracy = "HyperLogLog approximation accuracy";
-constexpr auto kDSampleSize = "Size of a table sample";
+constexpr auto kDHllAccuracy =
+        "HyperLogLog approximation accuracy. Should be positive\n"
```
BUYT-1, line 90: nit:
```suggestion
        "HyperLogLog approximation accuracy. Must be positive\n"
```

BUYT-1 (changes requested):

cli/cli.py
```diff
@@ -523,6 +608,18 @@ def get_option_type_info() -> dict[str, Any]:
     return option_type_info
 
 
+def process_tables_options(opts: dict[str, Any], algo_name: str) -> dict[str, Any]:
+    if algo_name not in TASK_INFO[Task.ind].algos:
+        return opts
```
BUYT-1, line 613: There are going to be other algorithms that accept multiple tables. I was thinking of changing print_unused... so that it works with the actual kwargs and handles those options itself or something better, if possible.

cli/cli.py
```diff
@@ -523,6 +608,18 @@ def get_option_type_info() -> dict[str, Any]:
...
+
+    for option_name, parse_func in ((TABLES_LIST, parse_tables_list_file),
+            (TABLES_DIRECTORY, parse_tables_directory)):
+        opts[TABLES] = list(opts[TABLES]) + parse_func(opts[option_name])
+        opts.pop(option_name)
```
BUYT-1, line 618:
```suggestion
        value = opts.pop(option_name)
        opts[TABLES] = list(opts[TABLES]) + parse_func(value)
```
`dict.pop` [returns the value](https://docs.python.org/3.8/library/stdtypes.html#dict.pop). If the value is not needed (which is *not* the case here), `del dict[key]` should be used.

BUYT-1 (changes requested):

cli/cli.py
BUYT-1: I think whoever is going to add another algorithm with this option would like being able to just add whatever is relevant to their algorithm into the globals and have things work, while having to understand as little as possible of the CLI.
On the contrary, of course, it may lead to problems if there are other options like it, but I believe it'd be better to make a general system that handles that case later and bear with the additional complexity right now.

BUYT-1 (approved):
LGTM👍

PR #372, opened 2024-03-11
Add verifying FD and AFD

Sneper-Breeze:
Add verifying_fd_afd.py in examples/ and DnD.csv in examples/datasets.
This example shows how you can verify FD and AFD using Desbordante.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ lib/
 .vscode/
 **/myeasylog.log
 dist
+venv
--- a/examples/datasets/DnD.csv
+++ b/examples/datasets/DnD.csv
@@ -1,0 +1,8 @@
+Creature,Strength,HaveMagic
+Ogre,9,false
+Ogre,6,false
+Elf,6,true
+Elf,6,true
+Elf,1,true
+Dwarf,9,false
+Dwarf,6,false
--- a/examples/verifying_fd_afd.py
+++ b/examples/verifying_fd_afd.py
@@ -1,0 +1,100 @@
+import desbordante
+import pandas as pd
+
+
+GREEN_CODE = "\033[1;42m"
+RED_CODE = "\033[1;41m"
+BLUE_CODE = "\033[1;46m"
+DEFAULT_COLOR_CODE = "\033[1;49m"
+
+
+def print_clusters(verifier, data, lhs, rhs):
+    print(f"Number of clusters violating FD: {verifier.get_num_error_clusters()}")
+    for i, highlight in enumerate(verifier.get_highlights(), start=1):
+        print(f"{BLUE_CODE} #{i} cluster: {DEFAULT_COLOR_CODE}")
+        for el in highlight.cluster:
+            print(f"{el}: {data[data.columns[lhs]][el]} -> {data[data.columns[rhs]][el]}")
+        
+        print(f"Most frequent rhs value proportion: {highlight.most_frequent_rhs_value_proportion}")
+        print(f"Num distinct rhs values: {highlight.num_distinct_rhs_values}\n")
+
+
+def print_results_for_fd(verifier, data, lhs, rhs):
+    if verifier.fd_holds():
+        print(GREEN_CODE, "FD holds", DEFAULT_COLOR_CODE)
+    else:
+        print(RED_CODE, "FD does not hold", DEFAULT_COLOR_CODE)
+        print_clusters(verifier, data, lhs, rhs)
+        # print(f"But the same {GREEN_CODE} AFD with error threshold = {verifier.get_error()} holds{DEFAULT_COLOR_CODE}")
+
+
+def print_results_for_afd(verifier, error):
+    if verifier.get_error() < error:
+        print(GREEN_CODE, "AFD with this error threshold holds", DEFAULT_COLOR_CODE)
+    else:
+        print(RED_CODE, "AFD with this error threshold does not hold", DEFAULT_COLOR_CODE)
+        print(f"But the same {GREEN_CODE} AFD with error threshold = {verifier.get_error()} holds{DEFAULT_COLOR_CODE}")
+
+
+def exact_scenario(table='examples/datasets/duplicates_short.csv'):
+    print("First, let's look at the duplicates_short.csv table and try to verify the functional dependency in it.\n")
+
+    data = pd.read_csv(table, header=[0])
+    print(data)
+
+    algo = desbordante.afd_verification.algorithms.Default()
+    algo.load_data(table=data)
+    
+    print(DEFAULT_COLOR_CODE)
+    # Verifying exact FD (holds)
+    print("Checking whether [id] -> [name] FD holds")
+
+    algo.execute(lhs_indices=[0], rhs_indices=[2])
+    print_results_for_fd(algo, data, 0, 2)
+
+    # Verifying exact FD (does not hold)
+    print("Checking whether [name] -> [credit_score] FD holds")
+
+    algo.execute(lhs_indices=[1], rhs_indices=[2])
+    print_results_for_fd(algo, data, 1, 2)
+    print("We learned that in this case the specified FD does not hold and there are two "
+          "clusters of rows that contain values that prevent our FD from holding. "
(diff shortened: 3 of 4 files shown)


chernishev:
I've looked it through, but focusing only on example quality.

Sneper-Breeze:
I've made edits

.gitignore
```diff
@@ -11,3 +11,4 @@ lib/
 .vscode/
 **/myeasylog.log
 dist
+venv
```
BUYT-1, line 14: Seems unrelated to the rest of the changes, should be a separate commit

examples/datasets/DnD.csv
BUYT-1, line 1: I don't think there should be spaces here

examples/verifying_fd_afd.py
```diff
@@ -0,0 +1,63 @@
+import desbordante
+import pandas as pd
+
+
+TABLE = 'examples/datasets/DnD.csv'
+COLOR_CODES = {'green': '\033[1;42m', 'red': '\033[1;41m', 'default': '\033[1;49m'}
```
BUYT-1, line 6: Should be constants in the code

examples/verifying_fd_afd.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def print_results_for_fd(verifier):
+    if verifier.fd_holds():
+        print(COLOR_CODES["green"], 'FD holds', COLOR_CODES["default"])
+
```
BUYT-1, line 12: nit: extra space

examples/verifying_fd_afd.py
```diff
@@ -0,0 +1,63 @@
...
+    else:
+        print(COLOR_CODES["red"], 'FD does not hold', COLOR_CODES["default"])
+        print(f"But the same {COLOR_CODES['green']} AFD with error threshold = {verifier.get_error()} holds{COLOR_CODES['default']}")
+
+
+def print_results_for_afd(verifier, error):
+    if verifier.get_error() < error:
+        print(COLOR_CODES["green"], 'AFD with this error threshold holds', COLOR_CODES["default"])
+
```
BUYT-1, line 21: nit:extra space

BUYT-1 (changes requested):

examples/verifying_fd_afd.py
Sneper-Breeze: I didn't understand it a bit. It's a python, there are no constants, aren't there? Or is it meant to put it somewhere in a separate file or something?

BUYT-1: There are no actual constants, yes. I meant that those color codes should be their own variables. See also [here](https://peps.python.org/pep-0008/#constants)

examples/verifying_fd_afd.py
```diff
@@ -0,0 +1,93 @@
...
+        print(f"Num distinct rhs values: {highlight.num_distinct_rhs_values}\n")
+
+
+def print_results_for_fd(verifier, lhs, rhs):
+    if verifier.fd_holds():
+        print(GREEN_CODE, 'FD holds', DEFAULT_COLOR_CODE)
+    else:
+        print(RED_CODE, 'FD does not hold', DEFAULT_COLOR_CODE)
+        print_clusters(verifier, DATA_FOR_FD, lhs, rhs)
+        # print(f"But the same {GREEN_CODE} AFD with error threshold = {verifier.get_error()} holds{DEFAULT_COLOR_CODE}")
```
BUYT-1, line 30: Commented code

examples/verifying_fd_afd.py
```diff
@@ -0,0 +1,93 @@
...
+def print_results_for_afd(verifier, error):
+    if verifier.get_error() < error:
+        print(GREEN_CODE, 'AFD with this error threshold holds', DEFAULT_COLOR_CODE)
+    else:
+        print(RED_CODE, 'AFD with this error threshold does not hold', DEFAULT_COLOR_CODE)
+        print(f"But the same {GREEN_CODE} AFD with error threshold = {verifier.get_error()} holds{DEFAULT_COLOR_CODE}")
+
+
+def exact_scenario():
+    print("First, let's look at the duplicates_short.csv table and try to verify the functional dependence in it.\n")
```
BUYT-1, line 42:
```suggestion
    print("First, let's look at the duplicates_short.csv table and try to verify the functional dependency in it.\n")
```

examples/verifying_fd_afd.py
```diff
@@ -0,0 +1,93 @@
...
+    print_results_for_fd(algo, 0, 2)
+
+    # Verifying exact FD (does not hold)
+    print("Checking whether [name] -> [credit_score] FD holds")
+
+    algo.execute(lhs_indices=[1], rhs_indices=[2])
+    print_results_for_fd(algo, 1, 2)
+    print("We learned that in this case the specified FD does not hold and there are two clusters of rows that contain values that prevent our FD from holding."
+        +f"A {BLUE_CODE}cluster{DEFAULT_COLOR_CODE} (with respect to a fixed FD) is a collection of rows that share the same left-hand side part but differ on the right-hand side one.")
+    print('Let\'s take a closer look at them.\n')
```
BUYT-1, line 62:
```suggestion
    print("Let's take a closer look at them.\n")
```

examples/verifying_fd_afd.py
```diff
@@ -0,0 +1,93 @@
...
+    print('In the first cluster, three values are "0" and a single one is "nan".'
+        +'This suggests us that this single entry with the "nan" value is a result of a mistake by someone who is not familiar with the table population policy. Therefore, probably, it is right to change it to "0".\n')
```
BUYT-1, line 64:
```suggestion
         'This suggests that this single entry with the "nan" value is a result of a mistake by someone who is not familiar with the table population policy. Therefore, it should probably be changed to "0".\n')
```

examples/verifying_fd_afd.py
```diff
@@ -0,0 +1,93 @@
...
+    print("Now let's take a look at the second cluster."
+        + 'There are two entries: "27" and "28". In this case it is probably a typo, since buttons 7 and 8 are located close to each other on the keyboard.\n')
```
BUYT-1, line 66:
```suggestion
          'There are two entries: "27" and "28". In this case, it is probably a typo, since buttons 7 and 8 are located close to each other on the keyboard.\n')
```

examples/verifying_fd_afd.py
```diff
@@ -0,0 +1,93 @@
+import desbordante
+import pandas as pd
+
+
+GREEN_CODE = '\033[1;42m'
+RED_CODE = '\033[1;41m'
+BLUE_CODE = '\033[1;46m'
+DEFAULT_COLOR_CODE = '\033[1;49m'
+DATA_FOR_FD = pd.read_csv('examples/datasets/duplicates_short.csv', header=[0])
+DATA_FOR_AFD = pd.read_csv('examples/datasets/DnD.csv', header=[0])
```
BUYT-1, line 10: It'd be preferable if these variables were function parameters instead of globals.

BUYT-1 (changes requested):
Minor things.

Be consistent with your quotes. They should be the same every time, other than the cases where the same quote symbol is contained inside, then they should be changed.
The changes in commits are still not right. One commit should contain the example, the other `.gitignore`.
Your lines are sometimes too long. String literals can be concatenated at compile time, see [here](https://docs.python.org/3/reference/lexical_analysis.html#string-literal-concatenation).

examples/verifying_fd_afd.py
chernishev: We decided to leave it as this since this invocation clutters the output. However, it is useful to show that clusters can be obtained while using the AFD primitive too.

Sneper-Breeze:
My commits are broken due to the fact that I added venv to .gitignor at the very beginning of all the changes and made a commit with it much later along with other changes. Which makes it difficult for me to separate it somehow now. That's why I made a fake commit with venv added. Please forgive me for this, I will be more careful with this in the future.

BUYT-1 (approved):
Quotes are still inconsistent.

PR #373, opened 2024-03-11
Remove paths-ignore from core tests workflow

BUYT-1:
CI can't pass if required workflows don't run


chernishev (approved):
This is a partial rollback of https://github.com/Mstrutov/Desbordante/pull/363 

We have to do this since otherwise CI stalls.

PR #376, opened 2024-03-18
Display table and column names in INDs

Petua41:
Add methods and bindings for displaying table and
column names instead of indices in IND algorithms' output

### DIFF
--- a/src/core/algorithms/ind/ind.h
+++ b/src/core/algorithms/ind/ind.h
@@ -4,6 +4,8 @@
 #include <string>
 
 #include "model/table/column_combination.h"
+#include "model/table/relational_schema.h"
+#include "model/table/vertical.h"
 
 namespace model {
 
@@ -13,10 +15,12 @@ class IND {
 private:
     std::shared_ptr<ColumnCombination> lhs_;
     std::shared_ptr<ColumnCombination> rhs_;
+    std::shared_ptr<std::vector<RelationalSchema>> schemas_;
 
 public:
-    IND(std::shared_ptr<ColumnCombination> lhs, std::shared_ptr<ColumnCombination> rhs)
-        : lhs_(std::move(lhs)), rhs_(std::move(rhs)) {}
+    IND(std::shared_ptr<ColumnCombination> lhs, std::shared_ptr<ColumnCombination> rhs,
+        std::shared_ptr<std::vector<RelationalSchema>> schemas)
+        : lhs_(std::move(lhs)), rhs_(std::move(rhs)), schemas_(std::move(schemas)) {}
 
     ColumnCombination const& GetLhs() const {
         return *lhs_;
@@ -26,7 +30,9 @@ public:
         return *rhs_;
     }
 
-    std::string ToString() const;
+    std::string ToShortString() const;
+
+    std::string ToLongString() const;
 };
 
 }  // namespace model
--- a/src/core/model/table/relational_schema.h
+++ b/src/core/model/table/relational_schema.h
@@ -30,6 +30,12 @@ public:
     std::unique_ptr<Vertical> empty_vertical_;
 
     RelationalSchema(std::string name);
+
+    RelationalSchema(RelationalSchema const& other) = delete;
+    RelationalSchema& operator=(RelationalSchema const& rhs) = delete;
+    RelationalSchema(RelationalSchema&& other) noexcept = default;
+    RelationalSchema& operator=(RelationalSchema&& rhs) noexcept = default;
+
     void Init();
 
     std::string GetName() const {
--- a/src/core/algorithms/ind/faida/faida.cpp
+++ b/src/core/algorithms/ind/faida/faida.cpp
@@ -30,7 +30,7 @@ void Faida::MakeExecuteOptsAvailable() {
                           config::kThreadNumberOpt.GetName()});
 }
 
-void Faida::LoadDataInternal() {
+void Faida::LoadINDAlgorithmDataInternal() {
     auto start_time = std::chrono::system_clock::now();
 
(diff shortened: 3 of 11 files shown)


vs9h (changes requested):
I don't really like this solution where we need to update the implementation of algorithms, instead of modifying the general class of ind algorithms. In fact, we just need to modify the `INDAlgorithm` class.

I can suggest two possible modifications:
1) **Modification of `RegisterIND` method and the `IND` class.** 

We add a type (further we call this type `Data`) that contains information about the names of all tables and their columns. Inside the `INDAlgorithm` class we store `std::shared_ptr<Data>`. It must be initialized inside the `INDAlgorithm::LoadDataInternal()`, using the same idea from the `INDAlgorithm::ResetState`.
```cpp
class INDAlgorithm {
    ...
    void LoadDataInternal() final {
        /* init std::shared_ptr<Data> field */
        LoadINDAlgorithmDataInternal();
    }
    virtual void LoadINDAlgorithmDataInternal() = 0;
    ...
};
```
Next we modify the register ind method, internally we also pass the `std::shared_ptr<Data>` to the `IND` constructor (as third param).

The last thing to do is to add serialization to the `IND` class using `Data`. Use this struct only within `IND` class. Define a lambda inside `ToLongString` that will serialize column combinations by calling `GetTableIndex()` and `GetColumnIndices()` and, using a `Data` to resolve indices into names.

Note: The type should not be called the `Data`, the name should be more meaningful.

2) Add a new `INDString` type

Add a new type with strings, similar to how it is implemented in the `ARAlgorithm`. And we add a method that can create a new list (`std::list<INDString>`) from the `ind_collection_`. Then we add bindings to Python using `INDString` instead of `IND`.

I would prefer the first method, but the second is also acceptable.

src/core/algorithms/ind/ind.h
```diff
@@ -13,10 +14,14 @@ class IND {
 private:
     std::shared_ptr<ColumnCombination> lhs_;
     std::shared_ptr<ColumnCombination> rhs_;
+    std::shared_ptr<ColumnNamesData> column_names_data_;
```
vs9h, line 17: Let's reuse existing class `RelationalSchema`, that describes a single relation. And we will simply use a vector of these schemas:
```suggestion
    std::shared_ptr<std::vector<RelationalSchema>> schemas_;
```
Initialization of schema looks like:
```cpp
    schema = std::make_shared<RelationalSchema>(input_table->GetRelationName());
    for (size_t col_id = 0; col_id < input_table->GetNumberOfColumns(); ++col_id) {
        schema->AppendColumn(input_table->GetColumnName(col_id));
    }
```

vs9h (changes requested):
Firstly, fix the comment in the `ind.h` file about `RelationalSchema`, then fix other small comments and we will be very close to merging

src/core/algorithms/ind/ind.h
```diff
@@ -4,6 +4,7 @@
 #include <string>
 
 #include "model/table/column_combination.h"
+#include "model/table/relational_schema.h"
```
vs9h, line 7: Add include for `Vertical`, because `RelationalSchema` uses forward declaration:
```cpp
#inlcude "model/table/vertical.h"
```

src/core/model/table/relational_schema.h
```diff
@@ -30,6 +30,7 @@ class RelationalSchema {
     std::unique_ptr<Vertical> empty_vertical_;
 
     RelationalSchema(std::string name);
+    RelationalSchema(RelationalSchema&& other);
```
vs9h, line 33: It's enough to use default move constructor. And let's define other constructors/operators from rule of 5 (destructor is already defined):
```suggestion
    RelationalSchema(RelationalSchema const& other) = delete;
    RelationalSchema& operator=(RelationalSchema const& rhs) = delete;
    RelationalSchema(RelationalSchema&& other) noexcept = default;
    RelationalSchema& operator=(RelationalSchema&& rhs) noexcept = default;
```

vs9h (changes requested):
Don't forget to fix commit history (maybe squash to a single commit).

The last thing I want to discuss is the serialization format. Now the output will look like this:
```
"1.[1], 1.[3] -> 2.[3], 2.[5]" /* for ToShortString() */
"Table1.Attr1, Table1.Attr3 -> Table2.Attr3, Table2.Attr5" /* ToLongString() */
```
Maybe it's better to have something like this:
```
"(1, [1, 3]) -> (2, [3, 5])" /* for ToShortString() */
"(Table1, [Attr1, Attr3) -> (Table2, [Attr3, Attr5])" /* ToLongString() */
```
It seems to me that the second option is better, but before making changes to the code I would like to hear the opinion of @chernishev.

chernishev:
> Don't forget to fix commit history (maybe squash to a single commit).
> 
> The last thing I want to discuss is the serialization format. Now the output will look like this:
> 
> ```
> "1.[1], 1.[3] -> 2.[3], 2.[5]" /* for ToShortString() */
> "Table1.Attr1, Table1.Attr3 -> Table2.Attr3, Table2.Attr5" /* ToLongString() */
> ```
> 
> Maybe it's better to have something like this:
> 
> ```
> "(1, [1, 3]) -> (2, [3, 5])" /* for ToShortString() */
> "(Table1, [Attr1, Attr3) -> (Table2, [Attr3, Attr5])" /* ToLongString() */
> ```
> 
> It seems to me that the second option is better, but before making changes to the code I would like to hear the opinion of @chernishev.

Discussed with Anton, 2nd is definitely better

vs9h:
Also regarding the history of commits, I would prefer to squash them into one. We try to make sure that within the same PR, newer commits do not fix older commits of this PR. 
After changing `ToString` formats we can merge it

vs9h (approved):

PR #377, opened 2024-03-21
Clang tidy add naming check

vs9h:
This pull request adds rules for checking naming, which will be useful for both contributors and reviewers.
Also in this pull request, warnings from `clang-tidy` about naming and warnings from `clang-format` have been fixed.


src/core/algorithms/ind/faida/hashing/murmur_hash_3.h
```diff
@@ -9,6 +9,7 @@
 // Microsoft Visual Studio
 
 // clang-format off
+// NOLINTBEGIN(readability-identifier-naming)
```
BUYT-1, line 12: Is there a problem with reformatting these files as well?

BUYT-1, line 12: I think linting shouldn't be disabled for the entire file here.

polyntsov (changes requested):
Overall I think it's a good idea to use naming rules in clang-tidy, however, there are several issues with this. Firstly, I'm not sure we want to use it for `python_bindings` subdirectory, since there are a lot of exceptions for naming rules and adding everywhere `NOLINT` seems very cumbersome.
Secondly, I'm not sure we want to make constants naming rules (with `k` prefix) obligatory , because it forces, for example, `k` prefix for local constexpr variables and for constexpr lambdas which IMO is strange.
Nevertheless, perhaps it makes sense to discuss this with other developers, because all this is rather opinion-based.

vs9h:
> Overall I think it's a good idea to use naming rules in clang-tidy, however, there are several issues with this. Firstly, I'm not sure we want to use it for `python_bindings` subdirectory, since there are a lot of exceptions for naming rules and adding everywhere `NOLINT` seems very cumbersome. Secondly, I'm not sure we want to make constants naming rules (with `k` prefix) obligatory , because it forces, for example, `k` prefix for local constexpr variables and for constexpr lambdas which IMO is strange. Nevertheless, perhaps it makes sense to discuss this with other developers, because all this is rather opinion-based.

I removed the check for constexpr variables, now for local constexpr variables without static storage duration we do not use a prefix `k` and name them as a regular local variable.

Also, it turns out that `clang-tidy` does not issue warnings for errors in `python_bindinds/` (to check this, I locally added changes to the lines on which `clang-tidy` locally generates warnings, run CI on these changes and did not receive any warnings). So for now I just removed `NOLINT` lines. Now there are only 3 places in the code where we suppress errors for this rule (murmurhash and hyperloglog)

For now, I would suggest not touching the static constexpr lambdas and leaving them with prefixes, because this is fully consistent with code style. Or, if you are sure, then we can add `NOLINT` line

polyntsov (approved):

PR #381, opened 2024-03-26
Add Statistic ctor with model::TypeId

xJoskiy:
Implement a ctor to create a Statistic object without passing a pointer to Type object

### DIFF
--- a/src/core/algorithms/statistics/statistic.h
+++ b/src/core/algorithms/statistics/statistic.h
@@ -13,6 +13,7 @@ class Statistic {
 public:
     Statistic() noexcept = default;
     Statistic(std::byte const *, model::Type const *, bool clone_data);
+    Statistic(std::byte const *data, model::TypeId id, bool clone_data = false);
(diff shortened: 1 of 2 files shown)


src/core/algorithms/statistics/statistic.h
```diff
@@ -13,6 +13,7 @@ class Statistic {
 public:
     Statistic() noexcept = default;
     Statistic(std::byte const *, model::Type const *, bool clone_data);
+    Statistic(std::byte const *, model::TypeId, bool clone_data = false);
```
polyntsov, line 16: Don't omit parameter names, this impairs readability.

polyntsov: Tbh I don't remember at all what all this is about, is this constructor going to be used anywhere?

xJoskiy: For some reason new implemented method doesn't pass python bindings tests, though passes all cpp tests.

PR #374, opened 2024-03-11
Implement Split algorithm

MichaelS239:
Implement Split algorithm for mining of differential dependencies

PR #389, opened 2024-04-05
Add ND verification

Petua41:
Add numerical dependencies verification algorithm

### DIFF
--- a/src/core/algorithms/nd/nd.cpp
+++ b/src/core/algorithms/nd/nd.cpp
@@ -1,0 +1,31 @@
+#include "algorithms/nd/nd.h"
+
+#include <numeric>
+
+#include "algorithms/nd/nd_verifier/util/vector_to_string.h"
+
+namespace model {
+
+using namespace algos::nd_verifier::util;
+
+std::string ND::ToShortString() const {
+    return VectorToString(GetLhsIndices()) + " -> " + VectorToString(GetRhsIndices());
+}
+
+std::string ND::ToLongString() const {
+    auto cols_to_long_string = [](std::vector<Column const*> const& cols) -> std::string {
+        if (cols.size() == 0) {
+            return "[]";
+        }
+
+        return std::accumulate(std::next(cols.begin()), cols.end(), cols[0]->GetName(),
+                               [](std::string&& a, Column const* b) {
+                                   return std::move(a) + ", " + b->GetName();
+                               });
+    };
+
+    return cols_to_long_string(GetLhs().GetColumns()) + " -> " +
+           cols_to_long_string(GetRhs().GetColumns());
+}
+
+}  // namespace model
--- a/src/core/algorithms/nd/nd.h
+++ b/src/core/algorithms/nd/nd.h
@@ -1,0 +1,45 @@
+#pragma once
+
+#include "model/table/column.h"
+#include "model/table/vertical.h"
+
+namespace model {
+
+using WeightType = unsigned int;
+
+class ND {
+private:
+    Vertical lhs_;
+    Vertical rhs_;
+    WeightType weight_;
+
+public:
+    ND(Vertical const& lhs, Vertical const& rhs, WeightType weight)
+        : lhs_(lhs), rhs_(rhs), weight_(weight) {}
+
+    [[nodiscard]] Vertical const& GetLhs() const {
+        return lhs_;
+    }
+
+    [[nodiscard]] Vertical const& GetRhs() const {
+        return rhs_;
+    }
+
+    [[nodiscard]] WeightType GetWeight() const {
+        return weight_;
+    }
+
+    [[nodiscard]] std::vector<ColumnIndex> GetLhsIndices() const {
+        return lhs_.GetColumnIndicesAsVector();
+    }
+
+    [[nodiscard]] std::vector<ColumnIndex> GetRhsIndices() const {
+        return rhs_.GetColumnIndicesAsVector();
+    }
+
+    [[nodiscard]] std::string ToShortString() const;
+
+    [[nodiscard]] std::string ToLongString() const;
+};
+
+}  // namespace model
--- a/src/core/algorithms/nd/nd_verifier/nd_verifier.cpp
+++ b/src/core/algorithms/nd/nd_verifier/nd_verifier.cpp
@@ -1,0 +1,196 @@
+#include "algorithms/nd/nd_verifier/nd_verifier.h"
+
+#include <chrono>
+#include <cstddef>
+#include <sstream>
+#include <stdexcept>
+#include <string>
+#include <vector>
+
+#include <easylogging++.h>
+
+#include "algorithms/nd/nd_verifier/util/stats_calculator.h"
+#include "algorithms/nd/nd_verifier/util/value_combination.h"
+#include "algorithms/nd/nd_verifier/util/vector_to_string.h"
+#include "config/descriptions.h"
+#include "config/equal_nulls/option.h"
+#include "config/indices/option.h"
+#include "config/names.h"
+#include "config/option.h"
+#include "config/option_using.h"
+#include "config/tabular_data/input_table/option.h"
+#include "model/table/column_layout_typed_relation_data.h"
+#include "model/table/typed_column_data.h"
+#include "model/types/builtin.h"
+#include "model/types/type.h"
+#include "util/timed_invoke.h"
+
+namespace algos::nd_verifier {
+
+NDVerifier::NDVerifier() : Algorithm({}) {
+    RegisterOptions();
+    MakeOptionsAvailable({config::kTableOpt.GetName(), config::kEqualNullsOpt.GetName()});
+}
+
+void NDVerifier::RegisterOptions() {
+    DESBORDANTE_OPTION_USING;
+
+    auto get_schema_cols = [this]() { return typed_relation_->GetSchema()->GetNumColumns(); };
+
+    RegisterOption(config::kTableOpt(&input_table_));
+    RegisterOption(config::kEqualNullsOpt(&is_null_equal_null_));
+    RegisterOption(config::kLhsIndicesOpt(&lhs_indices_, get_schema_cols));
+    RegisterOption(config::kRhsIndicesOpt(&rhs_indices_, get_schema_cols));
+    RegisterOption(Option<model::WeightType>(&weight_, kWeight, kDNDWeight, 1));
+}
+
+void NDVerifier::LoadDataInternal() {
+    typed_relation_ =
+            model::ColumnLayoutTypedRelationData::CreateFrom(*input_table_, is_null_equal_null_);
+    input_table_->Reset();
+    if (typed_relation_->GetColumnData().empty()) {
+        throw std::runtime_error("Got an empty dataset: ND verifying is meaningless.");
+    }
+}
+
+void NDVerifier::MakeExecuteOptsAvailable() {
+    MakeOptionsAvailable({config::kLhsIndicesOpt.GetName(), config::kRhsIndicesOpt.GetName(),
+                          config::names::kWeight});
+}
+
+unsigned long long NDVerifier::ExecuteInternal() {
+    LOG(INFO) << "Parameters of NDVerifier:";
+    LOG(INFO) << "\tInput table: " << input_table_->GetRelationName();
+    LOG(INFO) << "\tNull equals null: " << is_null_equal_null_;
+    LOG(INFO) << "\tLhs indices: " << util::VectorToString(lhs_indices_);
+    LOG(INFO) << "\tRhs indices: " << util::VectorToString(rhs_indices_);
+    LOG(INFO) << "\tWeight: " << weight_;
+
+    auto verification_time = ::util::TimedInvoke(&NDVerifier::VerifyND, this);
+
+    LOG(DEBUG) << "ND verification took " << std::to_string(verification_time) << "ms";
+
+    auto stats_calculation_time = ::util::TimedInvoke(&NDVerifier::CalculateStats, this);
+
+    LOG(DEBUG) << "Statistics calculation took " << std::to_string(stats_calculation_time) << "ms";
+
+    return verification_time + stats_calculation_time;
+}
+
+bool NDVerifier::NDHolds() const {
+    return stats_calculator_.GetRealWeight() <= weight_;
+}
+
+void NDVerifier::ResetState() {
+    stats_calculator_ = util::StatsCalculator{};
+}
+
+void NDVerifier::AddVCToValues(
+        std::shared_ptr<std::vector<util::ValueCombination>> values,
+        std::shared_ptr<std::vector<size_t>> row,
+        std::vector<std::pair<model::TypeId, std::byte const*>> const& typed_data,
+        bool is_null) const {
+    util::ValueCombination vc{typed_data};
+
+    size_t index{values->size()};
+    if (is_null && !is_null_equal_null_) {
+        // If not null_eq_null, every value with null will be unique
+        values->push_back(vc);
+    } else {
+        for (size_t i{0}; i < values->size(); ++i) {
+            if ((*values)[i] == vc) {
+                index = i;
+                break;
+            }
+        }
+
+        if (index == values->size()) {  // wasn`t in values
+            values->push_back(vc);
+        }
+    }
+    row->push_back(index);
+}
+
+NDVerifier::CombinedValuesType NDVerifier::CombineValues(
+        config::IndicesType const& col_idxs) const {
+    auto values = std::make_shared<std::vector<util::ValueCombination>>();
+    auto row = std::make_shared<std::vector<size_t>>();
+
+    for (size_t row_idx{0}; row_idx < typed_relation_->GetNumRows(); ++row_idx) {
+        std::vector<std::pair<model::TypeId, std::byte const*>> typed_data;
+        bool was_null = false;
+        for (auto col_idx_pt{col_idxs.begin()}; col_idx_pt != col_idxs.end(); ++col_idx_pt) {
+            model::TypedColumnData const& col_data = typed_relation_->GetColumnData(*col_idx_pt);
+            std::vector<std::byte const*> const& byte_data = col_data.GetData();
+            auto type_id = col_data.GetTypeId();
+
+            std::byte const* bytes_ptr = byte_data[row_idx];
+            if (bytes_ptr == nullptr) {
+                LOG(INFO) << "WARNING: Cell (" << *col_idx_pt << ", " << row_idx << ") is empty";
+                was_null = true;
+            }
+
+            typed_data.emplace_back(std::move(type_id), bytes_ptr);
+        }
+
+        AddVCToValues(values, row, typed_data, was_null);
+    }
+    return std::make_pair(std::move(values), std::move(row));
+}
+
+void NDVerifier::VerifyND() {
+    auto local_start_time = std::chrono::system_clock::now();
+    auto [lhs_values, combined_lhs] = CombineValues(lhs_indices_);
+    auto [rhs_values, combined_rhs] = CombineValues(rhs_indices_);
+
+    LOG(DEBUG) << "Values combination took "
+               << std::to_string(std::chrono::duration_cast<std::chrono::milliseconds>(
+                                         std::chrono::system_clock::now() - local_start_time)
+                                         .count())
+               << "ms";
+
+    local_start_time = std::chrono::system_clock::now();
+    std::unordered_map<size_t, std::unordered_set<size_t>> value_deps{};
+
+    for (size_t i{0}; i < combined_lhs->size(); ++i) {
+        auto lhs_code = (*combined_lhs)[i];
+        auto rhs_code = (*combined_rhs)[i];
+
+        if (value_deps.find(lhs_code) == value_deps.end()) {
+            value_deps.emplace(lhs_code, std::unordered_set<size_t>({rhs_code}));
+        } else {
+            value_deps[lhs_code].insert(rhs_code);
+        }
+    }
+    LOG(DEBUG) << "Value deps calculation took "
+               << std::to_string(std::chrono::duration_cast<std::chrono::milliseconds>(
+                                         std::chrono::system_clock::now() - local_start_time)
+                                         .count())
+               << "ms";
+
+    stats_calculator_ = util::StatsCalculator(std::move(value_deps), std::move(lhs_values),
+                                              std::move(rhs_values), std::move(combined_lhs),
+                                              std::move(combined_rhs));
+}
+
+[[nodiscard]] std::vector<util::Highlight> const& NDVerifier::GetHighlights() const {
+    return stats_calculator_.GetHighlights();
+}
+
+[[nodiscard]] model::WeightType NDVerifier::GetGlobalMinWeight() const {
+    return stats_calculator_.GetGlobalMinWeight();
+}
+
+[[nodiscard]] model::WeightType NDVerifier::GetRealWeight() const {
+    return stats_calculator_.GetRealWeight();
+}
+
+[[nodiscard]] std::unordered_map<std::string, size_t> NDVerifier::GetLhsFrequencies() const {
+    return stats_calculator_.GetLhsFrequencies();
+}
+
+[[nodiscard]] std::unordered_map<std::string, size_t> NDVerifier::GetRhsFrequencies() const {
+    return stats_calculator_.GetRhsFrequencies();
+}
+
+}  // namespace algos::nd_verifier
--- a/src/core/algorithms/nd/nd_verifier/nd_verifier.h
+++ b/src/core/algorithms/nd/nd_verifier/nd_verifier.h
@@ -1,0 +1,67 @@
+#pragma once
+
+#include <memory>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#include "algorithms/algorithm.h"
+#include "algorithms/nd/nd.h"
+#include "algorithms/nd/nd_verifier/util/highlight.h"
+#include "algorithms/nd/nd_verifier/util/stats_calculator.h"
+#include "algorithms/nd/nd_verifier/util/value_combination.h"
+#include "config/equal_nulls/type.h"
+#include "config/indices/type.h"
+#include "config/tabular_data/input_table_type.h"
+#include "model/table/column_layout_typed_relation_data.h"
+#include "model/types/builtin.h"
+
+namespace algos::nd_verifier {
+
+/// @brief Algorithm for verifying if ND holds with given weight
+class NDVerifier : public Algorithm {
+private:
+    using CombinedValuesType = std::pair<std::shared_ptr<std::vector<util::ValueCombination>>,
+                                         std::shared_ptr<std::vector<size_t>>>;
+
+    config::InputTable input_table_;
+    config::IndicesType lhs_indices_;
+    config::IndicesType rhs_indices_;
+    model::WeightType weight_;
+    config::EqNullsType is_null_equal_null_;
+
+    std::shared_ptr<model::ColumnLayoutTypedRelationData> typed_relation_;
+
+    util::StatsCalculator stats_calculator_;
+
+    void RegisterOptions();
+    void ResetState() override;
+    void VerifyND();
+
+    CombinedValuesType CombineValues(config::IndicesType const& col_idxs) const;
+
+    void CalculateStats() {
+        stats_calculator_.CalculateStats();
+    }
+
+    void AddVCToValues(std::shared_ptr<std::vector<util::ValueCombination>> values,
+                       std::shared_ptr<std::vector<size_t>> row,
+                       std::vector<std::pair<model::TypeId, std::byte const*>> const& typed_data,
+                       bool is_null) const;
+
+protected:
+    void LoadDataInternal() override;
+    void MakeExecuteOptsAvailable() override;
+    unsigned long long ExecuteInternal() override;
+
+public:
+    NDVerifier();
+    [[nodiscard]] bool NDHolds() const;
+    [[nodiscard]] std::vector<util::Highlight> const& GetHighlights() const;
+    [[nodiscard]] model::WeightType GetGlobalMinWeight() const;
+    [[nodiscard]] model::WeightType GetRealWeight() const;
+    [[nodiscard]] std::unordered_map<std::string, size_t> GetLhsFrequencies() const;
+    [[nodiscard]] std::unordered_map<std::string, size_t> GetRhsFrequencies() const;
+};
+
+}  // namespace algos::nd_verifier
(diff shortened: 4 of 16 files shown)


yakovypg (commented):
After correcting your code you will need to make a new commit history. For example, commit [Implement NDVerifier::VerifyND method](https://github.com/Desbordante/desbordante-core/pull/389/commits/e8e64e7ae44e62ad3779601e9b5034ca65f58ef2) contains a lot of changes and additions (e.g. tests).

src/core/algorithms/nd/nd_verifier/nd_verifier.cpp
```diff
@@ -0,0 +1,228 @@
+#include "algorithms/nd/nd_verifier/nd_verifier.h"
+
+#include <chrono>
+#include <numeric>
+#include <set>
```
yakovypg, line 5: Remove unused includes and add necessary explicit includes. For example, I don't see using std::set. Furthermore, I see that you use std::unordered_set, however, it isn't included explicitly.

src/core/algorithms/nd/nd_verifier/nd_verifier.h
```diff
@@ -0,0 +1,82 @@
...
+class NDVerifier : public Algorithm {
+private:
+    config::InputTable input_table_;
+    config::IndicesType lhs_indices_;
+    config::IndicesType rhs_indices_;
+    model::WeightType weight_;
+    config::EqNullsType is_null_equal_null_;
+
+    using ValueType = std::string;
+    using CodeType = size_t;
```
yakovypg, line 31: I haven't found any use for these usings. If it is really not used, it should be removed.

src/core/algorithms/nd/nd.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+namespace model {
+
+std::string ND::ToShortString() const {
+    auto indices_to_short_string =
+            [](std::vector<model::ColumnIndex> const& indices) -> std::string {
+        return '[' +
+               std::accumulate(std::next(indices.begin()), indices.end(),
+                               std::to_string(indices[0]),
+                               [](std::string&& a, unsigned b) {
```
yakovypg, line 14: Use `ColumnIndex` instead of `unsigned`.

src/core/algorithms/nd/nd.h
```diff
@@ -0,0 +1,45 @@
+#pragma once
+
+#include "model/table/column.h"
+#include "model/table/vertical.h"
+
+namespace model {
+
+using WeightType = unsigned;
```
yakovypg, line 8: It is better to specify the full name of the type. In this case it is better to use `unsigned int`.

src/core/algorithms/nd/nd_verifier/util/highlight.cpp
```diff
@@ -0,0 +1,117 @@
...
+#include <numeric>
+#include <string>
+#include <vector>
+
+namespace algos::nd_verifier::util {
+
+std::vector<size_t> const& Highlight::GetMostFrequentRhsValueCodes() {
+    if (most_frequent_rhs_value_codes_ == nullptr) {
+        most_frequent_rhs_value_codes_ = std::make_unique<std::vector<size_t>>();
+        unsigned num_most_frequent_rhs_values{0};
```
yakovypg, line 14: `freq` is `size_t`, so `num_most_frequent_rhs_values` should has the same type.

src/core/algorithms/nd/nd_verifier/util/highlight.h
```diff
@@ -0,0 +1,84 @@
...
+    /// @brief Get cached indices of all occurences of this highlight of calculate it (heavy
+    /// operation)
+    [[nodiscard]] std::vector<size_t> const& GetOccurencesIndices();
+
+    /// @brief Get number of occurences of this highlight
+    [[nodiscard]] size_t GetOccurencesNumber() const;
+
+    /// @brief Get Lhs value
+    [[nodiscard]] std::string GetLhsValue() const {
+        return lhs_values_->at(highlight_lhs_);
```
yakovypg, line 60: Use `operator[]`.

src/core/algorithms/nd/nd_verifier/util/highlight.h
```diff
@@ -0,0 +1,84 @@
...
+
+    /// @brief Get most frequent rhs values -- heavy operation
+    [[nodiscard]] std::vector<std::string> GetMostFrequentRhsValues();
+
+    [[nodiscard]] std::string ToIndicesString();
+
+    [[nodiscard]] std::string ToValuesString();
+};
+
+std::ostream& operator<<(std::ostream& os, Highlight& hl);
```
yakovypg, line 82: `hl` should be const reference.

src/core/algorithms/nd/nd_verifier/util/highlight.h
```diff
@@ -0,0 +1,84 @@
...
+
+    /// @brief Get indices of most frequent rhs values -- heavy operation
+    [[nodiscard]] std::vector<size_t> GetMostFrequentRhsValueIndices();
+
+    /// @brief Get most frequent rhs values -- heavy operation
+    [[nodiscard]] std::vector<std::string> GetMostFrequentRhsValues();
+
+    [[nodiscard]] std::string ToIndicesString();
+
+    [[nodiscard]] std::string ToValuesString();
```
yakovypg, line 79: Why this method isn't const?

src/core/algorithms/nd/nd_verifier/util/highlight.h
```diff
@@ -0,0 +1,84 @@
...
+        return highlight_rhs_.size();
+    }
+
+    /// @brief Get indices of most frequent rhs values -- heavy operation
+    [[nodiscard]] std::vector<size_t> GetMostFrequentRhsValueIndices();
+
+    /// @brief Get most frequent rhs values -- heavy operation
+    [[nodiscard]] std::vector<std::string> GetMostFrequentRhsValues();
+
+    [[nodiscard]] std::string ToIndicesString();
```
yakovypg, line 77: Why this method isn't const?

src/core/algorithms/nd/nd_verifier/util/highlight.cpp
```diff
@@ -0,0 +1,117 @@
...
+        }
+    }
+
+    return *most_frequent_rhs_value_codes_;
+}
+
+[[nodiscard]] std::vector<size_t> const& Highlight::GetOccurencesIndices() {
+    if (occurences_indices_ == nullptr) {
+        occurences_indices_ = std::make_unique<std::vector<size_t>>();
+        for (size_t i{0}; i < encoded_lhs_->size(); ++i) {
```
yakovypg, line 33:
```cpp
std::copy_if(encoded_lhs_->begin(), encoded_lhs_->end(), std::back_inserter(*occurences_indices_), [highlight_lhs_](size_t v) {
    return v == highlight_lhs_;
});
```

src/core/algorithms/nd/nd_verifier/util/highlight.cpp
```diff
@@ -0,0 +1,117 @@
...
+            }
+        }
+    }
+
+    return *occurences_indices_;
+}
+
+[[nodiscard]] size_t Highlight::GetOccurencesNumber() const {
+    // This operation doesn't calculate occurences if they're not cached, so it's pretty fast
+    if (occurences_indices_ == nullptr) {
```
yakovypg, line 45: Use ternary operator.

src/core/algorithms/nd/nd_verifier/util/highlight.cpp
```diff
@@ -0,0 +1,117 @@
...
+        return std::count(encoded_lhs_->begin(), encoded_lhs_->end(), highlight_lhs_);
+    } else {
+        return occurences_indices_->size();
+    }
+}
+
+[[nodiscard]] std::vector<std::string> const& Highlight::GetRhsValues() {
+    if (highlight_rhs_values_ == nullptr) {
+        highlight_rhs_values_ = std::make_unique<std::vector<std::string>>();
+        for (auto rhs_code : highlight_rhs_) {
```
yakovypg, line 55: Use `std::copy`.

src/core/algorithms/nd/nd_verifier/util/highlight.cpp
```diff
@@ -0,0 +1,117 @@
...
+    return *highlight_rhs_values_;
+}
+
+[[nodiscard]] std::vector<size_t> Highlight::GetMostFrequentRhsValueIndices() {
+    std::vector<size_t> most_freq_rhs_inds;
+
+    auto const& most_freq_rhs_codes = GetMostFrequentRhsValueCodes();
+    for (size_t idx{0}; idx < encoded_rhs_->size(); ++idx) {
+        auto candidate_code = encoded_rhs_->at(idx);
+        for (auto code : most_freq_rhs_codes) {
```
yakovypg, line 69: Use `std::copy_if`.

src/core/algorithms/nd/nd_verifier/util/highlight.cpp
```diff
@@ -0,0 +1,117 @@
...
+        }
+    }
+
+    return *highlight_rhs_values_;
+}
+
+[[nodiscard]] std::vector<size_t> Highlight::GetMostFrequentRhsValueIndices() {
+    std::vector<size_t> most_freq_rhs_inds;
+
+    auto const& most_freq_rhs_codes = GetMostFrequentRhsValueCodes();
```
yakovypg, line 66: Use explicit type.

src/core/algorithms/nd/nd_verifier/util/highlight.cpp
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    return most_freq_rhs_inds;
+}
+
+[[nodiscard]] std::vector<std::string> Highlight::GetMostFrequentRhsValues() {
+    std::vector<std::string> most_freq_rhs_vals;
+
+    auto const& most_freq_rhs_codes = GetMostFrequentRhsValueCodes();
+    for (auto code : most_freq_rhs_codes) {
```
yakovypg, line 83: Use `std::copy`.

src/core/algorithms/nd/nd_verifier/util/highlight.h
```diff
@@ -0,0 +1,84 @@
...
+    /// @brief Find most frequent rhs values and cache them
+    std::vector<size_t> const& GetMostFrequentRhsValueCodes();
+
+public:
+    Highlight(std::shared_ptr<std::vector<std::string>> lhs_codes,
+              std::shared_ptr<std::vector<std::string>> rhs_codes,
+              std::shared_ptr<std::vector<size_t>> encoded_lhs,
+              std::shared_ptr<std::vector<size_t>> encoded_rhs,
+              std::shared_ptr<std::vector<size_t>> lhs_frequencies,
+              std::shared_ptr<std::vector<size_t>> rhs_frequencies, size_t highlight_lhs_code,
```
yakovypg, line 40: `size_t highlight_lhs_code` should be in new line.

src/core/algorithms/nd/nd_verifier/util/stats_calculator.cpp
```diff
@@ -0,0 +1,118 @@
...
+    lhs_frequencies_ =
+            CalculateFrequencies(lhs_values_->size(), encoded_lhs_->begin(), encoded_lhs_->end());
+    rhs_frequencies_ =
+            CalculateFrequencies(rhs_values_->size(), encoded_rhs_->begin(), encoded_rhs_->end());
+
+    std::vector<size_t> highlights_lhs_codes;
+    model::WeightType max_weight{0};
+    model::WeightType min_weight{UINT_MAX};
+    for (size_t code{0}; code < lhs_values_->size(); ++code) {
+        auto weight = value_deps_->at(code).size();
```
yakovypg, line 36: Use explicit type.

yakovypg, line 36: Use `operator[]`.

src/core/algorithms/nd/nd_verifier/util/stats_calculator.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+        assert(weight > 0);
```
yakovypg, line 38: `assert` macro makes use only for debug build. I think that it is better to use exceptions here if necessary.

src/core/algorithms/nd/nd_verifier/util/stats_calculator.cpp
```diff
@@ -0,0 +1,118 @@
...
+        } else if (weight == max_weight) {
+            highlights_lhs_codes.push_back(code);
+        }
+
+        if (weight < min_weight) {
+            min_weight = weight;
+        }
+    }
+
+    assert(max_weight > 0);
```
yakovypg, line 53: `assert` macro makes use only for debug build. I think that it is better to use exceptions here if necessary.

src/core/algorithms/nd/nd_verifier/util/stats_calculator.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+    for (auto highlight_lhs_code : highlights_lhs_codes) {
+        auto& rhs_set = value_deps_->at(highlight_lhs_code);
```
yakovypg, line 56: Use explicit type.

yakovypg, line 56: Use `operator[]`.

src/core/algorithms/nd/nd_verifier/util/stats_calculator.h
```diff
@@ -0,0 +1,85 @@
...
+    [[nodiscard]] auto GetRealWeight() const {
+        return real_weight_;
+    }
+
+    [[nodiscard]] auto GetLhsFrequencies() const {
+        assert(lhs_values_->size() == lhs_frequencies_->size());
+
+        std::unordered_map<std::string, size_t> result;
+        for (size_t i{0}; i < lhs_values_->size(); ++i) {
+            result.emplace(lhs_values_->at(i), lhs_frequencies_->at(i));
```
yakovypg, line 65: Use `operator[]`.

src/core/algorithms/nd/nd_verifier/util/stats_calculator.h
```diff
@@ -0,0 +1,85 @@
...
+
+        return result;
+    }
+
+    [[nodiscard]] auto GetRhsFrequencies() const {
+        assert(rhs_values_->size() == rhs_frequencies_->size());
+
+        std::unordered_map<std::string, size_t> result;
+        for (size_t i{0}; i < rhs_values_->size(); ++i) {
+            result.emplace(rhs_values_->at(i), rhs_frequencies_->at(i));
```
yakovypg, line 76: Use `operator[]`.

src/core/algorithms/nd/nd_verifier/util/stats_calculator.h
```diff
@@ -0,0 +1,85 @@
...
+    [[nodiscard]] auto GetGlobalMinWeight() const {
+        return global_min_weight_;
+    }
+
+    [[nodiscard]] auto GetRealWeight() const {
+        return real_weight_;
+    }
+
+    [[nodiscard]] auto GetLhsFrequencies() const {
+        assert(lhs_values_->size() == lhs_frequencies_->size());
```
yakovypg, line 61: `assert` macro makes use only for debug build. I think that it is better to use exceptions here if necessary.

src/core/algorithms/nd/nd_verifier/util/stats_calculator.h
```diff
@@ -0,0 +1,85 @@
...
+        std::unordered_map<std::string, size_t> result;
+        for (size_t i{0}; i < lhs_values_->size(); ++i) {
+            result.emplace(lhs_values_->at(i), lhs_frequencies_->at(i));
+        }
+
+        return result;
+    }
+
+    [[nodiscard]] auto GetRhsFrequencies() const {
+        assert(rhs_values_->size() == rhs_frequencies_->size());
```
yakovypg, line 72: `assert` macro makes use only for debug build. I think that it is better to use exceptions here if necessary.

src/core/algorithms/nd/nd_verifier/util/stats_calculator.h
```diff
@@ -0,0 +1,85 @@
...
+
+        std::unordered_map<std::string, size_t> result;
+        for (size_t i{0}; i < lhs_values_->size(); ++i) {
+            result.emplace(lhs_values_->at(i), lhs_frequencies_->at(i));
+        }
+
+        return result;
+    }
+
+    [[nodiscard]] auto GetRhsFrequencies() const {
```
yakovypg, line 71: Use explicit type.

yakovypg, line 71: Big methods should be placed in .cpp file.

src/core/algorithms/nd/nd_verifier/util/stats_calculator.h
```diff
@@ -0,0 +1,85 @@
...
+
+    [[nodiscard]] auto GetGlobalMinWeight() const {
+        return global_min_weight_;
+    }
+
+    [[nodiscard]] auto GetRealWeight() const {
+        return real_weight_;
+    }
+
+    [[nodiscard]] auto GetLhsFrequencies() const {
```
yakovypg, line 60: Big methods should be placed in .cpp file.

src/tests/test_nd_verifier.cpp
```diff
@@ -0,0 +1,71 @@
...
+#include "config/names.h"
+
+namespace tests {
+namespace onam = config::names;
+
+struct NDVerifyingParams {
+    algos::StdParamsMap params;
+
+    NDVerifyingParams(config::IndicesType lhs_indices, config::IndicesType rhs_indices,
+                      model::WeightType weight = -1, CSVConfig const& csv_config = kTestND,
```
yakovypg, line 16: `model::WeightType` is `unsigned`. So you should be careful with `-1`.

If you are looking to get the maximum value of a certain type, it is better to use the std::numeric_limits class:

```cpp
#include <limits>
T max = std::numeric_limits<T>::max()
```

You can also use `static_cast<unsigned>(-1)`.

src/tests/test_nd_verifier.cpp
```diff
@@ -0,0 +1,71 @@
...
+        ));
+
+INSTANTIATE_TEST_CASE_P(
+    NDVerifierTestNullEqualNull, TestNDVerifying,
+    ::testing::Values(  // 6-th column contains 2 values and 7 empty cells
+        NDVerifyingParams({0}, {6}, 3, kTestND, true),
+        NDVerifyingParams({0}, {6}, 7, kTestND, false)
+    ));
+
+#if 0
```
yakovypg, line 61: I think that the final code should only contain what will actually be used.

src/core/algorithms/nd/nd_verifier/util/highlight.cpp
Petua41: I can't use std::copy_if here, because I need to copy indices, not values

src/core/algorithms/nd/nd_verifier/nd_verifier.cpp
```diff
@@ -0,0 +1,228 @@
...
+        ss << ')';
+
+        auto string_data = ss.str();
+        size_t index{values->size()};
+        if (is_null && !is_null_equal_null_) {
+            // If not null_eq_null, every value with null will be unique
+            values->push_back(string_data);
+        } else {
+            for (size_t i{0}; i < values->size(); ++i) {
+                if (values->at(i) == string_data) {
```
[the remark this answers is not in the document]
Petua41: Here I don't need to compare two given values. I need to check if the result (vector of **strings**) contain given value (that needs to be converted to string later)

yakovypg (commented):
I returned to reviewing the code and have the following thoughts:
- Method `.at()` provides bound check, but this can be useful during the implementation of the algorithm to catch errors during debugging. In other situations we don't want to have overhead for these checks, so it is preferable to use `operator[]`. In theory, method `.at()` can be used when the index comes from the user-side to avoid undefined behavior in case of an invalid index. However, in this case it is preferable to use special checks and throw special exceptions.
- The previous arguments are also true for smart pointers. In this case, you should write `(*ptr)[index]` instead of `ptr->operator[](index)`. If the expression is very long, you can define additional variable (e.g. reference to `vector` in `shared_ptr`) and use it without pointer dereference.

src/core/algorithms/nd/nd_verifier/nd_verifier.cpp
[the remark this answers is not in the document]
yakovypg: - Method .at() provides bound check, but this can be useful during the implementation of the algorithm to catch errors during debugging. In other situations we don't want to have overhead for these checks, so it is preferable to use operator[]. In theory, method .at() can be used when the index comes from the user-side to avoid undefined behavior in case of an invalid index. However, in this case it is preferable to use special checks and throw special exceptions.
- The previous arguments are also true for smart pointers. In this case, you should write (*ptr)[index] instead of ptr->operator[](index). If the expression is very long, you can define additional variable (e.g. reference to vector in shared_ptr) and use it without pointer dereference.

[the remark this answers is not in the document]
yakovypg: But why you store string representation instead of custom structs?

src/core/algorithms/nd/nd_verifier/util/highlight.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+#include "algorithms/nd/nd_verifier/util/vector_to_string.h"
+
+namespace algos::nd_verifier::util {
+
+std::unordered_set<size_t> const& Highlight::GetMostFrequentRhsValueCodes() {
+    if (most_frequent_rhs_value_codes_ == nullptr) {
+        most_frequent_rhs_value_codes_ = std::make_unique<std::unordered_set<size_t>>();
+        size_t num_most_frequent_rhs_values{0};
+        for (auto rhs_code : highlight_rhs_) {
```
yakovypg, line 18: Use explicit type.

src/core/algorithms/nd/nd_verifier/nd_verifier.cpp
```diff
@@ -0,0 +1,210 @@
...
+    return std::make_pair(std::move(values), std::move(row));
+}
+
+NDVerifier::EncodedValuesType NDVerifier::EncodeSingleValues(config::IndexType col_idx) const {
+    auto values = std::make_shared<std::vector<std::string>>();
+    auto row = std::make_shared<std::vector<size_t>>();
+
+    auto const& col_data = typed_relation_->GetColumnData(col_idx);
+    auto const& byte_data = col_data.GetData();
+    auto const& type = col_data.GetType();
```
yakovypg, line 138: It is better to use explicit type here.

src/core/algorithms/nd/nd_verifier/nd_verifier.cpp
```diff
@@ -0,0 +1,210 @@
...
+
+    for (auto const* bytes_ptr : byte_data) {
```
yakovypg, line 140: It is better to use explicit type here.

src/core/algorithms/nd/nd_verifier/util/highlight.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+    return *highlight_rhs_values_;
+}
+
+[[nodiscard]] std::vector<size_t> Highlight::GetMostFrequentRhsValueIndices() {
+    std::vector<size_t> most_freq_rhs_inds;
+
+    std::unordered_set<size_t> const& most_freq_rhs_codes = GetMostFrequentRhsValueCodes();
+    for (size_t idx{0}; idx < encoded_rhs_->size(); ++idx) {
+        auto candidate_code = encoded_rhs_->at(idx);
```
yakovypg, line 58: It is better to use explicit type here.

src/core/algorithms/nd/nd_verifier/util/highlight.cpp
```diff
@@ -0,0 +1,94 @@
...
+    return most_freq_rhs_vals;
+}
+
+[[nodiscard]] std::string Highlight::ToIndicesString() {
+    return util::VectorToString(GetOccurencesIndices());
+}
+
+[[nodiscard]] std::string Highlight::ToValuesString() {
+    auto const& lhs = GetLhsValue();
+    auto const& rhs = GetRhsValues();
```
yakovypg, line 84: It is better to use explicit type here.

src/core/algorithms/nd/nd_verifier/util/highlight.cpp
```diff
@@ -0,0 +1,94 @@
...
+            most_freq_rhs_inds.push_back(idx);
+        }
+    }
+
+    return most_freq_rhs_inds;
+}
+
+[[nodiscard]] std::vector<std::string> Highlight::GetMostFrequentRhsValues() {
+    std::vector<std::string> most_freq_rhs_vals;
+    auto const& most_freq_rhs_codes = GetMostFrequentRhsValueCodes();
```
yakovypg, line 69: It is better to use explicit type here.

src/core/algorithms/nd/nd_verifier/util/highlight.h
```diff
@@ -0,0 +1,86 @@
...
+          highlight_lhs_(highlight_lhs_code),
+          highlight_rhs_(std::move(highlight_rhs_codes)) {}
+
+    /// @brief Get cached indices of all occurences of this highlight of calculate it (heavy
+    /// operation)
+    [[nodiscard]] std::vector<size_t> const& GetOccurencesIndices();
+
+    /// @brief Get number of occurences of this highlight
+    [[nodiscard]] size_t GetOccurencesNumber() {
+        return GetOccurencesIndices().size();
```
yakovypg, line 57: I think the implementation of this method should be placed in `.cpp` file due to dependence of other method.

src/core/algorithms/nd/nd_verifier/nd_verifier.cpp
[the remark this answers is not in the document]
Petua41: The only purpose of theese strings is for user to see, which values violate the ND. Algorithm doesn't use them anywhere, it works with internal indices. So there is no point in creating custom structs for each row (and consume a lot of memory)

src/core/algorithms/nd/nd_verifier/util/stats_calculator.h
```diff
@@ -0,0 +1,68 @@
...
+    std::shared_ptr<std::vector<size_t>> encoded_rhs_;
+
+    // Cached data:
+    std::vector<Highlight> highlights_;
+    std::shared_ptr<std::vector<size_t>> lhs_frequencies_;
+    std::shared_ptr<std::vector<size_t>> rhs_frequencies_;
+    model::WeightType global_min_weight_{UINT_MAX};
+    model::WeightType real_weight_{0};
+
+    template <typename It>
```
yakovypg, line 31: Template methods shouldn't be sepated into header file and .cpp file.

src/core/algorithms/nd/nd_verifier/util/stats_calculator.cpp
```diff
@@ -0,0 +1,99 @@
...
+            LOG(INFO) << "WARNING: Lhs value '" << (*lhs_values_)[code]
+                      << "''s weight is 0, but it's present in value_deps";
+        }
+
+        if (weight > max_weight) {
+            max_weight = weight;
+            highlights_lhs_codes.clear();
+            highlights_lhs_codes.push_back(code);
+        } else if (weight == max_weight) {
+            highlights_lhs_codes.push_back(code);
```
yakovypg, line 76: I see a possible large number of unnecessary operations. For example, consider sequence [1,...,1,2,...,2,...,n,...,n]. You will perform a lot of operations (`clear()`, `push_back()`), but only codes that correspond to the maximum weight are important. So we should avoid unnecessary operations. For example, we can calculate minimum and maximum (with their indices) and only then update `highlights_lhs_codes`. If possible, you can try to use `std::minmax_element()` or `std::max_element`.

src/core/algorithms/nd/nd_verifier/util/stats_calculator.cpp
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        if (weight < min_weight) {
+            min_weight = weight;
+        }
+    }
+
+    for (auto highlight_lhs_code : highlights_lhs_codes) {
```
yakovypg, line 84: Use explicit type.

src/core/algorithms/nd/nd_verifier/util/highlight.cpp
```diff
@@ -0,0 +1,102 @@
...
+        most_frequent_rhs_value_codes_ = std::make_unique<std::unordered_set<size_t>>();
+        size_t num_most_frequent_rhs_values{0};
+        for (size_t rhs_code : highlight_rhs_) {
+            size_t freq = (*rhs_frequencies_)[rhs_code];
+            if (freq > num_most_frequent_rhs_values) {
+                num_most_frequent_rhs_values = freq;
+                most_frequent_rhs_value_codes_->clear();
+                most_frequent_rhs_value_codes_->insert(rhs_code);
+            } else if (freq == num_most_frequent_rhs_values) {
+                most_frequent_rhs_value_codes_->insert(rhs_code);
```
yakovypg, line 25: Likewise. See comment in stats_calculator.cpp.

src/core/algorithms/nd/nd_verifier/nd_verifier.cpp
```diff
@@ -0,0 +1,234 @@
...
+                    index = i;
+                    break;
+                }
+            }
+
+            if (index == values->size()) {  // wasn't in codes
+                values->push_back(string_data);
+            }
+        }
+        row->push_back(index);
```
yakovypg, line 131: I see duplication of this code. Can we separate big loop into two parts (move each part into a separate method) and then use one of these methods here and in `EncodeSingleValues()` method?

yakovypg (commented):
I think it isn't a good idea to store string representation of NDs and lhs/rhs values/indices as strings (even if these values will ultimately only be used for output to the user). It is better to use a custom structure to store this.
Some reasons:
- Strings take up a lot of memory.
- Working with strings is time-consuming.
- Using a custom structure makes the code more scalable.
- There will be no dependency on a string representation (it may change over time).

yakovypg (commented):
I think you should create more tests for different situations. I see that a lot of tests use attribute 0 as lhs, but it is a constant attribute (1,1,...,1). It would be better to consider situations where lhs is a non-constant attribute and a little more situations where lhs/rhs consists of several attributes. Furthermore, it would be nice to test situations where ND isn't hold.

yakovypg:
Now it's time to see your commit history. It isn't good. For example, you add `ValueCombination` in your last commit, but this class is necessary for `NDVerifier` that appeared earlier in the commit history. I can say the same things about `VectorToString` that is necessary for `ND` class. Furthermore, there is no need to create a separate commit where some small class is added. Each commit should reflect some complete logical part of innovations or corrections. Since you only added the verifier and tests, your commit history should probably consist of only two or three commits.

See examples of good commit history: [PR304](https://github.com/Desbordante/desbordante-core/pull/304/commits), [PR374](https://github.com/Desbordante/desbordante-core/pull/374/commits).

yakovypg (approved):

PR #396, opened 2024-04-21
Fix TANE-based algorithms

iliya-b:


src/core/algorithms/fd/tane/tane_common.cpp
```diff
@@ -43,46 +43,40 @@ void TaneCommon::Prune(model::LatticeLevel* level) {
...
-                                // for each outer rhs: if there is a sibling s.t. it doesn't
-                                // have this rhs, there is no FD: vertex->rhs
-                            }
-                            // Found fd: vertex->rhs => register it
-                            if (is_rhs_candidate) {
-                                RegisterAndCountFd(columns, schema->GetColumn(rhs_index));
+                for (std::size_t rhs_index = vertex->GetRhsCandidates().find_first();
+                     rhs_index != boost::dynamic_bitset<>::npos;
+                     rhs_index = vertex->GetRhsCandidates().find_next(rhs_index)) {
+                    Vertical rhs = static_cast<Vertical>(*schema->GetColumn((int)rhs_index));
```
vs9h, line 51: nit: remove `(int)` cast, since we should pass `std::size_t`

src/core/algorithms/fd/tane/tane_common.cpp
```diff
@@ -43,46 +43,40 @@ void TaneCommon::Prune(model::LatticeLevel* level) {
...
         }
-        // if we seek for exact FDs then SetInvalid
-        if (max_fd_error_ == 0 && max_ucc_error_ == 0) {
-            for (auto key_vertex : key_vertices) {
-                key_vertex->GetRhsCandidates() &= key_vertex->GetVertical().GetColumnIndices();
-                key_vertex->SetInvalid(true);
-            }
+        for (auto key_vertex : key_vertices) {
+            key_vertex->GetRhsCandidates() &= key_vertex->GetVertical().GetColumnIndices();
+            key_vertex->SetInvalid(true);
```
vs9h, line 79: Can you explain why we can remove this additional `max_fd_error_` check?

vs9h (commented):
I can't say that I fully remember the essence of the algorithm and can fully verify the changes.
Can you provide useful links? I would like to study in more detail what is happening in these methods.

Also, is it known how much these changes affect performance?

In general, I think the changes are adequate and most likely after I study in more detail what is happening here, it will be possible to merge.

PR #399, opened 2024-04-22
Add script for testing examples

DaniilGoncharov:
Add script for testing examples (without gfd).  Bindings now return sorted list of fds (only for algos inherited from FDAlgorithm). Update README to the current output


src/core/model/table/vertical.h
```diff
@@ -42,7 +42,9 @@ class Vertical {
      * it treats bitsets little endian during comparison and this is not
      * suitable for this case, check out operator< for Columns.
      */
-    bool operator<(Vertical const& rhs) const;
+    bool operator<(Vertical const& rhs) const {
+        return column_indices_ < rhs.column_indices_;
+    }
```
polyntsov, line 47: Why do we change this operator< behavior?
Given the comment above (`@brief` section), we are not using operator< of bitsets intentionally

examples/testing/test_examples.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
```
polyntsov, line 1: Looks like we reinventing cram tests: https://bitheap.org/cram/

DaniilGoncharov:
Moved to #525

PR #400, opened 2024-04-23
Add AR to console

Sneper-Breeze:


cli/cli.py
```diff
@@ -779,12 +803,14 @@ def desbordante_cli(**kwargs: Any) -> None:
...
     error_opt = kwargs[ERROR]
     error_measure_opt = kwargs[ERROR_MEASURE]
+    input_format_opt = kwargs[INPUT_FORMAT]
     verbose = kwargs[VERBOSE]
     filename = kwargs[FILENAME]
 
     check_mismatch(curr_algo_name, curr_task)
     check_error_option_presence(curr_task, error_opt)
     check_error_measure_option_presence(curr_task, error_measure_opt)
+    check_input_format_option_presence(curr_task, input_format_opt)
```
BUYT-1, line 813: Isn't the algorithm complaining about this option not being present enough?

cli/cli.py
```diff
@@ -443,6 +457,8 @@ class Algorithm(StrEnum):
     Task.pfd: TaskInfo([Algorithm.pfdtane], Algorithm.pfdtane),
     Task.ind: TaskInfo([Algorithm.spider, Algorithm.faida],
                        Algorithm.spider),
+    Task.ar: TaskInfo([Algorithm.apriori],
+                       Algorithm.apriori),
```
BUYT-1, line 461:
```suggestion
                      Algorithm.apriori),
```

cli/cli.py
Sneper-Breeze: I little bit dont understand what about talking. If it is needed, then yes, the algorithm cannot work without this option.

BUYT-1: I'm saying that because the algorithm itself errors out if one of the options without a default value, which I presume 'input_format' is, is not set, there shouldn't be a need for an additional check in the CLI.

cli/cli.py
```diff
@@ -64,6 +66,7 @@ class Algorithm(StrEnum):
 TABLES = 'tables'
 TABLES_LIST = 'tables_list'
 TABLES_DIRECTORY = 'tables_directory'
+INPUT_FORMAT = "input_format"
```
BUYT-1, line 69:
```suggestion
INPUT_FORMAT = 'input_format'
```

cli/cli.py
```diff
@@ -554,7 +571,6 @@ def parse_tables_directory(tp: tuple[click.Path, str, bool]) \
         return [(dir_entry.path, separator, has_header) for dir_entry in entries]
     except OSError as exc:
         click.echo(exc)
-        sys.exit(1)
```
BUYT-1, line 557: shouldn't be removed

Sneper-Breeze: Sorry, probably missclick.

cli/cli.py
```diff
@@ -779,6 +798,7 @@ def desbordante_cli(**kwargs: Any) -> None:
     curr_algo = ALGOS[curr_algo_name]()
     error_opt = kwargs[ERROR]
     error_measure_opt = kwargs[ERROR_MEASURE]
+    input_format_opt = kwargs[INPUT_FORMAT]
```
BUYT-1, line 801: unused

BUYT-1 (approved):

PR #401, opened 2024-04-23
Implement `Mind` algorithm

vs9h:
This pull request adds an algorithm for mining nary approximate inclusion dependencies `Mind`.
The algorithm is described in the article "Unary and n-ary inclusion dependency discovery in relational databases" (Fabien De Marchi, Stéphane Lopes, Jean-Marc Petit)

Commits contain comments, and there are also quite a lot of explanatory comments left in the code itself. I would advise reviewers not to look at the entire diff at once, but to go through the commits sequentially.


src/core/algorithms/ind/mind/mind.h
```diff
@@ -0,0 +1,63 @@
+/** \file
+ * \brief Mind algorithm
+ *
+ * Spider algorithm class definition
```
vs9h, line 4:
```suggestion
 * Mind algorithm class definition
```

src/core/algorithms/ind/mind/mind.cpp
```diff
@@ -0,0 +1,271 @@
+/** \file
+ * \brief Spider algorithm
+ *
+ * Spider algorithm class methods definition
```
vs9h, line 4: same

src/core/algorithms/ind/ind.h
```diff
@@ -30,6 +30,14 @@ class IND {
         return *rhs_;
     }
 
+    bool StartsWith(IND const& other) const noexcept {
+        return GetLhs().StartsWith(other.GetLhs()) && GetRhs().StartsWith(other.GetRhs());
+    }
+
+    ArityIndex GetArity() const {
```
slesarev-hub, line 37:
```suggestion
    ArityIndex GetArity() const noexcept {
```

src/core/algorithms/ind/mind/mind.cpp
```diff
@@ -0,0 +1,271 @@
...
+
+void Mind::MakeExecuteOptsAvailable() {
+    MakeOptionsAvailable({config::kErrorOpt.GetName(), config::kMaxArityOpt.GetName()});
+}
+
+bool Mind::SetExternalOption(std::string_view option_name, boost::any const& value) {
+    try {
+        auind_algo_->SetOption(option_name, value);
+        return true;
+    } catch (config::ConfigurationError&) {
```
slesarev-hub, line 47: Should we handle this error somehow? Or just return false?

src/core/model/table/block_dataset_stream.h
```diff
@@ -46,18 +41,27 @@ class BlockDatasetStream {
...
-    BlockDatasetStream(std::shared_ptr<IDatasetStream> const& stream, size_t capacity)
-        : stream_(stream), capacity_(capacity) {
-        TryStoreNextRow();
-    }
+    template <typename Stream>
+    BlockDatasetStream(Stream&& stream, size_t capacity)
+        : stream_(std::forward<Stream>(stream)), capacity_(capacity) {}
 
     /// check if there is a next block available in the dataset stream
     bool HasNextBlock() const {
```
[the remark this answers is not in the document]
vs9h: I've added `[[nodiscard]]` here and elsewhere, similar to how they are defined in the `IDatasetStream` class.

src/core/model/table/column_combination.h
```diff
@@ -28,6 +30,52 @@ class ColumnCombination {
...
+        std::vector<model::ColumnIndex> const& indices = GetColumnIndices();
+        return std::accumulate(indices.begin(), indices.end(), indices.size(),
+                               [](std::size_t seed, model::ColumnIndex i) {
+                                   return seed ^ (i + 0x9e3779b9 + (seed << 6) + (seed >> 2));
+                               });
+    }
+
+    bool StartsWith(ColumnCombination const& other) const {
+        return this->table_index_ == other.table_index_ &&
+               std::equal(this->column_indices_.begin(), this->column_indices_.end() - 1,
```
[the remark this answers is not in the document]
vs9h: We do not check for equality of the last index

src/core/algorithms/ind/mind/mind.cpp
vs9h: This function is needed to try setting the option for the unary algorithm. If it doesn’t work out, then we’ll return the false here and then (in the `Algorithm::SetOption`) try to set it up for the mind. If it doesn’t work there, we’ll throw a configuration error to the user.
At the same time, we call this function for all passed options, including those that may be specific to the `Mind`. In this case, when trying to set it for a unary algorithm, we may get an error that we should simply suppress.

src/core/algorithms/ind/mind/mind.cpp
```diff
@@ -0,0 +1,271 @@
...
+     * Stop INDs mining if no new dependencies were found at the previous lattice level
+     * (from this condition it follows that no more dependencies can be found).
+     */
+    while (prev_it != INDList().end() && INDList().back().GetArity() != max_arity_) {
+        for (auto p_it = prev_it; p_it != INDList().end(); ++p_it) {
+            std::for_each(std::next(p_it), INDList().end(), [&](const IND& q) {
+                std::optional<RawIND> candidate_opt =
+                        mind::GetCandidateIfValid(*p_it, q, prev_raw_inds);
+                if (candidate_opt) {
+                    candidates.push_back(candidate_opt.value());
```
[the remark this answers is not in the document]
vs9h: i agree with `std::move`, but not with `emplace_back`. We should use `emplace_back`, when we want to construct the element in-place, by passing parameters to create an object. In this case, we already have a `RawIND` object - for this case it is semantically better to use `push_back`.

slesarev-hub (approved):

vs9h:
I pushed a commit adding a `Mind` to the python bindings.
I think I’ll add the cli and examples in a separate pull request.
I might also try to speed up this algorithm and add `error` field to the `IND` class (it will be much more informative for users).

slesarev-hub (approved):

PR #402, opened 2024-04-25
Use full header names in Python bindings

BUYT-1:
I've made this error when remaking the bindings and now see it copied by others. Hopefully, not after this.


BUYT-1:
Superseded by #425

PR #405, opened 2024-05-01
Add tests for cli

DaniilGoncharov:
Add tests for cli parsing options and help pages


cli/cli_tests/test_cli_help_pages.py
polyntsov, line 1: At first glance looks good, but note that we also want to include these tests (along with example tests) into CI pipeline. Please think how to do it :)

PR #407, opened 2024-05-03
Add example for AR

Sneper-Breeze:


examples/mining_ar.py
```diff
@@ -0,0 +1,110 @@
...
+          'That happened since minsup=0.5, which basically says that association rule needs '
+          'to present in 50% of all transactions in order to be included in the result. '
+          'Unfortunately, you can\'t obtain support for each individual association rule with Desbordante.\n'
+          '\nA typical approach to controlling the algorithm is to employ "usefulness", which is defined as confidence * support. '
+          'In last example we set up min "usefulness" = 0.5 * 0.5 = 0.25. \n\nNow lets try with '
+          f'{COLOR_CODES["green"]}minsup=0.7{COLOR_CODES["default"]} {COLOR_CODES["green"]}minconf=0.7{COLOR_CODES["default"]}, '
+          f'so {COLOR_CODES["bold_yellow"]}"usefulness"=0.49{COLOR_CODES["default"]}\n')
+    algo.execute(minsup=0.7, minconf=0.7)
+    print_ars(algo.get_ars())
+    print('\nSo, now the total number of returned association rules is only four. '
```
Darderion, line 80: Inconsistent usage of "ARs" and "association rules"

examples/mining_ar.py
```diff
@@ -0,0 +1,110 @@
...
+          'We efficiently reduced the amount of "noisy" information in our output. '
```
Darderion, line 81: How is it "efficient"?

examples/mining_ar.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def scenario_singular():
+    algo = desbordante.ar.algorithms.Default()
+    algo.load_data(table=(TABLE_SINGULAR, ',', True), input_format='tabular')
+    algo.execute()
+    table = pandas.read_csv(TABLE_SINGULAR, header=None, index_col=0)
+    #help(algo)
+
+    print('\nFor the second example lets take a dataset with the same receipts from '
```
Darderion, line 94: Same as other comments

examples/mining_ar.py
```diff
@@ -0,0 +1,110 @@
...
+          'the same supermarket with input_format="singular". This is two-column format, where '
```
Darderion, line 95: Article

examples/mining_ar.py
```diff
@@ -0,0 +1,110 @@
...
+          'second column denotes item and the first one denotes the transaction it belongs to.\n')
```
Darderion, line 96: Article and order of items

examples/mining_ar.py
```diff
@@ -0,0 +1,110 @@
...
+    print_table(table)
+    print('\nThis format is just a different table representation. Desbordante is perfectly capable '
+          'of association rule discovery from such tables, and the found rules are the same '
```
Darderion, line 99:
```
the discovered rules
```

examples/mining_ar.py
```diff
@@ -0,0 +1,110 @@
...
+          'objects as in previous examples (i.e., they possess the same methods and fields).\n')
```
Darderion, line 100: Article

examples/mining_ar.py
```diff
@@ -0,0 +1,110 @@
...
+    print('In this example we show you how to list all unique items in the dataset.')
```
Darderion, line 101: Same as other comments

examples/mining_ar.py
```diff
@@ -0,0 +1,110 @@
...
+    print(f'For this you can use {COLOR_CODES["bold_blue"]}get_itemnames{COLOR_CODES["default"]} method:\n')
+    print_itemnames(algo.get_itemnames())
```
Darderion, line 103: No such method

examples/mining_ar.py
```diff
@@ -0,0 +1,110 @@
...
+    print('\nNow you have all items used in this dataset, so you can filter it and then print.\n')
```
Darderion, line 104: Missing a noun

Darderion (changes requested):

examples/mining_ar.py
Sneper-Breeze: But it is there and it works.

examples/mining_ar.py
Sneper-Breeze: I didn`t realize which word in this line I needed to replace with "article".

Darderion: The sentence lacks article, as in "This is a two-column format"

examples/mining_ar.py
```diff
@@ -0,0 +1,128 @@
...
+
+    print("\n\nNow, let's examine the same dataset with "
+          f'{COLOR_CODES["yellow"]}minconf=0.7{COLOR_CODES["default"]}')
+    algo.execute(minconf=0.7)
+    print_ars(algo.get_ars())
+    print("\n['Milk'] -> ['Eggs'] with confidence 0.75 means that when milk "
+          'is found in the receipt, the chance of eggs being '
+          'present amounts to 75%. So, customers are '
+          f'{COLOR_CODES["bold_yellow"]}likely{COLOR_CODES["default"]} to buy eggs '
+          f'with milk, but {COLOR_CODES["bold_red"]}not always.{COLOR_CODES["default"]}')
```
Darderion, line 71: "but not always" is unnecessary here as it directly follows from context

examples/mining_ar.py
```diff
@@ -0,0 +1,128 @@
...
+
+    print('\nLet us turn to the next issue. You can observe that there are too many '
```
Darderion, line 73: "too many" can be changed to "a lot of"

examples/mining_ar.py
```diff
@@ -0,0 +1,128 @@
...
+          'association rules found in this small dataset. This happens because we did not '
+          'set up the support value. Due to this fact, the system discovers '
```
Darderion, line 75: "support value" can be highlighted as it is a term that was not yet introduced. There's also no explanation that would directly follow said term

"set up the so called support value. This is [support value's definition]"

Adding this would also require you to change the next sentence to accommodate for the change in flow of text here

examples/mining_ar.py
```diff
@@ -0,0 +1,128 @@
...
+          'all association rules, even those which are present only '
```
Darderion, line 76: You could change "which" to "that"

examples/mining_ar.py
```diff
@@ -0,0 +1,128 @@
...
+          'once in the dataset. Now let\'s see the results with '
```
Darderion, line 77: Quotation marks as well as a missing comma

examples/mining_ar.py
```diff
@@ -0,0 +1,128 @@
...
+          f'{COLOR_CODES["yellow"]}minsup=0.5{COLOR_CODES["default"]} and '
+          f'{COLOR_CODES["yellow"]}minconf=0.5{COLOR_CODES["default"]}.\n')
+    algo.execute(minsup=0.5, minconf=0.5)
+    print_ars(algo.get_ars())
+    print('\nNow you can see that the number of association rules have decreased '
+          'significantly. This happened due to minsup being set to 0.5, which basically '
+          'means that association rule needs to are present in 50% of all transactions '
```
Darderion, line 84: Adding a definition to the support value line would necessitate changes here to remove redundant information

examples/mining_ar.py
```diff
@@ -0,0 +1,128 @@
...
+          "in order to be included in the result. Unfortunately, you can't get support "
```
Darderion, line 85: What does it mean to "get" support values? Earlier the explanation had expanded on how this support value is supposed to be SET, meaning you do not acquire it from data, but instead choose it. If it was about setting minimum support value, then adding that information would be useful as well

If it's instead "setting support for individual association rules", then correct this line accordingly

examples/mining_ar.py
```diff
@@ -0,0 +1,128 @@
...
+          'for each individual association rule with Desbordante.\n'
+          '\nA typical approach to controlling the algorithm is to employ "usefulness", '
```
Darderion, line 87: "usefulness" could be highlighted

examples/mining_ar.py
```diff
@@ -0,0 +1,128 @@
...
+          'which is defined as confidence * support. In the last example, we set up min '
+          '"usefulness" = 0.5 * 0.5 = 0.25. \n\nNow let\'s try with '
```
Darderion, line 89: You could add comma after "Now"

examples/mining_ar.py
```diff
@@ -0,0 +1,128 @@
...
+          f'{COLOR_CODES["green"]}minsup=0.7{COLOR_CODES["default"]}, '
+          f'{COLOR_CODES["green"]}minconf=0.7{COLOR_CODES["default"]} '
+          f'and {COLOR_CODES["bold_yellow"]}"usefulness"=0.49{COLOR_CODES["default"]}\n')
+    algo.execute(minsup=0.7, minconf=0.7)
+    print_ars(algo.get_ars())
+    print('\nSo, now the total number of returned association rules is only four. '
+          'We reduced the amount of "noisy" information in our output. However, '
+          'this is just a single example of how you can evaluate the usefulness of data. '
```
Darderion, line 97: You typically do not "evaluate" a parameter, as you are setting it to get some other results. You could omit this sentence or rewrite it if you deem this information to be of use to the user

examples/mining_ar.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def scenario_singular():
+    algo = desbordante.ar.algorithms.Default()
+    algo.load_data(table=(TABLE_SINGULAR, ',', True), input_format='tabular')
+    algo.execute()
+    table = pandas.read_csv(TABLE_SINGULAR, header=None, index_col=0)
+
+    print("\nFor the second example, let's take a dataset with the same receipts "
+          'from the same supermarket with input_format="singular". '
```
Darderion, line 109: "from the same supermarket, changing the input format to" could work better here as it highlights the difference between two examples

examples/mining_ar.py
```diff
@@ -0,0 +1,128 @@
...
+          'This is a two-column format, where the second column denotes '
```
Darderion, line 110: Depending on the meaning of "order" here, this sentence still requires a rewrite- partially because it would be preferable to describe columns in order of their arrival

examples/mining_ar.py
```diff
@@ -0,0 +1,128 @@
...
+          'item and the first one denotes the order of items it belongs to.\n')
+    print_table(table)
+    print('\nThis format is just a different table representation. Desbordante '
+          'is perfectly capable of the discovered rules from such tables, '
```
Darderion, line 114: It's "association rules discovery" here though, the comment was referring to rewriting "found rules"

examples/mining_ar.py
```diff
@@ -0,0 +1,128 @@
...
+          'and the found rules are the same objects as in the previous '
```
Darderion, line 115: Depending on the context, you could write it as "represented by the same type of objects"

examples/mining_ar.py
```diff
@@ -0,0 +1,128 @@
...
+          'examples (i.e., they possess the same methods and fields).\n')
```
Darderion, line 116: You could write "have" instead of "possess" here as it is a bit more common to use in the context of classes

examples/mining_ar.py
```diff
@@ -0,0 +1,128 @@
...
+    print('In this example, we show you how to list all of the unique items in the dataset.')
+    print(f'For this you can use {COLOR_CODES["bold_blue"]}'
```
Darderion, line 118: You could write "In order to do this, you can"

examples/mining_ar.py
```diff
@@ -0,0 +1,128 @@
...
+          f'get_itemnames{COLOR_CODES["default"]} method:\n')
+    print_itemnames(algo.get_itemnames())
+    print('\nNow you have all items used in this dataset, '
```
Darderion, line 121: You could write "all of the items" here

examples/mining_ar.py
```diff
@@ -0,0 +1,128 @@
...
+          'so you can filter it and then print it.\n')
```
Darderion, line 122: Depending on the context, you could write "allowing you to filter them and then print individual entries"

Darderion (changes requested):

examples/mining_ar.py
```diff
@@ -0,0 +1,132 @@
...
+        print(title)
+
+    print(tabulate(table, tablefmt='pipe', stralign='center'))
+
+
+def print_ars(ars):
+    print('Total count of ARs:', len(ars))
+    print('First 10 ARs:')
+    for ar in ars[:10]:
+        print(COLOR_CODES['bold_green' if ar.confidence > 0.9 else \
```
Darderion, line 30: You could just select "colour_code" separately instead of using ternary here

examples/mining_ar.py
```diff
@@ -0,0 +1,132 @@
...
+def print_table(table, title = None):
+    if title is not None:
+        print(title)
+
+    print(tabulate(table, tablefmt='pipe', stralign='center'))
+
+
+def print_ars(ars):
+    print('Total count of ARs:', len(ars))
+    print('First 10 ARs:')
```
Darderion, line 28: It's preferable to address those rows with an article

"The first 10 ARs"

examples/mining_ar.py
```diff
@@ -0,0 +1,132 @@
...
+    print("\n['Milk'] -> ['Eggs'] with confidence 0.75 means that when milk "
+          'is found in the receipt, the chance of eggs being '
+          'present amounts to 75%. So, customers are '
+          f'{COLOR_CODES["bold_yellow"]}likely{COLOR_CODES["default"]} '
+          'to buy eggs with milk.')
+
+    print('\nLet us turn to the next issue. You can observe that there are a lot of '
+          'association rules found in this small dataset. This happens because we did not '
+          f'set up the so called {COLOR_CODES["green"]}support{COLOR_CODES["default"]} '
+          'value. This is an indication of how frequently the itemset appears '
```
Darderion, line 76: "This value specifies how frequently the itemset should appear in the dataset"

Continue with "to be considered / to qualify for" or something else, filling in the description of requirements imposed by this parameter

"In this particular case, the frequency is defined as the ration between the number of transactions containing A and B, and the total number of transactions in a dataset"

examples/mining_ar.py
```diff
@@ -0,0 +1,132 @@
...
+          'in the dataset, expressed as the ratio of the number of transactions '
+          'containing A and B to the total number of transactions. '
+          'Since the default support value is 0, the system discovers '
+          'all association rules, even those that are present only '
```
Darderion, line 80: You could rewrite it as "even those that only occur once"

examples/mining_ar.py
```diff
@@ -0,0 +1,132 @@
...
+    algo.execute()
+    table = pandas.read_csv(TABLE_SINGULAR, header=None, index_col=0)
+
+    print("\nFor the second example, let's take a dataset with the same receipts "
+          'from the same supermarket, changing the input format to input_format='
+          '"singular". This is a two-column format, where the first column is the order '
+          'of the items, and the second column is the item that belongs to that order.\n')
+    print_table(table)
+    print('\nThis format is just a different table representation. Desbordante '
+          'is perfectly capable of the discovered rules from such tables, '
```
Darderion, line 117: "is perfectly capable of rules discovery in such tables, and the discovered objects are represented by the same type of objects as in the previous examples"

examples/mining_ar.py
```diff
@@ -0,0 +1,132 @@
...
+          'from the same supermarket, changing the input format to input_format='
+          '"singular". This is a two-column format, where the first column is the order '
+          'of the items, and the second column is the item that belongs to that order.\n')
+    print_table(table)
+    print('\nThis format is just a different table representation. Desbordante '
+          'is perfectly capable of the discovered rules from such tables, ' 
+          'and the association rules discovery objects represented by the same '
+          'type of objects as in the previous examples '
+          '(i.e., they have the same methods and fields).\n')
+    print('In this example, we show you how to list all of the unique items in the dataset.')
```
Darderion, line 121: You could separate the examples as to not induce confusion by having "in this example" act as a reference to either the following example or the previous example

examples/mining_ar.py
```diff
@@ -0,0 +1,132 @@
...
+def scenario_tabular():
+    algo = desbordante.ar.algorithms.Default()
+    algo.load_data(table=(TABLE_TABULAR, ',', True), input_format='tabular')
+    algo.execute(minconf=1)
+    table = pandas.read_csv(TABLE_TABULAR, header=None)
+
+    print("As the first example, let's look at the dataset containing receipts "
+          'from some supermarket using input_format="tabular". In this '
+          'format, each table row lists all items participating in the same '
+          'transaction. Note that in this table some rows may be shorter than others.\n')
```
Darderion, line 51: Line length (check all lines)

Darderion (changes requested):

Darderion (approved):

Darderion (approved):

Darderion (approved):

PR #408, opened 2024-05-04
Nd bindings and examples

Petua41:
Add python bindings and examples for ND verification algorithm


yakovypg (commented):
I returned to reviewing the code and have the following thoughts:
- It's not a good idea to delete conversation history.
- Method `.at()` provides bound check, but this can be useful during the implementation of the algorithm to catch errors during debugging. In other situations we don't want to have overhead for these checks, so it is preferable to use `operator[]`. In theory, method `.at()` can be used when the index comes from the user-side to avoid undefined behavior in case of an invalid index. However, in this case it is preferable to use special checks and throw special exceptions.
- The previous arguments are also true for smart pointers. In this case, you should write `(*ptr)[index]` instead of `ptr->operator[](index)`. If the expression is very long, you can define additional variable (e.g. reference to `vector` in `shared_ptr`) and use it without pointer dereference.

PR #409, opened 2024-05-05
Multi-threaded implementation of HyFD algorithm

cone-forest:
The HyFD algorithm was originally single-threaded,
which goes against its intended implementation.
If ThreadNumber option is specified in the algorithm configuration step,
the algorithm will utilize the specified number of threads.

This implementation was benchmarked on several large datasets such as:
- iowa1kk
- EpicMeds
- flight

The general trend is displayed in the following graph:
hyfd_bench_threads_iowa1kk
(data for iowa1kk dataset on i9-13905h)

Alternative implementations showing worse performance:
- Desbordante's utils::ParallelForEach
- std::future + std::async
- std::execution::par
- OpenMP parallel for macro
- TBB parallel for function

The final implementation showed high time consumption by memory allocation/deallocation.
In attempt of optimizing them the following allocators were benchmarked:
- tbb::scalable_allocator
- tcmalloc
- jemalloc
- mimalloc
All of which showed similar/worse performance,
therefore implementing custom allocator is probably useless.

### DIFF
--- a/src/core/algorithms/fd/hyfd/hyfd.cpp
+++ b/src/core/algorithms/fd/hyfd/hyfd.cpp
@@ -25,7 +25,7 @@ HyFD::HyFD(std::optional<ColumnLayoutRelationDataManager> relation_manager)
     RegisterOption(config::kThreadNumberOpt(&threads_num_));
 }
 
-void HyFD::MakeExecuteOptsAvailable() {
+void HyFD::MakeExecuteOptsAvailableFDInternal() {
     MakeOptionsAvailable({config::names::kThreads});
 }
 
--- a/src/core/algorithms/fd/hyfd/hyfd.cpp
+++ b/src/core/algorithms/fd/hyfd/hyfd.cpp
@@ -12,6 +12,8 @@
 
 #include "algorithms/fd/hycommon/preprocessor.h"
 #include "algorithms/fd/hycommon/util/pli_util.h"
+#include "config/names.h"
+#include "config/thread_number/option.h"
 #include "inductor.h"
 #include "sampler.h"
 #include "validator.h"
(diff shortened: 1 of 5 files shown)


src/core/algorithms/fd/hyfd/hyfd.cpp
```diff
@@ -14,12 +14,16 @@
...
+#include "thread_number/option.h"
 #include "validator.h"
 
 namespace algos::hyfd {
 
 HyFD::HyFD(std::optional<ColumnLayoutRelationDataManager> relation_manager)
-    : PliBasedFDAlgorithm({}, relation_manager) {}
+    : PliBasedFDAlgorithm({}, relation_manager) {
+    RegisterOption(config::kThreadNumberOpt(&threads_num_));
+    MakeOptionsAvailable({config::kThreadNumberOpt.GetName()});
```
BUYT-1, line 25: Logically, this is an execution stage option, so override `MakeExecuteOptsAvailable` and do this there.

src/core/algorithms/fd/hyfd/hyfd.h
```diff
@@ -9,6 +9,7 @@
 #include "algorithms/fd/pli_based_fd_algorithm.h"
 #include "algorithms/fd/raw_fd.h"
 #include "model/table/position_list_index.h"
+#include "thread_number/type.h"
```
BUYT-1, line 12:
```suggestion
#include "config/thread_number/type.h"
```

BUYT-1 (changes requested):

src/core/algorithms/fd/hyfd/validator.cpp
```diff
@@ -249,6 +250,30 @@ Validator::FDValidations Validator::ValidateAndExtendSeq(std::vector<LhsPair> co
...
+
+    for (auto const& vertex : vertices) {
+        std::packaged_task<FDValidations()> task(
+                [this, &vertex]() { return GetValidations(vertex); });
+        validation_futures.push_back(task.get_future());
+        boost::asio::post(pool, std::move(task));
+    }
+
+    pool.join();
+    pool.wait();
```
BUYT-1, line 267: Looking at the documentation, these seem to do the same thing, why is it written this way?

BUYT-1 (approved):

PR #417, opened 2024-05-22
Add Denial Constraint structures that are necessary to implement DC verification

ol-imorozko:


src/core/algorithms/dc/base_provider.h
```diff
@@ -0,0 +1,55 @@
...
+class BaseProvider {
+protected:
+    static std::shared_ptr<Derived> instance_;
+
+    BaseProvider() = default;
+
+    // FIXME: this is wrong, there should be one class representing the algorithm,
+    // which will create/clear singletons and call PredicateBuilder, PliShardBuiler,
+    // SomeOtherBuilder in order..
+    friend PredicateBuilder;
```
polyntsov, line 20: Why is declared friend though? This class doesn't have private fields at all

src/core/algorithms/dc/predicate_provider.h
```diff
@@ -0,0 +1,45 @@
...
+ * operands involved.
+ */
+class PredicateProvider : public BaseProvider<PredicateProvider> {
+private:
+    using ColumnMap = std::unordered_map<ColumnOperand, Predicate>;
+    using OperatorMap = std::unordered_map<ColumnOperand, ColumnMap>;
+    // predicates_[op][col1][col2] corresponds to the related Predicate object
+    std::unordered_map<Operator, OperatorMap> predicates_;
+
+    friend BaseProvider<PredicateProvider>;
```
polyntsov, line 30: why making it firend? `BaseProvider` doesn't use any private fields of this class as far as I can see

polyntsov (changes requested):

PR #418, opened 2024-05-26
Dynamic FDVerifier

rpuwa:


src/core/model/table/dynamic_position_list_index.cpp
```diff
@@ -0,0 +1,223 @@
...
+        for (int i : cluster.first) {
+            fprintf(stderr, "%d,", i);
+        }
+        fprintf(stderr, "} -> [");
+        for (int i : cluster.second) {
+            fprintf(stderr, "%d,", i);
+        }
+        fprintf(stderr, "]\n");
+    }
+}
```
BUYT-1, line 60: Remove or move to `LOG(TRACE)`

BUYT-1 (changes requested):

src/core/model/table/dynamic_position_list_index.cpp
```diff
@@ -0,0 +1,207 @@
...
+void ProcessDeleteToCluster(model::DynamicPositionListIndex::Cluster& cluster_tuple,
+                            std::vector<size_t> const& delete_data, std::vector<bool>& is_deleted) {
+    std::stack<size_t> to_delete;
+    for (size_t i = 0; i < delete_data.size(); ++i) {
+        if (is_deleted[i]) continue;
+        auto idx = std::lower_bound(cluster_tuple.begin(), cluster_tuple.end(), delete_data[i]) -
+                   cluster_tuple.begin();
+        if (static_cast<size_t>(cluster_tuple[idx]) == delete_data[i]) {
+            to_delete.push(idx);
+            is_deleted[i] = true;
```
BUYT-1, line 25:
```suggestion
        std::vector<bool>::reference is_deleted_ref = is_deleted[i];
        if (is_deleted_ref) continue;
        size_t deleted_record_id = deleted_record_ids[i];
        auto cluster_begin = cluster.begin();
        auto cluster_end = cluster.end()
        auto found = std::lower_bound(cluster_begin, cluster_end, deleted_record_id);
        if (found == cluster_end) continue;
        if (static_cast<size_t>(*found) == deleted_record_id) {
            auto idx = std::distance(cluster_begin, found);
            to_delete.push(idx);
            is_deleted_ref = true;
```

src/core/model/table/dynamic_position_list_index.cpp
```diff
@@ -0,0 +1,207 @@
...
+std::unique_ptr<DynamicPositionListIndex> DynamicPositionListIndex::CreateFor(
+        std::vector<std::vector<int>>& data) {
+    std::unordered_map<std::vector<int>, Cluster> clusters;
+    for (size_t position = 0; position < data.size(); ++position) {
+        std::vector<int> values = data[position];
+        clusters[values].push_back(position);
+    }
+
+    unsigned int size = 0;
+    for (auto& cluster : clusters) {
```
BUYT-1, line 64:
```suggestion
    for (auto& [lhs, cluster] : clusters) {
```

src/core/model/table/dynamic_position_list_index.cpp
```diff
@@ -0,0 +1,207 @@
...
+                partial_index[probing_table_value_id].second.emplace_back(position);
+            } else {
+                std::vector<int> value = cluster.first;
+                value.insert(value.end(), cluster_values[probing_table_value_id].begin(),
+                             cluster_values[probing_table_value_id].end());
+                partial_index[probing_table_value_id] = {std::move(value), {position}};
+            }
+        }
+
+        for (auto& iter : partial_index) {
```
BUYT-1, line 179:
```suggestion
        for (auto& [<yournamehere>, new_cluster] : partial_index) {
```

src/core/model/table/dynamic_position_list_index.h
```diff
@@ -0,0 +1,108 @@
...
+    bool IsConstant() const {
+        return relation_size_ <= 1 || (GetNumNonSingletonCluster() == 1 && size_ == relation_size_);
+    }
+
+    std::unique_ptr<DynamicPositionListIndex> Intersect(DynamicPositionListIndex const* that) const;
+    std::unique_ptr<DynamicPositionListIndex> Probe(DynamicPositionListIndex const* that) const;
+    std::string ToString() const;
+};
+
+using DynPLI = DynamicPositionListIndex;
```
BUYT-1, line 106: Not useful alias. [Relevant](https://google.github.io/styleguide/cppguide.html#Aliases)

src/python_bindings/dynamic/bind_dynamic_fd_verification.cpp
```diff
@@ -0,0 +1,34 @@
+#include "bind_dynamic_fd_verification.h"
```
BUYT-1, line 1:
```suggestion
#include "dynamic/bind_dynamic_fd_verification.h"
```

BUYT-1 (changes requested):
Use aliases for very long and complicated types. You may want an alias for `std::unordered_map<std::vector<int>, Cluster>`, for example.

src/core/model/table/dynamic_position_list_index.cpp
rpuwa: i decided to rewrite it not so expanded

src/core/config/tabular_data/crud_operations/operations.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+#include "config/names.h"
+#include "delete/option.h"
+#include "insert/option.h"
+#include "update/option.h"
+
+const std::vector<std::string_view> kCrudOptions{
+    config::names::kInsertStatements, config::names::kDeleteStatements, 
+    config::names::kUpdateStatements};
```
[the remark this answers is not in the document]
rpuwa: `CommonOption::GetName` throws segfault, because you need to call constructor before this method

src/core/config/tabular_data/crud_operations/insert/option.cpp
```diff
@@ -0,0 +1,10 @@
+#include "config/tabular_data/crud_operations/insert/option.h"
+
+#include "config/names_and_descriptions.h"
+#include "parser/csv_parser/csv_parser.h"
+
+namespace config {
+using names::kInsertStatements, descriptions::kDInsertStatements;
+extern CommonOption<InputTable> const kInsertStatementsOpt = {kInsertStatements, kDInsertStatements,
+                                                              std::make_shared<CSVParser>()};
```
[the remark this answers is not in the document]
rpuwa: `InputTable{}` constructor throws segfault. I suppose, the reason is that `InputTable = std::shared_ptr<IDatasetStream>`, and `IDatasetStream` is a virtual class without constructor

src/core/config/tabular_data/crud_operations/update/option.cpp
```diff
@@ -0,0 +1,10 @@
+#include "config/tabular_data/crud_operations/update/option.h"
+
+#include "config/names_and_descriptions.h"
+#include "parser/csv_parser/csv_parser.h"
+
+namespace config {
+using names::kUpdateStatements, descriptions::kDUpdateStatements;
+extern CommonOption<InputTable> const kUpdateStatementsOpt = {kUpdateStatements, kDUpdateStatements,
+                                                              std::make_shared<CSVParser>()};
```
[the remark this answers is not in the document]
rpuwa: Same as in insert statements

examples/dynamic_verifying_fd.py
```diff
@@ -0,0 +1,113 @@
...
+        insert_df.set_index(pd.Index(range(max_data_index + 1, max_data_index + len(insert_df) + 1)), inplace=True)
+        data = pd.concat([data, insert_df])
+        max_data_index = data.index.max()
+    # delete
+    if (delete_array is not None):
+        data = data.drop(delete_array)
+    # update
+    if (update_df is not None):
+        df = update_df.drop(['_id'], axis = 1)
+        if (update_in_place):
```
BUYT-1, line 52:
```suggestion
        if update_in_place:
```

examples/dynamic_verifying_fd.py
```diff
@@ -0,0 +1,113 @@
...
+    data = pd.read_csv(table, header=[0])
+    max_data_index = data.index.max()
+    print(data)
+
+    algo = desbordante.dynamic_fd_verification.algorithms.Default()
+    algo.load_data(table=data, lhs_indices=[0, 2], rhs_indices=[1], update_in_place=True)
+    print(DEFAULT_COLOR_CODE)
+    print_results_for_fd(algo, data, [0, 2], [1])
+
+    print(f" Then, let's try to insert a row into the table, and check FD again.")
```
BUYT-1, line 75:
```suggestion
    print(f" Then, let's try to insert a row into the table, and check whether FD holds again.")
```

examples/dynamic_verifying_fd.py
```diff
@@ -0,0 +1,113 @@
...
+    insert_df1 = pd.DataFrame({'Creature': ['Elf'], 'Strength': [6], 'HaveMagic': [True]})
+    algo.execute(insert=insert_df1)
+    data = update_data_table(data=data, insert_df=insert_df1)
+    print('Current table state:')
+    print(data)
+    print('\nFD verification result:')
+    print_results_for_fd(algo, data, [0, 2], [1])
+
+    print(f" Now we are going to delete rows, that violating FD.")
```
BUYT-1, line 84:
```suggestion
    print(f"Now we are going to delete rows that violate the FD.")
```

src/core/algorithms/fd/fd_verifier/dynamic_fd_verifier.cpp
```diff
@@ -0,0 +1,247 @@
...
+}
+
+unsigned long long DynamicFDVerifier::ExecuteInternal() {
+    auto start_time = std::chrono::system_clock::now();
+    std::vector<std::pair<std::optional<size_t>, std::vector<int>>> lhs_inserts{}, rhs_inserts{};
+    std::vector<size_t> lhs_deletes{delete_batch_}, rhs_deletes{delete_batch_};
+    if (insert_batch_ != nullptr) {
+        while (insert_batch_->HasNextRow()) {
+            std::vector<std::string> row = insert_batch_->GetNextRow();
+            if (row.size() != input_table_->GetNumberOfColumns()) {
```
BUYT-1, line 122: Display a warning here

src/core/algorithms/fd/fd_verifier/dynamic_fd_verifier.cpp
```diff
@@ -0,0 +1,247 @@
...
+std::vector<int> DynamicFDVerifier::ParseRowForPLI(
+        model::IDatasetStream::Row::iterator const& row_begin,
+        std::vector<unsigned int> const& indices) {
+    std::vector<int> result{};
+    for (size_t index : indices) {
+        std::string const& field = *(row_begin + index);
+        if (field.empty()) {
+            result.emplace_back(kNullValueId);
+        } else {
+            auto location = value_dictionary_.find(field);
```
BUYT-1, line 200: Use `try_emplace`

src/core/config/tabular_data/crud_operations/update_in_place/option.h
```diff
@@ -0,0 +1,8 @@
+#pragma once
+
+#include "config/common_option.h"
+#include "config/equal_nulls/type.h"
+
+namespace config {
+extern CommonOption<bool> const kUpdateInPlaceOpt;
```
BUYT-1, line 7: Is it really an option that can be reused?

rpuwa: Yes, for example: DynamicFD algo needs to represent update operation as combination of delete and insert (insert to the end of table), and DynamicFDVerifier doesn't need. It's even better to to in-place update here, because it will be simpler and clearer to user, if we will process update as a single operation

src/core/algorithms/fd/fd_verifier/dynamic_fd_verifier.cpp
```diff
@@ -0,0 +1,247 @@
...
+    stats_calculator_ =
+            std::make_unique<DynamicStatsCalculator>(input_table_, lhs_indices_, rhs_indices_);
+    VerifyFD();
+    SortHighlightsByProportionDescending();
+}
+
+unsigned long long DynamicFDVerifier::ExecuteInternal() {
+    auto start_time = std::chrono::system_clock::now();
+    std::vector<std::pair<std::optional<size_t>, std::vector<int>>> lhs_inserts{}, rhs_inserts{};
+    std::vector<size_t> lhs_deletes{delete_batch_}, rhs_deletes{delete_batch_};
```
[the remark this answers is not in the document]
BUYT-1: `lhs_deletes` and `rhs_deletes` always have the same contents, don't separate them. In fact, passing `delete_batch_` straight to the methods might work even better.

src/core/model/table/dynamic_position_list_index.cpp
```diff
@@ -0,0 +1,200 @@
...
+#include <utility>
+
+#include <boost/dynamic_bitset.hpp>
+#include <easylogging++.h>
+
+namespace {
+// O(delete_data * log(cluster))
+void ProcessDeleteToCluster(model::DynamicPositionListIndex::Cluster& cluster,
+                            std::vector<size_t> const& deleted_record_ids,
+                            std::vector<bool>& is_deleted) {
```
BUYT-1, line 18: Make `deleted_record_ids` an `unordered_set`, use `std::erase_if` with `deleted_record_ids.erase(record_id) == deleted_record_ids.end()` predicate. `std::erase_if` will maintain the sort invariant too.

src/core/model/table/dynamic_position_list_index.cpp
```diff
@@ -0,0 +1,200 @@
...
+            size_t idx = std::distance(cluster.begin(), found_it);
+            to_delete.push(idx);
+            is_deleted[i] = true;
+        }
+    }
+    while (!to_delete.empty()) {  // deletion separated because of lower_bound usage
+        size_t idx = to_delete.top();
+        to_delete.pop();
+        if (idx < cluster.size() - 1) {
+            std::swap(cluster[idx], cluster[cluster.size() - 1]);
```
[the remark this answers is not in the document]
BUYT-1: (may become irrelevant)

src/core/model/table/dynamic_position_list_index.cpp
```diff
@@ -0,0 +1,200 @@
...
+      size_(size),
+      relation_size_(relation_size),
+      probing_table_cache_() {}
+
+std::unique_ptr<DynamicPositionListIndex> DynamicPositionListIndex::CreateFor(
+        std::vector<std::vector<int>>& data) {
+    ClusterCollection clusters;
+    for (size_t position = 0; position < data.size(); ++position) {
+        std::vector<int> values = data[position];
+        clusters[values].push_back(position);
```
BUYT-1, line 57:
```suggestion
        std::vector<int> const& values = data[position];
        clusters[values].push_back(position);
```

src/core/model/table/dynamic_position_list_index.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+DynamicPositionListIndex::DynamicPositionListIndex(ClusterCollection clusters, size_t size,
+                                                   size_t relation_size)
+    : clusters_(std::move(clusters)),
+      size_(size),
+      relation_size_(relation_size),
+      probing_table_cache_() {}
+
+std::unique_ptr<DynamicPositionListIndex> DynamicPositionListIndex::CreateFor(
+        std::vector<std::vector<int>>& data) {
```
BUYT-1, line 53:
```suggestion
        std::vector<std::vector<int>>& records) {
```

src/core/model/table/dynamic_position_list_index.cpp
```diff
@@ -0,0 +1,200 @@
...
+    ClusterCollection clusters;
+    for (size_t position = 0; position < data.size(); ++position) {
```
BUYT-1, line 55:
```suggestion
    for (size_t record_id = 0; record_id < data.size(); ++record_id) {
```

src/core/model/table/dynamic_position_list_index.cpp
```diff
@@ -0,0 +1,200 @@
...
+        std::vector<int> values = data[position];
+        clusters[values].push_back(position);
+    }
+
+    for (auto& [_, cluster] : clusters) {
+        sort(cluster.begin(), cluster.end());
+    }
```
BUYT-1, line 62: Does nothing, you're already inserting record IDs in sorted order.

src/core/model/table/dynamic_position_list_index.h
```diff
@@ -0,0 +1,93 @@
...
+namespace model {
+class DynamicPositionListIndex {
+public:
+    using Cluster = std::vector<int>;
+    using ClusterCollection = std::unordered_map<std::vector<int>, Cluster>;
+
+private:
+    ClusterCollection clusters_;
+    size_t size_;
+    size_t relation_size_;
```
BUYT-1, line 31:
```suggestion
    size_t valid_records_number_;
    size_t next_record_id_;
```

src/core/model/table/dynamic_position_list_index.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+    return frequencies;
+}
+
+std::shared_ptr<std::vector<int> const> DynamicPositionListIndex::CalculateAndGetProbingTable()
+        const {
+    if (probing_table_cache_ != nullptr) return probing_table_cache_;
+    cluster_values_cache_.clear();
+    std::vector<int> probing_table = std::vector<int>(relation_size_);
```
BUYT-1, line 122: `assert(kNullValueId == 0)`

BUYT-1 (changes requested):
Just the things I caught for now.

Look for similar things in the rest of the code yourself:
* Repeated lookups (ex: same things repeating inside `[]` calls for attention)
* Fields with generic names (ex: `a.first`, `a.second`, `a.second`) used many times in a row is suspicious
* Invariants not spelled out explicitly. If you just see the function and want to write an `assert`, write it or make it such that it holds "automatically". Another class won't hurt.
* Consider using more type aliases. It's hard to know which `std::vector<int>` is a collection of value IDs and which is a collection of record IDs.
The less context you need to understand to figure out what's happening in a function, the better. Be more thorough with names.

src/core/algorithms/fd/fd_verifier/dynamic_fd_verifier.cpp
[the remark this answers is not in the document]
rpuwa: > `delete_batch_` should be an `unordered_set` too.
I think it's shouldn't and we don't need `unordered_set` at all, because we don't do any lookup through `delete_batch_`, we only iterate over it

src/core/model/table/dynamic_position_list_index.cpp
rpuwa: where and why? also, we got `static constexpr int kNullValueId = -1;` expression in dynamic_fd_verifier.h

src/core/model/table/dynamic_position_list_index.cpp
rpuwa: > Make `deleted_record_ids` an `unordered_set`

Why `unordered_set`? We do not process any lookup over `deleted_record_ids` (at least now, with `std::erase_if `realization)

rpuwa: > Make `deleted_record_ids` an `unordered_set`

When i tried this, I had this error:
`ConfigurationError: Unable to cast Python object to C++ type for option "delete"`
As temporary solution, we construct `unordered_set` from `vector`, and pass it to `DynPLI::UpdateWith` Method

BUYT-1: That's because `std::unordered_set` is missing from `py_to_any.h`

src/core/model/table/dynamic_position_list_index.cpp
BUYT-1: Oh, I misunderstood something, then

examples/dynamic_verifying_fd.py
```diff
@@ -0,0 +1,113 @@
...
+max_data_index = 0
+
+
+def print_clusters(verifier, data, lhs, rhs):
+    print(f"Number of clusters violating FD: {verifier.get_num_error_clusters()}")
+    for i, highlight in enumerate(verifier.get_highlights(), start=1):
+        print(f"{BLUE_CODE} #{i} cluster: {DEFAULT_COLOR_CODE}")
+        for el in highlight.cluster:
+            arr1 = [data[data.columns[i]][el] for i in lhs]
+            arr2 = [data[data.columns[i]][el] for i in rhs]
```
BUYT-1:
```suggestion
def print_clusters(verifier, data, lhs_indices, rhs_indices):
    print(f"Number of clusters violating FD: {verifier.get_num_error_clusters()}")
    lhs_columns = [data[data.columns[i]] for i in lhs_indices]
    rhs_columns = [data[data.columns[i]] for i in rhs_indices]
    for i, highlight in enumerate(verifier.get_highlights(), start=1):
        print(f"{BLUE_CODE} #{i} cluster: {DEFAULT_COLOR_CODE}")
        for record_index in highlight.cluster:
            lhs_values = [lhs_column[record_index] for lhs_column in lhs_columns]
            rhs_values = [rhs_column[record_index] for rhs_column in rhs_columns]
```

src/core/model/table/dynamic_position_list_index.cpp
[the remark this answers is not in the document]
rpuwa: became irrelevant

src/core/algorithms/fd/fd_verifier/dynamic_fd_verifier.cpp
```diff
@@ -0,0 +1,243 @@
...
+    auto elapsed_milliseconds = std::chrono::duration_cast<std::chrono::milliseconds>(
+            std::chrono::system_clock::now() - start_time);
+    return elapsed_milliseconds.count();
+}
+
+void DynamicFDVerifier::VerifyFD() const {
+    std::unique_ptr<model::DynPLI const> intersection_pli = lhs_pli_->Intersect(rhs_pli_.get());
+
+    if (lhs_pli_->GetNumCluster() == intersection_pli->GetNumCluster()) {
+        return;
```
BUYT-1, line 168: Statistics might have irrelevant values if they aren't updated. It makes sense for `get_error` to return `0.0` if the FD holds, for example.

src/core/model/table/dynamic_position_list_index.cpp
```diff
@@ -0,0 +1,176 @@
...
+    std::vector<ClusterValue> const& cluster_values = that->GetClusterValuesCache();
+    assert(this->next_record_id_ == probing_table->size());
+    ClusterCollection new_clusters;
+    unsigned int new_size = 0;
+
+    std::unordered_map<int, std::pair<ClusterValue, Cluster>> partial_clusters;
+
+    for (auto& [value, cluster] : clusters_) {
+        for (int position : cluster) {
+            int probing_table_value_id = (*probing_table)[position];
```
BUYT-1, line 137: UB if the probing table is old and the record is one of the newly inserted?
That is:
1. The initial FD verification is done. `Intersect` was called, the probing table cache is now active.
2. A new record was inserted. Not accounted for in the probing table.
3. A follow-up check is performed. One of the clusters contains the new record, leading to UB here.

Maybe the caching is not necessary and we should unconditionally create and update the probing table along with the clusters?

src/core/model/table/dynamic_position_list_index.cpp
```diff
@@ -0,0 +1,176 @@
...
+            if (partial_clusters.contains(probing_table_value_id)) {
+                partial_clusters[probing_table_value_id].second.emplace_back(position);
+            } else {
+                ClusterValue key = std::move(value);
+                key.insert(key.end(), cluster_values[probing_table_value_id].begin(),
+                           cluster_values[probing_table_value_id].end());
+                partial_clusters[probing_table_value_id] = {std::move(key), {position}};
```
BUYT-1, line 144: Repeated lookup for `partial_clusters` and `cluster_values`. Using `try_emplace` and storing references will help avoid this.

src/core/model/table/dynamic_position_list_index.cpp
```diff
@@ -0,0 +1,176 @@
...
+    return frequencies;
+}
+
+std::shared_ptr<std::vector<int> const> DynamicPositionListIndex::CalculateAndGetProbingTable()
+        const {
+    if (probing_table_cache_ != nullptr) return probing_table_cache_;
+    cluster_values_cache_.clear();
+    std::vector<int> probing_table = std::vector<int>(next_record_id_);
+    int next_cluster_id = 1;
+    cluster_values_cache_.push_back({});
```
BUYT-1, line 102: Explain what this does in a comment, or make it apparent from the nearby code.

src/core/model/table/dynamic_position_list_index.cpp
```diff
@@ -0,0 +1,176 @@
...
+                partial_clusters[probing_table_value_id].second.emplace_back(position);
+            } else {
+                ClusterValue key = std::move(value);
+                key.insert(key.end(), cluster_values[probing_table_value_id].begin(),
+                           cluster_values[probing_table_value_id].end());
+                partial_clusters[probing_table_value_id] = {std::move(key), {position}};
+            }
+        }
+
+        for (auto& iter : partial_clusters) {
```
BUYT-1, line 148:
```suggestion
        for (auto& [_, new_cluster] : partial_clusters) {
```

src/core/model/table/dynamic_table_data.h
```diff
@@ -0,0 +1,79 @@
...
+    std::vector<std::vector<std::string>> full_table_;
+    std::unordered_set<size_t> deleted_rows_{};
+
+public:
+    DynamicTableData(config::InputTable input_table) {
+        full_table_.resize(input_table->GetNumberOfColumns());
+        while (input_table->HasNextRow()) {
+            std::vector<std::string> row = input_table->GetNextRow();
+            for (size_t i = 0; i < row.size(); ++i) {
+                full_table_[i].emplace_back(std::move(row[i]));
```
BUYT-1, line 24: Check for invalid size

src/core/model/table/dynamic_table_data.h
```diff
@@ -0,0 +1,79 @@
...
+    std::string const& GetValue(size_t row_index, size_t col_index) const {
+        assert(col_index < full_table_.size() && row_index < full_table_[col_index].size());
+        return full_table_[col_index][row_index];
+    }
+
+    void Update(config::InputTable insert_data, config::InputTable update_data,
+                std::unordered_set<size_t> const& delete_data) {
+        for (size_t row_id : delete_data) {
+            for (size_t i = 0; i < full_table_.size(); ++i) {
+                full_table_[i][row_id].clear();
```
BUYT-1, line 46: If you want to free memory, this can work
```suggestion
                full_table_[i][row_id] = {};
```

src/core/model/table/dynamic_table_data.h
```diff
@@ -0,0 +1,79 @@
...
+            }
+            deleted_rows_.emplace(row_id);
+        }
+        if (insert_data != nullptr) {
+            while (insert_data->HasNextRow()) {
+                std::vector<std::string> row = insert_data->GetNextRow();
+                for (size_t i = 0; i < row.size(); ++i) {
+                    full_table_[i].emplace_back(std::move(row[i]));
```
BUYT-1, line 54: `emplace_back` doesn't do anything special for a move constructor.
```suggestion
                    full_table_[i].push_back(std::move(row[i]));
```

src/core/model/table/dynamic_table_data.h
```diff
@@ -0,0 +1,79 @@
...
+            }
+        }
+    }
+
+    std::unordered_set<size_t> const& GetDeletedRows() const {
+        return deleted_rows_;
+    }
+};
+
+}  // namespace model
```
BUYT-1, line 79: No newline

src/core/model/table/dynamic_table_data.h
```diff
@@ -0,0 +1,79 @@
...
+#include <vector>
+
+#include "config/exceptions.h"
+#include "config/tabular_data/input_table_type.h"
+
+namespace model {
+
+struct DynamicTableData {
+private:
+    std::vector<std::vector<std::string>> full_table_;
```
BUYT-1, line 15:
```suggestion
    std::vector<std::vector<std::string>> columns_;
```

src/core/model/table/dynamic_table_data.h
```diff
@@ -0,0 +1,79 @@
...
+            while (update_data->HasNextRow()) {
+                std::vector<std::string> row = update_data->GetNextRow();
+                size_t row_id = row.empty() ? 0 : std::stoull(row[0]);
+                if (deleted_rows_.contains(row_id)) {
+                    throw config::ConfigurationError(
+                            "Attempt to update a deleted row during processing of update "
+                            "operations");
+                }
+                for (size_t i = 1; i < row.size(); ++i) {
+                    full_table_[i - 1][row_id] = std::move(row[i]);
```
BUYT-1, line 68: No check for row size

BUYT-1 (changes requested):

src/core/algorithms/fd/fd_verifier/dynamic_fd_verifier.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+void DynamicFDVerifier::MakeExecuteOptsAvailable() {
+    using namespace config::names;
+    MakeOptionsAvailable(kCrudOptions);
+}
+
+void DynamicFDVerifier::LoadDataInternal() {
+    if (!input_table_->HasNextRow()) {
+        throw std::runtime_error("Got an empty dataset: FD verifying is meaningless.");
+    }
```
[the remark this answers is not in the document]
rpuwa: tests will be soon

src/core/algorithms/fd/fd_verifier/dynamic_fd_verifier.cpp
rpuwa: this case handles during ResetState() call

BUYT-1: Please add a comment for that, then.
Isn't there going to be an optimized version that won't have to redo all the calculations, by the way? I'd expect it to be updating statistics directly. You should probably move resetting statistics here instead, since it is kind of a crutch before the "proper" method is implemented. That is, statistics are going to be reused, not reset every time.

src/core/model/table/dynamic_position_list_index.cpp
BUYT-1: By "this" I mean the exact line this comment references. Why are we adding an empty value here?

src/core/algorithms/fd/fd_verifier/dynamic_fd_verifier.cpp
```diff
@@ -0,0 +1,247 @@
...
+            return;
+        }
+        if (update_batch->GetNumberOfColumns() != input_table_->GetNumberOfColumns() + 1) {
+            throw config::ConfigurationError(
+                    "Update statements has differ column number from input table");
+        }
+        for (size_t i = 0; i < input_table_->GetNumberOfColumns(); ++i) {
+            if (update_batch->GetColumnName(i + 1) != input_table_->GetColumnName(i)) {
+                throw config::ConfigurationError(
+                        "Update statements has differ columns from input table");
```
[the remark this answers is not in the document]
rpuwa: Yep

src/core/model/table/dynamic_position_list_index.cpp
```diff
@@ -0,0 +1,172 @@
...
+// O(cluster)
+void ProcessDeleteToCluster(model::DynamicPositionListIndex::Cluster& cluster,
+                            std::unordered_set<size_t>& deleted_records_ids) {
+    auto records_pred = [&](size_t element) {
+        auto it = deleted_records_ids.find(element);
+        if (it != deleted_records_ids.end()) {
+            deleted_records_ids.erase(it);
+            return true;
+        }
+        return false;
```
BUYT-1, line 24:
```suggestion
        return static_cast<bool>(deleted_records_ids.erase(element));
```

src/core/model/table/dynamic_position_list_index.cpp
```diff
@@ -0,0 +1,172 @@
...
+        return;
+    }
+    for (auto it = clusters_.begin(); it != clusters_.end();) {
+        auto& [cluster_value, cluster] = *it;
+        ProcessDeleteToCluster(cluster, deleted_records_ids);
+        if (cluster.empty()) {
+            it = clusters_.erase(it);
+        } else {
+            ++it;
+        }
```
BUYT-1, line 67:
```suggestion
        }
        if (deleted_records_ids.empty()) break;
```

src/core/model/table/dynamic_position_list_index.cpp
```diff
@@ -0,0 +1,172 @@
...
+    }
+    for (auto const& [encoded_value_opt, cluster_value] : inserted_records) {
+        if (encoded_value_opt.has_value()) {
+            clusters_[cluster_value].emplace_back(*encoded_value_opt);
+        } else {
+            clusters_[cluster_value].emplace_back(next_record_id_++);
+            ++valid_records_number_;
+        }
+    }
+    valid_records_number_ -= deleted_records_ids.size();
```
BUYT-1, line 77: Save `deleted_records_ids.size()` into a variable before the deletion loop, this always subtracts zero.

src/core/model/table/dynamic_table_data.h
```diff
@@ -0,0 +1,82 @@
...
+
+namespace model {
+
+struct DynamicTableData {
+private:
+    std::vector<std::vector<std::string>> columns_;
+    std::unordered_set<size_t> deleted_rows_{};
+
+public:
+    DynamicTableData(config::InputTable input_table) {
```
BUYT-1, line 19: This function doesn't take ownership of the `shared_ptr`, use `IDatasetStream&` instead.

src/core/model/table/dynamic_table_data.h
```diff
@@ -0,0 +1,82 @@
...
+        columns_.resize(input_table->GetNumberOfColumns());
+        while (input_table->HasNextRow()) {
+            std::vector<std::string> row = input_table->GetNextRow();
+            assert(row.size() == columns_.size());
```
BUYT-1, line 23: `assert` is only used for debugging; this is data from the user, so it should be checked properly.

src/core/model/table/dynamic_table_data.h
```diff
@@ -0,0 +1,82 @@
...
+    size_t GetNumRowsTotal() const {
+        return columns_[0].size();
+    }
+
+    std::string const& GetValue(size_t row_index, size_t col_index) const {
+        assert(col_index < columns_.size() && row_index < columns_[col_index].size());
+        return columns_[col_index][row_index];
+    }
+
+    bool ValidateRowIndex(size_t idx) {
```
BUYT-1, line 43:
```suggestion
    bool IsRowIndexValid(size_t row_index) {
```

src/core/model/table/dynamic_table_data.h
```diff
@@ -0,0 +1,82 @@
...
+        for (size_t row_id : delete_data) {
+            for (size_t i = 0; i < columns_.size(); ++i) {
+                columns_[i][row_id] = {};
+            }
+            deleted_rows_.emplace(row_id);
+        }
+        if (insert_data != nullptr) {
+            while (insert_data->HasNextRow()) {
+                std::vector<std::string> row = insert_data->GetNextRow();
+                assert(row.size() == columns_.size());
```
BUYT-1, line 58: `assert` is only used for debugging; this is data from the user, so it should be checked properly.

src/core/model/table/dynamic_table_data.h
```diff
@@ -0,0 +1,82 @@
...
+        if (update_data != nullptr) {
+            while (update_data->HasNextRow()) {
+                std::vector<std::string> row = update_data->GetNextRow();
+                size_t row_id = row.empty() ? 0 : std::stoull(row.front());
+                if (deleted_rows_.contains(row_id)) {
+                    throw config::ConfigurationError(
+                            "Attempt to update a deleted row during processing of update "
+                            "operations");
+                }
+                assert(row.size() == columns_.size() + 1);
```
BUYT-1, line 73: `assert` is only used for debugging; this is data from the user, so it should be checked properly.

src/core/model/table/dynamic_table_data.h
```diff
@@ -0,0 +1,82 @@
...
+                assert(row.size() == columns_.size());
+                for (size_t i = 0; i < row.size(); ++i) {
+                    columns_[i].push_back(std::move(row[i]));
+                }
+            }
+        }
+        if (update_data != nullptr) {
+            while (update_data->HasNextRow()) {
+                std::vector<std::string> row = update_data->GetNextRow();
+                size_t row_id = row.empty() ? 0 : std::stoull(row.front());
```
BUYT-1, line 67: Returning 0 if the row is empty is not obvious, it should be explained in a comment if that's what's to be expected. Shouldn't all rows have the size of `columns_.size() + 1` anyway?

BUYT-1 (changes requested):

src/core/model/table/dynamic_table_data.h
rpuwa: yes, but this case we process during value checking of the `insert` option

src/core/model/table/dynamic_position_list_index.cpp
rpuwa: I don't think, that try_emplace is a good idea as in this case we will construct key every time

BUYT-1: `partial_clusters.try_emplace(probing_table_value_id, std::piecewise_construct, std::forward_as_tuple(std::move(v)), std::forward_as_tuple(std::initializer_list{position}));`

src/tests/test_dynamic_fd_verifier.cpp
```diff
@@ -0,0 +1,173 @@
+#include <algorithm>
+#include <memory>
+
+#include <gtest/gtest.h>
+
+#include "algo_factory.h"
+#include "all_csv_configs.h"
+#include "builtin.h"
```
BUYT-1, line 8:
```suggestion
#include "algorithms/algo_factory.h"
#include "all_csv_configs.h"
#include "model/types/builtin.h"
```

examples/dynamic_verifying_fd.py
```diff
@@ -0,0 +1,130 @@
...
+        print(INSERT_CODE + 'Rows to insert:')
+        print(insert_df.to_string(index=False), end=DEFAULT_COLOR_CODE+'\n')
+    if delete_set is not None:
+        print(DELETE_CODE + 'Rows indices to delete: ', ', '.join([str(i) for i in delete_set]) + DEFAULT_COLOR_CODE)
+    if update_df is not None:
+        update_indices = set(update_df.index)
+        print(UPDATE_CODE + 'Rows to update:')
+        update_rows = update_df.to_string(index=True).split('\n')
+        # deleting row with _id column name
+        del update_rows[1]
```
BUYT-1, line 54:
```suggestion
        update_rows = update_df.to_string(index=True, index_names=False).split('\n')
```

examples/dynamic_verifying_afd.py
```diff
@@ -0,0 +1,116 @@
...
+        print(INSERT_CODE + 'Rows to insert:')
+        print(insert_df.to_string(index=False), end=DEFAULT_COLOR_CODE+'\n')
+    if delete_set is not None:
+        print(DELETE_CODE + 'Rows indices to delete: ', ', '.join([str(i) for i in delete_set]) + DEFAULT_COLOR_CODE)
+    if update_df is not None:
+        update_indices = set(update_df.index)
+        print(UPDATE_CODE + 'Rows to update:')
+        update_rows = update_df.to_string(index=True).split('\n')
+        # deleting row with _id column name
+        del update_rows[1]
```
BUYT-1, line 51:
```suggestion
        update_rows = update_df.to_string(index=True, index_names=False).split('\n')
```

BUYT-1 (approved):

PR #421, opened 2024-05-29
Implement new column type deduction methods

### LINKED ISSUE #343: Regex for double type values doesn't support scientific notation
This regex doesn't match `1.1E10` or `1E+10`, which are valid numbers in scientific notation. So if we're going to support scientific notation, let's support it fully (and add corresponding tests). However, this could be done in a separate PR (in this case, please, create an issue).

_Originally posted by @polyntsov in https://github.com/Mstrutov/Desbordante/pull/294#discussion_r1457985766_

MichaelS239:
- Add new method for column type deduction
- Add more complex type checking (also closes #343)
- Add undelimited dates support
- Change some old tests and add new ones


src/core/model/table/typed_column_data.cpp
```diff
@@ -18,27 +21,85 @@ size_t GetNextAlignedOffset(size_t cur_offset, size_t align) {
...
+
+            std::bitset<5> new_candidate_types_bitset("00000");
+            bool matched = false;
+            for (auto const& [type_id, type_check] : kTypeIdToChecker) {
+                if (type_id != first_type_id && type_check(unparsed_[i])) {
+                    if (first_type_id == +TypeId::kUndefined && !matched) {
+                        first_type_id = type_id;
+                    }
+                    matched = true;
+                    if (type_id != +TypeId::kDate) {
```
Firsov62121, line 46: Add kDate to kTypeIdToBitset.

src/core/model/table/typed_column_data.cpp
```diff
@@ -18,27 +21,85 @@ size_t GetNextAlignedOffset(size_t cur_offset, size_t align) {
...
+                        new_candidate_types_bitset |= kTypeIdToBitset.at(type_id);
+                        break;
```
Firsov62121, line 48: Add some comment here why do we use this break.

Firsov62121 (approved):

PR #426, opened 2024-06-07
Add HyMD initial implementation paper

BUYT-1:


chernishev (changes requested):
I believe there is an issue in the file name: autumn 2024 has not yet arrived, and the name should refer to spring 2024.

chernishev (approved):

PR #436, opened 2024-07-09
Fix compilation errors

Aviu00:
Compilation failed due to warnings being treated as errors. gcc 14.1.1.


src/core/model/table/dynamic_position_list_index.cpp
```diff
@@ -132,7 +132,7 @@ std::unique_ptr<DynamicPositionListIndex> DynamicPositionListIndex::Probe(
                 auto& [key, positions] = partial_cluster->second;
                 positions.emplace_back(position);
             } else {
-                ClusterValue key = std::move(value);
+                ClusterValue key = value;
```
BUYT-1, line 135: I suggest also changing line 127 (`auto&` -> `auto const&`) for clarity

BUYT-1 (approved):

PR #440, opened 2024-07-15
Add descriptions and group examples

Darderion:
Add descriptions for each example
Add categories and sort examples accordingly
Add description for categories


examples/expert/README.md
```diff
@@ -0,0 +1,13 @@
...
+  <dd>A scenario demonstrating the process of monitoring the quality of newly-arrived data. The idea of this scenario is as follows: first, a user explores a dataset D1 (which is considered a “clean” dataset) by discovering all exact functional dependencies. Next, when the D2 dataset arrives, its exact functional dependencies are discovered and compared with D1’s. There are no missing dependencies, so this portion is considered clean. Then, D3 arrives, the same comparison is performed and the user learns that a single dependency is missing. Afterwards, the user makes various efforts to understand what has happened with data by trying to approximate functional and, finally, metric functional dependencies. Finally, the user learns that item_id -> item_weight became a metric dependency with range 2. This means that the weight has changed for some items, due to typo or due to changes in the actual weight. The idea of this approach is sketched in this <a href="https://arxiv.org/abs/2307.14935">paper</a>. There is also a streamlit version of it <a href="https://desbordante.streamlit.app/">here</a>.</dd>
```
BUYT-1, line 8:
```suggestion
  <dd>A scenario demonstrating the process of monitoring the quality of newly-arrived data. The idea of this scenario is as follows: first, a user explores a dataset D1 (which is considered a “clean” dataset) by discovering all exact functional dependencies. Next, when the D2 dataset arrives, its exact functional dependencies are discovered and compared with D1’s. There are no missing dependencies, so this portion is considered clean. Then, D3 arrives, the same comparison is performed and the user learns that a single dependency is missing. Afterwards, the user makes various efforts to understand what has happened with data by trying to use approximate functional and, finally, metric functional dependencies. Finally, the user learns that item_id -> item_weight became a metric dependency with range 2. This means that the weight has changed for some items, due to a typo or due to changes in the actual weight. The idea of this approach is sketched in this <a href="https://arxiv.org/abs/2307.14935">paper</a>. There is also a streamlit version of it <a href="https://desbordante.streamlit.app/">here</a>.</dd>
```

BUYT-1 (changes requested):
The `.md` files here are using full HTML for some reason when it's not needed.

examples/expert/README.md
```diff
@@ -0,0 +1,8 @@
...
++ [mine_typos.py](https://github.com/Desbordante/desbordante-core/tree/main/examples/expert/mine_typos.py) — a scenario demonstrating typo discovery inside tables using functional and approximate functional dependencies, which are discovered from a dataset. A user is presented with a number of “almost holding” dependencies, and each of them may indicate the presence of a typo. The script then offers means to easily check them by presenting rows with exceptions (clusters where the exact functional dependency is violated). The idea of this approach is sketched in this [paper](https://arxiv.org/abs/2307.14935). There is also a streamlit version of it [here](https://desbordante.streamlit.app/).
```
BUYT-1, line 8:
```suggestion
+ [mine_typos.py](https://github.com/Desbordante/desbordante-core/tree/main/examples/expert/mine_typos.py) — a scenario demonstrating typo discovery inside tables using functional and approximate functional dependencies, which are discovered from a dataset. The user is presented with a number of “almost holding” dependencies, and each of them may indicate the presence of a typo. The script then offers means to easily check them by presenting rows with exceptions (clusters where the exact functional dependency is violated). The idea of this approach is sketched in this [paper](https://arxiv.org/abs/2307.14935). There is also a streamlit version of it [here](https://desbordante.streamlit.app/).
```

BUYT-1 (changes requested):
Please move the changes from the last commit to the others.

BUYT-1 (approved):
LGTM

PR #443, opened 2024-07-29
Bump boost version, change caching strategy

BUYT-1:
1.81.0 contains unordered flat structures

### DIFF
--- a/.github/composite-actions/download-libraries/action.yml
+++ b/.github/composite-actions/download-libraries/action.yml
@@ -11,6 +11,11 @@ inputs:
     description: 'Download googletest'
     default: true
 
+  install-boost:
+    type: boolean
+    description: 'Install boost'
+    default: true
+
 runs:
   using: 'composite'
   steps:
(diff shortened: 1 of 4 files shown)


.github/composite-actions/download-libraries/action.yml
```diff
@@ -22,7 +22,10 @@ runs:
       shell: bash
 
     - name: Install Boost
-      run: sudo apt-get install libboost-all-dev -y
+      run: |
+        sudo add-apt-repository ppa:mhier/libboost-latest
+        sudo apt update
+        sudo apt-get install libboost1.81-all-dev -y
```
polyntsov, line 28: It'll probably work fine, but installing latest version from sources is a better approach IMO, because this PPA is unofficial by some completely unknown third-party and we can't actually trust it

polyntsov (approved):

.github/composite-actions/download-libraries/action.yml
BUYT-1: I'm having trouble building boost from source, the wheel workflows seem to be failing for some reason that I can't understand right now.

BUYT-1: Figured it out: the build files were cached along with the library and the build script erroneously detected the presence of compression libraries on the wheel container when they weren't there.

polyntsov (approved):

PR #444, opened 2024-08-01
Add DC verification algorithm

xJoskiy:
Add DC verification algorithm and corresponding tests as well as python bindings


Petua41 (changes requested):
Check newlines at the ends of files

src/core/algorithms/dc/verifier/dc_verifier.cpp
```diff
@@ -0,0 +1,536 @@
...
+    auto start = std::chrono::system_clock::now();
+    DC dc;
+    try {
+        dc = SplitDC(dc_string_);
+    } catch (std::exception const& e) {
+        LOG(INFO) << e.what();
+        return 0;
+    }
+
+    bool has_header = relation_->GetSchema()->GetColumns()[0].get()->GetName() != "";
```
Petua41, line 67:
```suggestion
    bool has_header = relation_->GetSchema()->GetColumn(0)->GetName() != "";
```

src/core/algorithms/dc/verifier/dc_verifier.cpp
```diff
@@ -0,0 +1,536 @@
...
+            {DCType::kAllEquality, &DCVerifier::VerifyAllEquality},
+            {DCType::kOneInequality, &DCVerifier::VerifyOneInequality},
+            {DCType::kOneTuple, &DCVerifier::VerifyOneTuple},
+            {DCType::kTwoTuples, &DCVerifier::VerifyTwoTuples},
+            {DCType::kMixed, &DCVerifier::VerifyMixed}};
+    result_ = check(this, map.at(dc_type));
+
+    auto elapsed_time = std::chrono::duration_cast<std::chrono::milliseconds>(
+            std::chrono::system_clock::now() - start);
+    LOG(DEBUG) << "";
```
Petua41, line 89:
```suggestion
```

src/core/algorithms/dc/verifier/dc_verifier.h
```diff
@@ -0,0 +1,76 @@
+#pragma once
+
+#include "algorithms/algorithm.h"
+#include "config/tabular_data/input_table/option.h"
+#include "config/tabular_data/input_table_type.h"
+#include "dc/model/dc.h"
+#include "dc/model/point.h"
+#include "table/typed_column_data.h"
+#include "util/kdtree.h"
```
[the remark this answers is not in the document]
xJoskiy: Can't find usage of `utility`

src/core/algorithms/dc/verifier/dc_verifier.cpp
```diff
@@ -0,0 +1,523 @@
...
+
+    return res;
+}
+
+// FIX: DC may contain braces, numbers or spaces, naive separation doesn't belong here.
+std::vector<Predicate> DCVerifier::SplitDC(std::string const& dc_string) {
+    size_t ind, start = 0;
+    std::string token, sep = " and ";
+    std::vector<Predicate> predicates;
+    while (true) {
```
Petua41, line 434: Maybe it's better to use
```suggestion
    std::vector<std::string> tokens;
    boost::split(tokens, dc_string, boost::is_any_of(" and "));
    for (auto& token : tokens) {
```

Petua41 (changes requested):

src/core/algorithms/dc/verifier/dc_verifier.h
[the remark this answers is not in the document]
Petua41: `std::pair` defined in `utility`

src/core/algorithms/dc/verifier/dc_verifier.cpp
```diff
@@ -0,0 +1,523 @@
...
+
+    return predicates;
+}
+
+std::vector<std::byte const*> DCVerifier::GetTuple(size_t row) {
+    std::vector<std::byte const*> res;
+    res.reserve(data_.size());
+    for (auto const& col : data_) {
+        res.push_back(col.GetValue(row));
+    }
```
Petua41, line 461:
```suggestion
    std::vector<std::byte const*> res;
    std::transform(data_.begin(), data_.end(), std::back_inserter(res), [&row](model::TypedColumnData& col) { return col.GetValue(row); });
```
or, if you want to pre-allocate memory,
```suggestion
    std::vector<std::byte const*> res(data_.size());
    std::transform(data_.begin(), data_.end(), res.begin(), [&row](model::TypedColumnData& col) { return col.GetValue(row); });
```

src/core/algorithms/dc/verifier/dc_verifier.cpp
xJoskiy: Unfortunately, it splits the string be each of the specified characters, not by a whole word

xJoskiy: Unfortunately I haven't found any intelligible way to split the string

src/core/algorithms/dc/verifier/dc_verifier.cpp
```diff
@@ -0,0 +1,536 @@
...
+bool DCVerifier::VerifyAllEquality(const DC& dc) {
+    bool res = true;
+    std::unordered_set<size_t> res_tuples;
+    std::vector<uint> const eq_cols = dc.GetColumnIndices();
+    for (size_t i = 0; i < data_[0].GetNumRows(); i++) {
+        if (ContainsNullOrEmpty(eq_cols, i)) continue;
+        std::vector<std::byte const*> tuple = GetTuple(i);
+        size_t key = HashTuple(tuple, eq_cols);
+        if (res_tuples.find(key) != res_tuples.end()) {
+            res = false;
```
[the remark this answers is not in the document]
xJoskiy: Yes, that's a typo

src/core/algorithms/dc/verifier/dc_verifier.cpp
```diff
@@ -0,0 +1,523 @@
...
+    for (auto const& pred : preds) {
+        auto left = pred.GetLeftOperand();
+        auto right = pred.GetRightOperand();
+        if (pred.IsCrossColumn() and pred.IsCrossTuple() and
+            pred.GetOperator().GetType() == OperatorType::kEqual) {
+            auto left_op = ColumnOperand(left.GetColumn(), left.GetTuple());
+            auto right_op = ColumnOperand(right.GetColumn(), right.GetTuple());
+
+            auto less_equal = Operator(OperatorType::kLessEqual);
+            auto greater_equal = Operator(OperatorType::kGreaterEqual);
```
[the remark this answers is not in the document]
xJoskiy: As for me, it is just a matter of choice. I find copy initialization here more readable and informative rather than direct list initialization

src/tests/all_csv_configs.cpp
```diff
@@ -120,4 +120,9 @@ CSVConfig const kTestDynamicFDUpdateBad3 =
 CSVConfig const kTestDynamicFDUpdateBad4 =
         CreateCsvConfig("dynamic_fd/TestDynamicUpdateBad4.csv", ',', true);
 CSVConfig const kLineItem = CreateCsvConfig("LineItem.csv", '|', true);
+CSVConfig const kTmpDC = CreateCsvConfig("tmp_dc.csv", ',', true);
```
Petua41, line 123: I can't find `tmp_dc.csv` dataset

src/tests/test_dc_verifier.cpp
Petua41, line 1: These tests differ only in the parameter, dataset and expected result. I think it's better to use TEST_P:
You'll need to:
- Define struct for parameters:
```
struct DCTestParams {
    std::string dc_string;
    CSVConfig csv_config;
    bool expected;
    DCTestParams(std::string dc_string, CSVConfig csv_config, bool expected) ...
};
```
- Define testing class:
`class TestDC : public ::testing::TestWithParam<DCTestParams> {};`
- Define standard test:
```
TEST_P(TestDC, DefaultTest) {
    auto const& p = GetParam();
    auto mp = algos::StdParamsMap(p.params);
    auto verifier = algos::CreateAndLoadAlgorithm<...>(mp);
    verifier->Execute();
    EXPECT_EQ(verifier->DCHolds(), p.expected)
}
```
- Create test suites using `INSTANTIATE_TEST_SUTIE_P` macro:
```
// clang-format off
INSTANTIATE_TEST_SUITE_P(
        TestSuiteName, TestDC,
        ::testing::Values(
            DCTestParams("predicate", kTestDC, true)
            ));
// clang-format on
```
or single tests using `INSTANTIATE_TEST_CASE_P` macro with the same syntax
See https://github.com/Desbordante/desbordante-core/blob/main/src/tests/test_nd_verifier.cpp for more examples

Petua41 (changes requested):

src/core/algorithms/dc/model/dc.h
```diff
@@ -0,0 +1,86 @@
...
+    std::vector<Predicate> predicates_;
+
+public:
+    DC(std::vector<Predicate>&& predicates) : predicates_(std::move(predicates)) {};
+    DC(std::vector<Predicate> const& predicates) : predicates_(predicates) {};
+    DC() = default;
+
+    template <class Iter>
+    DC(Iter first, Iter last) {
+        while (first != last) predicates_.push_back(*(first++));
```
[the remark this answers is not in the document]
xJoskiy: By the way, it won't compile since we have already declared predicates and we can only assign either std::initializer_list or std::vector

src/core/algorithms/dc/verifier/dc_verifier.cpp
```diff
@@ -0,0 +1,523 @@
...
+        Component right(tuple[right_ind], &data_[right_ind].GetType());
+        if (!Component::Eval(left, right, pred.GetOperator())) return false;
+    }
+
+    return true;
+}
+
+point DCVerifier::MakePoint(std::vector<std::byte const*> const& vec,
+                            std::vector<uint> const& indices, size_t point_ind /* = 0 */,
+                            ValType val_type /* = kFinite */) {
```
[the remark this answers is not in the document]
xJoskiy: It's a common practice to leave a comment about default values (at least in desbordante)

src/core/algorithms/dc/verifier/dc_verifier.cpp
```diff
@@ -0,0 +1,514 @@
...
+                assert(false);
+                __builtin_unreachable();
+        }
+    }
+
+    DC mixed_dc(mixed_predicates.begin(), mixed_predicates.end());
+    util::KDTree<point> s_tree, t_tree;
+    bool res = true;
+    std::vector<uint> all_cols = dc.GetColumnIndices();
+    for (size_t i = 0; i < data_[0].GetNumRows(); i++) {
```
Petua41, line 122: nit:
```suggestion
    for (size_t i = 0; i < data_[0].GetNumRows(); ++i) {
```
Style guide recommends using prefix operators unless you need postfix semantics

src/core/algorithms/dc/model/dc.h
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    std::string DCToString() const {
+        std::string const k_not = "!";
+        std::string const k_and = " and ";
+
+        std::stringstream ss;
+        for (size_t i = 0; i < predicates_.size(); ++i) {
+            ss << k_and << predicates_[i].ToString();
+        }
```
Petua41, line 82:
```suggestion
        for (size_t i = 0; i < predicates_.size(); ++i) {
            if (i != 0) {
                ss << k_and;
            }
            ss << predicates_[i].ToString();
        }
```

src/tests/test_dc_verifier.cpp
```diff
@@ -0,0 +1,66 @@
...
+    CSVConfig csv_config;
+    bool expected;
+};
+
+class TestDCVerifier : public ::testing::TestWithParam<DCTestParams> {};
+
+TEST_P(TestDCVerifier, DefaultTest) {
+    DCTestParams const& p = GetParam();
+    algos::StdParamsMap params = GetParamMap(p.csv_config, p.dc_string);
+    std::unique_ptr<DCVerifier> dc_verifier = algos::CreateAndLoadAlgorithm<DCVerifier>(params);
```
Petua41, line 34: nit:
```suggestion
    auto dc_verifier = algos::CreateAndLoadAlgorithm<DCVerifier>(params);
```
As for me, `dc_verifier` type here is obvious

src/core/algorithms/dc/verifier/dc_verifier.cpp
Petua41: This suggestion isn't resolved

examples/datasets/taxes.csv
Petua41, line 1: Shouldn't csv files contain a newline at the end?

examples/basic/verifying_dc.py
```diff
@@ -0,0 +1,59 @@
...
+TABLE_1 = "examples/datasets/taxes.csv"
+TABLE_2 = "examples/datasets/taxes_2.csv"
+
+print("This is an example of how to use Denial Constraint (DC) verification for checking hypotheses on data.")
+print("DC verification is perfomed by Rapidash algorithm: https://arxiv.org/abs/2309.12436\n")
+print("DC φ is an conjunction of predicates of the following form:")
+print("∀s, t ∈ R, s ≠ t: ¬(p_1 ∧ . . . ∧ p_m)\n")
+print("DCs involve comparisons between pairs of rows within a dataset.\n"
+      "A typical DC example, derived from a Functional Dependency such as A -> B,\n"
+      "is expressed as: `∀s, t ∈ R, s ≠ t, ¬(t.A == s.A and t.B ≠ s.B)`.\n"
```
Petua41, line 19:
```suggestion
      "is expressed as: \"∀s, t ∈ R, s ≠ t, ¬(t.A == s.A and t.B ≠ s.B)\".\n"
```

examples/datasets/taxes.csv
xJoskiy: Some time ago they explicitly forbidden newlines on the end. Now it has probably changed

examples/basic/README.md
```diff
@@ -25,4 +25,4 @@ These scenarios showcase a single pattern by discussing its definition and provi
 + [verifying_nd](https://github.com/Desbordante/desbordante-core/tree/main/examples/basic/verifying_nd) — a scenario showing how to verify numerical dependencies.
 + [verifying_mfd.py](https://github.com/Desbordante/desbordante-core/tree/main/examples/basic/verifying_mfd.py) — a scenario showing how to verify metric functional dependencies.
 + [verifying_ucc.py](https://github.com/Desbordante/desbordante-core/tree/main/examples/basic/verifying_ucc.py) — a scenario showing how to verify a unique column combination.
-
++ [verifying_dc.py](https://github.com/Desbordante/desbordante-core/tree/main/examples/basic/verifying_dc.py) — a scenarion showing how to verify a denial constraint.
```
Petua41, line 28:
```suggestion
+ [verifying_dc.py](https://github.com/Desbordante/desbordante-core/tree/main/examples/basic/verifying_dc.py) — a scenario showing how to verify a denial constraint.
```

Petua41 (approved):

BUYT-1 (changes requested):

src/core/algorithms/dc/verifier/dc_verifier.cpp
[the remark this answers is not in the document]
xJoskiy: Yep, for some reason it doesn't format it

src/core/algorithms/dc/model/component.h
```diff
@@ -0,0 +1,76 @@
+#pragma once
+
+#include <cstddef>
+#include <string>
+
+#include "algorithms/dc/model/operator.h"
+#include "model/types/builtin.h"
+#include "types.h"
```
BUYT-1, line 8:
```suggestion
#include "model/types/types.h"
```

src/tests/test_dc_verifier.cpp
```diff
@@ -0,0 +1,67 @@
+#include <memory>
+#include <string>
+
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
+#include "algorithms/algo_factory.h"
+#include "algorithms/dc/verifier/dc_verifier.h"
+#include "all_csv_configs.h"
+#include "names_and_descriptions.h"
```
BUYT-1, line 10:
```suggestion
#include "config/names_and_descriptions.h"
```

BUYT-1 (changes requested):
Some header names I missed. [Use full header names](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes).

src/core/algorithms/dc/verifier/dc_verifier.cpp
```diff
@@ -0,0 +1,521 @@
...
+#include <easylogging++.h>
+
+#include "algorithms/dc/model/component.h"
+#include "algorithms/dc/model/point.h"
+#include "algorithms/dc/model/predicate.h"
+#include "config/names_and_descriptions.h"
+#include "config/option_using.h"
+#include "config/tabular_data/input_table/option.h"
+#include "model/table/column_layout_relation_data.h"
+#include "util/kdtree.h"
```
BUYT-1, line 26:
```suggestion
#include "algorithms/dc/util/kdtree.h"
```

src/core/util/kdtree.h
```diff
@@ -0,0 +1,239 @@
...
+    PointType lower_bound_, upper_bound_;
+    std::vector<bound_type> lower_bound_type_, upper_bound_type_;
+
+    Rect() = default;
+    Rect(Rect<PointType> const&) = default;
+    Rect(Rect<PointType>&&) = default;
+    Rect<PointType>& operator=(Rect<PointType> const&) = default;
+
+    Rect(PointType const& lower_bound, PointType const& upper_bound)
+        : lower_bound_(lower_bound), upper_bound_(upper_bound) {
```
BUYT-1, line 36:
```suggestion
    Rect(PointType lower_bound, PointType upper_bound)
        : lower_bound_(std::move(lower_bound)), upper_bound_(std::move(upper_bound)) {
```

src/core/util/kdtree.h
```diff
@@ -0,0 +1,239 @@
...
+        upper_bound_type_ = std::vector<bound_type>(upper_bound.GetDim(), kOpen);
+    }
+
+    void SetBoundType(std::vector<bound_type> lower_bound_type,
+                      std::vector<bound_type> upper_bound_type) {
+        if (lower_bound_.GetDim() != upper_bound_.GetDim())
+            throw std::logic_error("Dimensionalities of given points don't match");
+
+        lower_bound_type_ = lower_bound_type;
+        upper_bound_type_ = upper_bound_type;
```
BUYT-1, line 50:
```suggestion
        lower_bound_type_ = std::move(lower_bound_type);
        upper_bound_type_ = std::move(upper_bound_type);
```

src/core/algorithms/dc/verifier/dc_verifier.h
```diff
@@ -0,0 +1,111 @@
...
+
+#include "algorithms/algorithm.h"
+#include "config/tabular_data/input_table/option.h"
+#include "config/tabular_data/input_table_type.h"
+#include "dc/model/dc.h"
+#include "dc/model/point.h"
+#include "frozen/unordered_map.h"
+#include "model/table/column_layout_relation_data.h"
+#include "table/typed_column_data.h"
+#include "util/kdtree.h"
```
BUYT-1, line 17:
```suggestion
#include "algorithms/algorithm.h"
#include "algorithms/dc/model/dc.h"
#include "algorithms/dc/model/point.h"
#include "algorithms/dc/util/kdtree.h"
#include "config/tabular_data/input_table/option.h"
#include "config/tabular_data/input_table_type.h"
#include "model/table/column_layout_relation_data.h"
#include "model/table/typed_column_data.h"

#include "frozen/unordered_map.h"
```

BUYT-1 (changes requested):
minimum: fix `clang-format` comments, maximum: fix all these

src/core/algorithms/dc/verifier/dc_verifier.cpp
xJoskiy: `kdtree.h` is located in `core/util/`

src/core/algorithms/dc/verifier/dc_verifier.cpp
```diff
@@ -0,0 +1,521 @@
...
+        if (!(this->*check)(cur_dc)) return false;
+    }
+
+    return true;
+}
+
+dc::DC DCVerifier::GetDC(std::vector<dc::Predicate> const& no_diseq_preds,
+                         std::vector<dc::Predicate> const& diseq_preds, size_t cur_signs) {
+    size_t preds_count = diseq_preds.size();
+    std::vector<dc::Operator> operators = {dc::Operator("<"), dc::Operator(">")};
```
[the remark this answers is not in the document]
xJoskiy: Since I've implemented constructor from `std::string`, I find initialization with "<" and ">" more informative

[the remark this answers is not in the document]
BUYT-1: Alright, I just thought there could be issues, but if it works fine, then good

src/core/algorithms/dc/verifier/dc_verifier.cpp
BUYT-1: Got it

src/core/algorithms/dc/verifier/dc_verifier.cpp
```diff
@@ -0,0 +1,521 @@
...
+
+bool DCVerifier::VerifyAllEquality(dc::DC const& dc) {
+    std::unordered_set<size_t> res_tuples;
+    std::vector<Column::IndexType> const eq_cols = dc.GetColumnIndices();
+    for (size_t i = 0; i < data_.front().GetNumRows(); ++i) {
+        if (ContainsNullOrEmpty(eq_cols, i)) continue;
+        std::vector<std::byte const*> tuple = GetTuple(i);
+        size_t key = HashTuple(tuple, eq_cols);
+        if (res_tuples.find(key) != res_tuples.end()) return false;
+        res_tuples.insert(key);
```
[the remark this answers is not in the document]
xJoskiy: Yes, that would work fine

src/core/algorithms/dc/verifier/dc_verifier.cpp
```diff
@@ -0,0 +1,521 @@
...
+        }
+    }
+
+    return res;
+}
+
+// FIX: DC may contain braces, numbers or spaces, naive separation doesn't belong here.
+std::vector<dc::Predicate> DCVerifier::SplitDC(std::string const& dc_string) {
+    size_t ind, start = 0;
+    std::string token, sep = " and ";
```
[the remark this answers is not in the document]
xJoskiy: Why static?

src/core/algorithms/dc/verifier/dc_verifier.cpp
```diff
@@ -0,0 +1,521 @@
...
+
+    return true;
+}
+
+dc::DC DCVerifier::GetDC(std::vector<dc::Predicate> const& no_diseq_preds,
+                         std::vector<dc::Predicate> const& diseq_preds, size_t cur_signs) {
+    size_t preds_count = diseq_preds.size();
+    std::vector<dc::Operator> operators = {dc::Operator("<"), dc::Operator(">")};
+    std::vector<dc::Predicate> cur_predicatates(no_diseq_preds);
+    cur_predicatates.reserve(preds_count);
```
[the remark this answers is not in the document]
xJoskiy: So, why don't just
```
std::vector<dc::Predicate> cur_predicatates(no_diseq_preds);
cur_predicatates.reserve(diseq_preds_count + no_diseq_preds.size());
```
?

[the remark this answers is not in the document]
BUYT-1: A vector owns a memory region of some size (this is capacity). When this region is resized, like after a `reserve` call, it acquires a new memory region and moves the previous elements into it, one-by-one.
Your way would first allocate a memory chunk to copy `no_diseq_preds` into, then `reserve` would allocate a new memory chunk, to where the copied elements would be moved. Reserving first avoids that: it allocates the chunk first, then copies the elements.
`reserve` doesn't increase the size of the memory chunk, it tries to set it: if the vector has capacity 5 and `reserve(2)` is called, it won't increase its capacity to 7 (or greater), it will try to (in a typical implementation) set it to 2, see that `2 < 5`, and refuse (do nothing).

src/core/algorithms/dc/verifier/dc_verifier.cpp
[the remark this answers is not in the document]
BUYT-1: I doubt it would matter much in this case, but `static` makes it so the variable doesn't occupy space on the stack during a call in this context.

src/core/algorithms/dc/verifier/dc_verifier.cpp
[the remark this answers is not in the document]
BUYT-1: If two different sets if values hash to the same number, this would falsely consider them equal and DC would be determined to not hold(?), even if it actually does (?).

src/core/algorithms/dc/verifier/dc_verifier.cpp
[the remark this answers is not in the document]
xJoskiy: That's fair, haven't considered case when the size is bigger then requested reserve size

src/core/algorithms/dc/verifier/dc_verifier.h
xJoskiy: It is formatted by clang-format. So it makes sense adding frozen to clang-format regex?

src/core/algorithms/dc/verifier/dc_verifier.cpp
```diff
@@ -0,0 +1,521 @@
...
+    for (size_t i = 0; i < preds_count; ++i) {
+        bool cur_sign = cur_signs & (1 << i);
+        dc::Operator cur_operator = operators[cur_sign];
+        auto left_op = diseq_preds[i].GetLeftOperand();
+        auto right_op = diseq_preds[i].GetRightOperand();
```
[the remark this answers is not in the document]
xJoskiy: I suppose, compiler could make such optimization itself

src/core/algorithms/dc/verifier/dc_verifier.h
BUYT-1: Yeah, it's a third-party library, so it should be before project files. Made an error in the suggestion, here is the right one.
```suggestion
#include "frozen/unordered_map.h"

#include "algorithms/algorithm.h"
#include "algorithms/dc/model/dc.h"
#include "algorithms/dc/model/point.h"
#include "algorithms/dc/util/kdtree.h"
#include "config/tabular_data/input_table/option.h"
#include "config/tabular_data/input_table_type.h"
#include "model/table/column_layout_relation_data.h"
#include "model/table/typed_column_data.h"
```

src/core/algorithms/dc/verifier/dc_verifier.cpp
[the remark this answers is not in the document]
xJoskiy: Hm, that actually may lead to hash collisions. We perhaps need to create a hashable structure that consists of `tuple` and `eq_cols` to be able to compare actual keys in hash table not simply their hash values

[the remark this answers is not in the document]
BUYT-1: Alternatively, write a custom hasher using the function you already have and plug it into the standard `std::unordered_set` as a template parameter

examples/basic/verifying_dc.py
```diff
@@ -0,0 +1,65 @@
...
+verifier = db.dc_verification.algorithms.Default()
+verifier.load_data(table=(TABLE_1, ',', True))
+
+# Algorithm execution
+verifier.execute(denial_constraint=DC)
+
+# Obtaining the result
+result: bool = verifier.dc_holds()
+
+print("DC " + DC + " holds: " + str(result) + "\n")
```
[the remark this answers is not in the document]
BUYT-1: Not done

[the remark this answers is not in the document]
BUYT-1:
```suggestion
print(f"DC {DC} holds: {result}\n")
```

src/core/util/kdtree.h
```diff
@@ -0,0 +1,239 @@
...
+        size_t axis_;
+
+        Node() = default;
+
+        Node(PointType const& point, Node* left, Node* right, size_t axis)
+            : point_(point), left_(left), right_(right), axis_(axis) {}
+
+        ~Node() {
+            if (left_ != nullptr) delete left_;
+            if (right_ != nullptr) delete right_;
```
BUYT-1, line 91: Just make them `std::unique_ptr<Node>`

src/core/util/kdtree.h
```diff
@@ -0,0 +1,239 @@
...
+}
+
+template <SubscriptableOrder PointType>
+void KDTree<PointType>::Erase(KDTree<PointType>::Node* dest) {
+    delete dest;
+}
+
+template <SubscriptableOrder PointType>
+KDTree<PointType>::~KDTree() {
+    Erase(root_);
```
BUYT-1, line 236: Use `std::unique_ptr<Node>`. Do you actually need a public `Erase` method?

src/core/util/kdtree.h
BUYT-1, line 1: Bring `if`s in line with [the style guide](https://google.github.io/styleguide/cppguide.html#Formatting_Looping_Branching). Use brackets in `if`s with an `else`, for example.

src/core/util/kdtree.h
```diff
@@ -0,0 +1,248 @@
...
+        Node(Node&&) = delete;
+        Node(Node const&) = delete;
+        Node& operator=(Node const&) = delete;
+        Node& operator=(Node&&) = delete;
+    };
+
+    std::unique_ptr<Node> root_;
+    size_t size_;
+
+    void Erase(std::unique_ptr<Node>& dest);
```
BUYT-1, line 108: `Erase` is only used in the destructor and the default destructor does the same thing. I think it should be removed.

src/core/util/kdtree.h
```diff
@@ -0,0 +1,241 @@
...
+                         Rect<PointType> const& box) const;
+
+public:
+    KDTree() : root_(nullptr), size_(0) {};
+
+    KDTree(std::vector<PointType> const& points);
+    KDTree(std::initializer_list<PointType> const& points);
+    KDTree& operator=(KDTree&& tree);
+    KDTree(KDTree&& tree);
+    ~KDTree();
```
BUYT-1, line 121:
```suggestion
    ~KDTree() = default;
```

BUYT-1 (approved):
👍

xJoskiy:
Removed unnecessary file configs

PR #449, opened 2024-08-15
Add ND verification bindings and examples

Petua41:
Add python bindings and examples for ND-Verifier algorithm

### DIFF
--- a/examples/basic/verifying_nd/verifying_nd_1.py
+++ b/examples/basic/verifying_nd/verifying_nd_1.py
@@ -1,0 +1,82 @@
+import desbordante
+from csv import reader
+
+RED = '\033[31m'
+GREEN = '\033[32m'
+CYAN = '\033[1m\033[36m'
+ENDC = '\033[0m'
+
+def row_to_padded_string(row: list[str], widths: list[int]) -> str:
+    return ''.join(field.ljust(width) for width, field in zip(widths, row))
+
+def print_table(filename: str, row_numbers: list[int] = []) -> None:
+    with open(filename, newline='') as table:
+        rows = list(reader(table, delimiter=','))
+
+    needed_rows = [rows[0]]
+    if len(row_numbers) > 0:
+        for row_num in row_numbers:
+            needed_rows.append(rows[row_num + 1])
+    else:
+        needed_rows = rows
+
+    column_widths = []
+    for col_num in range(len(needed_rows[0])):
+        max_len = max(len(row[col_num]) for row in needed_rows)
+        column_widths.append(max_len + 3)
+
+    header, *data_rows = needed_rows
+    print(row_to_padded_string(header, column_widths))
+    print('-' * sum(column_widths))
+    print('\n'.join(row_to_padded_string(row, column_widths) for row in data_rows))
+
+EXPIRED_PASSPORT_TABLE = 'examples/datasets/nd_verification_datasets/expired_passport.csv'
+VALID_PASSPORTS_TABLE = 'examples/datasets/nd_verification_datasets/valid_passports_1.csv'
+
+print(f'''{CYAN}This example illustrates the usage of Numerical Dependencies (NDs).
+Intuitively, given two sets of attributes X and Y, there is an ND from X to Y (denoted X -> Y, weight = k)
+if each value of X can never be associated to more than k distinct values of Y.
+For more information consult "Efficient derivation of numerical dependencies" by Paolo Ciaccia et al.
+{ENDC}''')
+
+print('Citizens of Arstozka can have no more than two documents: one Arstozka passport and one exit permit.')
+print('The following table contains records of some citizens\' documents:')
+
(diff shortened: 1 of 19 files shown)


src/python_bindings/nd/bind_nd.cpp
```diff
@@ -0,0 +1,26 @@
...
+    auto nd_module = main_module.def_submodule("nd");
+    py::class_<ND>(nd_module, "ND")
+            .def("__str__", &ND::ToLongString)
+            .def("to_short_string", &ND::ToShortString)
+            .def("to_long_string", &ND::ToLongString)
+            .def("get_lhs", &ND::GetLhs)
+            .def("get_rhs", &ND::GetRhs)
+            .def("get_lhs_indices", &ND::GetLhsIndices)
+            .def("get_rhs_indices", &ND::GetRhsIndices)
+            .def("get_weight", &ND::GetWeight);
```
yakovypg, line 24: I think it is better to use `.def_property_readonly(...)`.

examples/basic/verifying_nd/verifying_nd_1.py
```diff
@@ -0,0 +1,82 @@
...
+
+algo = desbordante.nd_verification.algorithms.NDVerifier()
+algo.load_data(table=(EXPIRED_PASSPORT_TABLE, ',', True))
+algo.execute(lhs_indices=[0], rhs_indices=[1], weight=2)
+
+highlights = algo.highlights
+
+print('We need to validate these data')
+print('Let\'s run ND verification algorithm to check that every citizen has no more than two records:')
+print('\tND: [Name] -> [ID] with weight 2')
```
yakovypg, line 57: lhs and rhs are sets, so I think that it is better to print '{Name} -> {ID}' instead of '[Name] -> [ID]' because it can be understood as a list.

chernishev:
> Both examples are related to dependency '{Name} -> {ID} with weight 2', but NDs can contain more than one attribute in its lhs and rhs. These examples don't fully reflect the capabilities of NDs, so you should add an overview of a complex dependency such as '{A, B} -> {C}' or similar.

Good idea, I totally forgot about this, when discussed it.

src/core/algorithms/nd/nd.h
```diff
@@ -25,6 +27,14 @@ class ND {
         return rhs_;
     }
 
+    [[nodiscard]] std::vector<std::string> GetLhsNames() const {
+        return GetVerticalNames(GetLhs());
+    }
+
+    [[nodiscard]] std::vector<std::string> GetRhsNames() const {
+        return GetVerticalNames(GetRhs());
+    }
```
yakovypg, line 36: After moving the `GetVerticalNames()` implementation of these methods should be placed in .cpp file.

src/python_bindings/nd/bind_nd.cpp
```diff
@@ -0,0 +1,44 @@
+#include "nd/bind_nd.h"
+
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+
+#include "algorithms/nd/nd.h"
+#include "py_util/bind_primitive.h"
+
+namespace py = pybind11;
+using namespace model;
```
yakovypg, line 10: Don't place `using namespace` here. Place it in methods if necessary.

src/core/algorithms/nd/nd.h
```diff
@@ -40,6 +50,14 @@ class ND {
     [[nodiscard]] std::string ToShortString() const;
 
     [[nodiscard]] std::string ToLongString() const;
+
+    [[nodiscard]] std::tuple<std::vector<std::string>, std::vector<std::string>, WeightType>
+    ToNameTuple() const {
+        return std::make_tuple(GetLhsNames(), GetRhsNames(), GetWeight());
+    }
```
yakovypg, line 57: After moving the `GetVerticalNames()` implementation of this method should be placed in .cpp file.

examples/basic/verifying_nd/verifying_nd_1.py
```diff
@@ -0,0 +1,82 @@
+import desbordante
+from csv import reader
+
+RED = '\033[31m'
+GREEN = '\033[32m'
+CYAN = '\033[1m\033[36m'
+ENDC = '\033[0m'
+
+def row_to_padded_string(row: list, widths: list):
```
yakovypg, line 9: I see you added type hints. But don't forget to add a hint for the return type.
```py
def foo() -> int:
    return 0;

def bar(n: int) -> None:
    print(n)
```

yakovypg (commented):
This is no longer necessary in the current pull request, but I will say some words for the future. During development you shouldn't do a lot of force-pushing. The full commit history is more convenient for the reviewer. Only after the reviewer has finished reviewing the pull request you should correct the commit history and do a force-push.

I have already looked at all the files, all you have to do is correct the last comments.

yakovypg (approved):

PR #456, opened 2024-09-08
Optimize hymd

BUYT-1:


BUYT-1:
`clang-tidy` reports false positives.

chernishev (approved):
Unfortunately:
1) we don't have anyone who can properly review this huge and complex PR
2) we are time-constrained, as we need matching dependencies for the upcoming release
Therefore, I approve and merge this PR without review. Alexey is a skilled programmer and he did his best here. I hope this should be sufficient.

PR #458, opened 2024-09-10
Add new measures to Tane

egshnov:
This pr implements mining of afds based on error measures pdep, tau, mu_plus, rho as in "Measuring Approximate Functional Dependencies: A Comparative Study"


src/core/algorithms/fd/tane/tane.cpp
```diff
@@ -1,27 +1,170 @@
...
+
+    config::ErrorType sum = 0;
+
+    std::unordered_map<int, unsigned> x_frequencies;
+
+    int x_value_id = 1;
+    for (Cluster const& x_cluster : x_index) {
+        x_frequencies[x_value_id++] = x_cluster.size();
+    }
+    x_frequencies[0] = 1;
```
Sched71, line 88: Shouldn't here be: `std::unordered_map<unsigned, std::size_t> x_frequencies`?

src/core/algorithms/fd/tane/tane.cpp
```diff
@@ -1,27 +1,170 @@
...
+    config::ErrorType mu = 1 - (1 - pdep_xy) / (1 - pdep_y) * (n - 1) / (n - k);
+    config::ErrorType mu_plus = std::max(0., mu);
+    return mu_plus;
+}
+
+config::ErrorType Tane::CalculateRhoMeasure(model::PositionListIndex const* x_pli,
+                                            model::PositionListIndex const* xa_pli) {
+    auto calculate_dom = [](model::PositionListIndex const* pli) {
+        auto index = pli->GetIndex();
+        int dom = index.size();
```
Sched71, line 154: Maybe not int, but std::size_t or something? We could be in trouble if index size can be bigger than int can hold

Sched71 (changes requested):

Sched71 (approved):

chernishev (approved):
Dmitriy Shcheka looked into and approved

PR #459, opened 2024-09-11
Add UCC and AUCC examples

Petua41:


examples/advanced/comparison_ucc_and_aucc_1.py
```diff
@@ -0,0 +1,89 @@
...
+else:
+    print(f'{RED}UCC does not hold{ENDC}, but AUCC holds with threshold = {ver_algo.get_error():1.3f}')
+
+print('Threshold is small. It means that, possibly, there is an error in this column.')
+print('Let\'s look at the table again:')
+print(CYAN, end='')
+print_table(TABLE)
+print(ENDC)
+
+print(f'''There are two {CYAN}Harrys{ENDC}. They have different work experience, so it\'s
```
MichaelS239, line 71:
```suggestion
print(f'''There are two {CYAN}Harrys{ENDC}. They have different work experience, so they are
```

examples/basic/mining_aucc.py
```diff
@@ -0,0 +1,99 @@
...
+print()
+
+print(f'''Out of single-column UCCs, {CYAN}Name{ENDC} requires the smallest threshold to be "unique".
+Let\'s look at the table again, paying a special attention to the {CYAN}Name{ENDC} column:''')
+print(CYAN, end='')
+print_table(TABLE)
+print(ENDC)
+
+print(f'''There are two {CYAN}Harrys{ENDC}. They have different work experience and salary,
+therefore it\'s two different employees. This is most likely an error/oversight in data.
```
MichaelS239, line 83:
```suggestion
therefore they are two different employees. This is most likely an error/oversight in data.
```

src/python_bindings/ucc/bind_ucc.cpp
```diff
@@ -1,17 +1,29 @@
 #include "bind_ucc.h"
 
+#include <algorithm>
+
```
MichaelS239, line 4: Add `#include <cstddef>` (that's where `std::size_t` is defined)

MichaelS239 (commented):
Everything else is OK. You only need to fix the last two comments and squash the commits with fixes

MichaelS239 (approved):

PR #461, opened 2024-09-13
Fix: add support output for ARs

Petua41:
Add output of the support value for Association Rules

### DIFF
--- a/src/tests/test_apriori.cpp
+++ b/src/tests/test_apriori.cpp
@@ -47,16 +47,16 @@ void CheckSupportAndConfidence(std::list<model::ARStrings> const& actual,
         std::set<std::string> actual_rhs(rule.right.begin(), rule.right.end());
 
         if (lhs == actual_lhs && rhs == actual_rhs) {
-            ASSERT_DOUBLE_EQ(rule.support, expected_support)
+            EXPECT_DOUBLE_EQ(rule.support, expected_support)
                     << "supports don't match: expected " << expected_support
                     << ", got: " << rule.support;
-            ASSERT_DOUBLE_EQ(rule.confidence, expected_confidence)
+            EXPECT_DOUBLE_EQ(rule.confidence, expected_confidence)
                     << "confidences don't match: expected " << expected_confidence
                     << ", got: " << rule.confidence;
             return;
         }
     }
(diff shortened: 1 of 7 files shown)


src/tests/test_apriori.cpp
```diff
@@ -318,4 +338,28 @@ TEST_F(ARAlgorithmTest, RepeatedExecutionConsistentResult) {
     }
 }
 
+TEST_F(ARAlgorithmTest, SupportAndConfidenceSingular) {
+    auto algorithm = CreateAlgorithmInstance(kRulesBook, 0.2, 0.5, 0, 1);
```
Sched71, line 342: I guess last parameter should be true

src/tests/test_apriori.cpp
```diff
@@ -39,6 +39,26 @@ void CheckAssociationRulesListsEquality(
...
+        std::set<std::string> actual_lhs(rule.left.begin(), rule.left.end());
+        std::set<std::string> actual_rhs(rule.right.begin(), rule.right.end());
+
+        if (lhs == actual_lhs && rhs == actual_rhs) {
+            ASSERT_DOUBLE_EQ(rule.support, expected_support)
+                    << "supports don't match: expected " << expected_support
+                    << ", got: " << rule.support;
+            ASSERT_DOUBLE_EQ(rule.confidence, expected_confidence)
+                    << "confidences don't match: expected " << expected_confidence
+                    << ", got: " << rule.confidence;
```
Sched71, line 55: If I remember correctly Assert will make test fail immedeatly. So you'll never get an output from the second one. I guess u can change them to Expect and than make it fail manually by Fail instead of return

Sched71 (changes requested):
Everything good in my opinion. Just fix two minor moments

src/tests/test_apriori.cpp
Petua41: No, it shouldn't. The last parameter is `itemColumnIndex`, which is `unsigned int`.

Sched71 (approved):
Looks fine to me!

chernishev (approved):
Approve on behalf of Dmitriy Shcheka.

PR #467, opened 2024-09-28
Implement `INDVerifier` and add an error threshold field to the `IND` class.

vs9h:
Description:
- Add an error threshold field to the `IND` class
- Introduce a field in the `IND` class to store an error threshold that represents the degree of `IND` satisfaction.
- Adjust the registration methods in the approximate `INDAlgorithm`s to provide the error threshold.
- Implement `INDVerifier`, add python bindings and tests
- Add examples for `AIND` primitive


alexandrsmirn (commented):
Вроде всё ок, только несколько минорных комментов оставил. Посмотрел всё, кроме последнего коммита. Чуть позже дополню если появятся комментарии.

chernishev:
> Вроде всё ок, только несколько минорных комментов оставил. Посмотрел всё, кроме последнего коммита. Чуть позже дополню если появятся комментарии.

Спасибо большое!

src/core/algorithms/ind/mind/mind.h
```diff
@@ -44,7 +45,13 @@ class Mind final : public INDAlgorithm {
...
     void LoadINDAlgorithmDataInternal() override;
 
-    bool TestCandidate(RawIND const& raw_ind);
+    ///
+    /// Test a given IND candidate to determine if it should be registered.
+    ///
+    /// \return `false` with `std::nullopt` if the candidate should not be registered,
+    ///         otherwise return `true` along with the associated error threshold.
+    ///
+    std::pair<bool, std::optional<config::ErrorType>> TestCandidate(RawIND const& raw_ind);
```
ol-imorozko, line 54: So `false` always means no value, and `true` means that the value will be present. I feel like we can simplify this logic by simply utilizing `std::optional<config::ErrorType>`, the interface will be much simplier

src/core/algorithms/ind/mind/mind.cpp
```diff
@@ -246,8 +250,9 @@ void Mind::MineNaryINDs() {
         prev_it = std::prev(INDList().end()); /*< last element of the previous lattice level */
         prev_raw_inds.clear();
         for (RawIND const& candidate : candidates) {
-            if (TestCandidate(candidate)) {
-                RegisterIND(candidate.lhs, candidate.rhs);
+            auto const& [is_valid, error_opt] = TestCandidate(candidate);
+            if (is_valid) {
+                RegisterIND(candidate.lhs, candidate.rhs, error_opt.value());
                 prev_raw_inds.insert(candidate);
             }
```
ol-imorozko, line 256:
```suggestion
            if (auto error_opt = TestCandidate(candidate)) {
                RegisterIND(candidate.lhs, candidate.rhs, error_opt.value());
                prev_raw_inds.insert(candidate);
            }
```

src/core/algorithms/ind/spider/attribute.h
```diff
@@ -84,6 +84,12 @@ class AINDAttribute final : public details::Attribute {
 
     /// get referenced attribute indices
     std::vector<AttributeIndex> GetRefIds(config::ErrorType max_error) const;
+
+    /// get error threshold
+    config::ErrorType GetError(AttributeIndex ref_id) const {
+        auto const dep_count = static_cast<config::ErrorType>(occurrences_[id_]);
+        return 1 - occurrences_[ref_id] / dep_count;
+    }
```
ol-imorozko, line 92: I don't know the logic behind filling `occurrences_` but when I see a variable with name `count`something I assume it _can_ be zero, in which case we will be dividing by it. `dep_count` cannot be zero, right?

src/core/model/table/dataset_stream_wrapper.h
ol-imorozko, line 1: `Fix support for raw stream pointers in DatasetStreamWrapper`:
Isn't that commit just fixes codestyle? I don't see any changes in behavior

src/python_bindings/data/bind_data_types.cpp
```diff
@@ -23,6 +34,11 @@ void BindDataTypes(py::module_& main_module) {
     using namespace model;
     py::class_<ColumnCombination>(data_module, "ColumnCombination")
             .def("__str__", &ColumnCombination::ToString)
+            .def("to_index_tuple",
+                 [](ColumnCombination const& cc) {
+                     return py::make_tuple(cc.GetTableIndex(),
+                                           VectorToTuple(cc.GetColumnIndices()));
+                 })
```
ol-imorozko, line 41: As far as I know, `py::cast()` can convert a `std::vector` directly to a python tuple without the need to manually iterate over the elements.

ol-imorozko:
```cpp
return py::make_tuple(cc.GetTableIndex(), py::cast(cc.GetColumnIndices()));
```

src/core/algorithms/ind/ind_verifier/ind_verifier.cpp
```diff
@@ -0,0 +1,114 @@
+#include "ind_verifier.h"
+
+#include <unordered_set>
+
+#include <boost/functional/hash.hpp>
+
+#include "config/indices/option.h"
+#include "config/option_using.h"
```
ol-imorozko, line 8: feels like `option_using` is unused header

src/core/algorithms/ind/ind_verifier/ind_verifier.h
```diff
@@ -0,0 +1,86 @@
...
+
+    using Cluster = std::vector<model::TupleIndex>;
+    using Error = config::ErrorType;
+
+private:
+    config::InputTables input_tables_;
+
+    /* IND to verify */
+    RawIND ind_;
+    config::EqNullsType is_null_equal_null_;
```
ol-imorozko, line 28: `is_null_equal_null_` field is unused

src/core/algorithms/ind/ind_verifier/ind_verifier.cpp
```diff
@@ -0,0 +1,114 @@
...
+    std::unordered_map<Row, ValueIndex, boost::hash<Row>> lhs_violating_row_to_id_map;
+    std::unordered_map<ValueIndex, Cluster> lhs_id_to_cluster_map;
+
+    ValueIndex next_value_id = 0;
+    model::TupleIndex current_row_id = 0;
+
+    std::unordered_set<Row, boost::hash<Row>> lhs_rows;
+
+    DatasetStreamFixedProjection lhs_projected_stream{lhs_table, ind_.lhs};
+    while (lhs_projected_stream.HasNextRow()) {
```
ol-imorozko, line 80: what if `HasNextRow` will return false? Or is it guaranteed there to be at least one row? 

If not, then `lhs_rows` will be empty and we will divide by zero later:
```cpp
    model::TupleIndex lhs_cardinality = lhs_rows.size();
    error_ = static_cast<Error>(violating_unique_rows) / lhs_cardinality;
```

src/core/algorithms/ind/ind_verifier/ind_verifier.cpp
```diff
@@ -0,0 +1,114 @@
...
+        rhs_rows.insert(rhs_stream.GetNextRow());
+    }
+
+    if (lhs_table == rhs_table) {
+        /* We would iterate through the same table, so we should reset this table */
+        lhs_table->Reset();
+    }
+
+    std::unordered_map<Row, ValueIndex, boost::hash<Row>> lhs_violating_row_to_id_map;
+    std::unordered_map<ValueIndex, Cluster> lhs_id_to_cluster_map;
```
ol-imorozko, line 72: why not just one map from Row to a Cluster?, smth like
```cpp
std::unordered_map<Row, Cluster, boost::hash<Row>> violating_clusters;
```

src/core/algorithms/ind/ind_verifier/ind_verifier.cpp
```diff
@@ -0,0 +1,114 @@
...
+        auto const& [it, inserted] = lhs_violating_row_to_id_map.try_emplace(row, next_value_id);
+        if (inserted) {
+            ++next_value_id;
+        }
+
+        ValueIndex value_id = it->second;
+        lhs_id_to_cluster_map[value_id].push_back(current_row_id);
+
+        ++violating_rows_;
+        ++current_row_id;
```
ol-imorozko, line 98: I feel like just
```cpp
        if (!rhs_rows.contains(row)) {
            violating_clusters[row].push_back(current_row_id);
            ++violating_rows_;
        }

        ++current_row_id;
```
is enough

src/core/algorithms/ind/ind_verifier/ind_verifier.cpp
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    for (auto& [value_id, cluster] : lhs_id_to_cluster_map) {
+        clusters_.push_back(std::move(cluster));
+    }
```
ol-imorozko, line 103:
```suggestion
    for (auto& [row, cluster] : violating_clusters) {
        clusters_.push_back(std::move(cluster));
    }
```

ol-imorozko (changes requested):

src/core/algorithms/ind/spider/attribute.h
vs9h: Yes, there is a guarantee that the `dep_count` is not zero. This is actually the number of unique values ​​for the dependent attribute (`this->id_`). And we know that at least one value exists, since we do not process empty tables.

src/core/model/table/dataset_stream_wrapper.h
vs9h: The main thing here is adding `else`, otherwise the code simply won't compile in the case where `DatasetStream = IDatasetStream *`.

The compiler will see this code as:
```cpp
            if (true) {
                return stream_;
            }
            return stream_.get();
```

So, it will fail because `IDatasetStream *stream_` don't have `get()` method.

src/tests/test_ind_algorithms.cpp
```diff
@@ -43,7 +44,15 @@ class INDAlgorithmTest : public ::testing::Test {
...
                 auto ind_algo = CreateAlgorithmInstance(csv_configs, test_config);
                 ind_algo->Execute();
-                EXPECT_EQ(HashVec(ToSortedINDTestVec(ind_algo->INDList()), HashPair), hash)
+
+                std::list<model::IND> const& inds = ind_algo->INDList();
+                // in this test we're mining only exact INDs
+                for (auto const& ind : inds) {
+                    EXPECT_EQ(ind.GetError(), config::ErrorType{0.0})
+                            << "Error threshold mismatch for IND: " << ind.ToShortString();
+                }
```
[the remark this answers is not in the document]
vs9h: Но эта информация уже есть в `EXPECT_EQ(ind.GetError(), config::ErrorType{0.0}`, в выводе будет показано полученное и ожадаемое значения.

И мы придерживаемся общения на английском на гитхабе

src/core/config/tabular_data/input_tables/option.cpp
```diff
@@ -1,8 +1,35 @@
...
+std::string_view InputTablesOption::GetName() const {
+    return common_option_.GetName();
+}
+
+Option<InputTables> InputTablesOption::operator()(InputTables* value_ptr,
+                                                  TableIndex tables_limit) const {
+    Option<InputTables> option = common_option_(value_ptr);
+    option.SetValueCheck([tables_limit](InputTables const& tables) {
+        assert(tables_limit != 0);
+        assert(!tables.empty());
```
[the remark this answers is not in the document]
vs9h: В `algo_factory`, что как минимум одна таблица есть. Но кажется, что лучше все же сюда перенести эту проверку.

rpuwa (commented):
there are some minor issues, but overall the algorithm is correct

src/core/algorithms/algo_factory.cpp
```diff
@@ -55,10 +55,6 @@ void LoadAlgorithm(Algorithm& algorithm, StdParamsMap const& options) {
             return boost::any{table};
         } else if (option_name == kTables && options.find(std::string{kTables}) == options.end()) {
             auto csv_configs = GetOptionValue<std::vector<CSVConfig>>(options, kCsvConfigs);
-            if (csv_configs.empty()) {
-                throw config::ConfigurationError("Expected collection of csv configs");
-            }
-
```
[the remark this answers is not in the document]
vs9h: For now, we do this check in the `InputTablesOption::operator()` operator.
It's more reasonable to do this check in `value_check_func`, where we should check if the option value is set correctly.

And it became more obvious after I started adding the option to limit the number of tables. It makes sense to do both checks in one place.

src/core/algorithms/ind/faida/preprocessing/preprocessor.cpp
```diff
@@ -1,15 +1,15 @@
...
 
 namespace algos::faida {
 
 std::unique_ptr<Preprocessor> Preprocessor::CreateHashedStores(
         std::string const& dataset_name,
         std::vector<std::shared_ptr<model::IDatasetStream>> const& data_streams, int sample_goal) {
-    if (data_streams.empty()) {
-        throw std::runtime_error("Got an empty dataset: IND mining is meaningless.");
-    }
+    assert(!data_streams.empty());
```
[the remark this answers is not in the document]
vs9h: This check is performed when configuring the algorithm, so there is no need to do this check here and try to throw an error.

But we can leave an assert to explicitly tell the reader that we have some precondition that zero tables should not be passed to this function.

src/core/algorithms/ind/ind_verifier/ind_verifier.h
vs9h: marked as todo

alexandrsmirn (approved):

vs9h:
@BUYT-1, can you please look at the [latest commit](https://github.com/Desbordante/desbordante-core/pull/467/commits/a0828467e71ef15e235c47f8b8db1d520cc3c165)? It's a small fix related to options, namely resolving to type index. This is needed for pipeline algorithms with bindings.

BUYT-1:
> @BUYT-1, can you please look at the [latest commit](https://github.com/Desbordante/desbordante-core/pull/467/commits/a0828467e71ef15e235c47f8b8db1d520cc3c165)? It's a small fix related to options, namely resolving to type index. This is needed for pipeline algorithms with bindings.

Looks okay, though I'm not sure you want to do that. Is it not possible to move common code into functions and call them?

vs9h:
> > @BUYT-1, can you please look at the [latest commit](https://github.com/Desbordante/desbordante-core/pull/467/commits/a0828467e71ef15e235c47f8b8db1d520cc3c165)? It's a small fix related to options, namely resolving to type index. This is needed for pipeline algorithms with bindings.
> 
> Looks okay, though I'm not sure you want to do that. Is it not possible to move common code into functions and call them?

The current idea is that this algorithm can use any of the algorithms for mining unary dependencies. In the future it is expected that the user will be able to choose which algorithm to use for mining unary ones. For this reason, I did not duplicate the algorithm code or take out the general implementation, etc.

BUYT-1:
> > > @BUYT-1, can you please look at the [latest commit](https://github.com/Desbordante/desbordante-core/pull/467/commits/a0828467e71ef15e235c47f8b8db1d520cc3c165)? It's a small fix related to options, namely resolving to type index. This is needed for pipeline algorithms with bindings.
> > 
> > 
> > Looks okay, though I'm not sure you want to do that. Is it not possible to move common code into functions and call them?
> 
> The current idea is that this algorithm can use any of the algorithms for mining unary dependencies. In the future it is expected that the user will be able to choose which algorithm to use for mining unary ones. For this reason, I did not duplicate the algorithm code or take out the general implementation, etc.

Okay, that makes sense

BUYT-1 (approved):
The `Algorithm` changes are okay

ol-imorozko (approved):

PR #468, opened 2024-09-30
Implement set based aod verifier, support aod mining in fastod

polyntsov:


src/core/config/column_index/option.cpp
```diff
@@ -0,0 +1,31 @@
...
+    option.SetValueCheck(
+            [get_col_count = std::move(get_col_count),
+             value_check_func = std::move(value_check_func)](config::IndexType const index) {
+                config::ValidateIndex(index, get_col_count());
+                if (value_check_func) {
+                    value_check_func(index);
+                }
+            });
+    return option;
+}
```
ol-imorozko, line 29: We're already in `config` namespace so things like `config::IndexType` or `config::ValidateIndex` are redundant, just `IndexType`/`ValidateIndex` is enough

src/core/config/column_index/option.cpp
```diff
@@ -0,0 +1,31 @@
+#include "config/column_index/option.h"
+
+#include "config/column_index/validate_index.h"
+
+namespace config {
+
+ColumnIndexOption::ColumnIndexOption(
+        std::string_view name, std::string_view description,
+        typename Option<config::IndexType>::DefaultFunc get_default_value)
```
ol-imorozko, line 9: what's the point of `typename` here? It's already a fully resolved type due to `<config::IndexType>`

src/core/config/column_index/option.cpp
```diff
@@ -0,0 +1,31 @@
...
+    : common_option_(name, description, std::move(get_default_value)) {}
+
+std::string_view ColumnIndexOption::GetName() const {
+    return common_option_.GetName();
+}
+
+Option<config::IndexType> ColumnIndexOption::operator()(
+        config::IndexType* value_ptr, std::function<config::IndexType()> get_col_count,
+        typename Option<config::IndexType>::ValueCheckFunc value_check_func) const {
```
ol-imorozko, line 18: same, `typename` not needed

src/core/config/indices/option.cpp
```diff
@@ -27,14 +32,17 @@ Option<config::IndicesType> IndicesOption::operator()(
...
-                    throw ConfigurationError("Indices cannot be empty");
+                if (!allow_empty && indices.empty()) {
+                    throw ConfigurationError(std::string{option_name} + " cannot be empty");
                 }
                 static_assert(std::is_unsigned_v<config::IndexType>);
                 assert(std::is_sorted(indices.begin(), indices.end()));
-                config::ValidateIndex(indices.back(), get_col_count());
+                if (!indices.empty()) {
+                    config::ValidateIndex(indices.back(), get_col_count());
+                }
```
ol-imorozko, line 45: redundant `config`

src/core/algorithms/od/fastod/partitions/complex_stripped_partition.cpp
```diff
@@ -93,6 +93,115 @@ void ComplexStrippedPartition::ToStrippedPartition() {
...
+            }
+        } else {
+            for (size_t i = group_begin; i < group_end; ++i) {
+                DataFrame::Range const range = (*rb_indexes_)[i];
+
+                for (size_t j = range.first; j <= range.second; ++j) {
+                    values.emplace_back(data_->GetValue(j, left), data_->GetValue(j, right));
+                }
+            }
+        }
```
ol-imorozko: IDK if I should comment on this code since you've just moved it to another place, but the `else` branch does not have the `values.reserve`

src/core/algorithms/od/fastod/partitions/complex_stripped_partition.cpp
```diff
@@ -93,6 +93,115 @@ void ComplexStrippedPartition::ToStrippedPartition() {
...
+                prev_group_max_index = current_group_max_index;
+                current_group_max_index = i;
+            } else if (values[current_group_max_index].second <= second) {
+                current_group_max_index = i;
+            }
+
+            if (!is_first_group && values[prev_group_max_index].second > second) {
+                return true;
+            }
+        }
```
ol-imorozko: This loop constantly checking for `i != 0`.
Moreover, the first iteration of this loop will always look like
```cpp
if (values[0].second <= values[0].second) //current_group_max_index is zero
    current_group_max_index = 0;
```
since `i == 0` and `!is_first_group` is false.

So we can just start from `i = 1` and remove that check

src/core/algorithms/od/fastod/partitions/complex_stripped_partition.cpp
```diff
@@ -93,6 +93,115 @@ void ComplexStrippedPartition::ToStrippedPartition() {
...
+
+        if constexpr (Ascending) {
+            std::sort(values.begin(), values.end(),
+                      [](auto const& p1, auto const& p2) { return p1.first < p2.first; });
+        } else {
+            std::sort(values.begin(), values.end(),
+                      [](auto const& p1, auto const& p2) { return p2.first < p1.first; });
+        }
```
ol-imorozko: Lambdas looks identical, I would do
```cpp
std::sort(values.begin(), values.end(),
          [](auto const& p1, auto const& p2) {
              return Ascending ? p1.first < p2.first : p1.first > p2.first;
          });
```
Though I'm not sure whether we can use constexpr values with ternary operator, but I think that with lambdas we can

ol-imorozko (changes requested):
First three commits:

ol-imorozko (changes requested):

src/core/algorithms/algorithm.cpp
```diff
@@ -112,6 +116,18 @@ void Algorithm::SetOption(std::string_view option_name, boost::any const& value)
...
+bool Algorithm::OptionIsRequired(std::string_view option_name) const {
+    if (bool ext_opt_is_required = ExternalOptionIsRequired(option_name); ext_opt_is_required) {
+        return true;
+    }
+
+    auto it = possible_options_.find(option_name);
+    if (it == possible_options_.end()) {
+        return false;
+    }
+    return it->second->IsRequired();
```
ol-imorozko: nit:
```suggestion
    return it != possible_options_.end() && it->second->IsRequired();
```

ol-imorozko (changes requested):

src/core/algorithms/od/set_based_verifier/verifier.cpp
```diff
@@ -0,0 +1,134 @@
...
+            config::ColumnIndexOption{kOcLeftIndex, kDOcLeftIndex}(&oc_->left, get_cols_num));
+    RegisterOption(
+            config::ColumnIndexOption{kOcRightIndex, kDOcRightIndex}(&oc_->right, get_cols_num));
+    RegisterOption(Option<Ordering>{&oc_->left_ordering, kOcLeftOrdering, kDODLeftOrdering,
+                                    Ordering::ascending});
+    RegisterOption(
+            config::IndicesOption{kOFDContext, kDOFDContext, true}(&ofd_->context, get_cols_num)
+                    .SetBeforeSetCallback(ofd_before_set)
+                    .SetConditionalOpts({{nullptr, {kOFDRightIndex}}})
+                    .SetIsRequiredFunc([this]() { return !oc_.has_value(); }));
```
ol-imorozko, line 55: IndicesOption regestiring looks very similar. Maybe we can extract it to a separate method, like `RegisterIndicesOption`

and use it like
```cpp
    RegisterIndicesOption(kOcContext, oc_, oc_before_set,
                          {kOcLeftIndex, kOcRightIndex, kOcLeftOrdering},
                          [this]() { return !ofd_.has_value(); });

    RegisterIndicesOption(kOFDContext, ofd_, ofd_before_set,
                          {kOFDRightIndex},
                          [this]() { return !oc_.has_value(); });
```

src/core/algorithms/od/set_based_verifier/verifier.h
```diff
@@ -0,0 +1,74 @@
...
+        config::IndicesType context;
+        config::IndexType right;
+    };
+
+    config::InputTable input_table_;
+    std::optional<OC> oc_;
+    std::optional<OFD> ofd_;
+
+    Error error_;
+    RemovalSet removal_set_;
```
ol-imorozko, line 37: mb `error_ = 0.0`?

src/core/config/descriptions.h
```diff
@@ -109,4 +112,10 @@ constexpr auto kDInsertStatements = "Rows to be inserted into the table using th
 constexpr auto kDDeleteStatements = "Rows to be deleted from the table using the delete operation";
 constexpr auto kDUpdateStatements = "Rows to be replaced in the table using the update operation";
 constexpr auto kDNDWeight = "Weight of ND to verify (positive integer)";
+constexpr auto kDOcContext = "Context of the OC to verify";
+constexpr auto kDOcLeftIndex = "Index of the left attribute of the OC to verify";
+constexpr auto kDOcRightIndex = "Index of the right attribute of the OC to verify";
+auto const kDODLeftOrdering = details::kDODLeftOrdering;
+constexpr auto kDOFDContext = "Context of the OFD to verify";
```
ol-imorozko, line 119: shouldn't it be `details::kDODLeftOrdering.data()` as all the similar lines here that defines `char *`?

src/tests/test_set_based_verifier.cpp
```diff
@@ -0,0 +1,202 @@
...
+#include "algorithms/od/fastod/model/removal_set.h"
+#include "algorithms/od/set_based_verifier/verifier.h"
+#include "all_csv_configs.h"
+
+namespace tests {
+
+namespace {
+
+using namespace algos::od;
+
```
ol-imorozko, line 15: add `using namespace config::names;` and do `:%s/config::names:://g`

src/tests/test_set_based_verifier.cpp
```diff
@@ -0,0 +1,202 @@
...
+                                                    /*ordering=*/Ordering::ascending,
+                                                    /*ofd_context=*/{5}, /*ofd_right=*/1,
+                                                    /*removal_set=*/{}),
+                          /* {}: Col5 -> Col2 */
+                          SetBasedAodVerifierParams(kTestFD, /*oc_context=*/{},
+                                                    /*oc_left=*/5, /*oc_right=*/2,
+                                                    /*ordering=*/Ordering::ascending,
+                                                    /*ofd_context=*/{5}, /*ofd_right=*/2,
+                                                    /*removal_set=*/{2, 5, 8}),
+                          /* {Col3}: Col4 -> Col1 */
```
ol-imorozko, line 79: Idk if this is possible, but can we add some descriptive names to individual tests in this test suite? Like make those comments `{}: Col5 -> Col1` into names? If it's hard, then feel free to ignore

src/tests/test_set_based_verifier.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+    RemovalSet actual_removal_set = verifier->GetRemovalSet();
+    RemovalSetAsVec actual_removal_set_vec(actual_removal_set.begin(), actual_removal_set.end());
+    std::sort(actual_removal_set_vec.begin(), actual_removal_set_vec.end());
+    RemovalSetAsVec expected_removal_set_vec = params.GetExpectedRemovalSet();
+    std::sort(expected_removal_set_vec.begin(), expected_removal_set_vec.end());
+    EXPECT_EQ(actual_removal_set_vec, expected_removal_set_vec);
+    SetBasedAodVerifier::Error expected_error =
+            static_cast<SetBasedAodVerifier::Error>(expected_removal_set_vec.size()) /
+            verifier->GetTupleCount();
```
ol-imorozko, line 61: since we're specifying type in `static_cast`, we can just use `auto expected_error`

src/tests/test_set_based_verifier.cpp
```diff
@@ -0,0 +1,202 @@
...
+               std::vector<algos::fastod::CanonicalOD<Ordering>> const& ocs) {
+    for (algos::fastod::CanonicalOD<Ordering> const& oc : ocs) {
+        std::vector<model::ColumnIndex> oc_context = oc.GetContext().AsVector();
+        model::ColumnIndex oc_left = oc.GetLeftColumn();
+        model::ColumnIndex oc_right = oc.GetRightColumn();
+        algos::StdParamsMap execute_params{{config::names::kOcContext, oc_context},
+                                           {config::names::kOcLeftIndex, oc_left},
+                                           {config::names::kOcRightIndex, oc_right},
+                                           {config::names::kOcLeftOrdering, Ordering}};
+        algos::ConfigureFromMap(verifier, execute_params);
```
ol-imorozko, line 141: separate with empty line

src/tests/test_set_based_verifier.cpp
```diff
@@ -0,0 +1,202 @@
...
+}
+
+void VerifyOFDs(algos::od::SetBasedAodVerifier& verifier,
+                std::vector<algos::fastod::SimpleCanonicalOD> const& ofds) {
+    for (algos::fastod::SimpleCanonicalOD const& ofd : ofds) {
+        std::vector<model::ColumnIndex> ofd_context = ofd.GetContext().AsVector();
+        model::ColumnIndex ofd_right = ofd.GetRightColumn();
+        algos::StdParamsMap execute_params{{config::names::kOFDContext, ofd_context},
+                                           {config::names::kOFDRightIndex, ofd_right}};
+        algos::ConfigureFromMap(verifier, execute_params);
```
ol-imorozko, line 163: separate with empty line

ol-imorozko (changes requested):

src/core/config/indices/option.cpp
```diff
@@ -14,8 +14,13 @@ static void NormalizeIndices(config::IndicesType& indices) {
 }
 
 IndicesOption::IndicesOption(std::string_view name, std::string_view description,
-                             typename Option<config::IndicesType>::DefaultFunc calculate_default)
-    : common_option_(name, description, std::move(calculate_default), NormalizeIndices, nullptr) {}
+                             typename Option<config::IndicesType>::DefaultFunc calculate_default,
+                             bool allow_empty)
+    : common_option_(name, description, std::move(calculate_default), NormalizeIndices, nullptr),
+      allow_empty_(allow_empty) {}
```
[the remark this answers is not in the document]
polyntsov: I am not the author of this code, but fixed

src/core/algorithms/od/fastod/model/attribute_set.h
```diff
@@ -159,11 +160,11 @@ struct boost::hash<algos::fastod::AttributeSet> {
 
 namespace algos::fastod {
 
-inline AttributeSet CreateAttributeSet(std::initializer_list<model::ColumnIndex> attributes,
+inline AttributeSet CreateAttributeSet(std::ranges::input_range auto const& attributes,
                                        model::ColumnIndex size) {
```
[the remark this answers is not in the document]
polyntsov: Span and input_range are not the same and here we actually want `input_range`

src/core/algorithms/od/fastod/partitions/complex_stripped_partition.cpp
polyntsov: That's because the exact number of elements in the `else` branch is not known beforehand, so that's why no reserve. It's possible to add it here, but since I am only moving this code, I am not going to do it

src/core/algorithms/od/fastod/partitions/complex_stripped_partition.cpp
polyntsov: Not sure I understand why we can remove this check. It is true that on first iteration it always fires and does nothing, but it might be important for the next iterations?

src/core/algorithms/od/fastod/partitions/complex_stripped_partition.cpp
polyntsov: > Though I'm not sure whether we can use constexpr values with ternary operator

We can, but it'll move computations to runtime. Also this is already refactored in later commits

src/core/algorithms/od/fastod/partitions/complex_stripped_partition.cpp
ol-imorozko: Sorry, I meant to remove only `i != 0` check.
Like go from:
```cpp
for (size_t i = 0; i < values.size(); i++) {
    auto const& [first, second] = values[i];

    if (i != 0 && values[i - 1].first != first) {
```
to
```cpp
for (size_t i = 1; i < values.size(); i++) {
    auto const& [first, second] = values[i];

    if (values[i - 1].first != first) {
```
This simplifies the logic IMHO

src/core/algorithms/algorithm.cpp
```diff
@@ -64,8 +65,15 @@ void Algorithm::MakeOptionsAvailable(std::vector<std::string_view> const& option
     }
 }
 
+bool Algorithm::AllRequiredOptionsAreSet() const noexcept {
+    std::unordered_set<std::string_view> needed = GetNeededOptions();
+    return std::none_of(needed.begin(), needed.end(), [this](std::string_view option_name) {
+        return possible_options_.at(option_name)->IsRequired();
+    });
+}
```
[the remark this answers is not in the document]
ol-imorozko: nvm. 
use ranges though :)

```suggestion
bool Algorithm::AllRequiredOptionsAreSet() const noexcept {
    std::unordered_set<std::string_view> needed = GetNeededOptions();
    return std::ranges::none_of(needed, [this](std::string_view option_name) {
        return possible_options_.at(option_name)->IsRequired();
    });
}
```

src/core/algorithms/od/fastod/model/attribute_set.cpp
```diff
@@ -30,4 +30,11 @@ void AttributeSet::Iterate(std::function<void(model::ColumnIndex)> callback) con
     }
 }
 
+std::vector<model::ColumnIndex> AttributeSet::AsVector() const {
+    std::vector<model::ColumnIndex> attrs;
+    attrs.reserve(Count());
+    Iterate([&attrs](model::ColumnIndex attr) { attrs.push_back(attr); });
+    return attrs;
+}
```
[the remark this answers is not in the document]
ol-imorozko: Ugh.. I mean we can do it, but probably we shouldn't :)
you can resolve this conversation

src/core/algorithms/od/fastod/model/canonical_od.cpp
```diff
@@ -40,31 +41,37 @@ std::string SimpleCanonicalOD::ToString() const {
...
 
-bool operator!=(CanonicalOD<false> const& x, CanonicalOD<false> const& y) {
+bool operator!=(CanonicalOD<od::Ordering::descending> const& x,
+                CanonicalOD<od::Ordering::descending> const& y) {
     return !(x == y);
 }
 
-bool operator<(CanonicalOD<false> const& x, CanonicalOD<false> const& y) {
+bool operator<(CanonicalOD<od::Ordering::descending> const& x,
+               CanonicalOD<od::Ordering::descending> const& y) {
```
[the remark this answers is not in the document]
polyntsov: Not sure if it's worth it. If I change it here, method definitions will have different parameter types comparing to the declarations

src/core/util/range_to_string.h
```diff
@@ -0,0 +1,27 @@
...
+    requires Printable<std::ranges::range_value_t<Range>>
+std::string RangeToString(Range const& range) {
+    std::stringstream sstream;
+    sstream << '[';
+    for (auto pt{range.begin()}; pt != range.end(); ++pt) {
+        if (pt != range.begin()) {
+            sstream << ", ";
+        }
+        sstream << *pt;
+    }
```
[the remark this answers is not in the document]
polyntsov: Not sure why this is better

src/core/algorithms/od/fastod/partitions/complex_stripped_partition.cpp
```diff
@@ -147,6 +147,145 @@ template bool ComplexStrippedPartition::Swap<od::Ordering::ascending>(
...
+        for (size_t i = subseq_len_to_index[longest_subseq];
+             i != std::numeric_limits<model::TupleIndex>::max(); i = predecessors[i]) {
+            longest_subseq_elements.insert(values[i].tuple_index);
+        }
+
+        for (auto const& [index, left_value, right_value] : values) {
+            if (!longest_subseq_elements.contains(index)) {
+                removal_set.push_back(index);
+            }
+        }
```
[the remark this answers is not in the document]
polyntsov: We sure can, but I am against premature optimization by replacing unordered sets or maps with vectors. It might be faster or might not be, but IMO it definitely makes the code less readable

src/core/algorithms/od/fastod/partitions/complex_stripped_partition.cpp
```diff
@@ -147,6 +147,145 @@ template bool ComplexStrippedPartition::Swap<od::Ordering::ascending>(
...
+        for (size_t i = group_begin; i < group_end; ++i) {
+            DataFrame::Range const range = (*rb_indexes_)[i];
+
+            for (size_t j = range.first; j <= range.second; ++j) {
+                values.emplace_back(data_->GetValue(j, right), j);
+            }
+        }
+        od::RemovalSetAsVec cur_removal_set = SplitRemovalSet(values);
+        removal_set.insert(removal_set.end(), cur_removal_set.begin(), cur_removal_set.end());
+    }
```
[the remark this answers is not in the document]
polyntsov: I really don't think that it makes the code any better with callback, lambda and std::function

Also these are vectors of ints, so move does nothing on their elements

src/core/algorithms/od/fastod/partitions/complex_stripped_partition.cpp
```diff
@@ -147,6 +147,145 @@ template bool ComplexStrippedPartition::Swap<od::Ordering::ascending>(
...
+        std::vector<ValueAndIndex> values;
+        for (size_t i = group_begin; i < group_end; ++i) {
+            DataFrame::Range const range = (*rb_indexes_)[i];
+
+            for (size_t j = range.first; j <= range.second; ++j) {
+                values.emplace_back(data_->GetValue(j, right), j);
+            }
+        }
+        od::RemovalSetAsVec cur_removal_set = SplitRemovalSet(values);
+        removal_set.insert(removal_set.end(), cur_removal_set.begin(), cur_removal_set.end());
```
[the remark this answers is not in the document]
polyntsov: move iterators do nothing here because these are vectors of ints

src/core/algorithms/od/fastod/partitions/complex_stripped_partition.cpp
```diff
@@ -147,6 +147,145 @@ template bool ComplexStrippedPartition::Swap<od::Ordering::ascending>(
...
+        size_t const group_end = (*rb_begins_)[begin_pointer + 1];
+
+        std::vector<ValueAndIndex> values;
+        for (size_t i = group_begin; i < group_end; ++i) {
+            DataFrame::Range const range = (*rb_indexes_)[i];
+
+            for (size_t j = range.first; j <= range.second; ++j) {
+                values.emplace_back(data_->GetValue(j, right), j);
+            }
+        }
```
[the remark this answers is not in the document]
polyntsov: I don't think it’s worth it

src/core/algorithms/od/set_based_verifier/verifier.cpp
polyntsov: Introducing a function for this seems like a bit of overengineering

src/tests/unit/test_set_based_verifier.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+    RemovalSet actual_removal_set = verifier->GetRemovalSet();
+    RemovalSetAsVec actual_removal_set_vec(actual_removal_set.begin(), actual_removal_set.end());
+    std::sort(actual_removal_set_vec.begin(), actual_removal_set_vec.end());
+    RemovalSetAsVec expected_removal_set_vec = params.GetExpectedRemovalSet();
+    std::sort(expected_removal_set_vec.begin(), expected_removal_set_vec.end());
+    EXPECT_EQ(actual_removal_set_vec, expected_removal_set_vec);
+    SetBasedAodVerifier::Error expected_error =
+            static_cast<SetBasedAodVerifier::Error>(expected_removal_set_vec.size()) /
+            verifier->GetTupleCount();
```
polyntsov: No, because `static_cast` is used only for the numerator of the expression in right hand side

src/tests/unit/test_set_based_verifier.cpp
```diff
@@ -0,0 +1,202 @@
...
+#include "algorithms/od/fastod/model/removal_set.h"
+#include "algorithms/od/set_based_verifier/verifier.h"
+#include "all_csv_configs.h"
+
+namespace tests {
+
+namespace {
+
+using namespace algos::od;
+
```
polyntsov: Not sure how useful is this. It'll make the code a little more compact, but at the cost of making constants like `kCsvConfig` more ambiguous. At current code it's obvious that ` config::names::kCsvConfig` is a name

src/tests/unit/test_set_based_verifier.cpp
```diff
@@ -0,0 +1,202 @@
...
+                                                    /*ordering=*/Ordering::ascending,
+                                                    /*ofd_context=*/{5}, /*ofd_right=*/1,
+                                                    /*removal_set=*/{}),
+                          /* {}: Col5 -> Col2 */
+                          SetBasedAodVerifierParams(kTestFD, /*oc_context=*/{},
+                                                    /*oc_left=*/5, /*oc_right=*/2,
+                                                    /*ordering=*/Ordering::ascending,
+                                                    /*ofd_context=*/{5}, /*ofd_right=*/2,
+                                                    /*removal_set=*/{2, 5, 8}),
+                          /* {Col3}: Col4 -> Col1 */
```
polyntsov: I don't think it's possible unfortunately

src/core/util/range_to_string.h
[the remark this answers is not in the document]
ol-imorozko: Agree, adding extra state `bool first` is strange.
Now that would be nice, but `join_with` is C++23 :black_bird: 
`return range | std::views::join_with(", ") | std::ranges::to<std::string>();`

src/core/algorithms/od/set_based_verifier/verifier.cpp
ol-imorozko: :black_bird:

src/core/algorithms/algorithm.cpp
```diff
@@ -152,4 +174,13 @@ std::string_view Algorithm::GetDescription(std::string_view option_name) const {
     }
 }
 
+[[nodiscard]] bool Algorithm::OptionIsSet(std::string_view option_name) const {
+    auto it = possible_options_.find(option_name);
+    if (it == possible_options_.end()) {
+        return false;
+    } else {
+        return it->second->IsSet();
+    }
```
ol-imorozko: nit:
We do
```cpp
    auto it = possible_options_.find(option_name);
    return it != possible_options_.end() && it->second->IsRequired();
```
In the `OptionIsRequired`. Might to it as well in `OptionIsSet` to stay consistent

src/core/algorithms/od/fastod/model/canonical_od.h
```diff
@@ -73,11 +95,11 @@ class SimpleCanonicalOD {
...
-        const size_t context_hash = hash<algos::fastod::AttributeSet>{}(od.context_);
-        const size_t ap_hash = hash<algos::fastod::AttributePair>{}(od.ap_);
+template <algos::od::Ordering Ordering>
+struct hash<algos::fastod::CanonicalOD<Ordering>> {
+    size_t operator()(algos::fastod::CanonicalOD<Ordering> const& od) const noexcept {
+        size_t const context_hash = hash<algos::fastod::AttributeSet>{}(od.context_);
+        size_t const ap_hash = hash<algos::fastod::AttributePair>{}(od.ap_);
 
         return algos::fastod::hashing::CombineHashes(context_hash, ap_hash);
     }
```
ol-imorozko: nit:
I understand that it's not your code, but we might fix it as well.
We should use `std::hash` rather than opening a `namespace std` since technically it's an undefined behavior. :nerd_face: 
From https://en.cppreference.com/w/cpp/language/extending_std.html:

```cpp
// Get the declaration of the primary std::hash template.
// We are not permitted to declare it ourselves.
// <typeindex> is guaranteed to provide such a declaration, 
// and is much cheaper to include than <functional>.
 
#include <typeindex> 
 
// Specialize std::hash so that MyType can be used as a key in 
// std::unordered_set and std::unordered_map.  Opening namespace
// std can accidentally introduce undefined behavior, and is not
// necessary for specializing class templates.
template<>
struct [std::hash](https://en.cppreference.com/w/cpp/utility/hash.html)<MyType>
{
    [std::size_t](https://en.cppreference.com/w/cpp/types/size_t.html) operator()(const MyType& t) const { return t.hash(); }
};
```

src/core/algorithms/od/fastod/storage/cache_with_limit.h
```diff
@@ -29,15 +29,19 @@ class CacheWithLimit {
         return entries_.at(key);
     }
 
-    void Set(const K& key, const V& value) {
-        if (!entries_.try_emplace(key, value).second) return;
+    V const& Set(K const& key, V const& value) {
+        auto [it, inserted] = entries_.try_emplace(key, value);
+        if (!inserted) {
+            return it->second;
+        }
```
ol-imorozko: Is is expected that if the key is already present, we do not update the value?
```cpp
cache.Set("a", 1);
//...
cache.Set("a", 99);  //value at "a" is 1, not 99
```

ol-imorozko: If not, maybe it's better to call it smth like `GetOrInsert`

src/core/algorithms/od/fastod/model/canonical_od.h
polyntsov: The reopening itself is not ub and the code here is perfectly fine. 

Although specializing `std::hash` instead of reopening namespace std is a cleaner way to do this of course

PR #471, opened 2024-10-06
Optimize type deduction

MichaelS239:
- Add quick regex pre-checks before time-consuming checks for double and date types
- Use boost::regex instead of std::regex


src/core/model/table/typed_column_data.h
```diff
@@ -204,32 +209,38 @@ class TypedColumnDataFactory {
...
+                try {
+                    std::size_t pos = 0;
+                    std::stod(val, &pos);
+                    if (pos == val.size()) {
+                        is_double = true;
+                    }
+                } catch (...) {
+                }
+                return is_double;
+            };
```
ol-imorozko, line 224: I understand that this is not your code, but since you're optimizing type deduction, we can make it better:
- try block with a catch-all and not handling the exception is generally not a good practice, looks very strange
 
 We can make it better by utilizing `std::from_chars`, which is a modern way of converting strings to numbers, plus it's exception-free.
 Smth like
 ```cpp
 [](std::string const& val) {
        double result;
        auto [p, ec] = std::from_chars(val.data(), val.data() + val.size(), result);
        return ec == std::errc{} && p == val.data() + val.size();
    };
```

**Review by ol-imorozko — CHANGES_REQUESTED** [review]

File: src/core/model/table/typed_column_data.h [hunk 2]
```diff
@@ -173,16 +174,20 @@ class TypedColumnDataFactory {
...
             +TypeId::kDate, +TypeId::kInt, +TypeId::kBigInt, +TypeId::kDouble, +TypeId::kString};
-    inline static std::unordered_map<TypeId, std::regex> const kTypeIdToRegex = {
-            {+TypeId::kBigInt, std::regex(R"(^(\+|-)?\d{20,}$)")},
-            {+TypeId::kInt, std::regex(R"(^(\+|-)?\d{1,19}$)")},
-            {+TypeId::kNull, std::regex(Null::kValue.data())},
-            {+TypeId::kEmpty, std::regex(R"(^$)")}};
+    inline static std::unordered_map<TypeId, boost::regex> const kTypeIdToRegex = {
+            {TypeId::kDate,
+             boost::regex(
+                     R"(^([0-9]{4})[-.\/]?(1[0-2]|0[1-9]|[1-9])[-.\/]?(3[0-1]|0[1-9]|[1-9]|[1-2][0-9])$)")},
```
ol-imorozko, line 180: IDK if this is intentional, but this regex:
- matches strings with different delimeters like `2024/12-01`
- does not match USA date format, that is, mm-dd-yyyy
check it yourself: https://regex101.com/r/MY1jCD/1

We can use `(\d{n})` to match a n digits and with `([-./])` we can capture a delimeter in a group and use it with `\group_number`.

I suggest using `^(?:(?!0000)(\d{4})([-.\/])(0[1-9]|1[0-2])\2(0[1-9]|[1-2][0-9]|3[0-1])|(0[1-9]|1[0-2])([-.\/])(0[1-9]|[1-2][0-9]|3[0-1])\6(?!0000)(\d{4}))$`
https://regex101.com/r/sOhFqg/1

src/core/model/table/typed_column_data.h
```diff
@@ -173,16 +174,20 @@ class TypedColumnDataFactory {
...
+            {TypeId::kDouble, boost::regex(R"(^([0-9]|\+|-|\.|e|E|i|I|n|N|f|F|a|A)+$)")},
```
ol-imorozko, line 181: Use `^[+-]?(\d+(\.\d*)?|\.\d+)([eE][+-]?\d+)?$|^[+-]?(inf|nan)$` cause this matches too many invalid numbers:
https://regex101.com/r/vV26nJ/1

src/core/model/table/typed_column_data.h
MichaelS239: Our date type is based on [`boost::gregorian::date`](https://www.boost.org/doc/libs/1_86_0/doc/html/date_time/gregorian.html). So USA date format is not supported.
Also, the only proper way to check whether a string is a correct date is trying to create `boost::gregorian::date` object (which is done in `kDelimitedDateCheck` and `kUndelimitedDateCheck`). The regex is needed to quickly eliminate strings that are not dates and to pass anything that looks like `boost::gregorian::date`.

Should we use `^(\d{4})([-.\/]?)(1[0-2]|0[1-9]|[1-9])\2(3[0-1]|0[1-9]|[1-9]|[1-2][0-9])$`?
https://regex101.com/r/rF63Ft/1
The only difference from the original regex is that it checks for the delimiters.

src/core/model/table/typed_column_data.h
MichaelS239: This regex doesn't support `INF`, `NAN` and hexadecimal floating-point values, which are supported by [`std::strtod`](https://en.cppreference.com/w/cpp/string/byte/strtof).

Should we use `^[+-]?(\d+(\.\d*)?|\.\d+)([eE][+-]?\d+)?$|^[+-]?(?i)(inf|nan)(?-i)$|^[+-]?0[xX](((\d|[a-f]|[A-F]))+(\.(\d|[a-f]|[A-F])*)?|\.(\d|[a-f]|[A-F])+)([pP][+-]?\d+)?$`?
https://regex101.com/r/k4ZhhJ/2
I'm not sure if it's equivalent to `std::strtod`, but if it is, then `kDoubleCheck` is not needed anymore.

src/core/model/table/typed_column_data.h
ol-imorozko: Understood. Yes, we should, thanks

src/core/model/table/typed_column_data.h
ol-imorozko: yes, we should, thanks

PR #472, opened 2024-10-07
Fix macOS building, CMake policies and finding Boost behaviour

Vdaleke:
Fix compilation issue related to new MacOSX15.0.sdk ([issue description](https://gist.github.com/scivision/d69faebbc56da9714798087b56de925a)). 

Set CMake [CMP0144](https://cmake.org/cmake/help/latest/policy/CMP0144.html) policy as new to match the macOS installation guide. 

Change CMake behaviour of finding Boost. Now it's trying to find BoostConfig.cmake which provided in Boost since 1.70 version. This avoids the warning about [CMP0167](https://cmake.org/cmake/help/latest/policy/CMP0167.html) policy and define behaviour of searching in all available CMake versions. 

Change boost toolset to Darwin which is more suitable for macOS.

Fix Boost compatibility issues with building under different compilers. Now we use `--layout=versioned` flag in`b2 install` option which allows to use different boost compilers changing the default boost library names. 

Set g{cc/++}-14 compiler as default. Add step with Xcode Command Line Tools installation that are needed in boost building and in CMake environment setting.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,5 @@
-cmake_minimum_required(VERSION 3.10)
+cmake_minimum_required(VERSION 3.13)
+cmake_policy(SET CMP0144 NEW)
 project(Desbordante)
 
 option(COPY_PYTHON_EXAMPLES "Copy Python examples" OFF)
@@ -102,7 +103,7 @@ endif()
 
 # configuring boost
 set(Boost_USE_STATIC_LIBS OFF)
-find_package(Boost 1.81.0 REQUIRED COMPONENTS container thread graph)
+find_package(Boost 1.81.0 REQUIRED COMPONENTS container thread graph CONFIG)
 include_directories(${Boost_INCLUDE_DIRS})
 message(${Boost_INCLUDE_DIRS})
 
--- a/README.md
+++ b/README.md
@@ -219,8 +219,8 @@ However, as Desbordante core uses C++, additional requirements on the machine ar
 
 ## Build instructions
 
-### Ubuntu and MacOS
-The following instructions were tested on Ubuntu 20.04+ LTS and MacOS 14.0+ (Apple Silicon).
+### Ubuntu and macOS
+The following instructions were tested on Ubuntu 20.04+ LTS and macOS Sonoma 14.7 (Apple Silicon).
 ### Dependencies
 Prior to cloning the repository and attempting to build the project, ensure that you have the following software:
 
@@ -244,39 +244,49 @@ You can also add them to the end of `~/.profile` to set this by default in all s
 
 #### MacOS dependencies installation
 
-To install GCC and CMake on MacOS we recommend to use [Homebrew](https://brew.sh/) package manager. With Homebrew
+Install Xcode Command Line Tools if you don't have them. Run:
+```sh
+xcode-select --install
+```
+Follow the prompts to continue.
+
+To install GCC and CMake on macOS we recommend to use [Homebrew](https://brew.sh/) package manager. With Homebrew
 installed, run the following commands:
 ```sh
-brew install gcc cmake 
+brew install gcc@14 cmake
 ```
 After installation, check `cmake --version`. If command is not found, then you need to add to environment path to
 homebrew installed packages. To do this open `~/.zprofile` (for Zsh) or
 `~/.bash_profile` (for Bash) and add to the end of the file the output of `brew shellenv`.
 After that, restart the terminal and check the version of CMake again, now it should be displayed.
 
-Then, check the installed version of GCC:`brew info gcc`. You must see something like `==> gcc: stable X.Y.Z ...`. 
-Check that `gcc-X` and `g++-X` commands work, where X is the version from this output 
-(this designation continues to be used further).
-
-After you need to install Boost library. Please, don't use Homebrew for this, as it may not work correctly.
-Instead, download the latest version of Boost from the [official website](https://www.boost.org/users/download/).
-After downloading, unpack the archive to the `/usr/local` directory or another directory of your
-choice.
-
-Go to unpacked boost directory in the terminal and run the following commands:
+Then you need to install Boost library built with GCC. Please avoid using Homebrew for this, as the Boost version provided by Homebrew
+is built with Clang, which has a different ABI. Instead, download the latest version of Boost from the [official website](https://www.boost.org/users/download/) and unpack the archive to the `/usr/local/` directory or another directory of your choice:
+```sh
+cd /usr/local/
+curl https://archives.boost.io/release/1.86.0/source/boost_1_86_0.tar.bz2 --output "boost_1_86_0.tar.bz2"
+tar xvjf boost_1_86_0.tar.bz2
+rm boost_1_86_0.tar.bz2
+cd boost_1_86_0
+```
+Navigate to the unpacked Boost directory in the terminal and run the following commands:
 ```sh
 ./bootstrap.sh 
-echo "using gcc : : g++-X ;" > user-config.jam
-./b2 --user-config=user-config.jam
-export BOOST_ROOT=$(pwd)
+echo "using darwin : : g++-14 ;" > user-config.jam
+./b2 install --user-config=user-config.jam --layout=versioned
+export BOOST_ROOT=$(pwd) # export Boost_ROOT=$(pwd) for CMake 3.26 and below.
 ``` 
-You can also add last export with current path to `~/.zprofile` or `~/.bash_profile` to set this boost path by default.
+You can also add the last export with current path to `~/.zprofile` or `~/.bash_profile` to set this boost path by default.
 
-Before building project you must set GCC as default compiler by changing the following environment variables:
+Before building the project you must set locally or in the above-mentioned dotfiles the following CMake environment variables:
 ```sh
-export CC=gcc-X
-export CXX=g++-X
+export CC=gcc-14
+export CXX=g++-14
+export SDKROOT=/Library/Developer/CommandLineTools/SDKs/MacOSX14.sdk/
 ```
+The first two lines set GCC as the default compiler in CMake. The last export is also necessary due to issues with GCC 14 and
+the last MacOSX15.0.sdk used by CMake by default, you can read more about this [here](https://gist.github.com/scivision/d69faebbc56da9714798087b56de925a)
+and [here](https://github.com/iains/gcc-14-branch/issues/11).
 
 ### Building the project
 #### Building the Python module using pip


README.md
```diff
@@ -244,39 +244,44 @@ You can also add them to the end of `~/.profile` to set this by default in all s
 
 #### MacOS dependencies installation
 
-To install GCC and CMake on MacOS we recommend to use [Homebrew](https://brew.sh/) package manager. With Homebrew
+First of all, install Xcode Command Line Tools if you don't have them. Open terminal and run:
+```sh
+xcode-select --install
+```
+You'll see a panel that asks you to install Xcode Command Line Tools. Click 'Install' to begin the installation process and wait for it to complete.
```
ol-imorozko, line 251: IMHO this feels like spoon-feeding.
I feel like
```md
#### Install Xcode Command Line Tools

Open Terminal and run:

```sh
xcode-select --install

Follow the prompts to install.
```

is sufficient

README.md
```diff
@@ -244,39 +244,44 @@ You can also add them to the end of `~/.profile` to set this by default in all s
...
 `~/.bash_profile` (for Bash) and add to the end of the file the output of `brew shellenv`.
 After that, restart the terminal and check the version of CMake again, now it should be displayed.
 
-Then, check the installed version of GCC:`brew info gcc`. You must see something like `==> gcc: stable X.Y.Z ...`. 
-Check that `gcc-X` and `g++-X` commands work, where X is the version from this output 
-(this designation continues to be used further).
-
-After you need to install Boost library. Please, don't use Homebrew for this, as it may not work correctly.
+Then you need to install Boost library built with GCC. Please, don't use Homebrew for this, because brew boost is built with Clang.
 Instead, download the latest version of Boost from the [official website](https://www.boost.org/users/download/).
```
ol-imorozko, line 264: And why is it bad that brew boost is built with clang? This needs some explanation
I.e 

> Please avoid using Homebrew for this, as the Boost version provided by Homebrew is built with Clang, which..

README.md
```diff
@@ -244,39 +244,44 @@ You can also add them to the end of `~/.profile` to set this by default in all s
...
-echo "using gcc : : g++-X ;" > user-config.jam
+echo "using gcc : : g++-14 ;" > user-config.jam
 ./b2 --user-config=user-config.jam
 export BOOST_ROOT=$(pwd)
 ``` 
 You can also add last export with current path to `~/.zprofile` or `~/.bash_profile` to set this boost path by default.
 
-Before building project you must set GCC as default compiler by changing the following environment variables:
+Before building project you must set localy or in the above mentioned files the following CMake environment variables:
 ```sh
```
ol-imorozko, line 277: above-mentioned

ol-imorozko: before building _the_ project

README.md
```diff
@@ -244,39 +244,44 @@ You can also add them to the end of `~/.profile` to set this by default in all s
...
-export CC=gcc-X
-export CXX=g++-X
+export CC=gcc-14
+export CXX=g++-14
+export SDKROOT=/Library/Developer/CommandLineTools/SDKs/MacOSX14.sdk/   
 ```
+The first two lines set GCC as default compiler in CMake. The last export is needed also because of issues with GCC 14 and 
+the last MacOSX15.0.sdk used by CMake by default, you can read about this [here](https://gist.github.com/scivision/d69faebbc56da9714798087b56de925a) 
+and [here](https://github.com/iains/gcc-14-branch/issues/11).
```
ol-imorozko, line 284: The last export is also necessary due to issues
you can read _more_ about this

ol-imorozko (approved):
I can't verify if this is correct since I don't own a Mac, but the Readme looks good to me.

README.md
```diff
@@ -244,39 +244,44 @@ You can also add them to the end of `~/.profile` to set this by default in all s
 
 #### MacOS dependencies installation
 
-To install GCC and CMake on MacOS we recommend to use [Homebrew](https://brew.sh/) package manager. With Homebrew
+Install Xcode Command Line Tools if you don't have them. Run:
+```sh
+xcode-select --install
+```
+Follow the prompts to contunue.
```
vs9h, line 251:
```suggestion
Follow the prompts to continue.
```

README.md
```diff
@@ -244,39 +244,44 @@ You can also add them to the end of `~/.profile` to set this by default in all s
...
-
-After you need to install Boost library. Please, don't use Homebrew for this, as it may not work correctly.
-Instead, download the latest version of Boost from the [official website](https://www.boost.org/users/download/).
-After downloading, unpack the archive to the `/usr/local` directory or another directory of your
-choice.
+Then you need to install Boost library built with GCC. Please avoid using Homebrew for this, as the Boost version provided by Homebrew 
+is built with Clang, which has a different ABI. Instead, download the latest version of Boost from the [official website](https://www.boost.org/users/download/).
+After downloading, unpack the archive to the `/usr/local` directory or another directory of your choice.
 
 Go to unpacked boost directory in the terminal and run the following commands:
```
vs9h, line 267: In this phrase the wording could be slightly more formal:
```suggestion
Navigate to the unpacked Boost directory in the terminal and run the following commands:
```
Also we should use `the` here and use `Boost` instead of `boost` for consistency

README.md
```diff
@@ -244,39 +244,44 @@ You can also add them to the end of `~/.profile` to set this by default in all s
...
 ```sh
 ./bootstrap.sh 
-echo "using gcc : : g++-X ;" > user-config.jam
+echo "using gcc : : g++-14 ;" > user-config.jam
 ./b2 --user-config=user-config.jam
 export BOOST_ROOT=$(pwd)
 ``` 
 You can also add last export with current path to `~/.zprofile` or `~/.bash_profile` to set this boost path by default.
```
vs9h, line 274: here we can also change `boost` to `Boost` for consistency

README.md
```diff
@@ -244,39 +244,44 @@ You can also add them to the end of `~/.profile` to set this by default in all s
...
+```sh
+xcode-select --install
+```
+Follow the prompts to contunue.
+
+To install GCC and CMake on macOS we recommend to use [Homebrew](https://brew.sh/) package manager. With Homebrew
 installed, run the following commands:
 ```sh
-brew install gcc cmake 
+brew install gcc@14 cmake
```
vs9h, line 256: remove trailing spaces (they are also on the lines 263, 280, 282, 283)

README.md
```diff
@@ -244,39 +244,44 @@ You can also add them to the end of `~/.profile` to set this by default in all s
...
-Check that `gcc-X` and `g++-X` commands work, where X is the version from this output 
-(this designation continues to be used further).
-
-After you need to install Boost library. Please, don't use Homebrew for this, as it may not work correctly.
-Instead, download the latest version of Boost from the [official website](https://www.boost.org/users/download/).
-After downloading, unpack the archive to the `/usr/local` directory or another directory of your
-choice.
+Then you need to install Boost library built with GCC. Please avoid using Homebrew for this, as the Boost version provided by Homebrew 
+is built with Clang, which has a different ABI. Instead, download the latest version of Boost from the [official website](https://www.boost.org/users/download/).
+After downloading, unpack the archive to the `/usr/local` directory or another directory of your choice.
```
vs9h, line 265: Since we are providing a very detailed guide, then perhaps we can also provide commands for downloading the boost (for example, use the current latest version of boost and the path `/usr/local` in the commands).

I would say that I would be inclined to do so, but I would not insist.

vs9h (changes requested):
Can you also fix you commit message? The rules for [formatting commits](https://github.com/Desbordante/desbordante-core/wiki/Development#%D0%BE%D1%84%D0%BE%D1%80%D0%BC%D0%BB%D0%B5%D0%BD%D0%B8%D0%B5-%D0%BA%D0%BE%D0%BC%D0%BC%D0%B8%D1%82%D0%BE%D0%B2) state that the body of a commit should not contain lines longer than 72 characters.

Like the previous reviewer, I can't verify that it works because I don't use a macOS.

README.md
vs9h: Seems, like this line wasn't updated

Vdaleke: Fixed

vs9h (approved):
I thought you'd just put line breaks in the commit message. But that's fine too.

And for the future, it's better to use the imperative mood. (not Simplified, but Simplify)

Looks ok

Vdaleke:
We found a bug related to new CMake policy https://cmake.org/cmake/help/latest/policy/CMP0167.html. Needs some improvement in CMakeLists.txt or somewhere.

Vdaleke:
I updated PR description to explain new CMake and guide changes.

README.md
```diff
@@ -244,39 +244,50 @@ You can also add them to the end of `~/.profile` to set this by default in all s
...
-echo "using gcc : : g++-X ;" > user-config.jam
-./b2 --user-config=user-config.jam
-export BOOST_ROOT=$(pwd)
+echo "using gcc : : g++-14 ;" > user-config.jam
+./b2 install --user-config=user-config.jam --layout=versioned
+export BOOST_ROOT=$(pwd) # export Boost_ROOT=$(pwd) for CMake 3.26 and below.
+export BOOST_COMPILER=gcc14
 ``` 
-You can also add last export with current path to `~/.zprofile` or `~/.bash_profile` to set this boost path by default.
+You can also add last 2 export with fixed Boost root path to `~/.zprofile` or `~/.bash_profile` to set them by default.
```
vs9h, line 280:
```suggestion
You can also add the last two exports with the fixed Boost root path to `~/.zprofile` or `~/.bash_profile` to set them by default.
```

README.md
```diff
@@ -244,39 +244,50 @@ You can also add them to the end of `~/.profile` to set this by default in all s
...
-Before building project you must set GCC as default compiler by changing the following environment variables:
+Before building the project you must set locally or in the above-mentioned dotfiles the following CMake environment variables:
 ```sh
-export CC=gcc-X
-export CXX=g++-X
+export CC=gcc-14
+export CXX=g++-14
+export SDKROOT=/Library/Developer/CommandLineTools/SDKs/MacOSX14.sdk/
 ```
+The first two lines set GCC as default compiler in CMake. The last export is also necessary due to issues with GCC 14 and
```
vs9h, line 288:
```suggestion
The first two lines set GCC as the default compiler in CMake. The last export is also necessary due to issues with GCC 14 and
```

vs9h (changes requested):
And split the commit into several. In one of the comments I wrote about the bump version of the cmake. But this generally applies to other changes in the cmake file. 

At the same time, you have a good description inside the pull request, it would be nice to reuse it to describe the commit. This will make the history more useful in the future for other developers. It is much less convenient to search for the pull request to which a specific commit belongs.

CMakeLists.txt
```diff
@@ -101,8 +102,12 @@ else()
 endif()
 
 # configuring boost
+if (DEFINED ENV{BOOST_COMPILER})
+    set(Boost_COMPILER $ENV{BOOST_COMPILER})
+endif()
```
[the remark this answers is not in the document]
Vdaleke: We build project and boost with the same gcc14 compiler. But CMake detect Desbordante compiler as xgcc14 when finding boost  for some reasons: 
```
➜  desbordante-core git:(fix/macos-build-with-sdk15) ✗ ./build.sh 
-- The C compiler identification is GNU 14.2.0
-- The CXX compiler identification is GNU 14.2.0
```
Logs are correct, but after this hapenned: 
```
CMake Error at /usr/local/lib/cmake/Boost-1.86.0/BoostConfig.cmake:141 (find_package):
  Found package configuration file:

    /usr/local/lib/cmake/boost_container-1.86.0/boost_container-config.cmake

  but it set boost_container_FOUND to FALSE so package "boost_container" is
  considered to be NOT FOUND.  Reason given by package:

  No suitable build variant has been found.
  
  The following variants have been tried and rejected:
    * libboost_container-gcc14-mt-a64-1_86.dylib (gcc14, detected xgcc14, set
  Boost_COMPILER to override)

  * libboost_container-gcc14-mt-a64-1_86.a (gcc14, detected xgcc14, set
  Boost_COMPILER to override)

Call Stack (most recent call first):
  /usr/local/lib/cmake/Boost-1.86.0/BoostConfig.cmake:262 (boost_find_component)
  CMakeLists.txt:110 (find_package)
```
I think this is a bug and it needs to be reported to CMake, which I will probably do later.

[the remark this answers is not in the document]
vs9h: I want to clarify my question a little. I see that when [searching for a compiler](https://github.com/Kitware/CMake/blob/8bb0281896847cda8b5bf0006691dce6efe1d557/Modules/FindBoost.cmake#L956) boost uses a compiler named `xgcc`. This decision was probably not made without reason and has some motivation. And most likely this is not a bug, but was done intentionally.

I don't know what the difference is between `gcc` and `xgcc`, why this name is used in the case of boost.

And my question is rather why we, for example, do not use `xgcc` to build desbordante in the case of macOS? Then there will be no need to set the `Boost_COMPILER` variable. Why is your solution more correct?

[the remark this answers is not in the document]
Vdaleke: Actually, your link is to CMake FindBoost module and we don't use it. With new policy CMake is trying to find BoostConfig.cmake which provided in Boost. But it doesn't matter, because in the [config](https://github.com/boostorg/boost_install/blob/d77d0e94ba3d78f2079313ecb6e4907a7e8618e9/BoostDetectToolset.cmake#L88) the default gcc toolset name for macos is xgcc.

Vdaleke:
So, I researched some boost sources and find the solution to use xgcc name. In Boost CMake [config](https://github.com/boostorg/boost_install/blob/d77d0e94ba3d78f2079313ecb6e4907a7e8618e9/BoostDetectToolset.cmake#L88) the default gcc toolset name for macOS is xgcc. And I found a reason in lib [tag building code](https://www.boost.org/doc/libs/1_66_0/tools/build/src/tools/common.jam), where xgcc name is default for gcc in [Darwin toolset](https://www.boost.org/doc/libs/1_31_0/tools/build/v1/darwin-tools.html) which, as they wright, is more compatible for macOS.

And now we use Darwin toolset by default and there is no necessity to redefine boost compiler.

vs9h (approved):
I have no more comments, it looks fine.

And I would like to highlight the commit descriptions separately, thanks for the detailed description.

PR #483, opened 2024-10-16
Fix CORDS tests

egshnov:
Added option fixed_sample that indicates that cords uses sample consisting of sample_size first rows of the given table, instead of random generated one, false by default. Add all related implementation fixes.

### DIFF
--- a/src/core/algorithms/fd/sfd/cords.cpp
+++ b/src/core/algorithms/fd/sfd/cords.cpp
@@ -67,6 +67,8 @@ void Cords::RegisterOptions() {
     RegisterOption(
             Option{&max_amount_of_categories_, kMaxAmountOfCategories, kDMaxAmountOfCategories}
                     .SetValueCheck(check_positive));
+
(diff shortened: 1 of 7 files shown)


src/core/algorithms/fd/sfd/cords.cpp
```diff
@@ -67,6 +67,8 @@ void Cords::RegisterOptions() {
     RegisterOption(
             Option{&max_amount_of_categories_, kMaxAmountOfCategories, kDMaxAmountOfCategories}
                     .SetValueCheck(check_positive));
+
+    RegisterOption(Option{&fixed_sample_, kFixedSample, kDFixedSample});
```
BUYT-1, line 71: Provide the default value to the option
```suggestion
    RegisterOption(Option{&fixed_sample_, kFixedSample, kDFixedSample, false});
```

BUYT-1 (changes requested):

BUYT-1 (approved):

PR #490, opened 2024-11-04
Add DES

VanyaVolgushev:
Add differential evolution solver algorithm for mining numerical association rules.


src/core/algorithms/nar/value_range.h
```diff
@@ -0,0 +1,87 @@
...
+};
+
+class StringValueRange : public ValueRange {
+public:
+    std::vector<String> domain;
+    explicit StringValueRange(TypedColumnData const& column);
+
+    StringValueRange(String value) : domain() {
+        domain.emplace_back(value);
+    };
```
alexandrsmirn, line 28:
```c++
StringValueRange(String value)
: domain{std::move(value)} {};
```

src/core/algorithms/nar/value_range.h
```diff
@@ -0,0 +1,87 @@
...
+
+    TypeId GetTypeId() const override {
+        return TypeId::kString;
+    }
+
+    bool Includes(std::byte const* value) const override {
+        String svalue = Type::GetValue<String>(value);
```
alexandrsmirn, line 35: `String const&`

alexandrsmirn (changes requested):
The first part of the review. I will check the rest of the code later.

src/core/algorithms/nar/des/differential_functions.cpp
```diff
@@ -0,0 +1,58 @@
...
+std::unordered_set<size_t> GetRandIndices(size_t except_index, size_t population,
+                                          size_t number_of_indices) {
+    std::unordered_set<size_t> indices;
+    indices.insert(except_index);
+    while (indices.size() < number_of_indices + 1) {
+        size_t random_index = (size_t)(RNG().Next() * population);
+        indices.insert(random_index);
+    }
+    indices.erase(except_index);
+    return indices;
```
ol-imorozko, line 17: Don't understand the idea here. We want to generate unique indices (number_of_indices amount) ranging from 0 to  `population_size - 1`?
If so, it's a good idea to add an assert: `assert(number_of_indices <= population_size - 1)`

And, well, the current solution seems strange.
Maybe just:
- use `std::vector` of `population_size`
- fill it with values from 0 to `population_size - 1` using `std::iota`
- remove `except_index` with `indices.erase(indices.begin() + except_index);`
- shuffle indices with `std::shuffle`, using a random device from some class, or just create it here
- resize to `number_of_indices` with `indices.resize(number_of_indices);`.
- return `indices`
?

src/core/algorithms/nar/des/differential_functions.cpp
```diff
@@ -0,0 +1,58 @@
...
+            double new_feature_val =
+                    sample1[i] + options.differential_scale * (sample2[i] - sample3[i]);
+            new_feature_val = std::clamp(new_feature_val, 0.0, 1.0);
+            new_individual[i] = new_feature_val;
+        }
+    }
+    return new_individual;
+}
+
+typedef EncodedNAR (*MutationFunction)(std::vector<EncodedNAR> population, size_t candidate_index,
```
ol-imorozko, line 44: don't use `typedef`, `using` is more modern way:
```suggestion
using MutationFunction = EncodedNAR(*)(const std::vector<EncodedNAR>&, size_t, const DifferentialOptions&);
```

src/core/algorithms/nar/des/differential_functions.cpp
```diff
@@ -0,0 +1,58 @@
...
+        size_t random_index = (size_t)(RNG().Next() * population);
+        indices.insert(random_index);
+    }
+    indices.erase(except_index);
+    return indices;
+}
+
+EncodedNAR Rand1Bin(std::vector<EncodedNAR> population, size_t candidate_index,
+                    DifferentialOptions options) {
+    auto candidate = population[candidate_index];
```
ol-imorozko, line 22: `candidate` is unused

src/core/algorithms/nar/des/differential_functions.cpp
```diff
@@ -0,0 +1,58 @@
...
+    auto sample_indices = GetRandIndices(candidate_index, population.size(), 3);
+    size_t sample_index1 = *std::next(sample_indices.begin(), 0);
+    size_t sample_index2 = *std::next(sample_indices.begin(), 1);
+    size_t sample_index3 = *std::next(sample_indices.begin(), 2);
```
ol-imorozko, line 26: since `sample_indices` will be vector, just use `[]` there

src/core/algorithms/nar/des/differential_functions.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+typedef EncodedNAR (*MutationFunction)(std::vector<EncodedNAR> population, size_t candidate_index,
+                                       DifferentialOptions options);
+
+// TODO: name is probably inconsistent with how it's called in the field.
+MutationFunction EnumToMutationStrategy(DifferentialStrategy strategy) {
+    switch (strategy) {
+        case DifferentialStrategy::rand1Bin:
+            return Rand1Bin;
+            break;
```
ol-imorozko, line 52: no need to `break` here, it's unreachable code

src/core/algorithms/nar/des/differential_functions.h
```diff
@@ -0,0 +1,25 @@
...
+struct DifferentialOptions {
+    double differential_scale;
+    double crossover_probability;
+    DifferentialStrategy differential_strategy = DifferentialStrategy::best1Exp;
+};
+
+EncodedNAR Rand1Bin(std::vector<EncodedNAR> population, size_t candidate_index,
+                    DifferentialOptions options);
+
+typedef EncodedNAR (*MutationFunction)(std::vector<EncodedNAR> population, size_t candidate_index,
```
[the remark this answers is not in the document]
ol-imorozko: we already have that in differential_functions.cpp. So remove that from there and here in .h file use `using`

src/core/algorithms/nar/des/encoded_nar.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+using model::NAR;
+
+size_t EncodedNAR::VectorSize() {
+    return encoded_value_ranges_.size() * EncodedValueRange().kFieldCount + 1;
+}
+
+size_t EncodedNAR::FeatureCount() {
+    return encoded_value_ranges_.size();
+}
```
ol-imorozko, line 16: those methods should be marked `const`

src/core/algorithms/nar/des/encoded_nar.h
```diff
@@ -0,0 +1,34 @@
...
+
+namespace algos::des {
+
+class EncodedNAR {
+    using TypedRelation = model::ColumnLayoutTypedRelationData;
+    using FeatureDomains = std::vector<std::shared_ptr<model::ValueRange>> const;
+
+public:
+    size_t VectorSize();
+    size_t FeatureCount();
```
ol-imorozko, line 15: those methods should be marked `const`

src/core/algorithms/nar/des/encoded_nar.cpp
```diff
@@ -0,0 +1,102 @@
...
+double const& EncodedNAR::operator[](size_t index) const {
+    if (index == 0) {
+        return implication_sign_pos_;
+    } else {
+        index--;
+        size_t feature = index / EncodedValueRange().kFieldCount;
+        size_t feature_field = index % EncodedValueRange().kFieldCount;
+        return encoded_value_ranges_[feature][feature_field];
+    }
+}
```
ol-imorozko, line 40: code duplication, extract insides to a private method, smth like:
```cpp
double& EncodedNAR::operator[](size_t index) {
    qualities_consistent_ = false;
    return GetElementAtIndex(index);
}

double const& EncodedNAR::operator[](size_t index) const {
    return GetElementAtIndex(index);
}
```

src/core/algorithms/nar/des/encoded_nar.cpp
```diff
@@ -0,0 +1,102 @@
...
+double& EncodedNAR::operator[](size_t index) {
+    qualities_consistent_ = false;
+    if (index == 0) {
+        return implication_sign_pos_;
+    } else {
+        index--;
+        size_t feature = index / EncodedValueRange().kFieldCount;
+        size_t feature_field = index % EncodedValueRange().kFieldCount;
+        return encoded_value_ranges_[feature][feature_field];
+    }
```
ol-imorozko, line 28: no need to have a branch here since we `return` in if. Just
```cpp
    if (index == 0) {
        return implication_sign_pos_;
    }
    index--;
    size_t feature = index / EncodedValueRange::kFieldCount;
    size_t feature_field = index % EncodedValueRange::kFieldCount;
    return encoded_value_ranges_[feature][feature_field];
```

src/core/algorithms/nar/des/encoded_value_range.h
```diff
@@ -0,0 +1,112 @@
...
+
+#include "algorithms/nar/value_range.h"
+#include "model/table/column_layout_typed_relation_data.h"
+#include "rng.h"
+
+namespace algos::des {
+
+class EncodedValueRange {
+public:
+    static size_t const kFieldCount = 4;
```
ol-imorozko, line 11: `constexpr` instead of `const`

src/core/algorithms/nar/des/encoded_nar.cpp
ol-imorozko: yea, use `::` to access static member variables

src/core/algorithms/nar/des/encoded_nar.cpp
```diff
@@ -0,0 +1,102 @@
...
+}
+
+NAR EncodedNAR::SetQualities(FeatureDomains domains, TypedRelation const* typed_relation) {
```
ol-imorozko, line 42: `FeatureDomains` by reference?

src/core/algorithms/nar/des/encoded_nar.cpp
```diff
@@ -0,0 +1,102 @@
...
+        throw std::logic_error("Getting uninitialized qualities from NAR.");
+    }
+    return qualities_;
+}
+
+NAR EncodedNAR::Decode(FeatureDomains domains) const {
+    auto resulting_nar = model::NAR();
+
+    std::vector<size_t> feature_order(encoded_value_ranges_.size());
+    std::iota(std::begin(feature_order), std::end(feature_order), 0);
```
ol-imorozko, line 61:
```suggestion
    std::iota(feature_order.begin(), feature_order.end(), 0);
```

src/core/algorithms/nar/des/encoded_nar.cpp
```diff
@@ -0,0 +1,102 @@
...
+    auto compare_by_permutation = [e = encoded_value_ranges_](size_t& a, size_t& b) -> bool {
+        return e[a].permutation > e[b].permutation;
+    };
+    std::sort(feature_order.begin(), feature_order.end(), compare_by_permutation);
+
+    uint implication_sign_after = implication_sign_pos_ * (encoded_value_ranges_.size() - 1);
+    uint handling_feat_num = -1;
```
ol-imorozko, line 68: `uint`? Use `size_t`

src/core/algorithms/nar/des/encoded_nar.cpp
```diff
@@ -0,0 +1,102 @@
...
+    }
+    return qualities_;
+}
+
+NAR EncodedNAR::Decode(FeatureDomains domains) const {
+    auto resulting_nar = model::NAR();
+
+    std::vector<size_t> feature_order(encoded_value_ranges_.size());
+    std::iota(std::begin(feature_order), std::end(feature_order), 0);
+    auto compare_by_permutation = [e = encoded_value_ranges_](size_t& a, size_t& b) -> bool {
```
ol-imorozko, line 62: No reason to pass `a` and `b` by reference, you are not changing them. Just `size_t a` and `size_t b`

ol-imorozko: Also no need to do `-> bool`, the type can be deduced fairly easily

src/core/algorithms/nar/des/encoded_nar.cpp
ol-imorozko: size of `encoded_value_ranges_` is not zero, right? Otherwise that's an integer overflow

src/core/algorithms/nar/des/encoded_nar.cpp
```diff
@@ -0,0 +1,102 @@
...
+    std::sort(feature_order.begin(), feature_order.end(), compare_by_permutation);
+
+    uint implication_sign_after = implication_sign_pos_ * (encoded_value_ranges_.size() - 1);
+    uint handling_feat_num = -1;
+    for (size_t feature_index : feature_order) {
+        handling_feat_num++;
+        EncodedValueRange encoded_feature = encoded_value_ranges_[feature_index];
+        if (encoded_feature.threshold < RNG().Next()) {
+            continue;
+        }
```
ol-imorozko, line 74: again, either add a random device in `EncodedNAR` structure as a field, or create it here

src/core/algorithms/nar/des/encoded_nar.cpp
ol-imorozko, line 68: `uint` **defined to** `-1`?? That's so wrong.
You have integer overflow here, then you increment it and it becomes zero again.
Just initialize with zero and do increment at the end of the loop

src/core/algorithms/nar/des/encoded_nar.cpp
```diff
@@ -0,0 +1,102 @@
...
+    auto compare_by_permutation = [e = encoded_value_ranges_](size_t& a, size_t& b) -> bool {
+        return e[a].permutation > e[b].permutation;
+    };
+    std::sort(feature_order.begin(), feature_order.end(), compare_by_permutation);
+
+    uint implication_sign_after = implication_sign_pos_ * (encoded_value_ranges_.size() - 1);
+    uint handling_feat_num = -1;
+    for (size_t feature_index : feature_order) {
+        handling_feat_num++;
+        EncodedValueRange encoded_feature = encoded_value_ranges_[feature_index];
```
ol-imorozko, line 71: you're copying it here. Use const ref

src/core/algorithms/nar/des/encoded_nar.cpp
```diff
@@ -0,0 +1,102 @@
...
+        if (encoded_feature.threshold < RNG().Next()) {
+            continue;
+        }
+        auto domain = domains[feature_index];
+        auto decoded = encoded_value_ranges_[feature_index].Decode(domain);
```
ol-imorozko, line 76: again, you copying the shared pointer when using `auto domain = domains[feature_index]`. Use const reference, no need to increase ref count of shared ptr for no reason

ol-imorozko: again, no need to create an additional copy of `decoded`.
I'm aware that  `InsertInCons` wants a shared pointer, not a const ref, but when passing const ref to it it will create a copy automatically, so no problems here.

src/core/algorithms/nar/des/encoded_nar.cpp
```diff
@@ -0,0 +1,102 @@
...
+        }
+    }
+    return resulting_nar;
+}
+
+EncodedNAR::EncodedNAR(FeatureDomains domains, TypedRelation const* typed_relation) {
+    size_t feature_count = domains.size();
+    for (size_t feature_index = 0; feature_index < feature_count; feature_index++) {
+        encoded_value_ranges_.emplace_back(EncodedValueRange());
+    }
```
ol-imorozko, line 90: That's a constructor, so we can just do
```suggestion
EncodedNAR::EncodedNAR(const FeatureDomains& domains, const TypedRelation* typed_relation)
    : encoded_value_ranges_(domains.size(), EncodedValueRange()) {
```

src/core/algorithms/nar/des/encoded_nar.cpp
```diff
@@ -0,0 +1,102 @@
...
+    implication_sign_pos_ = RNG().Next();
+    SetQualities(domains, typed_relation);
+}
+
+EncodedNAR::EncodedNAR(size_t feature_count) {
+    for (size_t feature_index = 0; feature_index < feature_count; feature_index++) {
+        encoded_value_ranges_.emplace_back(EncodedValueRange());
+    }
+    implication_sign_pos_ = RNG().Next();
+}
```
ol-imorozko, line 100: same thing with `RNG`, just create a `  std::random_device ` as a field in `EncodedNAR` class

src/core/algorithms/nar/des/encoded_nar.cpp
```diff
@@ -0,0 +1,102 @@
...
+        encoded_value_ranges_.emplace_back(EncodedValueRange());
+    }
+    implication_sign_pos_ = RNG().Next();
+    SetQualities(domains, typed_relation);
+}
+
+EncodedNAR::EncodedNAR(size_t feature_count) {
+    for (size_t feature_index = 0; feature_index < feature_count; feature_index++) {
+        encoded_value_ranges_.emplace_back(EncodedValueRange());
+    }
```
ol-imorozko, line 98:
```suggestion
EncodedNAR::EncodedNAR(size_t feature_count)
    : encoded_value_ranges_(feature_count, EncodedValueRange()) {
```

ol-imorozko (changes requested):

src/core/algorithms/nar/des/encoded_value_range.h
```diff
@@ -0,0 +1,112 @@
...
+                        std::static_pointer_cast<model::DoubleValueRange>(domain);
+                model::Double span = double_domain->upper_bound - double_domain->lower_bound;
+                model::Double resulting_lower = double_domain->lower_bound + span * this->bound1;
+                model::Double resulting_upper = double_domain->lower_bound + span * this->bound2;
+                if (resulting_lower > resulting_upper) {
+                    std::swap(resulting_lower, resulting_upper);
+                }
+                return std::make_shared<model::DoubleValueRange>(
+                        model::DoubleValueRange(resulting_lower, resulting_upper));
+            }
```
Petua41, line 81: Maybe it's better to use private template method here?
```
template <typename T, typename RangeT>
    std::shared_ptr<RangeT> DecodeTypedValueRange(std::shared_ptr<model::ValueRange> const& domain) const {
        std::shared_ptr<RangeT> int_domain = std::static_pointer_cast<RangeT>(domain);
        T span = int_domain->upper_bound - int_domain->lower_bound;
        T resulting_lower = int_domain->lower_bound + span * this->bound1;
        T resulting_upper = int_domain->lower_bound + span * this->bound2;
        if (resulting_lower > resulting_upper) {
            std::swap(resulting_lower, resulting_upper);
        }
        return std::make_shared<RangeT>(resulting_lower, resulting_upper);
    }
```
and call it here:
```suggestion
            case TypeId::kInt: {
            return DecodeTypedValueRange<Int, IntValueRange>(domain);
        }
        case TypeId::kDouble: {
            return DecodeTypedValueRange<Double, DoubleValueRange>(domain);
        }
```
Also you can define `Mixed` case as template specialization:
```
template <>
std::shared_ptr<model::StringValueRange>
EncodedValueRange::DecodeTypedValueRange<model::String, model::StringValueRange>
(std::shared_ptr<model::ValueRange> const& domain) const {
    using namespace model;
    std::shared_ptr<StringValueRange> string_domain =
            std::static_pointer_cast<StringValueRange>(domain);
    std::vector<String> string_vector = string_domain->domain;
    size_t span = string_vector.size();
    // upper_bound is not used, resulting NARs bind categorical values with a single
    // value.
    String result;
    if (bound1 == 1.0) {
        result = string_vector.back();
    } else {
        result = string_vector[(size_t)(span * this->bound1)];
    }
    return std::make_shared<StringValueRange>(StringValueRange(result));
}
```
Note that specialization should appear **after** template itself and **before** first instantiation. The easier way to do this is to place template specialization and `EncodeValueRange::Decode` definition in `.cpp` file -- in that order

Petua41, line 81: Anyway, `EncodedValueRange::Decode` should be placed in a `.cpp` file

src/core/algorithms/nar/des/encoded_value_range.h
```diff
@@ -0,0 +1,112 @@
...
+                std::shared_ptr<model::IntValueRange> int_domain =
+                        std::static_pointer_cast<model::IntValueRange>(domain);
+                model::Int span = int_domain->upper_bound - int_domain->lower_bound;
+                model::Int resulting_lower = int_domain->lower_bound + span * this->bound1;
+                model::Int resulting_upper = int_domain->lower_bound + span * this->bound2;
+                if (resulting_lower > resulting_upper) {
+                    std::swap(resulting_lower, resulting_upper);
+                }
+                return std::make_shared<model::IntValueRange>(
+                        model::IntValueRange(resulting_lower, resulting_upper));
```
Petua41, line 68:
```suggestion
                return std::make_shared<model::IntValueRange>(resulting_lower, resulting_upper);
```
The purpose of `make_shared` is to forward it's parameters to `IntValueRange`'s constructor and create `IntValueRange` in-place. In your code temporary `IntValueRange` is created and being copied (or moved) to `shared_ptr`. Also it's not a good idea to repeat type name.
This also applies to `Double` and `Mixed` cases

src/core/algorithms/nar/value_range.cpp
```diff
@@ -0,0 +1,109 @@
...
+            continue;
+        }
+        if (double_value > upper_bound) {
+            upper_bound = double_value;
+        }
+        if (double_value < lower_bound) {
+            lower_bound = double_value;
+        }
+    }
+}
```
Petua41, line 81: This method repeats `IntValueRange::IntValueRange`. The only difference is `int_value`/`double_value` type. I think, you can use template function like
```
template <typename T>
CreateValueRange(...) {
    ...
    T typed_value = model::Type::GetValue<T>(value);
    ...
```
and in constructor will be `CreateValueRange<model::Int>(...)` and `CreateValueRange<model::Double>(...)`.

src/core/algorithms/nar/value_range.cpp
```diff
@@ -0,0 +1,109 @@
...
+    result += std::to_string(upper_bound);
+    result += "]";
+    return result;
+}
+
+IntValueRange::IntValueRange(model::TypedColumnData const& column) {
+    bool initialized = false;
+    for (size_t row_index = 0; row_index < column.GetNumRows(); row_index++) {
+        std::byte const* value = column.GetValue(row_index);
+        long int int_value = model::Type::GetValue<model::Int>(value);
```
Petua41, line 45: You assign value of type `model::Int` to `int_value` of type `long`. Then you assign it to `lower_bound` and `upper_bound`, both of type `model::Int`. Why can't `int_value` be of type `model::Int`?
Moreover, `model::Int` is not the same type as `long` -- it's alias for `int_64_t`, which is twice as long on some machines (see https://en.cppreference.com/w/cpp/language/types).
```suggestion
        model::Int int_value = model::Type::GetValue<model::Int>(value);
```
The same for `DoubleValueRange`

src/core/algorithms/nar/value_range.h
```diff
@@ -0,0 +1,87 @@
...
+
+namespace model {
+
+class ValueRange {
+public:
+    virtual TypeId GetTypeId() const = 0;
+    virtual bool Includes(std::byte const* value) const = 0;
+    virtual std::string ToString() const = 0;
+
+protected:
```
Petua41, line 16: Style guide recommends that `protected` section precede `public`

src/tests/test_des.cpp
```diff
@@ -0,0 +1,37 @@
...
+    }
+};
+
+TEST_F(DESTest, LaunchTest) {
+    auto algorithm = CreateAlgorithmInstance(kAbalone, 0.5, 0.6, 100, 100, 0.9, 0.5,
+                                             algos::des::DifferentialStrategy::rand1Bin);
+    algorithm->Execute();
+    SUCCEED();
+}
+
```
Petua41, line 36: Add tests, checking that algorithm works fine:
- test that NARs that should hold are found
- test that their qualities are calculated correctly (fitness, support, confidence)

Petua41 (changes requested):

src/core/algorithms/nar/des/encoded_value_range.h
```diff
@@ -0,0 +1,112 @@
...
+            default:
+                throw std::out_of_range("Index out of range for value range.");
+        }
+    }
+
+    std::shared_ptr<model::ValueRange> Decode(std::shared_ptr<model::ValueRange> domain) const {
+        switch (domain->GetTypeId()) {
+            case model::TypeId::kInt: {
+                std::shared_ptr<model::IntValueRange> int_domain =
+                        std::static_pointer_cast<model::IntValueRange>(domain);
```
ol-imorozko, line 60: You can use `auto` when specifying type of a variable when you are performing a cast. The reason is that the type can be easily deduced by both a compiler and a person reading the code:
```suggestion
                auto int_domain = std::static_pointer_cast<model::IntValueRange>(domain);
```

src/core/algorithms/nar/des/encoded_value_range.h
```diff
@@ -0,0 +1,112 @@
...
+                model::Int resulting_upper = int_domain->lower_bound + span * this->bound2;
+                if (resulting_lower > resulting_upper) {
+                    std::swap(resulting_lower, resulting_upper);
+                }
+                return std::make_shared<model::IntValueRange>(
+                        model::IntValueRange(resulting_lower, resulting_upper));
+            }
+            case model::TypeId::kDouble: {
+                std::shared_ptr<model::DoubleValueRange> double_domain =
+                        std::static_pointer_cast<model::DoubleValueRange>(domain);
```
ol-imorozko, line 72: same here with an `auto`

src/core/algorithms/nar/des/encoded_value_range.h
ol-imorozko: also we should avoid using c-style casts: "`(size_t)`"

src/core/algorithms/nar/des/encoded_value_range.h
```diff
@@ -0,0 +1,112 @@
...
+                        std::static_pointer_cast<model::StringValueRange>(domain);
+                std::vector<model::String> string_vector = string_domain->domain;
+                size_t span = string_vector.size();
+                // upper_bound is not used, resulting NARs bind categorical values with a single
+                // value.
+                model::String result;
+                if (bound1 == 1.0) {
+                    result = string_vector.back();
+                } else {
+                    result = string_vector[(size_t)(span * this->bound1)];
```
ol-imorozko, line 94: why `this->bound1`? just `bound1` is enough

src/core/algorithms/nar/des/encoded_value_range.h
```diff
@@ -0,0 +1,112 @@
...
+                        std::string("ValueRange has invalid type_id in function: ") + __func__);
+        }
+    }
+
+    EncodedValueRange() {
+        permutation = RNG().Next();
+        threshold = RNG().Next();
+        bound1 = RNG().Next();
+        bound2 = RNG().Next();
+    }
```
ol-imorozko, line 109: Same thing with `RNG()`, I don't like this approach.

Here, since it's a constructor, we can initialize those fields using member initializer list:
```cpp
    EncodedValueRange()
        : permutation(RNG().Next()),
          threshold(RNG().Next()),
          bound1(RNG().Next()),
          bound2(RNG().Next()) {}
```

src/core/algorithms/nar/nar.cpp
```diff
@@ -0,0 +1,119 @@
...
+    size_t antecounter = 0;
+    for (auto const& [key, value] : ante_) {
+        if (antecounter > 0) {
+            result += ", ";
+        }
+        result += std::to_string(key);
+        result += ": ";
+        result += value->ToString();
+        antecounter++;
+    }
```
ol-imorozko, line 17: Incorporate changes suggested by [Petua41](https://github.com/Petua41), but also use `std::ostringstream` for the `result`. It makes consecutive creation of strings more readable and efficient:
```cpp
std::string NAR::ToString() const {
    std::ostringstream result;
    result << qualities_.fitness << " {";
    for (auto it = ante_.begin(); it != ante_.end(); ++it) {
        if (it != ante_.begin()) {
            result << ", ";
        }
        result << it->first << ": " << it->second->ToString();
    }
    result << "} ===> {";
    for (auto it = cons_.begin(); it != cons_.end(); ++it) {
        if (it != cons_.begin()) {
            result << ", ";
        }
        result << it->first << ": " << it->second->ToString();
    }
    result << "} s: " << qualities_.support << " c: " << qualities_.confidence;
    return result.str();
}
```

src/core/algorithms/nar/nar.cpp
```diff
@@ -0,0 +1,119 @@
...
+NARQualities CalcQualities(size_t num_rows_fit_ante, size_t num_rows_fit_ante_and_cons,
+                           size_t included_features, size_t feature_count, size_t num_rows) {
+    NARQualities result;
+    if (num_rows_fit_ante == 0) {
+        result.fitness = 0.0;
+        result.confidence = 0.0;
+        return result;
+    } else {
+        result.confidence = num_rows_fit_ante_and_cons / (double)num_rows_fit_ante;
+    }
```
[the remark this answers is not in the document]
ol-imorozko: Also `(double)` is a c-style cast. Prefer using `static_cast`

src/core/algorithms/nar/nar.cpp
```diff
@@ -0,0 +1,119 @@
...
+    result += std::to_string(qualities_.confidence);
+    return result;
+}
+
+NARQualities CalcQualities(size_t num_rows_fit_ante, size_t num_rows_fit_ante_and_cons,
+                           size_t included_features, size_t feature_count, size_t num_rows) {
+    NARQualities result;
+    if (num_rows_fit_ante == 0) {
+        result.fitness = 0.0;
+        result.confidence = 0.0;
```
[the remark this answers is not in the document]
ol-imorozko: Yes, I also think that `{0.0, 0.0, 0.0}` will be better. In fact, we can do that for all cases:
```cpp
    if (num_rows_fit_ante == 0) {
        return {0.0, 0.0, 0.0};
    }

    double confidence = static_cast<double>(num_rows_fit_ante_and_cons) / num_rows_fit_ante;
    double support = static_cast<double>(num_rows_fit_ante_and_cons) / num_rows;

    if (support == 0.0) {
        return {0.0, 0.0, 0.0};
    }

    double inclusion = static_cast<double>(included_features) / feature_count;
    double fitness = (confidence + support + inclusion) / 3.0;

    return {fitness, confidence, support};
```

src/core/algorithms/nar/nar.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+    double inclusion = included_features / (double)feature_count;
+    result.fitness = (result.confidence + result.support + inclusion) / 3.0;
+    return result;
+}
+
+// TODO: this function is way too big and cluttered
+void NAR::SetQualities(TypedRelation const* typed_relation) {
+    if (ante_.size() == 0 || cons_.size() == 0) {
+        qualities_.fitness = 0.0;
```
ol-imorozko, line 61: what about other fields?
```cpp
        qualities_ = {0.0, 0.0, 0.0};
```

src/core/algorithms/nar/nar.cpp
```diff
@@ -0,0 +1,119 @@
...
+        qualities_consistent_ = true;
+        return;
+    }
+
+    size_t num_rows_fit_ante = 0;
+    size_t num_rows_fit_ante_and_cons = 0;
+    for (size_t rowi = 0; rowi < typed_relation->GetNumRows(); rowi++) {
```
ol-imorozko, line 68: extract `typed_relation->GetNumRows()` to a variable since you will be using it later

src/core/algorithms/nar/nar.cpp
```diff
@@ -0,0 +1,119 @@
...
+        bool row_fits_ante = true;
+        bool row_fits_cons = true;
+        for (size_t coli = 0; coli < typed_relation->GetNumColumns(); coli++) {
```
ol-imorozko, line 71: same with `GetNumColumns()`

ol-imorozko (changes requested):

ol-imorozko (changes requested):

src/core/algorithms/nar/des/des.cpp
```diff
@@ -0,0 +1,89 @@
...
+}
+
+void DES::MakeExecuteOptsAvailable() {
+    using namespace config::names;
+    MakeOptionsAvailable({kPopulationSize, kMaxFitnessEvaluations, kDifferentialScale,
+                          kCrossoverProbability, kDifferentialStrategy});
+}
+
+FeatureDomains DES::FindFeatureDomains(TypedRelation const* typed_relation) {
+    auto feature_domains = std::vector<std::shared_ptr<ValueRange>>();
```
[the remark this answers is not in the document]
VanyaVolgushev: FeatureDomains are of const type (domain of any given feature doesn't change during execution) so it can't be done like this, but i can do `std::vector<std::shared_ptr<ValueRange>> feature_domains`

src/core/algorithms/nar/des/encoded_nar.cpp
ol-imorozko: Also use `::` to access static members: `EncodedValueRange::kFieldCount`

src/core/algorithms/nar/des/des.cpp
```diff
@@ -0,0 +1,89 @@
...
+    for (int i = 0; i < num_evaluations_;
+         i++) {  // TODO: change num_evaluations type to something unsigned
+        size_t candidate_i = i % population_size_;
+        EncodedNAR mutant = MutateIndividual(population, candidate_i);
+        NAR mutant_decoded = mutant.SetQualities(feature_domains, typed_relation_.get());
+        double candidate_fitness = population[candidate_i].GetQualities().fitness;
+
+        if (mutant.GetQualities().fitness > candidate_fitness) {
+            population[candidate_i] = mutant;
+            nar_collection_.emplace_back(mutant_decoded);
```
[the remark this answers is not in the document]
VanyaVolgushev: they are stack variables, shouldn't the compiler be able to figure out that its the last time we use them on its own and use `std::move`?

src/core/algorithms/nar/des/differential_functions.cpp
VanyaVolgushev: `std::unordered_set` is used to prevent duplicate indices from appearing. I did consider the solution you are suggesting, but if `population_size` is 100 (usually it's bigger) and number_of_indices is 3 (doesn't get much bigger than that with most mutation strategies) it's a lot of shuffling just to discard 97 indices and leave the remaining 3. Your solution is good when `number_of_indices` is comparable to `population_size`, but with this algorithm it's usually the opposite, so my solution doesn't run into many "collisions" and works out ok.

src/core/algorithms/nar/des/differential_functions.cpp
```diff
@@ -0,0 +1,58 @@
+#include "differential_functions.h"
+
+#include <algorithm>
+
+namespace algos::des {
+
+// gets slow if population ~= number_of_indices
+std::unordered_set<size_t> GetRandIndices(size_t except_index, size_t population,
```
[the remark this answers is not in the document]
VanyaVolgushev: I agree that the function doesn't need to return an `std::unordered_set` though, it's only needed for the internal implementation.

src/core/algorithms/nar/des/encoded_nar.cpp
VanyaVolgushev: `GetElementAtIndex(index)` then has to be a const function returning a non-const reference. This won't compile:

```
double& EncodedNAR::GetElementAtIndex(size_t index) const {
    if (index == 0) {
        return implication_sign_pos_;
    } else {
        index--;
        size_t feature = index / EncodedValueRange().kFieldCount;
        size_t feature_field = index % EncodedValueRange().kFieldCount;
        return encoded_value_ranges_[feature][feature_field];
    }
}
```
because "qualifiers dropped in binding reference of type "double &" to initializer of type "const double". I'm not sure how to fix this.

src/core/algorithms/nar/des/encoded_nar.cpp
VanyaVolgushev: `implication_sign_after` is not used if `encoded_value_ranges_` size is zero because then the loop is not entered because `feature_order` is of size 0.

VanyaVolgushev: > `uint`? Use `size_t`

Fixed. Could you tell me why size_t is better here?

VanyaVolgushev: There are two places where the loop can end, I did it so that I don't have to write the increment twice. Did fix though.

src/core/algorithms/nar/des/des.cpp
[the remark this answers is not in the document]
ol-imorozko: > they are stack variables, shouldn't the compiler be able to figure out that its the last time we use them on its own and use `std::move`?

No. 
Consider the first line:
```cpp
population[candidate_i] = mutant;
```
The compiler need to decide which `operator=` of `EncodedNar` class to use. There are two overloads  to it: : one is move-assignment, the other is copy-assignment. Compiler generated them by default for your class, since `EncodedNar` class does not define a destructor and/or move constructor and/or copy constructor and/or .... etc. There are some rules to this. Nevermind, the fact is it generated those overloads:
```cpp
//copy assignment
EncodedNAR& operator=(const EncodedNAR& other);
//move assignment
EncodedNAR& operator=(EncodedNAR&& other) noexcept;
```

src/core/algorithms/nar/des/encoded_nar.cpp
ol-imorozko: I believe smth like that will work:
```cpp
template <typename EncodedNARType>
auto& GetElementAtIndex(EncodedNARType& self, size_t index) {
    //body
}
```

and call like `return GetElementAtIndex(*this, index);`

src/core/algorithms/nar/des/encoded_nar.cpp
ol-imorozko: > > `uint`? Use `size_t`
> 
> Fixed. Could you tell me why size_t is better here?

I mean, it does it's job, but it's unconventional:
```bash
imorozko@112618560-nix:~/maga/VanyaVolgushev(add-des-release)$ git grep "uint "
src/core/algorithms/nar/des/encoded_nar.cpp:    uint implication_sign_after = implication_sign_pos_ * (encoded_value_ranges_.size() - 1);
src/core/algorithms/nar/des/encoded_nar.cpp:    uint handling_feat_num = -1;
src/core/model/types/null_type.h:        static std::atomic_uint hash = 0;
imorozko@112618560-nix:~/maga/VanyaVolgushev(add-des-release)$ 
imorozko@112618560-nix:~/maga/VanyaVolgushev(add-des-release)$ git grep "size_t " | wc -l
1372
```
There are two instances of `uint` usage and `1372` of `size_t`

src/core/algorithms/nar/des/des.cpp
```diff
@@ -0,0 +1,87 @@
...
+            EnumToMutationStrategy(differential_options_.differential_strategy);
+    return (*diff_func)(population, at, differential_options_);
+}
+
+unsigned long long DES::ExecuteInternal() {
+    FeatureDomains feature_domains = FindFeatureDomains(typed_relation_.get());
+    std::vector<EncodedNAR> population = GetRandomPopulationInDomains(feature_domains);
+
+    for (unsigned i = 0; i < num_evaluations_; i++) {
+        size_t candidate_i = i % population_size_;
```
Petua41, line 69: You mix `unsinged`, `size_t` and `int` in one statement. Generally, it's not safe (see, for example, https://github.com/Nekrolm/ubbook/blob/master/numeric/integer_promotion.md). Is '`int` the best type for `population_size_`? I guess, **size** cannot be negative.
The same is for `num_evaluations_` in this class.

src/core/algorithms/nar/des/encoded_nar.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+double const& EncodedNAR::operator[](size_t index) const {
+    if (index == 0) {
+        return implication_sign_pos_;
+    } else {
+        index--;
+        size_t feature = index / EncodedValueRange().kFieldCount;
+        size_t feature_field = index % EncodedValueRange().kFieldCount;
+        return encoded_value_ranges_[feature][feature_field];
+    }
```
Petua41, line 38:
```suggestion
    index--;
    size_t feature = index / EncodedValueRange().kFieldCount;
    size_t feature_field = index % EncodedValueRange().kFieldCount;
    return encoded_value_ranges_[feature][feature_field];
```

src/core/algorithms/nar/des/encoded_nar.cpp
```diff
@@ -0,0 +1,101 @@
...
+}
+
+model::NARQualities const& EncodedNAR::GetQualities() const {
+    if (!qualities_consistent_) {
+        throw std::logic_error("Getting uninitialized qualities from NAR.");
+    }
+    return qualities_;
+}
+
+NAR EncodedNAR::Decode(FeatureDomains& domains) const {
```
Petua41, line 56:
```suggestion
NAR EncodedNAR::Decode(FeatureDomains const& domains) const {
```
If `FeatureDomains` doesn't have `const operator[]` (some of the stl containers may have this problem), use `at`

src/core/algorithms/nar/des/des.cpp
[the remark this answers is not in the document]
VanyaVolgushev: Thank you for the in-depth reply

src/core/algorithms/nar/nar_algorithm.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+NARAlgorithm::NARAlgorithm(std::vector<std::string_view> phase_names)
+    : Algorithm(std::move(phase_names)) {
+    using namespace config::names;
+    RegisterOptions();
+    algos::Algorithm::MakeOptionsAvailable({kTable});
+}
+
+void NARAlgorithm::LoadDataInternal() {
+    typed_relation_ = model::ColumnLayoutTypedRelationData::CreateFrom(*input_table_, true);
```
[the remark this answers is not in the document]
VanyaVolgushev: Nulls are not compared in current implementation

src/core/algorithms/nar/des/des.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+        if (mutant.GetQualities().fitness > candidate_fitness) {
+            population[candidate_i] = mutant;
+            nar_collection_.emplace_back(mutant_decoded);
+        }
+    }
+
+    auto compare_by_fitness = [](const NAR& a, const NAR& b) -> bool {
+        return a.GetQualities().fitness > b.GetQualities().fitness;
+    };
```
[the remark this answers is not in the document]
VanyaVolgushev: They take in different types, first takes in two NARs, second takes two EncodedNARs.

src/core/algorithms/nar/des/encoded_nar.cpp
VanyaVolgushev: If I understood correctly we then have to reference `self.implication_sign_pos` and `implication_sign_pos` is private and inaccessible in that function 😭. GetElementAtIndex must be a private method of EncodedNAR. But then we have to either make it const, making it unable to return a non-const ref, or not const, making it unusable in a const method. I guess a macro would work.

https://stackoverflow.com/questions/15413831/c-two-versions-of-overloading-subscript-operator

one of the commenters said:

> Yes, you'll need to define both of them. You could define the non-const version in terms of the const version but then you'd have to do a const_cast. You're better off just having the duplicate implementation.

src/core/algorithms/nar/des/encoded_nar.cpp
VanyaVolgushev: If we want to avoid all copies why don't we use `std::move` everywhere and avoid copies altogether:
```c++
auto domain = std::move(domains[feature_index]);
auto decoded = std::move(encoded_value_ranges_[feature_index].Decode(std::move(domain)));
if (handling_feat_num > implication_sign_after) {
            resulting_nar.InsertInCons(feature_index, std::move(decoded);)
        } else {
            resulting_nar.InsertInAnte(feature_index, std::move(decoded);)
        }
. . .
```
my types only use basic C++ types so their generated move constructors should be valid? To my understanding we don't because a) it makes code less readable and b) compiler should be able to figure out where to move most basic objects on its own.

src/core/algorithms/nar/des/encoded_nar.cpp
VanyaVolgushev: Surprisingly it doesn't work. Reason is: if you initialize it that way it only calls the constructor once and then copies the constructed object `domains.size()` times. But we need to call the constructor for each new EncodedValueRange because they have to be random.

src/core/algorithms/nar/value_range.cpp
```diff
@@ -0,0 +1,109 @@
+#include "value_range.h"
+
+namespace model {
+
+StringValueRange::StringValueRange(model::TypedColumnData const& column) {
+    domain = std::vector<std::string>();
```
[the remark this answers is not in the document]
VanyaVolgushev: domain has all repetitions removed so we can't know it's size beforehand, why reserve space for every row?

src/tests/test_des.cpp
VanyaVolgushev: This algorithm converges on non-useful NARs as it is right now. Just added a test that checks if it mined an exact expected set of NARs. Might help to find out when a change unexpectedly affected algo behaviour.

src/core/algorithms/nar/des/des.cpp
```diff
@@ -0,0 +1,89 @@
+#include "des.h"
+
```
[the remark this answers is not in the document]
VanyaVolgushev: I think I'll wait for CI to tell me what to fix then.

src/core/algorithms/nar/nar.cpp
```diff
@@ -0,0 +1,102 @@
...
+            model::TypedColumnData const& column = typed_relation->GetColumnData(coli);
+            auto value = column.GetValue(rowi);
+            if (row_fits_ante) {
+                row_fits_ante = AnteFitsValue(coli, value);
+            } else {
+                break;
+            }
+            if (row_fits_ante) {
+                row_fits_cons &= ConsFitsValue(coli, value);
+            }
```
Petua41, line 62:
```suggestion
            if (row_fits_ante) {
                row_fits_ante = AnteFitsValue(coli, value);
                row_fits_cons &= ConsFitsValue(coli, value);
            } else {
                break;
            }
```

src/core/algorithms/nar/des/encoded_nar.cpp
ol-imorozko: Okay, makes sense, thanks

src/core/algorithms/nar/value_range.cpp
```diff
@@ -0,0 +1,45 @@
...
+StringValueRange::StringValueRange(TypedColumnData const& column) {
+    for (size_t row_index = 0; row_index < column.GetNumRows(); ++row_index) {
+        std::byte const* value = column.GetValue(row_index);
+        std::string string_value = Type::GetValue<std::string>(value);
+        bool first_occurrence =
+                std::find(domain.begin(), domain.end(), string_value) == domain.end();
+        if (first_occurrence) {
+            domain.push_back(std::move(string_value));
+        }
+    }
```
ol-imorozko, line 14: IMHO smth like this is easier:
```suggestion
    std::unordered_set<std::string> unique_values;
    for (size_t row_index = 0; row_index < column.GetNumRows(); ++row_index) {
        std::byte const* value = column.GetValue(row_index);
        auto string_value = Type::GetValue<std::string>(value);
        if (unique_values.insert(string_value).second) {
            domain.push_back(std::move(string_value));
        }
    }
```

src/core/algorithms/nar/value_range.cpp
```diff
@@ -0,0 +1,45 @@
...
+                                        __func__);
+    }
+}
+
+std::string StringValueRange::ToString() const {
+    std::ostringstream result;
+    result << "[";
+    if (domain.empty()) {
+        return "[]";
+    }
```
ol-imorozko, line 36: move that before 
```cpp
    std::ostringstream result;
    result << "[";
```

src/core/algorithms/nar/value_range.h
```diff
@@ -0,0 +1,94 @@
...
+                initialized = true;
+            } else {
+                if (numeric_value < lower_bound) {
+                    lower_bound = numeric_value;
+                }
+                if (numeric_value > upper_bound) {
+                    upper_bound = numeric_value;
+                }
+            }
+        }
```
ol-imorozko, line 65:
```suggestion
        std::vector<T> values(column.GetNumRows());

        for (std::byte const* raw_value : column.GetData()) {
            values.push_back(Type::GetValue<T>(raw_value));
        }

        auto [min, max] = std::ranges::minmax(values);
        lower_bound = min;
        upper_bound = max;
```
Smth like this?

src/core/algorithms/nar/value_range.h
```diff
@@ -0,0 +1,94 @@
...
+    explicit NumericValueRange(T lower_bound, T upper_bound)
+        : lower_bound(lower_bound), upper_bound(upper_bound) {}
+
+    [[nodiscard]] bool Includes(std::byte const* value) const override {
+        T numeric_value = Type::GetValue<T>(value);
+        return numeric_value >= lower_bound && numeric_value <= upper_bound;
+    }
+
+    TypeId GetTypeId() const override {
+        if (std::is_same<T, Double>::value) {
```
ol-imorozko, line 77:
```suggestion
        if constexpr (std::is_same_v<T, Double>) {
```

src/core/algorithms/nar/value_range.h
```diff
@@ -0,0 +1,94 @@
...
+            return TypeId::kDouble;
+        } else if (std::is_same<T, Int>::value) {
```
ol-imorozko, line 79:
```suggestion
        } else if constexpr (std::is_same_v<T, Int>) {
```

src/tests/test_des.cpp
```diff
@@ -0,0 +1,45 @@
...
+};
+
+TEST_F(DESTest, LaunchTest) {
+    auto algorithm = CreateAlgorithmInstance(kAbalone, 0.5, 0.6, 100u, 100u, 0.9, 0.5,
+                                             algos::des::DifferentialStrategy::rand1Bin);
+    algorithm->Execute();
+    std::vector<std::string> result;
+    for (auto i : algorithm->GetNARVector()) {
+        result.push_back(std::to_string(i.GetQualities().fitness));
+    }
```
ol-imorozko, line 37: I would add a method to `DESTest`, seems like it can be reused later for other tests:
```cpp
    static std::vector<std::string> ExtractFitnessValues(
            std::vector<algos::des::NAR> const& nar_vector) {
        std::vector<std::string> fitness_values;
        fitness_values.reserve(nar_vector.size());
        for (auto const& nar : nar_vector) {
            fitness_values.push_back(std::to_string(nar.GetQualities().fitness));
        }
        return fitness_values;
    }
```
```suggestion
    auto fitness_values = ExtractFitnessValues(algorithm->GetNARVector());
```

ol-imorozko (changes requested):
Oh, by the way, I feel like I left a comment somewhere that asks to change
```cpp
std::vector a;
loop (x : vector b) {a.push_back(foo(x))}
```
to
```cpp
std::vector a(b.size());
loop (x : vector b) {a.push_back(foo(x))}
```
instead of
```cpp
std::vector a;
a.reserve(b.size());
loop (x : vector b) {a.push_back(foo(x))}
```
which is, of course, wrong, but I don't remember where :)

src/core/algorithms/nar/value_range.h
ol-imorozko: sorry, I meant reserve here instead of a constructor from int 
```cpp
std::vector<T> values();
values.reserve(column.GetNumRows());
```

src/core/algorithms/nar/des/des.cpp
```diff
@@ -0,0 +1,95 @@
...
+                          kCrossoverProbability, kDifferentialStrategy});
+}
+
+FeatureDomains DES::FindFeatureDomains(TypedRelation const* typed_relation) {
+    std::vector<std::shared_ptr<ValueRange>> feature_domains;
+    feature_domains.reserve(typed_relation->GetNumColumns());
+    for (size_t i = 0; i < typed_relation->GetNumColumns(); ++i) {
+        std::shared_ptr<ValueRange> domain = CreateValueRange(typed_relation->GetColumnData(i));
+        feature_domains.push_back(std::move(domain));
+    }
```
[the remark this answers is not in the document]
VanyaVolgushev: Fixed and marked this as resolved but I think you meant `push_back`? We already construct a value range with `CreateValueRange`.

src/core/algorithms/nar/des/encoded_nar.cpp
```diff
@@ -0,0 +1,101 @@
...
+NAR EncodedNAR::Decode(FeatureDomains const& domains, RNG& rng) const {
+    NAR resulting_nar;
+    std::vector<size_t> feature_order(encoded_value_ranges_.size());
+    std::iota(std::begin(feature_order), std::end(feature_order), 0);
+    auto compare_by_permutation = [this](size_t a, size_t b) {
+        return encoded_value_ranges_[a].permutation > encoded_value_ranges_[b].permutation;
+    };
+    std::sort(feature_order.begin(), feature_order.end(), compare_by_permutation);
+
+    size_t implication_sign_after = implication_sign_pos_ * (encoded_value_ranges_.size() - 1);
```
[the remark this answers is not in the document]
VanyaVolgushev: Added a comment that clears it up.

src/core/algorithms/nar/des/encoded_nar.cpp
```diff
@@ -0,0 +1,101 @@
...
+        encoded_value_ranges_.emplace_back(rng);
+    }
+    implication_sign_pos_ = rng.Next();
+    SetQualities(domains, typed_relation, rng);
+}
+
+EncodedNAR::EncodedNAR(size_t feature_count, RNG& rng) {
+    for (size_t feature_index = 0; feature_index < feature_count; ++feature_index) {
+        encoded_value_ranges_.emplace_back(rng);
+    }
```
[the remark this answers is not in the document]
VanyaVolgushev: `resize` doesn't work in both cases because it constructs objects if previous size was smaller, did you mean `reserve`?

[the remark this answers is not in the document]
ol-imorozko: no, here we have to use `resize` to ensure that the vector contains exactly `feature_count` elements, which gets default-constructed, and then replaced by `ranges::generate` with one generated by the lambda.

If the constructor is trivial, performance difference will be negligible.

If we don't want to default construct them first, we can do
```cpp
encoded_value_ranges_.reserve(feature_count);
std::ranges::generate_n(std::back_inserter(encoded_value_ranges_), feature_count, [&rng]() { return EncodedValueRange(rng); });
```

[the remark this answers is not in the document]
ol-imorozko: yeah, maybe `generate_n` is more desirable

src/core/algorithms/nar/des/encoded_value_range.cpp
```diff
@@ -0,0 +1,92 @@
...
+std::shared_ptr<RangeT> EncodedValueRange::DecodeTypedValueRange(
+        std::shared_ptr<model::ValueRange> const& domain) const {
+    auto typed_domain = std::static_pointer_cast<RangeT>(domain);
+    T span = typed_domain->upper_bound - typed_domain->lower_bound;
+    T resulting_lower = typed_domain->lower_bound + span * bound1;
+    T resulting_upper = typed_domain->lower_bound + span * bound2;
+    if (resulting_lower > resulting_upper) {
+        std::swap(resulting_lower, resulting_upper);
+    }
+    return std::make_shared<RangeT>(resulting_lower, resulting_upper);
```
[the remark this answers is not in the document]
VanyaVolgushev: Produces "possibly dangling reference to a temporary" warning.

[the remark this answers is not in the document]
ol-imorozko: I still think that
```cpp
    T decoded_lower = typed_domain->lower_bound + span * bound1;
    T decoded_upper = typed_domain->lower_bound + span * bound2;

#if 1 // first option
    auto result = std::minmax(decoded_lower, decoded_upper);
    return std::make_shared<RangeT>(result.first, result.second);
#else
    auto [resulting_lower, resulting_upper] = std::minmax(decoded_lower, decoded_upper);
    return std::make_shared<RangeT>(resulting_lower, resulting_upper);
#endif
```

is better than
```cpp
    T resulting_lower = typed_domain->lower_bound + span * bound1;
    T resulting_upper = typed_domain->lower_bound + span * bound2;
    
    if (resulting_lower > resulting_upper) {
        std::swap(resulting_lower, resulting_upper);
    }
    return std::make_shared<RangeT>(resulting_lower, resulting_upper);
```

src/core/algorithms/nar/nar.cpp
```diff
@@ -0,0 +1,100 @@
...
+        for (size_t coli = 0; coli < num_columns; ++coli) {
+            model::TypedColumnData const& column = typed_relation->GetColumnData(coli);
+            auto value = column.GetValue(rowi);
+            if (row_fits_ante) {
+                row_fits_ante = AnteFitsValue(coli, value);
+                row_fits_cons &= ConsFitsValue(coli, value);
+            } else {
+                break;
+            }
+        }
```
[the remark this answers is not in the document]
VanyaVolgushev: This removes the `break` statement completely, which forces us to check the rest of the columns even though we can already tell that the row doesn't satisfy our antecedent.

src/core/algorithms/nar/nar.cpp
```diff
@@ -0,0 +1,100 @@
...
+model::NARQualities const& NAR::GetQualities() const {
+    if (!qualities_consistent_) {
+        throw std::logic_error("Getting uninitialized qualities from NAR.");
+    }
+    return qualities_;
+}
+
+void NAR::InsertInAnte(size_t feature_index, std::shared_ptr<ValueRange> range) {
+    qualities_consistent_ = false;
+    ante_.insert({feature_index, std::move(range)});
```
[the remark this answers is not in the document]
VanyaVolgushev: ChatGPT says `insert` doesn't call default constructor before assigning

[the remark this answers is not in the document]
ol-imorozko: Yea, you're right. But if the key already exists, we won't insert a new value though. If that's not a problem, use `insert`

src/core/algorithms/nar/nar.h
```diff
@@ -0,0 +1,45 @@
...
+    bool AnteFitsValue(size_t feature_index, std::byte const* value_of_feature) const {
+        return MapFitsValue(ante_, feature_index, value_of_feature);
+    }
+
+    bool ConsFitsValue(size_t feature_index, std::byte const* value_of_feature) const {
+        return MapFitsValue(cons_, feature_index, value_of_feature);
+    }
+
+    static bool MapFitsValue(std::map<size_t, std::shared_ptr<ValueRange>> const& map,
+                             size_t feature_index, std::byte const* value);
```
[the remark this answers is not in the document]
VanyaVolgushev: Why remove the static modifier? This method doesn't need access to any instance fields.

src/core/algorithms/nar/value_range.h
VanyaVolgushev: Duplicating the whole column into the `values` vector seems wasteful to me. What if we have a dataset that is 5 columns, each with 5000 rows. Then we are duplicating 20% of the dataset, when the rest of the algo can be made non-memory-demanding if population is set to some low value.

ol-imorozko: yeah, makes sense, thanks

ol-imorozko: hm, I think this will work then, cause we can just tell `std::minmax` how to extract values:
```cpp
    auto [min_ptr, max_ptr] = std::ranges::minmax(column.GetData(), {},
        [](std::byte const* value) { 
            return Type::GetValue<T>(value); 
        });

    lower_bound = Type::GetValue<T>(min_ptr);
    upper_bound = Type::GetValue<T>(max_ptr);
```

VanyaVolgushev: works but makes the test run 14 seconds instead of 1 :)

VanyaVolgushev: image

ol-imorozko: That's unfortunate :skull: 
Because all that is just one `std::minmax` algo, don't want to give up on that and use raw loop.. But, well, I see no other choice.

But still let's use `std::min` and `std::max` then instead of comparing manually.

Like this?
```cpp
    std::byte const* first_value = column.GetValue(0);
    T lower_bound = Type::GetValue<T>(first_value);
    T upper_bound = lower_bound;

    for (size_t row_index = 1; row_index < column.GetNumRows(); ++row_index) {
        std::byte const* value = column.GetValue(row_index);
        T numeric_value = Type::GetValue<T>(value);

        lower_bound = std::min(lower_bound, numeric_value);
        upper_bound = std::max(upper_bound, numeric_value);
    }
```
(don't want to add extra branch in a loop)

VanyaVolgushev: If we really wanted to make it more concise we could do something like this, getting rid of the `initialized` bool:

```C++
template <typename T>
struct NumericValueRange : public ValueRange {
    T lower_bound = std::numeric_limits<T>::max();
    T upper_bound = std::numeric_limits<T>::lowest();

    explicit NumericValueRange(TypedColumnData const& column) {
        for (size_t row_index = 0; row_index < column.GetNumRows(); ++row_index) {
            std::byte const* value = column.GetValue(row_index);
            T numeric_value = Type::GetValue<T>(value);
            if (numeric_value < lower_bound) {
                lower_bound = numeric_value;
            }
            if (numeric_value > upper_bound) {
                upper_bound = numeric_value;
            }
        }
    }
```

VanyaVolgushev: edited^

src/core/algorithms/nar/des/des.cpp
```diff
@@ -0,0 +1,95 @@
...
+                          kCrossoverProbability, kDifferentialStrategy});
+}
+
+FeatureDomains DES::FindFeatureDomains(TypedRelation const* typed_relation) {
+    std::vector<std::shared_ptr<ValueRange>> feature_domains;
+    feature_domains.reserve(typed_relation->GetNumColumns());
+    for (size_t i = 0; i < typed_relation->GetNumColumns(); ++i) {
+        auto const& column = typed_relation->GetColumnData(i);
+        feature_domains.push_back(CreateValueRange(column));
+    }
```
ol-imorozko, line 48: So sorry, don't know how I missed this, but that is just a 
```suggestion
    std::ranges::transform(typed_relation->GetColumnData(), std::back_inserter(feature_domains), CreateValueRange);
```

src/core/algorithms/nar/nar.cpp
```diff
@@ -0,0 +1,104 @@
...
+        }
+        return os.str();
+    };
+
+    std::ostringstream result;
+    result << qualities_.fitness << " {";
+    result << map_to_string(ante_) << "} ===> {" << map_to_string(cons_);
+    result << "} s: " << qualities_.support << " c: " << qualities_.confidence;
+    return result.str();
+    return result.str();
```
ol-imorozko, line 24:
```suggestion
    return result.str();
```

ol-imorozko (changes requested):

ol-imorozko:
LGTM apart from those last five requested changes

src/core/algorithms/nar/nar.cpp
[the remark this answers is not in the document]
ol-imorozko: No, it doesn't, I've added `row_fits_ante` to the loop check, which is basically the same as adding 
`if (row_fits_ante) else break` to the body:

```cpp
for (size_t coli = 0; coli < num_columns && row_fits_ante; ++coli) {
```

src/core/algorithms/nar/nar.cpp
VanyaVolgushev: as far as I understand range based iota exists in std since C++ 23 and we use 20

https://en.cppreference.com/w/cpp/algorithm/ranges/iota

ol-imorozko (approved):

Petua41 (approved):
LGTM. The one thing left is to fix commit history. We try to make sure that newer commits don't fix older commit within same PR.

src/core/algorithms/nar/nar.cpp
ol-imorozko: Yes, sorry, I meant to say that we should use `feature_order.begin()` instead of `std::begin(feature_order)`, as that's how we're doing it everywhere else.

alexandrsmirn (approved):

PR #492, opened 2024-11-11
Fix CMakeLists.txt, macOS guide and refactor .gitignore 

Vdaleke:
Increase CMake minimum required version to 3.15 due to [incompatibilities
with scikit-build-core](https://github.com/scikit-build/scikit-build-core/blob/acc69a86cd1995b8baff4e631e453906226801dd/README.md?plain=1#L72). Update README.md.

Add POLICY checks to avoid errors on policy-unsupported versions.

Change CMP0167 policy of boost searching to OLD due to non-compliance
with "[Modern CMake](https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1)" development patterns. Add "TODO" to fix this later.
Add useful logs to understand what Boost libraries were found by CMake.

Apply sort to .gitignore to improve readability.
Add .DS_Store, which is generated in macOS directories.
Add .cache, which is generated by clangd.

Add python3 to installation guide.
Fix macOS Boost installation guide to work with python module.
Fix default installation directories as root permissions were required.
Change macOS SDK version to compatibility with [new version](https://developer.apple.com/documentation/macos-release-notes/macos-15_1-release-notes).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,14 +1,15 @@
-cmake-build-debug/
+# Build directories, libraries and binary files
+**/myeasylog.log
 build/
-Build/
-BUILD/
-bUILD/
-Datasets/
-.sconsign.dblite
+cmake-build-debug/
+dist/
 lib/
-.csv
+venv/
+
+# IDE and related files
+.cache/
 .idea/
 .vscode/
-**/myeasylog.log
-dist
-venv
+
+# OS Generated file
+**/.DS_Store
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,13 @@
-cmake_minimum_required(VERSION 3.13)
-cmake_policy(SET CMP0144 NEW)
+cmake_minimum_required(VERSION 3.15)
+if (POLICY CMP0167)
+    # deprecated by definition, but we use old behaviour for find_package(Boost) to load CMake's FindBoost module
+    # TODO: port project to "Modern CMake" https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1
+    cmake_policy(SET CMP0167 OLD)
+endif()
+if (POLICY CMP0144)
+    cmake_policy(SET CMP0144 NEW)
+endif()
+
 project(Desbordante)
 
 option(COPY_PYTHON_EXAMPLES "Copy Python examples" OFF)
@@ -103,9 +111,16 @@ endif()
 
 # configuring boost
 set(Boost_USE_STATIC_LIBS OFF)
-find_package(Boost 1.81.0 REQUIRED COMPONENTS container thread graph CONFIG)
-include_directories(${Boost_INCLUDE_DIRS})
-message(${Boost_INCLUDE_DIRS})
+
+find_package(Boost 1.81.0 REQUIRED COMPONENTS container thread graph)
+if (Boost_FOUND)
+    message(STATUS "Found Boost include dirs: ${Boost_INCLUDE_DIRS}")
+    include_directories( ${Boost_INCLUDE_DIRS})
+    message(STATUS "Found Boost library dirs: ${Boost_LIBRARY_DIRS}")
+else ()
+    message(FATAL_ERROR "Boost not found. If boost is installed specify environment"
+        "variables like \"BOOST_ROOT\" for hint to CMake where to search")
+endif()
 
 # providing subdirectories for header inclusion
 include_directories(
--- a/README.md
+++ b/README.md
@@ -246,7 +246,7 @@ The following instructions were tested on Ubuntu 20.04+ LTS and macOS Sonoma 14.
 Prior to cloning the repository and attempting to build the project, ensure that you have the following software:
 
 - GNU GCC, version 10+
-- CMake, version 3.13+
+- CMake, version 3.15+
 - Boost library built with GCC, version 1.81.0+
 
 To use test datasets you will need:
--- a/README.md
+++ b/README.md
@@ -256,7 +256,7 @@ To use test datasets you will need:
 
 Run the following commands:
 ```sh 
-sudo apt install gcc g++ cmake libboost-all-dev git-lfs
+sudo apt install gcc g++ cmake libboost-all-dev git-lfs python3
 export CC=gcc
 export CXX=g++
 ```
@@ -271,10 +271,10 @@ xcode-select --install
 ```
 Follow the prompts to continue.
 
-To install GCC and CMake on macOS we recommend to use [Homebrew](https://brew.sh/) package manager. With Homebrew
+To install GCC, CMake and python on macOS we recommend to use [Homebrew](https://brew.sh/) package manager. With Homebrew
 installed, run the following commands:
 ```sh
-brew install gcc@14 cmake
+brew install gcc@14 cmake python3
 ```
 After installation, check `cmake --version`. If command is not found, then you need to add to environment path to
 homebrew installed packages. To do this open `~/.zprofile` (for Zsh) or
@@ -282,20 +282,19 @@ homebrew installed packages. To do this open `~/.zprofile` (for Zsh) or
 After that, restart the terminal and check the version of CMake again, now it should be displayed.
 
 Then you need to install Boost library built with GCC. Please avoid using Homebrew for this, as the Boost version provided by Homebrew
-is built with Clang, which has a different ABI. Instead, download the latest version of Boost from the [official website](https://www.boost.org/users/download/) and unpack the archive to the `/usr/local/` directory or another directory of your choice:
+is built with Clang, which has a different ABI. Instead, download the latest version of Boost from the [official website](https://www.boost.org/users/download/), open terminal and run:
 ```sh
-cd /usr/local/
+cd ~/Downloads
 curl https://archives.boost.io/release/1.86.0/source/boost_1_86_0.tar.bz2 --output "boost_1_86_0.tar.bz2"
-tar xvjf boost_1_86_0.tar.bz2
-rm boost_1_86_0.tar.bz2
+tar xvjf boost_1_86_0.tar.bz2 && rm boost_1_86_0.tar.bz2
 cd boost_1_86_0
 ```
 Navigate to the unpacked Boost directory in the terminal and run the following commands:
 ```sh
 ./bootstrap.sh 
 echo "using darwin : : g++-14 ;" > user-config.jam
-./b2 install --user-config=user-config.jam --layout=versioned
-export BOOST_ROOT=$(pwd) # export Boost_ROOT=$(pwd) for CMake 3.26 and below.
+sudo ./b2 install --user-config=user-config.jam --layout=versioned
+export BOOST_ROOT=/usr/local/ # export Boost_ROOT=/usr/local/ for CMake 3.26 and below.
 ``` 
 You can also add the last export with current path to `~/.zprofile` or `~/.bash_profile` to set this boost path by default.
 
@@ -304,10 +303,11 @@ Before building the project you must set locally or in the above-mentioned dotfi
 export CC=gcc-14
 export CXX=g++-14
 export SDKROOT=/Library/Developer/CommandLineTools/SDKs/MacOSX14.sdk/
+export DYLD_LIBRARY_PATH=/usr/local/lib:${DYLD_LIBRARY_PATH}
 ```
-The first two lines set GCC as the default compiler in CMake. The last export is also necessary due to issues with GCC 14 and
-the last MacOSX15.0.sdk used by CMake by default, you can read more about this [here](https://gist.github.com/scivision/d69faebbc56da9714798087b56de925a)
-and [here](https://github.com/iains/gcc-14-branch/issues/11).
+The first two lines set GCC as the default compiler in CMake. The `SDKROOT` export is also necessary due to issues with GCC 14 and
+the last macOS 15 SDK used by CMake by default, you can read more about this [here](https://gist.github.com/scivision/d69faebbc56da9714798087b56de925a)
+and [here](https://github.com/iains/gcc-14-branch/issues/11). The last export is the solution for dynamic linking with python module.
 
 ### Building the project
 #### Building the Python module using pip


Vdaleke:
By the way, I checked the build with the new SDK and nothing changed. So the need to use the 14 SDK remains the same.

Vdaleke:
I also comment `export DYLD_LIBRARY_PATH`. It's architecturally bad solution and not recommended to use.

The problem is that built python module can't find boost dynamic libraries:
```
>>> import desbordante
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: dlopen(/Users/vdaleke/Documents/desbordante-core/venv/lib/python3.12/site-packages/desbordante.cpython-312-darwin.so, 0x0002): Library not loaded: libboost_container-xgcc14-mt-a64-1_86.dylib
  Referenced from: <5DE266BA-6882-3835-BB89-3E13967FF861> /Users/vdaleke/Documents/desbordante-core/venv/lib/python3.12/site-packages/desbordante.cpython-312-darwin.so
  Reason: tried: 'libboost_container-xgcc14-mt-a64-1_86.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OSlibboost_container-xgcc14-mt-a64-1_86.dylib' (no such file), 'libboost_container-xgcc14-mt-a64-1_86.dylib' (no such file), '/usr/lib/libboost_container-xgcc14-mt-a64-1_86.dylib' (no such file, not in dyld cache), '/Users/vdaleke/Documents/desbordante-core/libboost_container-xgcc14-mt-a64-1_86.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/Users/vdaleke/Documents/desbordante-core/libboost_container-xgcc14-mt-a64-1_86.dylib' (no such file), '/Users/vdaleke/Documents/desbordante-core/libboost_container-xgcc14-mt-a64-1_86.dylib' (no such file), '/usr/lib/libboost_container-xgcc14-mt-a64-1_86.dylib' (no such file, not in dyld cache)
```

The reason is that CMake doesn't set 'LC_LOAD_DYLIB' cmd in ABI for boost libraries with correct path:
```
➜ otool -L /Users/vdaleke/Documents/desbordante-core/venv/lib/python3.12/site-packages/desbordante.cpython-312-darwin.so 
/Users/vdaleke/Documents/desbordante-core/venv/lib/python3.12/site-packages/desbordante.cpython-312-darwin.so:
	libboost_container-xgcc14-mt-a64-1_86.dylib (compatibility version 0.0.0, current version 0.0.0)
	libboost_thread-xgcc14-mt-a64-1_86.dylib (compatibility version 0.0.0, current version 0.0.0)
	libboost_graph-xgcc14-mt-a64-1_86.dylib (compatibility version 0.0.0, current version 0.0.0)
	/opt/homebrew/opt/gcc/lib/gcc/current/libstdc++.6.dylib (compatibility version 7.0.0, current version 7.33.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1345.120.2)
```
I don't know why it's like this and is it possible to force CMake use full paths or rpath-relative paths. I searched enough and couldn't find a solution. I suggest to stay with this solution just to have a working guide for the moment. In next related PR I suggest to focus on porting project to C++ standard to use clang and homebrew Boost on macOS.

CMakeLists.txt
```diff
@@ -103,9 +110,15 @@ endif()
...
 set(Boost_USE_STATIC_LIBS OFF)
-find_package(Boost 1.81.0 REQUIRED COMPONENTS container thread graph CONFIG)
-include_directories(${Boost_INCLUDE_DIRS})
-message(${Boost_INCLUDE_DIRS})
+
+find_package(Boost 1.81.0 REQUIRED COMPONENTS container thread graph)
+if (Boost_FOUND) 
+    message(STATUS "Found Boost include dirs: ${Boost_INCLUDE_DIRS}")
+    include_directories(${Boost_INCLUDE_DIRS})
+    message(STATUS "Found Boost library dirs: ${Boost_LIBRARY_DIRS}")
```
Vdaleke, line 118: As for "Modern CMake", I encountered a problem when decide to add this log. In 3c34d42 we set default behaviour to searching for BoostConfig.cmake. And if we look at [config](https://github.com/boostorg/boost_install/blob/d77d0e94ba3d78f2079313ecb6e4907a7e8618e9/BoostConfig.cmake#L226) there is a section "compatibility variables" that contains setting of Boost_INCLUDE_DIRS, but not Boost_LIBRARY_DIRS, so log was empty.

Why like this? I found a discussion there one of boost contributors [answered](https://lists.boost.org/Archives/boost/2021/07/251772.php): 
"""
> but Boost_LIBRARY_DIRS is not set anymore - but maybe its just not needed

All of them are not needed at all and not set by the CONFIG files. The
FindBoost module has explicit code to add those for legacy reasons.
In "good" CMake you are supposed to use the targets so all those
variables don't matter for ages now.
"""
So this is about "Modern CMake" approach. CMake forces to use targets, it can also be seen in [include_directories documentation](https://cmake.org/cmake/help/latest/command/include_directories.html). Our CMake script not use this approach and we should fix it at least for such compatibility reasons. 

So as this compatibility was included for legacy reasons I suggest to stay use Boost find module and fix this later.

README.md
```diff
@@ -250,31 +250,30 @@ xcode-select --install
...
+tar xvjf boost_1_86_0.tar.bz2 && rm boost_1_86_0.tar.bz2
 cd boost_1_86_0
 ```
 Navigate to the unpacked Boost directory in the terminal and run the following commands:
 ```sh
 ./bootstrap.sh 
 echo "using darwin : : g++-14 ;" > user-config.jam
 ./b2 install --user-config=user-config.jam --layout=versioned
-export BOOST_ROOT=$(pwd) # export Boost_ROOT=$(pwd) for CMake 3.26 and below.
+export BOOST_ROOT=/usr/local/ # export Boost_ROOT=/usr/local/ for CMake 3.26 and below.
```
Vdaleke, line 276: I didn't know it before, but Boost `./b2 install` installs boost libs to usr/local/ by default. It can be override by `--prefix`.

CMakeLists.txt
```diff
@@ -1,5 +1,12 @@
-cmake_minimum_required(VERSION 3.13)
-cmake_policy(SET CMP0144 NEW)
+cmake_minimum_required(VERSION 3.15)
+if (POLICY CMP0167)
+    # deprecated by definition, TODO: port project to "Modern CMake" https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1
+    cmake_policy(SET CMP0167 OLD)
```
vs9h, line 4: Can we add here small comment about the meaning of this policy? Briefly, in one sentence, about the fact that we use this behavior for find_package(Boost) to load CMake's FindBoost module.

CMakeLists.txt
```diff
@@ -1,5 +1,12 @@
-cmake_minimum_required(VERSION 3.13)
-cmake_policy(SET CMP0144 NEW)
+cmake_minimum_required(VERSION 3.15)
+if (POLICY CMP0167)
+    # deprecated by definition, TODO: port project to "Modern CMake" https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1
+    cmake_policy(SET CMP0167 OLD) 
+endif()
+if (POLICY CMP0144)
+    cmake_policy(SET CMP0144 NEW)
+endif()
```
vs9h, line 8: Am I right in understanding that this is just a fix, since we have this policy only since version 3.27?

At the same time, if someone tries to use an older version now, they will get a warning or even an error?

vs9h (changes requested):
Overall it looks good, I left a few comments and questions.

CMakeLists.txt
Vdaleke: > Am I right in understanding that this is just a fix, since we have this policy only since version 3.27?

Yes.

> At the same time, if someone tries to use an older version now, they will get a warning or even an error?

I used to think that the behavior of policies was like: if I know about this policy I will set the right one, if I don't know I don't care.  But when I set the CMP0167 policy as `OLD` I got an error in CI. I think with `NEW` there was a warning.

README.md
```diff
@@ -283,10 +282,11 @@ Before building the project you must set locally or in the above-mentioned dotfi
 export CC=gcc-14
 export CXX=g++-14
 export SDKROOT=/Library/Developer/CommandLineTools/SDKs/MacOSX14.sdk/
+export DYLD_LIBRARY_PATH=/usr/local/lib
```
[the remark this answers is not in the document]
Vdaleke: I added curly braces, I don't know the reason, but I had on mac this import generated for python where they used:
```
# Setting PATH for Python 3.12
# The original version is saved in .zprofile.pysave
PATH="/Library/Frameworks/Python.framework/Versions/3.12/bin:${PATH}"
export PATH
```

.gitignore
```diff
@@ -1,14 +1,16 @@
...
-venv
+BUILD/
+Build/
+Datasets/
+bUILD/
+build/
+cmake-build-debug/
+dist/
+lib/
+venv/
```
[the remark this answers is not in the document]
Vdaleke: Moved .cache/ to `# IDE and related files` as this folder is generated by `clangd` plugin.

vs9h (approved):
Unfortunately, I can't verify whether it is correct to add such an export for `DYLD_LIBRARY_PATH`. I have not encountered this before.

I have no more comments on the changes, overall it looks good.

BUYT-1:
After the library is compiled, does it work if `DYLD_LIBRARY_PATH` is not set?

Vdaleke:
> After the library is compiled, does it work if `DYLD_LIBRARY_PATH` is not set?

I'm not sure about the meaning of 'does it work', as for `Desbordante_test` it works fine and as for 'build/target/desbordante.cpython-312-darwin.so' which is built by sci-build-core not, the error is like this:
```
(venv) ➜  target git:(fix/refactor-gitignore-and-readme) ✗ python3                                                           
Python 3.12.5 (v3.12.5:ff3bc82f7c9, Aug  7 2024, 05:32:06) [Clang 13.0.0 (clang-1300.0.29.30)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import desbordante
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: dlopen(/Users/vdaleke/Documents/desbordante-core/build/target/desbordante.cpython-312-darwin.so, 0x0002): Library not loaded: libboost_container-xgcc14-mt-a64-1_86.dylib
  Referenced from: <2B68442A-C7AA-3FFB-A70C-C3B1D6F570A0> /Users/vdaleke/Documents/desbordante-core/build/target/desbordante.cpython-312-darwin.so
  Reason: tried: 'libboost_container-xgcc14-mt-a64-1_86.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OSlibboost_container-xgcc14-mt-a64-1_86.dylib' (no such file), 'libboost_container-xgcc14-mt-a64-1_86.dylib' (no such file), '/usr/lib/libboost_container-xgcc14-mt-a64-1_86.dylib' (no such file, not in dyld cache), '/Users/vdaleke/Documents/desbordante-core/build/target/libboost_container-xgcc14-mt-a64-1_86.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/Users/vdaleke/Documents/desbordante-core/build/target/libboost_container-xgcc14-mt-a64-1_86.dylib' (no such file), '/Users/vdaleke/Documents/desbordante-core/build/target/libboost_container-xgcc14-mt-a64-1_86.dylib' (no such file), '/usr/lib/libboost_container-xgcc14-mt-a64-1_86.dylib' (no such file, not in dyld cache)
```
I absolutely don't know why it works like this and what good ways to fix it. I attach `otool -l` output for both binaries, where presented cmds for dyld: 
`Desbordante_test` https://pastebin.com/rxwv8xs1 
`desbordante.cpython-312-darwin.so` https://pastebin.com/wq50BcXm

Vdaleke:
I found some interesting c make variables which may be help to force include @rpath to dynamic libraries paths which is boost, I will try it later  

image

Petua41:
> Unfortunately, I can't verify whether it is correct to add such an export for `DYLD_LIBRARY_PATH`. I have not encountered this before.

`DYLD_LIBRARY_PATH` is equivalent of Linux `LD_PATH`, so yes, it's needed to be exported like this for `ld` to find shared libraries installed into `/usr/local/lib` (which is default path where `brew` installs them. Also, Boost is installed here).

Another way to do this is to set CMake's `LDFLAGS`.

Vdaleke:
I made some research and suggest to stay like this for several reasons.

1) I noticed different behaviour of CMake with myself-built Boost and Homebrew Boost. With Homebrew boost there are full paths for Boost libs and, as I wrote above, for gcc-built boost it's not. I looked through CMake FindBoost module and some homebrew files and can't get the answer. [I asked about this on CMake discourse](https://discourse.cmake.org/t/how-to-set-dyld-path-via-findpackage-boost/13323) and waiting for some answers.

2) Due to last point I don't know how to set RPATH correctly if it's so unstable except the way of adding "" and "/usr/local/ there, but this strange solution as for me.

3) We have PR where it's possible to build with Apple Clang and Homebrew Boost and this is the easiest solution. 

4) We need this CMake fixes and there is no reason to spend more time on this.

Vdaleke:
As a addition to the previous comment, I had an idea to install boost built with gcc by Homebrew, I recently found out that this is can be possible with 'brew install boost -s' option which means 'build from source', and I have some success in it, but it was stopped by impossibility of setting MacOSX14.sdk instead of broke MacOSX15.sdk. I [created an issue](https://github.com/Homebrew/brew/issues/19028) to brew repo and waiting for reaction.

vs9h:
> I made some research and suggest to stay like this for several reasons.
> 
> 1. I noticed different behaviour of CMake with myself-built Boost and Homebrew Boost. With Homebrew boost there are full DYLD_PATHs for Boost libs and, as I wrote above, for gcc-built boost it's not. I looked through CMake FindBoost module and some homebrew files and can't get the answer. [I asked about this on CMake discourse](https://discourse.cmake.org/t/how-to-set-dyld-path-via-findpackage-boost/13323) and hope for some answers.
> 2. Due to last point I don't know how to set RPATH correctly if it's so unstable except the way of adding "" and "/usr/local/ there, but this is strange solution as for me.
> 3. We have PR where it's possible to build with Apple Clang and Homebrew Boost and this is the easiest solution.
> 4. We need these CMake fixes and there is no reason to spend more time on this.

In general, I agree that the current solution should be sufficient and we can stop here.

Regarding the first point, I don't know what the problem is.

Regarding hardcoding `rpath`, it seems to me that this solution may be even worse than the current one. I don't think anyone does this in existing projects. We can do this only if we are sure that the installation occurs along single specific path. But we probably don't have such guarantees.

vs9h (approved):

PR #496, opened 2024-11-19
Fix the problem of primitives depending on the algorithm object

Sneper-Breeze:
### The problem:

> The problem was with the connection between the object of the primitive and the lifetime of the algorithm object. I.e. if the algorithm object was deleted, the primitive was invalid. 
> This behaviour is wrong and causes additional memory consumption.

### The solution presented:

- Add `shared_ptr<RelationalSchema>` in FD so that the schema is not deleted when the algorithm object is deleted.
- Add `shared_ptr<RelationalSchema>` in UCC. Same as in FD.
- Add new method to `AbstractRelationData` so that `shared_ptr<RelationalSchema>` can be used to create primitive objects.
- Changed the code to interact properly with the changed FD and UCC classes.


src/core/algorithms/ucc/ucc.h
```diff
@@ -1,13 +1,23 @@
 #pragma once
 
 #include "model/table/vertical.h"
+#include <memory>
```
BUYT-1, line 4:
```suggestion
#include <memory>

#include "model/table/vertical.h"
```

STL, external libraries, project files is the header order here, see [the style guide](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes). I don't know why `clang-format` didn't catch it.

src/core/model/table/relational_schema.h
```diff
@@ -27,7 +27,7 @@ class RelationalSchema {
     std::string name_;
 
 public:
-    std::unique_ptr<Vertical> empty_vertical_;
+    std::shared_ptr<Vertical> empty_vertical_;
```
BUYT-1, line 30: Ownership of `empty_vertical_` is not shared, no need to change it.

BUYT-1 (changes requested):
Commit names should describe what was changed, everything else goes into commit descriptions, see the [wiki](https://github.com/Desbordante/desbordante-core/wiki/Development).
```
Add schema `shared_ptr` field in `FD`

Fixes the lifetime issue where FD objects became invalid after the
algorithm object containing their schema was deleted.
```

src/core/algorithms/ucc/ucc.h
```diff
@@ -1,13 +1,23 @@
...
 // Defining UCC as Vertical, because we don't really need more functionality than Vertical provides.
-using UCC = Vertical;
+class UCC : public Vertical {
+    private:
+    std::shared_ptr<RelationalSchema> schema_;
+    public:
+    UCC(std::shared_ptr<RelationalSchema> schema, boost::dynamic_bitset<> indices)
+    : Vertical(schema.get(), indices), schema_(schema) { }
+    UCC(std::shared_ptr<RelationalSchema> schema, Vertical ucc_val)
+    : Vertical(ucc_val), schema_(schema) { }
```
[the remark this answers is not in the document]
Sneper-Breeze: But in order for this to make sense, it is need to specify && in the constructor, don't it?

src/python_bindings/py_util/bind_primitive.h
```diff
@@ -56,7 +56,7 @@ auto BindPrimitive(pybind11::module_& module, auto result_method, char const* ba
                    // util::PrimitiveCollection<...>::AsList() call, the default should work,
                    // otherwise consult pybind11's docs.
                    pybind11::return_value_policy result_rv_policy =
-                           pybind11::return_value_policy::reference_internal) {
+                           pybind11::return_value_policy::copy) {
```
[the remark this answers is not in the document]
Sneper-Breeze: I think this should be left in case someone needs it in the future for specific algorithm, plus it doesn't seem to have any effect on the compiled code in terms of time and memory.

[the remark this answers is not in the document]
BUYT-1: I put that argument in as a crutch. Nobody will need it if all primitive objects are copyable, which they should be.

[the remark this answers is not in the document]
Sneper-Breeze: All right, I got it.

src/core/algorithms/ucc/ucc.h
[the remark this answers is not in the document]
BUYT-1: `ucc_val` is passed by value, so there is nothing that can be affected on the caller's side. We only want to take the resources of this value (no copying) by using the move constructor.
The expression `ucc_val`, a variable name, is an lvalue expression, which bind to lvalue references (`T&`). The value returned by `std::move`, by contrast, is an xvalue expression, and those bind to rvalue references (`T&&`).
What this means is if there is an overloaded function with overloads the signatures of which include `T&` and `T&&`, like a copy (`T const&`) and move (`T&&`) constructor, when `ucc_val` is passed to it, the one with the lvalue reference (i.e. the copy constructor) is called, but when `std::move(ucc_val)` is, the one with the rvalue reference (i.e. the move constructor) is called.
It's the standard pass-by-value-and-move pattern in C++.

src/python_bindings/py_util/bind_primitive.h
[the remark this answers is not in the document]
Sneper-Breeze: This argument is used in some other places, so I'll leave it as is for a while until I fix other places.

src/core/algorithms/ucc/hpivalid/hpivalid.cpp
```diff
@@ -70,7 +70,7 @@ unsigned long long HPIValid::ExecuteInternal() {
 
 void HPIValid::RegisterUCCs(hpiv::ResultCollector const& rc) {
     std::vector<model::RawUCC> ucc_vector = rc.GetUCCs();
-    auto const* const schema = relation_->GetSchema();
+    std::shared_ptr<RelationalSchema> const schema = relation_->GetSharedPtrSchema();
```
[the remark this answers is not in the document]
Sneper-Breeze: error: returning reference to temporary [-Werror=return-local-addr]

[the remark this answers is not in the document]
BUYT-1: Should work fine if `AbstractRelationData::schema_` is `std::shared_ptr<RelationalSchema const>`. I don't think it needs a non-const pointer there, try changing it. If it doesn't compile, then don't bother and return `std::shared_ptr<RelationalSchema> const&` instead.

src/python_bindings/py_util/bind_primitive.h
BUYT-1, line 1: Looks like a reformat. Put it into a separate commit.

BUYT-1 (changes requested):
Rebase, reformat, and it can be merged.

src/core/algorithms/fd/fd.h
```diff
@@ -10,9 +10,12 @@ class FD {
 private:
     Vertical lhs_;
     Column rhs_;
+    std::shared_ptr<RelationalSchema const> schema_;
 
 public:
-    FD(Vertical const& lhs, Column const& rhs) : lhs_(lhs), rhs_(rhs) {}
+    FD(Vertical const& lhs, Column const& rhs,
+       std::shared_ptr<RelationalSchema const> const& schema)
+        : lhs_(lhs), rhs_(rhs), schema_(schema) {}
```
BUYT-1, line 18:
```suggestion
       std::shared_ptr<RelationalSchema const> schema)
        : lhs_(lhs), rhs_(rhs), schema_(std::move(schema)) {}
```

src/core/algorithms/fd/fdep/fd_tree_element.cpp
```diff
@@ -243,15 +243,15 @@ void FDTreeElement::PrintDependencies(std::bitset<kMaxAttrNum>& active_path, std
...
+void FDTreeElement::FillFdCollection(std::shared_ptr<RelationalSchema> scheme,
+                                     std::list<FD>& fd_collection, unsigned int max_lhs) const {
     std::bitset<kMaxAttrNum> active_path;
     this->TransformTreeFdCollection(active_path, fd_collection, scheme, max_lhs);
 }
 
 void FDTreeElement::TransformTreeFdCollection(std::bitset<kMaxAttrNum>& active_path,
                                               std::list<FD>& fd_collection,
-                                              RelationalSchema const& scheme,
+                                              std::shared_ptr<RelationalSchema> scheme,
```
BUYT-1, line 254:
```suggestion
                                              std::shared_ptr<RelationalSchema> const& scheme,
```

src/core/algorithms/fd/fdep/fd_tree_element.cpp
```diff
@@ -243,15 +243,15 @@ void FDTreeElement::PrintDependencies(std::bitset<kMaxAttrNum>& active_path, std
     }
 }
 
-void FDTreeElement::FillFdCollection(RelationalSchema const& scheme, std::list<FD>& fd_collection,
-                                     unsigned int max_lhs) const {
+void FDTreeElement::FillFdCollection(std::shared_ptr<RelationalSchema> scheme,
```
BUYT-1, line 246:
```suggestion
void FDTreeElement::FillFdCollection(std::shared_ptr<RelationalSchema> const& scheme,
```

src/core/algorithms/ucc/hpivalid/hpivalid.cpp
```diff
@@ -70,7 +70,7 @@ unsigned long long HPIValid::ExecuteInternal() {
 
 void HPIValid::RegisterUCCs(hpiv::ResultCollector const& rc) {
     std::vector<model::RawUCC> ucc_vector = rc.GetUCCs();
-    auto const* const schema = relation_->GetSchema();
+    std::shared_ptr<RelationalSchema const> const schema = relation_->GetSharedPtrSchema();
```
BUYT-1, line 73:
```suggestion
    std::shared_ptr<RelationalSchema const> const& schema = relation_->GetSharedPtrSchema();
```

src/core/algorithms/ucc/hyucc/hyucc.cpp
```diff
@@ -67,7 +67,7 @@ unsigned long long HyUCC::ExecuteInternal() {
 
 void HyUCC::RegisterUCCs(std::vector<boost::dynamic_bitset<>>&& uccs,
                          std::vector<hy::ClusterId> const& og_mapping) {
-    auto const* const schema = relation_->GetSchema();
+    std::shared_ptr<RelationalSchema const> const schema = relation_->GetSharedPtrSchema();
```
BUYT-1, line 70:
```suggestion
    std::shared_ptr<RelationalSchema const> const& schema = relation_->GetSharedPtrSchema();
```

src/core/algorithms/ucc/hyucc/hyucc.h
```diff
@@ -13,7 +13,7 @@ namespace algos {
 
 class HyUCC : public UCCAlgorithm {
 private:
-    std::unique_ptr<ColumnLayoutRelationData> relation_;
+    std::shared_ptr<ColumnLayoutRelationData> relation_;
```
BUYT-1, line 16: No need to change this
```suggestion
    std::unique_ptr<ColumnLayoutRelationData> relation_;
```

src/core/algorithms/ucc/pyroucc/pyroucc.h
```diff
@@ -12,7 +12,7 @@ namespace algos {
 
 class PyroUCC : public DependencyConsumer, public UCCAlgorithm {
 private:
-    std::unique_ptr<ColumnLayoutRelationData> relation_;
+    std::shared_ptr<ColumnLayoutRelationData> relation_;
```
BUYT-1, line 15:
```suggestion
    std::unique_ptr<ColumnLayoutRelationData> relation_;
```

src/core/model/table/column_layout_relation_data.cpp
```diff
@@ -19,7 +19,7 @@ std::vector<int> ColumnLayoutRelationData::GetTuple(int tuple_index) const {
     return tuple;
 }
 
-std::unique_ptr<ColumnLayoutRelationData> ColumnLayoutRelationData::CreateFrom(
+std::shared_ptr<ColumnLayoutRelationData> ColumnLayoutRelationData::CreateFrom(
```
BUYT-1, line 22:
```suggestion
std::unique_ptr<ColumnLayoutRelationData> ColumnLayoutRelationData::CreateFrom(
```

BUYT-1 (changes requested):

src/core/algorithms/fd/fd_algorithm.h
```diff
@@ -43,8 +43,10 @@ class FDAlgorithm : public Algorithm {
     /* Registers new FD.
      * Should be overrided if custom behavior is needed
      */
-    virtual void RegisterFd(Vertical lhs, Column rhs) {
-        if (lhs.GetArity() <= max_lhs_) fd_collection_.Register(std::move(lhs), std::move(rhs));
+    virtual void RegisterFd(Vertical lhs, Column rhs,
+                            std::shared_ptr<RelationalSchema const> const& schema) {
+        if (lhs.GetArity() <= max_lhs_)
+            fd_collection_.Register(std::move(lhs), std::move(rhs), std::move(schema));
```
BUYT-1, line 49:
```suggestion
            fd_collection_.Register(std::move(lhs), std::move(rhs), schema);
```

BUYT-1 (changes requested):
Almost there
Rename commits:
Create UCC class -> Create `UCC` class
Fix the python bindings for FD -> Copy `FD` objects when returning results in Python
Fix the python bindings for UCC -> Copy `UCC` objects when returning results in Python

BUYT-1 (approved):

PR #500, opened 2024-12-02
Add ability to build with Clang on macOS

p-senichenkov:
Add ability to build with Clang compiler (Homebrew) on macOS:
- remove GCC-specific code
- add corresponding tests to CI


p-senichenkov:
Now it's a part of PR#507

PR #509, opened 2025-01-05
Fix Boost installation, macOS-GCC guide. Set project default language to C++

Vdaleke:
Fix CI and README.md Boost installing steps to install only 3 used libraries to reduce installation time.

Set CXX(C++) as default language in CMakeLists.txt.

Fix README.md due to fix of GCC & macOS SDK15 incompatibility. 
Increase Boost version to 1.87.0 in macOS dependency installation guide.
Remove C-compilers exports in dependencies installation guide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,11 +257,10 @@ To use test datasets you will need:
 Run the following commands:
 ```sh 
 sudo apt install gcc g++ cmake libboost-all-dev git-lfs python3
-export CC=gcc
 export CXX=g++
 ```
-The last 2 lines set gcc as CMake compiler in your terminal session.
-You can also add them to the end of `~/.profile` to set this by default in all sessions.
+The last line sets GCC as CMake compiler in your terminal session.
+You can also add it to the end of `~/.profile` to set this by default in all sessions.
 
 #### MacOS dependencies installation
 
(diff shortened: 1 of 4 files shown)


Vdaleke:
Looks like boost fix is very effective. Decreasing of time for some jobs up to 6 minutes:
image

Vdaleke:
I didn't notice before that the SDK bug was fixed because I had to reinstall GCC.
https://github.com/Homebrew/brew/issues/19028#issuecomment-2571714667

README.md
```diff
@@ -300,14 +299,10 @@ You can also add the last export with current path to `~/.zprofile` or `~/.bash_
...
 ```sh
-export CC=gcc-14
 export CXX=g++-14
-export SDKROOT=/Library/Developer/CommandLineTools/SDKs/MacOSX14.sdk/
 export DYLD_LIBRARY_PATH=/usr/local/lib:${DYLD_LIBRARY_PATH}
 ```
-The first two lines set GCC as the default compiler in CMake. The `SDKROOT` export is also necessary due to issues with GCC 14 and
-the last macOS 15 SDK used by CMake by default, you can read more about this [here](https://gist.github.com/scivision/d69faebbc56da9714798087b56de925a)
-and [here](https://github.com/iains/gcc-14-branch/issues/11). The last export is the solution for dynamic linking with python module.
+The first two lines set GCC as the default compiler in CMake. The last export is the solution for dynamic linking with python module.
```
BUYT-1, line 305:
```suggestion
The first line sets GCC as the default compiler in CMake. The last export is the solution for dynamic linking with python module.
```

PR #510, opened 2025-01-06
Constant DC verification

xJoskiy:
This is an extension of a previous work https://github.com/Desbordante/desbordante-core/blob/main/docs/papers/DC%20Verification%20-%20Anosov%20Pavel%20-%202024%20spring.pdf. This PR adds support for DC verification with constant predicactes, violations highlights and minimization algorithm. Original verification algorithm can be found here https://arxiv.org/pdf/2309.12436.


src/core/algorithms/dc/model/column_operand.h
```diff
@@ -3,79 +3,136 @@
...
+enum class Tuple { kS, kT, kMixed };
+
 //  @brief Represents a column operand for Predicate.
 //
 //  A predicate (e.g., t.A == s.A) comprises three elements:
 //  the column operand from the first tuple ("t.A"), the comparison operator
 //  ("=="), and the column operand from the second tuple ("s.A"). The `ColumnOperand` class
 //  encapsulates the column operand part of a predicate, such as "t.A" or "s.A".
 //
+//  A constant value also can be a column operand thus boost::optional is utilized
```
BUYT-1, line 27: Why not `std::optional`?

src/python_bindings/dc/bind_dc_verification.cpp
```diff
@@ -11,7 +19,9 @@ void BindDCVerification(py::module_& main_module) {
     auto dc_verification_module = main_module.def_submodule("dc_verification");
 
     BindPrimitiveNoBase<algos::DCVerifier>(dc_verification_module, "DCVerification")
-            .def("dc_holds", &algos::DCVerifier::DCHolds);
-}
+            .def("dc_holds", &algos::DCVerifier::DCHolds)
+            .def("get_violations", &algos::DCVerifier::GetViolations);
 
-}  // namespace python_bindings
+    py::bind_vector<std::vector<std::pair<size_t, size_t>>>(main_module, "vector_pair");
```
BUYT-1, line 25: Is there an issue with the standard conversion?

src/core/algorithms/dc/model/column_operand.h
ol-imorozko, line 27:
```suggestion
//  A constant value also can be a column operand thus std::optional is utilized
```

src/core/algorithms/dc/model/column_operand.h
```diff
@@ -3,79 +3,136 @@
...
+//
 //  The class distinguishes between operands derived from the first tuple (t) and those
```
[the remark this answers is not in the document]
ol-imorozko: Does not seem to be resolved, I still see the comment about `is_first_tuple_`

src/core/algorithms/dc/model/column_operand.h
```diff
@@ -3,79 +3,136 @@
...
 namespace algos::dc {
 
+enum class Tuple { kS, kT, kMixed };
+
 //  @brief Represents a column operand for Predicate.
 //
 //  A predicate (e.g., t.A == s.A) comprises three elements:
 //  the column operand from the first tuple ("t.A"), the comparison operator
 //  ("=="), and the column operand from the second tuple ("s.A"). The `ColumnOperand` class
 //  encapsulates the column operand part of a predicate, such as "t.A" or "s.A".
```
ol-imorozko, line 25: It's hard to me to grasp what `kMixed` represents here. I suggest you to add the explanation to this comment right here that explains what "t" and "s" means

src/core/algorithms/dc/model/column_operand.h
```diff
@@ -3,79 +3,136 @@
...
-            }
+    ColumnOperand() noexcept : column_(nullptr), val_(nullptr) {};
+
+    ColumnOperand(Column const* column, dc::Tuple tuple, model::Type const* type)
+        : column_(column), tuple_(tuple), type_(type), val_(nullptr) {}
+
+    ColumnOperand(std::string const& str_val, model::Type const* type)
+        : column_(nullptr), type_(type) {
+        std::byte* val = type->Allocate();
+        type->ValueFromStr(val, str_val);
```
ol-imorozko, line 49: Feel free to ignore:
I would use `type_` here. Makes more sense to me that in constructor we filled some private fields first and then called methods on _those_ fields, instead on what we filled them with (though it's the same pointer, naturally)

src/core/algorithms/dc/model/column_operand.h
```diff
@@ -3,79 +3,136 @@
...
+            return type_->Compare(GetVal(), rhs.GetVal()) == model::CompareResult::kEqual;
+        }
+
+        return column_ == rhs.column_ && tuple_ == rhs.tuple_;
     }
 
-    bool operator!=(ColumnOperand const& rhs) const noexcept {
+    bool operator!=(ColumnOperand const& rhs) const {
         return !(*this == rhs);
     }
```
ol-imorozko, line 92: I think C++20 can do this automatically with `bool operator!=(ColumnOperand const& rhs) = default` assuming `operator==` is provided

src/core/algorithms/dc/model/column_operand.h
```diff
@@ -3,79 +3,136 @@
...
+//
 //  The class distinguishes between operands derived from the first tuple (t) and those
 //  from the second tuple (s) using a boolean flag `is_first_tuple_`, where `true` indicates an
 //  operand from the first tuple (t), and `false` indicates an operand from the second
 //  tuple (s).
 class ColumnOperand {
 private:
     Column const* column_;
-    bool is_first_tuple_;
+    std::optional<dc::Tuple> tuple_;
```
ol-imorozko, line 36: what does `std::optional` here achieve? 
I might be wrong, but I think the only real usage of a `tuple_` is within a `GetTuple` method which **assert** that it has value anyway.

ol-imorozko (changes requested):

src/core/algorithms/dc/model/column_operand.h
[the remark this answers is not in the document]
xJoskiy: I've only resolved for my convenience. I haven't yet uploaded a new version that fixes this issue thus didn't ask for re-review. I think this is a common practice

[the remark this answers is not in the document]
ol-imorozko: Oh, yeah, my bad, sorry :)

src/core/algorithms/dc/model/column_operand.h
xJoskiy: Actually, a `ColumnOperand` may have a `tuple_` variable initialized or may not. If a `ColumnOperand` is a constant value, thus `tuple_` is not initialized. In this case accessing `tuple_` is not allowed. I should definitely write it above

src/core/algorithms/dc/model/column_operand.h
xJoskiy: `kMixed` is used for defining a tuple on `Predicate` since it may have `ColumnPredicate`'s with `kS` and `sT` tuples. We could also create two enums, for `Predicate` and `ColumnPredicate` but I'm not sure, couldn't come up with other solution.

src/python_bindings/dc/bind_dc_verification.cpp
xJoskiy: For some reason it didn't work without it in a previous version, now it doesn't work with it, but works perfectly without conversion 🤷‍♀️

src/core/algorithms/dc/model/column_operand.h
```diff
@@ -3,79 +3,134 @@
...
 #include "model/table/column.h"
 #include "model/table/relation_data.h"
+#include "model/types/type.h"
 
 namespace algos::dc {
 
+// kS and kT are used to tell which tuple is used in ColumnOperand.
+// To describe Predicate tuple also can be used kMixed since
+// ColumnOperand's may have different tuples.
+enum class Tuple { kS, kT, kMixed };
```
ol-imorozko, line 21: Sorry, I don't understand the English here: "To describe Predicate tuple also can be used kMixed".
And still I don't understand what kMixed is. Suppose we have two attributes: A and B, and two tuples: T and S.
We can form different types of predicates: A.t op A.t, A.t op A.s, A.s op A.s, A.s op A.t, A.s op B.t, A.t op B.t and so on.. Also any of the column operand could be a constant.
What does A.mixed means? Can we form predicate A.mixed op B.t, for example?

src/core/algorithms/dc/model/column_operand.h
ol-imorozko: Isn't the `column_` in that case should also be an `optional`? Since a constant column operand is, well, just a constant, it does not have a pair column/tuple

src/core/algorithms/dc/model/dc.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+namespace algos::dc {
+
+bool DC::CheckAllEquality() const {
+    auto check = [](auto const& pred) {
+        return pred.GetOperator() == dc::OperatorType::kEqual and pred.IsCrossTuple() and
+               pred.IsOneColumn();
+    };
+
+    return std::all_of(predicates_.begin(), predicates_.end(), check);
```
ol-imorozko, line 20: you can ignore it:
`std::ranges` allows to get rid of boilerplate:
```suggestion
    return std::ranges::all_of(predicates_, check);
```

src/core/algorithms/dc/model/dc.cpp
```diff
@@ -0,0 +1,104 @@
...
+    for (auto pred = std::next(predicates_.begin()); pred != predicates_.end(); ++pred) {
+        ss << kAnd << pred->ToString();
+    }
+    ss << ')';
+
+    return ss.str();
+}
+
+void DC::ConvertEqualities() {
+    std::vector<dc::Predicate> res;
```
ol-imorozko, line 88: you can ignore it:
`res.reserve(predicates_.size() * 2);` could make sense here

src/core/algorithms/dc/model/dc.cpp
```diff
@@ -0,0 +1,104 @@
...
+        return DCType::kTwoTuples;
+    else
+        return DCType::kMixed;
+}
+
+std::string DC::ToString() const {
+    if (predicates_.empty()) return {};
+
+    static constexpr char const* kNot = "!";
+    static constexpr char const* kAnd = " and ";
```
ol-imorozko, line 75: you can ignore it:
```suggestion
    static constexpr std::string_view kNot = "!";
    static constexpr std::string_view kAnd = " and ";
```
using `std::string_view` is better for non-owning string literals

src/core/algorithms/dc/model/dc.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+    std::stringstream ss;
+    ss << kNot << '(' << predicates_.front().ToString();
+    for (auto pred = std::next(predicates_.begin()); pred != predicates_.end(); ++pred) {
+        ss << kAnd << pred->ToString();
+    }
```
ol-imorozko, line 81: I would do
```suggestion
    for (Predicate pred : predicates_ | std::views::drop(1)) {
        ss << kAnd << pred->ToString();
    }
```
`std::views` is lightweight, no overhead here

src/core/algorithms/dc/parser/dc_parser.cpp
```diff
@@ -0,0 +1,168 @@
...
+namespace mo = model;
+
+namespace {
+
+using namespace algos::dc;
+
+std::unique_ptr<ColumnOperand> GetOperand(ColumnOperand const& operand, std::string str_val) {
+    mo::Type const* type = operand.GetType();
+    return std::make_unique<ColumnOperand>(str_val, type);
+}
```
ol-imorozko, line 22: is it correct to take `std::string` here by value?

src/core/algorithms/dc/parser/dc_parser.cpp
```diff
@@ -0,0 +1,168 @@
...
+                   std::vector<model::TypedColumnData> const& data)
+    : relation_(relation),
+      data_(data),
+      dc_string_(std::move(dc_string)),
+      has_next_predicate_(true),
+      cur_(0) {
+    str_operators_.reserve(str_operators_.size());
+    for (std::pair<frozen::string const, algos::dc::OperatorType> const& op_it :
+         Operator::kStringToOperatorType) {
+        std::string str_op = op_it.first.data();
```
ol-imorozko, line 38: maybe smth like
```suggestion
    for (auto const& [str, _] : Operator::kStringToOperatorType) {
        std::string str_op = str.data();
```
but with name better than just `str`

src/core/algorithms/dc/parser/dc_parser.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+    if (dc_string_.empty()) throw std::invalid_argument("Empty DC is not allowed");
+
+    std::vector<Predicate> res;
+    while (has_next_predicate_) {
+        Predicate pred = GetNextPredicate();
+        res.push_back(pred);
+    }
+
+    return {res};
```
ol-imorozko, line 64: we're copying that vector here. I thing there should be
`return DC{std::move(res)};`

src/core/algorithms/dc/parser/dc_parser.cpp
```diff
@@ -0,0 +1,168 @@
...
+    dc_string_.erase(dc_string_.begin());
+    dc_string_.erase(std::prev(dc_string_.end()));
+
+    if (dc_string_.empty()) throw std::invalid_argument("Empty DC is not allowed");
+
+    std::vector<Predicate> res;
+    while (has_next_predicate_) {
+        Predicate pred = GetNextPredicate();
+        res.push_back(pred);
+    }
```
ol-imorozko, line 62: You can ignore it.

IMHO `res.push_back(GetNextPredicate());` is better.

src/core/algorithms/dc/parser/dc_parser.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+    return {res};
+}
+
+Predicate DCParser::GetNextPredicate() {
+    size_t ind = dc_string_.find(sep_, cur_);
+    if (ind == std::string::npos) has_next_predicate_ = false;
+
+    std::string str_pred = dc_string_.substr(cur_, ind - cur_);
```
ol-imorozko, line 71: Is it correct? If the `ind` is equal to `std::string::npos`, can we subtract `cur_` from it?
Seems like a bug

src/core/algorithms/dc/parser/dc_parser.cpp
```diff
@@ -0,0 +1,168 @@
...
+    for (std::string const& op : str_operators_) {
+        std::string spaced_op = ' ' + op + ' ';
+        ind = pred.find(spaced_op);
+        if (ind != std::string::npos) {
+            str_operator = op;
+            break;
+        }
+    }
+
+    if (ind == std::string::npos) throw std::invalid_argument("Missing permissible operator");
```
ol-imorozko, line 89: This can be simplified:
```suggestion
auto some_name = [&pred](const std::string& op) {
    std::string spaced_op = ' ' + op + ' ';
    return pred.find(spaced_op) != std::string::npos;
};

auto it = std::ranges::find_if(str_operators_, some_name);
if (it == str_operators_.end()) {
    throw std::invalid_argument("Missing permissible operator");
}
```
we can extract the lambda to a variable, it will make for the

src/core/algorithms/dc/parser/dc_parser.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+    size_t right_start = ind + str_operator.size() + 1;
+    auto str_left_operand = std::string(pred, 0, ind);
+    auto str_right_operand = std::string(pred, right_start, pred.size() - right_start + 1);
```
ol-imorozko, line 93: I think that's better:
```cpp
    std::string str_left_operand = pred.substr(0, ind);
    std::string str_right_operand = pred.substr(right_start);
```

src/core/algorithms/dc/parser/dc_parser.cpp
ol-imorozko: Why do we need uniq pointers here? I see that the only usage is in `DCParser::GetOperands`, which uses it to get uniq pointers, but returns a pair of dereferenced ones aka a pair of `ColumnOperand` object.
Can't we just do
```cpp
ColumnOperand GetOperand(const ColumnOperand& operand, const std::string& str_val) {
    return ColumnOperand(str_val, operand.GetType());
}
```
It will be copy-elided.

src/core/algorithms/dc/parser/dc_parser.cpp
```diff
@@ -0,0 +1,168 @@
...
+    if (!left_op and !right_op)
+        throw std::invalid_argument("Pure constant predicate is not allowed");
+
+    if (!left_op) {
+        left_op = GetOperand(*right_op.get(), str_left_op);
+    } else if (right_op == nullptr) {
+        right_op = GetOperand(*left_op.get(), str_right_op);
+    }
+
+    return {std::move(*left_op), std::move(*right_op)};
```
ol-imorozko, line 128: Why do we need uniq pointers here?
Maybe just
```cpp
    std::optional<ColumnOperand> left_op, right_op;

    if (IsVarOperand(str_left_op)) {
        RemoveEscaping(str_left_op);
        left_op = ConvertToVariableOperand(str_left_op);
    }

    if (IsVarOperand(str_right_op)) {
        RemoveEscaping(str_right_op);
        right_op = ConvertToVariableOperand(str_right_op);
    }

    if (!left_op && !right_op) {
        throw std::invalid_argument("Pure constant predicate is not allowed");
    }

    if (!left_op) {
        left_op = GetOperand(*right_op, str_left_op);
    } else if (!right_op) {
        right_op = GetOperand(*left_op, str_right_op);
    }

    return {std::move(*left_op), std::move(*right_op)};
```

ol-imorozko: we can track whether left/right op have been initizlized with `std::optional`

src/core/algorithms/dc/parser/dc_parser.cpp
```diff
@@ -0,0 +1,168 @@
...
+}
+
+void DCParser::RemoveEscaping(std::string& str) const {
+    size_t ind = str.find_first_of('\\');
+    if (ind != std::string::npos) str.erase(ind, 1);
+}
+
+bool DCParser::IsVarOperand(std::string op) const {
+    return op.find("t.") == 0 or op.find("s.") == 0;
```
ol-imorozko, line 137:
```suggestion
    return op.starts_with("t.") || op.starts_with("s.");
```
I love C++20, didn't know we can do this :)

src/core/algorithms/dc/parser/dc_parser.cpp
```diff
@@ -0,0 +1,168 @@
...
+}
+
+ColumnOperand DCParser::ConvertToVariableOperand(std::string operand) const {
+    Column* column = nullptr;
+    dc::Tuple tuple = operand.front() == 't' ? dc::Tuple::kT : dc::Tuple::kS;
+    std::string name(operand.begin() + 2, operand.end());
```
ol-imorozko, line 143:
```suggestion
    std::string name= operand.substr(2);
```

src/core/algorithms/dc/parser/dc_parser.cpp
```diff
@@ -0,0 +1,168 @@
...
+    std::vector<std::unique_ptr<Column>> const& cols = relation_->GetSchema()->GetColumns();
+    if (!cols.front()->GetName().empty()) {  // Has header
+        for (auto const& col : cols) {
+            if (name == col->GetName()) {
+                column = col.get();
+                break;
+            }
+        }
```
ol-imorozko, line 151:
```cpp
        auto it = std::ranges::find_if(cols, [&name](const std::unique_ptr<Column>& col) {
            return name == col->GetName();
        });
        if (it != cols.end()) {
            column = it->get();
        }
```
extract lambda to variable if you want

src/core/algorithms/dc/parser/dc_parser.cpp
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    if (column == nullptr) {
+        try {
+            std::string str_ind(operand.begin() + 2, operand.end());
+            size_t ind = static_cast<mo::ColumnIndex>(std::stoi(str_ind));
```
ol-imorozko, line 157: do we need `str_ind`? I mean we already have `name`

src/core/algorithms/dc/parser/dc_parser.h
```diff
@@ -0,0 +1,39 @@
...
+#include "model/table/typed_column_data.h"
+
+namespace algos::dc {
+
+class DCParser {
+private:
+    std::vector<std::string> str_operators_;
+    ColumnLayoutRelationData const* relation_;
+    std::vector<model::TypedColumnData> const& data_;
+    std::string const sep_ = " and ";
```
ol-imorozko, line 18: mb `std::string_view`? I might be wrong, ignore if I am

ol-imorozko (changes requested):

src/tests/all_csv_configs.cpp
```diff
@@ -132,4 +132,6 @@ CSVConfig const kTestDC = CreateCsvConfig("TestDC.csv", ',', true);
 CSVConfig const kTestDC1 = CreateCsvConfig("TestDC1.csv", ',', true);
 CSVConfig const kTestDC2 = CreateCsvConfig("TestDC2.csv", ',', true);
 CSVConfig const kTestDC3 = CreateCsvConfig("TestDC3.csv", ',', true);
+CSVConfig const kTestDC4 = CreateCsvConfig("TestDC4.csv", ',', true);
+CSVConfig const kTestDC5 = CreateCsvConfig("TestDC5.csv", ',', true);
```
vs9h, line 136:
```suggestion
CSVConfig const kTestDC5 = CreateCsvConfig("TestDC5.csv", ',', false);
```

src/core/algorithms/dc/model/dc.cpp
```diff
@@ -0,0 +1,104 @@
...
+            res.emplace_back(dc::OperatorType::kGreaterEqual, left, right);
+        } else {
+            res.emplace_back(pred);
+        }
+    }
+
+    predicates_ = std::move(res);
+}
+
+}  // namespace algos::dc
```
vs9h, line 104: add a line break (in [test_input_data/TestDC5.csv](https://github.com/Desbordante/desbordante-core/pull/510/files#diff-b9d6044bade9ca7845eceef24327c0e9f365538e110d11742fd1e753ea404101) too)

src/core/algorithms/dc/verifier/dc_verifier.cpp
```diff
@@ -60,42 +65,57 @@ void DCVerifier::LoadDataInternal() {
 }
 
 unsigned long long int DCVerifier::ExecuteInternal() {
+    // Is it supposed to be used like that
+    // if we run the algorithms several times?
+    ResetState();
```
BUYT-1, line 70: This is run by `Algorithm` before `ExecuteInternal` already

src/core/algorithms/dc/model/dc.cpp
ol-imorozko: yes, it's just that cpp core guidelines advises to use `std::string_view` for such use of the string literals, but that's not necessary

src/core/algorithms/dc/parser/dc_parser.cpp
```diff
@@ -0,0 +1,174 @@
...
+    if (ind != std::string::npos) str.erase(ind, 1);
+}
+
+bool DCParser::IsVarOperand(std::string op) const {
+    return op.starts_with("t.") or op.starts_with("s.");
+}
+
+ColumnOperand DCParser::ConvertToVariableOperand(std::string const& operand) const {
+    dc::Tuple tuple = operand.front() == 't' ? dc::Tuple::kT : dc::Tuple::kS;
+    std::string name = operand.substr(2);
```
ol-imorozko, line 149: I think we can make this a `string_view` to avoid extra copies
```suggestion
    std::string_view name = operand.substr(2);
```

src/core/algorithms/dc/verifier/dc_verifier.cpp
```diff
@@ -153,51 +167,69 @@ bool DCVerifier::VerifyMixed(dc::DC const& dc) {
...
+
     for (size_t i = 0; i < data_.front().GetNumRows(); ++i) {
+        std::unordered_set<Point, Point::Hasher> res_points;
         if (ContainsNullOrEmpty(all_cols, i)) continue;
+        auto process =
+                std::bind(&DCVerifier::ProcessMixed, this, std::placeholders::_1,
+                          std::placeholders::_2, var_preds, all_cols, i, std::ref(res_points));
+
+        boost::optional<Point> t_point = process(s_tree, t_predicates);
+        boost::optional<Point> s_point = process(t_tree, s_predicates);
```
ol-imorozko, line 183: why not `std::optonal`?

ol-imorozko (approved):

src/core/algorithms/dc/verifier/dc_verifier.cpp
xJoskiy: Oops, forgot to change this one

ol-imorozko:
To be honest I did not give a thorough look at dc_parser.cpp, I'll do that by tomorrow evening

PR #517, opened 2025-01-31
Use libstdc++ implementation of std::bitset if other standard library is used

### LINKED ISSUE #516: Some tests work very slow on Clang in Debug mode
There's a test named "LightDatasetConsistentHash", which takes a while with all FD algorithms, but becames inadequately slow with FDep algo on LLVM (both Linux and macOS, LLVM and Apple Clang) in Debug mode without sanitizers. One test runs about an hour. 😱

Image

Image

Sanitizers somehow fix it, but run time is still no so good (about 10 minutes):

Image

On Linux with GCC in Debug mode it takes about 5 minutes:

Image

p-senichenkov:
Copy `libstdc++` implementation of `std::bitset` to use it when project is compiled with `libc++` (and possibly other STL implementations). Fixes #516.

Here are macOS Apple Clang run times after this fix:
Снимок экрана_20250201_214626
Снимок экрана_20250201_214756


src/core/model/types/bitset.h
```diff
@@ -0,0 +1,1190 @@
...
+// For hash to be named properly
+// NOLINTBEGIN(readability-identifier-naming)
+/// @note This hash is equal to hash of @b libc++ std::bitset with the same data.
+/// It's @b not equal to hash of @b libstdc++ std::bitset with the same data.
+template <size_t NumBits>
+struct hash<model::bitset_impl::BitsetImpl<NumBits>> {
+    size_t operator()(model::bitset_impl::BitsetImpl<NumBits> const& b) const noexcept {
+        return b.HashCode();
+    }
+};
```
ol-imorozko, line 1186: It is undefined behavior to add declarations or definitions to namespace std or to any namespace nested within std.

We can add template specializations, but this should be done without opening namespace std with `namespace` keyword:

```cpp
template<>
struct std::hash<MyType>
{
    std::size_t operator()(const MyType& t) const { return t.hash(); }
};
```

ol-imorozko (changes requested):

src/tests/test_bitset.cpp
```diff
@@ -0,0 +1,476 @@
...
+REGISTER_TYPED_TEST_SUITE_P(BitsetTest, DefaultConstructor, ULongConstructor, ULLongConstructor,
+                            StringConstructor, CharArrayConstructor, OperatorsEqNeq,
+                            SubscriptOperatorRead, SubscriptOperatorWrite, Test, TestInvalidIndex,
+                            All, Any, None, Count, kSize, AndAssign, OrAssign, XorAssign,
+                            NotOperator, LeftShift, LeftShiftAssign, RightShift, RightShiftAssign,
+                            SetAll, Set, SetInvalidIndex, ResetAll, Reset, ResetInvalidIndex,
+                            FlipAll, Flip, FlipInvalidIndex, ToString, ToStringCustomChars, ToULong,
+                            ToULLong, And, Or, Xor, ToStream, FromStream, FindFirst, FindNext);
+
+using BitsetSizes = ::testing::Types<BitsetSizeParam<8>, BitsetSizeParam<128>>;
```
Vdaleke, line 459: Why exactly these values? Are they used in our algorithms? Is 64 also used? Maybe we should put all our values ​​from all used bitsets there?

Vdaleke (changes requested):
Looks ok to me, but I can't say with much certainty that using Apache and GPL-3 with our AGPL is fully compatible, as I'm not a lawyer. But if you trust the information on the Internet, then everything seems ok. For example:

https://opensource.stackexchange.com/a/8271

https://en.wikipedia.org/wiki/GNU_Affero_General_Public_License#:~:text=By%20contrast%2C%20the%20GPLv3,combinations.%5B2%5D

src/tests/test_bitset.cpp
p-senichenkov: These are just convenient values for testing. 8-bit is a single-word bitset (with most zeros), 128-bit contains at least 2 words on any system. I believe that 64-bit will work fine too, considering these two work.

Vdaleke (changes requested):
LGTM, but squash commits into 1, please

Vdaleke (approved):
LGTM, but I can't guarantee that there will be no license violations, since I'm not a lawyer

chernishev:
> LGTM, but I can't guarantee that there will be no license violations, since I'm not a lawyer

Discussed this with Kirill, should be fine.

ol-imorozko (approved):

PR #520, opened 2025-02-01
Disable slow test to avoid #516, add concurrency-groups for workflows

Vdaleke:
- Make gtest_filter to exclude slow test on not-gcc toolsets in Debug mode, to avoid #516 
- Refactor build-matrix in core-tests
- Disable old actions that stay in one PR or github-reference(branch/tag)
- Refactor cibuildwheel job


BUYT-1 (approved):
looks okay

PR #525, opened 2025-02-11
Add tests for examples

DaniilGoncharov:
Add python script for testing examples (except sfd). Method get_fds() now returns sorted list of fds


examples/test_examples/requirements.txt
polyntsov, line 1: It's a bit weird that we have a separate `requirements.txt` for examples tests and don't have `requirements.txt` for examples themselves at all. Some subset of these requirements will intersect and it'll be hard to keep it consistent.

Ideally I think we should have one list of requirements to build desbordante, one list of requirements for all examples, and requirements for tests, which have `desbordante` package as a dependency.

This is important, but probably can be implemented in another PR

src/core/algorithms/fd/fd_algorithm.cpp
```diff
@@ -27,6 +27,20 @@ void FDAlgorithm::ResetState() {
...
+        if (l_fd.GetLhs().GetArity() != r_fd.GetLhs().GetArity()) {
+            return l_fd.GetLhs().GetArity() < r_fd.GetLhs().GetArity();
+        }
+        if (l_fd.GetLhs() != r_fd.GetLhs()) {
+            return l_fd.GetLhs() < r_fd.GetLhs();
+        }
+        return l_fd.GetRhsIndex() < r_fd.GetRhsIndex();
+    });
+
+    return fd_collection_.AsList();
```
polyntsov, line 41: nit:
I would use a variable for `fd_collection_.AsList()` to avoid repeating it. Also it will be a bit easier to read, since it'll be obvious that it's a lvalue ref
```cpp
std::list<FD>& fd_collection = fd_collection_.AsList();
fd_collection.sort();
return fd_collection;
```

examples/test_examples/test_examples_pytest.py
```diff
@@ -0,0 +1,89 @@
+import os
+import subprocess
+import pytest
+import snapshottest
+
+
+TEST_CASES = [
```
polyntsov, line 7: Maybe we can collect testcases dynamically to avoid changing this file everytime when new example is added

PR #528, opened 2025-02-22
feat: use Ninja instead make, bump CMake to 3.25

Vdaleke:
Change build.sh script:
- use Ninja instead of make
- add option for install dependencies only

Set CMake min version to 3.25 due to using of
`SYSTEM` option in add_subdirectory()

Update README.md:
- update Ubuntu dependencies installation guide
- fix building steps

Fix some typos and outdated links

### DIFF
--- a/.github/composite-actions/install-dependencies/action.yml
+++ b/.github/composite-actions/install-dependencies/action.yml
@@ -56,11 +56,11 @@ runs:
     - name: Install build tools using apt
       run: |
         sudo apt-get update -y
-        sudo apt-get install cmake build-essential -y
+        sudo apt install cmake ninja-build python3 python3-venv -y
       shell: bash
       if: env.OS == 'ubuntu'
     - name: Install build tools using brew
-      run: brew install make cmake --formula
+      run: brew install cmake ninja python3 --formula
       shell: bash
       if: env.OS == 'macos'
 
--- a/README.md
+++ b/README.md
@@ -253,7 +253,7 @@ The following instructions were tested on Ubuntu 20.04+ LTS and macOS Sonoma 14.
 Prior to cloning the repository and attempting to build the project, ensure that you have the following software:
 
 - GNU GCC, version 10+, LLVM Clang, version 16+, or Apple Clang, version 15+
-- CMake, version 3.15+
+- CMake, version 3.25+
 - Boost library built with compiler you're going to use (GCC or Clang), version 1.85.0+
 
 To use test datasets you will need:
@@ -264,13 +264,26 @@ Instructions for other supported compilers can be found in [Desbordante wiki](ht
 
 #### Ubuntu dependencies installation (GCC)
 
-Run the following commands:
+For Ubuntu versions earlier than 24.04, you need to add the Kitware APT repository to your system 
+by following their [official guide](https://apt.kitware.com) to install the latest version of CMake.
+
+Then run the following commands:
 ```sh 
-sudo apt install g++ cmake libboost-all-dev git-lfs python3
+sudo apt update && sudo apt upgrade
+sudo apt install g++ cmake ninja-build libboost-all-dev git-lfs python3 python3-venv
 export CXX=g++
 ```
 The last line sets g++ as CMake compiler in your terminal session.
-You can also add them to the end of `~/.profile` to set this by default in all sessions.
+You can also set it by default in all sessions: `echo 'export CXX=g++' >> ~/.profile`
+
+For Ubuntu 24.04 and above, you can skip to the build steps. For older versions the Ubuntu APT repository
+might not have a compatible version of Boost, so you'll need to install it manually:
+```sh
+wget https://archives.boost.io/release/1.87.0/source/boost_1_87_0.tar.gz
+tar xzvf boost_1_87_0.tar.gz
+cd boost_1_87_0 && ./bootstrap.sh
+sudo ./b2 install --prefix=/usr/
+```
 
 #### macOS dependencies installation (Apple Clang)
 
@@ -319,12 +332,12 @@ In order to build tests, pull the test datasets using the following command:
 ```
 then build the tests themselves:
 ```sh
-./build.sh -j$(nproc)
+./build.sh
 ```
 
 The Python module can be built by providing the `--pybind` switch:
 ```sh
-./build.sh --pybind -j$(nproc)
+./build.sh --pybind 
 ```
 
 See `./build.sh --help` for more available options.
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 [![Downloads](https://static.pepy.tech/badge/desbordante/month)](https://pepy.tech/project/desbordante)
 
 <p>
-   <img src="https://github.com/Mstrutov/Desbordante/assets/88928096/d687809b-5a3b-420e-a192-a1a2b6697b2a"/>
+   <img src="https://github.com/Desbordante/desbordante-core/assets/88928096/d687809b-5a3b-420e-a192-a1a2b6697b2a"/>
 </p>
 
 # General
@@ -304,9 +304,9 @@ You can also add them to the end of `~/.profile` to set this by default in all s
 Clone the repository, change the current directory to the project directory and run the following commands:
 
 ```bash
-./build.sh
-python3 -m venv venv
-source venv/bin/activate
+./build.sh --deps-only
+python3 -m venv .venv
+source .venv/bin/activate
 python3 -m pip install .
 ```
 
@@ -329,7 +329,7 @@ The Python module can be built by providing the `--pybind` switch:
 
 See `./build.sh --help` for more available options.
 
-The `./build.sh` script generates the following file structure in `/path/to/Desbordante/build/target`:
+The `./build.sh` script generates the following file structure in `/path/to/desbordante-core/build/target`:
 ```
 ├───input_data
 │   └───some-sample-csv\'s.csv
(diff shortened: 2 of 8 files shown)


.github/composite-actions/install-dependencies/action.yml
```diff
@@ -56,11 +56,11 @@ runs:
     - name: Install build tools using apt
       run: |
         sudo apt-get update -y
-        sudo apt-get install cmake build-essential -y
+        sudo apt-get install cmake ninja-build build-essential -y
       shell: bash
       if: env.OS == 'ubuntu'
     - name: Install build tools using brew
-      run: brew install make cmake --formula
+      run: brew install make cmake ninja --formula
```
p-senichenkov, line 63: `make` is not needed now

README.md
```diff
@@ -264,13 +264,26 @@ Instructions for other supported compilers can be found in [Desbordante wiki](ht
...
 #### Ubuntu dependencies installation (GCC)
 
-Run the following commands:
+For Ubuntu versions earlier than 24.04, you need to add the Kitware APT repository to your system 
+by following their [official guide](https://apt.kitware.com) to install the latest version of CMake.
+
+Then run the following commands:
 ```sh 
-sudo apt install g++ cmake libboost-all-dev git-lfs python3
+sudo apt update
```
p-senichenkov, line 272: Isn't `upgrade` also needed?

README.md
```diff
@@ -304,9 +317,9 @@ You can also add them to the end of `~/.profile` to set this by default in all s
 Clone the repository, change the current directory to the project directory and run the following commands:
 
 ```bash
-./build.sh
-python3 -m venv venv
-source venv/bin/activate
+./build.sh --deps-only
+python3 -m venv .venv
```
p-senichenkov, line 321: `.venv` is also needed to be added to `.gitignore`

build.sh
```diff
@@ -35,13 +38,10 @@ for i in "$@"
...
             NO_UNPACK=true
             ;;
-        -j*|--jobs=*) # Allow N jobs at once
-            JOBS_OPTION=$i
-            ;;
         -d|--debug) # Set debug build type
             DEBUG_MODE=true
             ;;
-        # It's a nightmare, we should use getopts for args parsing or even use something else
+        # It\'s a nightmare, we should use getopts for args parsing or even use something else
```
p-senichenkov, line 44: Is it necessary to escape quotes in comments?

Vdaleke: I had a bash static analyzer error in vs code without escaping this symbol. I think it's a bug, but it's easier to add escaping here and not figure it out.

p-senichenkov: I think, in this case it's better not to use quotes: `It is a nightmare`. At least, it doesn't look like a typo.

build.sh
```diff
@@ -9,10 +9,10 @@ Usage: ./build.sh [options]
 
 Possible options:
   -h,         --help                  Display help
+              --deps-only             Install only necessary dependencies
```
p-senichenkov, line 12: As far as I understand, `--deps-only` only installs dependencies and doesn't do build. In that case this description is not so clear. I think something like `Only install dependencies (don't build)` will be better

p-senichenkov (approved):
LGTM, only need to check description for `--deps-only` option

README.md
```diff
@@ -370,7 +383,7 @@ Error downloading object: datasets/datasets.zip (2085458): Smudge error: Error d
 ```
 delete the already cloned version, set `GIT_LFS_SKIP_SMUDGE=1` environment variable and clone the repo again:
 ```sh
-GIT_LFS_SKIP_SMUDGE=1 git clone git@github.com:Mstrutov/Desbordante.git
+GIT_LFS_SKIP_SMUDGE=1 git clone https://github.com/Desbordante/desbordante-core.git
```
vs9h, line 386: Let's also update the link to the image at the beginning of the `README.md` and `README_PYPI.md`
```
https://github.com/Mstrutov/Desbordante/assets/88928096/d687809b-5a3b-420e-a192-a1a2b6697b2a
# to
https://github.com/Desbordante/desbordante-core/assets/88928096/d687809b-5a3b-420e-a192-a1a2b6697b2a
```

build.sh
```diff
@@ -9,10 +9,10 @@ Usage: ./build.sh [options]
 
 Possible options:
   -h,         --help                  Display help
+              --deps-only             Only install dependencies (don't build)
```
vs9h, line 12:
```suggestion
  -o,         --deps-only             Install dependencies only (don't build)
```

1) Let's add short option alias for uniformity with other options

2) Placing "only" at the end aligns with natural English phrasing. Or we can write something like "Install dependencies without building".

3) And don't forget to sync description with description on line 29.

README.md
```diff
@@ -253,7 +253,7 @@ The following instructions were tested on Ubuntu 20.04+ LTS and macOS Sonoma 14.
 Prior to cloning the repository and attempting to build the project, ensure that you have the following software:
 
 - GNU GCC, version 10+, LLVM Clang, version 16+, or Apple Clang, version 15+
-- CMake, version 3.15+
+- CMake, version 3.25+
```
vs9h, line 256: Why do we want to bump the version for the cmake? I see that this option was added in this version. But why is it critical in general? Otherwise we get warnings?

vs9h (commented):
Overall it looks good.

It would be better to split the changes into commits, there are a lot of changes here that could easily be split into a few small commits. And it would be easier to review it consistently. Please split it into multiple commits if you can.

And we do not use such prefixes (`feat: `) in commit names

p-senichenkov:
I've just noticed that ninja supresses colors in compiler messages. To work it around CMake [environment variable](https://cmake.org/cmake/help/latest/envvar/CMAKE_COLOR_DIAGNOSTICS.html) `CMAKE_COLOR_DIAGNOSTICS` can be set to `ON` in `build.sh` or [CMake variable](https://cmake.org/cmake/help/latest/variable/CMAKE_COLOR_DIAGNOSTICS.html) with the same name can be set somewhere in `CMakeLists.txt`. It's said in [ninja FAQ](https://github.com/ninja-build/ninja/wiki/FAQ#why-does-my-program-with-colored-output-not-have-color-under-ninja) that it's safe and ninja will escape colors if needed.

README.md
Vdaleke: > Otherwise we get warnings?

The project could not be built at all.
image

build.sh
Vdaleke: I didn't want to add the short option o because in many unix tools it is fixed for output and can cause confusion

Vdaleke:
> It would be better to split the changes into commits, there are a lot of changes here that could easily be split into a few small commits. And it would be easier to review it consistently. Please split it into multiple commits if you can.

I agree that it is easier to review, but that is where the significant advantages end for me. When blaming, it generally does not matter what other changes were contained in the commit. And in other situations, I do not really understand why it is necessary to look at the commit history and meaningfully understand what was in what

Considering that we try to maintain a beautiful commit history, fixing the history with several commits at once on the PR turns into a nightmare. My usual strategy is to do a hard reset and then cherry-pick from the remote branch one by one with sequential ammend. This takes a decent amount of time and causes huge inconvenience if you need to fix something small in meaning related to one change.

I would generally make the default merge strategy in the project settings squash in the main. This is convenient for developing and writing a commit description only in the PR itself.

> And we do not use such prefixes (feat: ) in commit names

And here I also feel sorry that we don't use it, it seems to me that on average the commits in the branch are ugly. Here the prefix makes the main meaning of the changes clear and it's easy to see if something was added or fixed. And the main advantage is that there is good documentation about this commit convention https://www.conventionalcommits.org/en/v1.0.0/ and if everyone starts to adhere to it life will become better.

README.md
vs9h: It would be good to correct this in `README_PYPI.md` too

p-senichenkov (approved):

vs9h:
> I agree that it is easier to review, but that is where the significant advantages end for me. When blaming, it generally does not matter what other changes were contained in the commit. And in other situations, I do not really understand why it is necessary to look at the commit history and meaningfully understand what was in what

I understand that reviewing is just one aspect, but my concern is broader. Having well-structured commits makes it much easier to understand their content just by looking at the commit message. When a commit contains unrelated changes, it becomes harder to grasp its purpose at a glance.

Additionally, if I ever need to study a specific change, it's much more convenient to see only the relevant modifications instead of sorting through a mix of unrelated updates. This also benefits anyone who, at some point, wants to analyze which parts of the code were affected by a particular change.

> Considering that we try to maintain a beautiful commit history, fixing the history with several commits at once on the PR turns into a nightmare. My usual strategy is to do a hard reset and then cherry-pick from the remote branch one by one with sequential ammend. This takes a decent amount of time and causes huge inconvenience if you need to fix something small in meaning related to one change.
> 
> I would generally make the default merge strategy in the project settings to **squash** for the main. This is convenient for developing and writing a commit description only in the PR itself.

I'm not sure I fully understand the issue or why it takes so much time. You can keep these commits sequentially in your remote branch and cherry-pick this range of commits into another branch for the pull request. It would be unfortunate to sacrifice a meaningful commit history for this reason.

I also strongly disagree with squashing changes, as it makes the commit history impossible to study properly. Having all the details in the pull request might help during the review, but later, when someone examines the history, they are unlikely to open the PR just to check the description. This information should be preserved in the commits themselves.

> And here I also feel sorry that we don't use it, it seems to me that on average the commits in the branch are ugly. Here the prefix makes the main meaning of the changes clear and it's easy to see if something was added or fixed. And the main advantage is that there is good documentation about this commit convention https://www.conventionalcommits.org/en/v1.0.0/ and if everyone starts to adhere to it life will become better.

In theory, we could bring this up for discussion, then add it to the [wiki](https://github.com/Desbordante/desbordante-core/wiki/Development#%D0%BE%D1%84%D0%BE%D1%80%D0%BC%D0%BB%D0%B5%D0%BD%D0%B8%D0%B5-%D0%BA%D0%BE%D0%BC%D0%BC%D0%B8%D1%82%D0%BE%D0%B2) and force people to use the new naming rules. But it would have to be explicitly discussed and agreed upon.

vs9h (approved):
If possible, fix the last small comment about updating the link.

Looks good

PR #530, opened 2025-02-24
Ar validation

Tydik42:
### New AR Verification Functionality:

*   **`core/algorithms/association_rules/ar_verifier/ar_verifier.h`**: Introduced the `ARVerifier` class to handle AR verification, including methods for loading data, executing the verification, and retrieving various statistics.
*   **`core/algorithms/association_rules/ar_verifier/ar_stats_calculator.h`**: Added the `ARStatsCalculator` class to compute actual support, confidence, and analyze violating rule clusters.
*   **`core/algorithms/association_rules/model/rule_coverage.h`**: Defined the `RuleCoverage` class to calculate the proportion of a rule within a transaction, used for cluster prioritization.
*   **`core/algorithms/association_rules/ar_verifier/enums.h`**: Added `ClusterPriority` enums to define different strategies for cluster analysis.

### Core Refactoring:

*   **`core/config/`**: Refactored shared options (`min_support`, `min_confidence`) into reusable `CommonOption` objects with strict `[0, 1]` validation to reduce code duplication between the AR miner and verifier.
*   **`core/model/transactional_data.h`**: Improved `TransactionalData` creation by introducing a dedicated `Params` struct.


src/tests/test_ar_verifier.cpp
```diff
@@ -0,0 +1,64 @@
...
+        ARVerifyingParams(kRulesBook, {"Eggs"}, {"Milk"}, 0.3, 0.1),
+        ARVerifyingParams(kRulesBook, {"Yogurt"}, {"Eggs"}, 0.2, 0.1),
+        ARVerifyingParams(kRulesBook, {"Yogurt"}, {"Milk"}, 0.2, 0.1)
+    ));
+
+INSTANTIATE_TEST_SUITE_P(
+  ARVerifierRowsDataset, TestARVerifying,
+  ::testing::Values(
+      // test cases tabular
+     ARVerifyingParams(kRulesKaggleRows, {"MILK"}, {"BREAD"}, 0.2, 0.1, algos::InputFormat::tabular, 0, 0, 1),
```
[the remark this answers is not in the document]
vs9h: Also, I don't like passing a large number of numbers, which makes the tests unreadable. It is necessary to compare which number is in which position. This can be avoided by adding an auxiliary structure and using aggregate initialization.

Here is a small sketch of the solution:
```cpp
struct ARVerifyingParams {
    algos::StdParamsMap params;

    struct TabularCSV {
        CSVConfig csv_config;
        unsigned int tid_col;
        unsigned int item_col;
    };

    struct Config {
        std::list<std::string> left;
        std::list<std::string> right;
        double min_sup;
        double min_conf;
    };

    ARVerifyingParams(TabularCSV const& table, Config config)
      : params({{onam::kCsvConfig, table.csv_config},
                {onam::kInputFormat, +algos::InputFormat::singular},
                {onam::kMinimumSupport, config.min_sup},
                {onam::kMinimumConfidence, config.min_conf},
                {onam::kARuleLeft, std::move(config.left)},
                {onam::kARuleRight, std::move(config.right)},
                {onam::kTIdColumnIndex, table.tid_col},
                {onam::kItemColumnIndex, table.item_col}}) {}

};
```

And write here something like this:
```cpp
     ARVerifyingParams(kRulesKaggleRowsSingular, {.left = {"MILK"}, .right ={"BREAD"}, .min_sup = 0.2, .min_conf = 0.1})
```

This will be much more readable and visual.

If you need more examples, I've done something similar before (`test_ind_verifier.c`, you can see here how the code is organized and do something similar) ​​and I've definitely promoted similar solutions for other primitives a few times (I don't remember which ones specifically).

vs9h (changes requested):
I partially looked at the pull request, I'll look at some parts in more detail later.

src/core/algorithms/association_rules/ar_verifier/ar_verifier.cpp
```diff
@@ -0,0 +1,112 @@
...
+    }
+    if (transactional_data_->GetNumTransactions() == 0) {
+        throw std::runtime_error("Got an empty dataset: AR verifying is meaningless.");
+    }
+}
+
+unsigned long long ARVerifier::ExecuteInternal() {
+    std::vector<std::string> const& item_names_map = transactional_data_->GetItemUniverse();
+
+    std::vector<unsigned> ar_left_id;
```
vs9h, line 67: It seems more correct to use `model::ColumnIndex`. If this is true, then it should be here and in those places where we refer to the index inside the rule.

src/core/algorithms/association_rules/ar_verifier/ar_verifier.h
```diff
@@ -0,0 +1,83 @@
...
+    /* input options */
+    config::InputTable input_table_;
+    InputFormat input_format_ = InputFormat::singular;
+    unsigned int tid_column_index_;
+    unsigned int item_column_index_;
+    bool first_column_tid_;
+
+    std::shared_ptr<model::TransactionalData> transactional_data_;
+    std::list<std::string> string_rule_left_;
+    std::list<std::string> string_rule_right_;
```
vs9h, line 22: Why do we want to use a `std::list<>` here?

src/core/algorithms/association_rules/ar_verifier/ar_stats_calculator.cpp
```diff
@@ -0,0 +1,98 @@
...
+        if (is_right_full) return ClusterPriority::partial_left_full_right;
+        return ClusterPriority::partial_left_partial_right;
+    }
+    return ClusterPriority::partial_left_no_right;
+}
+
+void ARStatsCalculator::CalculateRuleCoverageCoefficients() {
+    for (auto const& [id, itemset] : data_->GetTransactions()) {
+        double coverage_left = CalculateTransactionCoverage(itemset.GetItemsIDs(), rule_.left);
+        double coverage_right = CalculateTransactionCoverage(itemset.GetItemsIDs(), rule_.right);
```
vs9h, line 51: nit: I see that you use `const` for local variables. But in some places (like this one) you don't write them. I suggest adding them in all places for consistency here and in similar places. (This is not necessary)

vs9h (changes requested):

ol-imorozko:
Take a look at https://github.com/Desbordante/desbordante-core/wiki/Development#оформление-коммитов -- there are guidelines on how we aim to write commit messages. 
I can see that your commits do not use imperative mood (which is an easy fix) and there aren't any descriptions on any of the commit messages. That's not necessarily a bad thing since usually there's no description needed anyway, but  that is something to note in the future

ol-imorozko:
Also I would suggest to avoid commits like "fix things". If a commit fixes something that was introduced in the previous comment, it's better to just fixup those changes. This way we will be left with one good commit instead of a flawed commit and a commit that fixes that.
Again, there are cases where this is acceptable. For example, a commit like `Apply clang-format` is generally acceptable.
But commits like [fix(AR): fixed bindings after merging main/origin branch](https://github.com/Desbordante/desbordante-core/pull/530/commits/e61fced1b86c66ee987ac391545c4a76d86bcc0f)
[e61fced](https://github.com/Desbordante/desbordante-core/pull/530/commits/e61fced1b86c66ee987ac391545c4a76d86bcc0f)
and
[fix(AR): fixed bindings (again)](https://github.com/Desbordante/desbordante-core/pull/530/commits/23a3843aedf3b578395ab25c2d870b0335bb83b5)
are not -- you should just write one good commit that adds bindings

src/tests/unit/test_ar_verifier.cpp
vs9h, line 1: Should we move this file to `tests/unit` directory?

src/tests/test_ar_verifier.cpp
```diff
@@ -0,0 +1,142 @@
...
+};
+
+struct RuleConfig {
+    std::list<std::string> left;
+    std::list<std::string> right;
+    double min_sup;
+    double min_conf;
+};
+
+struct ARVerifyingParams {
```
vs9h: Let's rename 'Verifying' to 'Verifier' here and in test names

src/core/algorithms/association_rules/ar_verifier/ar_verifier.cpp
```diff
@@ -0,0 +1,96 @@
...
+void ARVerifier::RegisterOptions() {
+    DESBORDANTE_OPTION_USING;
+
+    auto check_rule_part = [this](std::list<std::string> const& string_rule_part) {
+        std::vector<std::string> const& item_names_map = transactional_data_->GetItemUniverse();
+        for (auto const& item_name : string_rule_part) {
+            auto it = std::ranges::find(item_names_map, item_name);
+            if (it == item_names_map.end()) {
+                throw config::ConfigurationError(
+                        "Item in left rule part not found in item universe: " + item_name);
```
vs9h: Update message, as the lambda is used for both sides

src/core/algorithms/association_rules/ar_algorithm.cpp
```diff
@@ -45,20 +43,9 @@ void ARAlgorithm::MakeExecuteOptsAvailable() {
...
-            transactional_data_ =
-                    model::TransactionalData::CreateFromTabular(*input_table_, first_column_tid_);
-            break;
-        default:
-            assert(0);
-    }
+    transactional_data_ = transactional_data_params_.CreateFrom();
     if (transactional_data_->GetNumTransactions() == 0) {
-        throw std::runtime_error("Got an empty dataset: AR mining is meaningless.");
+        throw std::runtime_error("Got an empty dataset: AR verifying is meaningless.");
```
vs9h: But we're in the mining class, isn't it?

src/core/config/names.h
```diff
@@ -27,6 +27,8 @@ constexpr auto kItemColumnIndex = "item_column_index";
 constexpr auto kMinimumConfidence = "minconf";
 constexpr auto kMinimumSupport = "minsup";
 constexpr auto kTIdColumnIndex = "tid_column_index";
+constexpr auto kARuleLeft = "arule_left";
+constexpr auto kARuleRight = "arule_right";
```
vs9h: I'd prefer to use such naming (for consistency with `lhs_indices` and `rhs_indices`)
```suggestion
constexpr auto kLhsRule = "lhs_rule";
constexpr auto kRhsRule = "rhs_rule";
```

src/core/algorithms/association_rules/ar_verifier/ar_verifier.cpp
```diff
@@ -0,0 +1,96 @@
...
+        check_rule_not_empty(string_rule_part);
+        check_rule_part(string_rule_part);
+    };
+
+    auto opts = config::TransactionalDataOptions(&transactional_data_params_);
+    RegisterOption(std::move(opts.input_table));
+    RegisterOption(std::move(opts.input_format));
+    RegisterOption(std::move(opts.first_column_tid));
+    RegisterOption(std::move(opts.item_column));
+    RegisterOption(std::move(opts.tid_column));
```
vs9h: I'd prefer to change this approach quite a bit.

In an ideal world, we’d have the concept of complex options, so that registration could look like this:
```cpp
/* inside class: */
TransactionalData::Params data_params_;

/* here: */
RegisterOption(config::kTransactionalDataParams(&data_params_));
```

Inside `kTransactionalDataParams`, we would then register all related options. Currently, however, we don't have this functionality. I see two possible approaches:
1) Add support for such "complex" options.
2) Define the options in "config/transactional_data/" directory, similar to how other options are defined (e.g., the number of threads). Ideologically, these options all belong together, so I'd put them all in three files (`config/transactional_data/{types.h,options.h,options.cpp}`). In this case, each option would still be registered individually, for example:
```cpp
RegisterOption(config::kInputFormat(&data_params_.input_format)); 
RegisterOption(config::k...(&data_params_....)); 
...
```

I would say that any option would be adequate. As for the proposed variant with `std::move`, I'm not really in favor of it. It doesn't quite align with the current way we typically work with options.

src/core/algorithms/association_rules/ar_verifier/ar_verifier.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+    RegisterOption(config::kMinimumConfidenceOpt(&minconf_));
+    RegisterOption(config::kMinimumSupportOpt(&minconf_));
```
vs9h:
```suggestion
    RegisterOption(config::kMinimumSupportOpt(&minsup_));
```

vs9h: Let's define options right here, because currently we use them only here (like for `kARuleLeft`, `kARuleRight`).

src/core/algorithms/association_rules/ar_verifier/ar_verifier.cpp
```diff
@@ -0,0 +1,96 @@
+#include "ar_verifier.h"
+
+#include <chrono>
```
vs9h: unused

src/core/algorithms/ar/ar_verifier/model/rule_coverage.h
```diff
@@ -0,0 +1,64 @@
+#pragma once
+
+#include <algorithm>
+#include <set>
+#include <vector>
+
+namespace algos::ar_verifier::util {
+
+struct RuleCoverage {
```
vs9h, line 6: Should we use class here?

src/core/algorithms/ar/ar_verifier/model/rule_coverage.h
```diff
@@ -0,0 +1,64 @@
...
+        : left_coverage(left_coverage), right_coverage(right_coverage) {}
+
+    RuleCoverage(std::vector<unsigned> const& transaction_indices,
+                 std::vector<unsigned> const& left_rule_part,
+                 std::vector<unsigned> const& right_rule_part)
+        : left_coverage(CalculateCoverage(transaction_indices, left_rule_part)),
+          right_coverage(CalculateCoverage(transaction_indices, right_rule_part)) {}
+
+private:
+    static double CalculateCoverage(std::vector<unsigned> const& transaction_indices,
```
vs9h, line 39: Move to the source file?

src/core/algorithms/association_rules/ar.h
```diff
@@ -36,8 +54,8 @@ struct ARStrings {
 
     ARStrings() = default;
 
-    ARStrings(std::list<std::string> left, std::list<std::string> right, double confidence,
-              double support)
+    ARStrings(std::list<std::string> left, std::list<std::string> right, double const confidence,
+              double const support)
```
vs9h: Usually we don't use const for params, could you leave it as it was?

src/core/config/ar_data/transactional_data_config.h
```diff
@@ -0,0 +1,62 @@
...
+#include "config/names_and_descriptions.h"
+#include "config/option.h"
+#include "config/tabular_data/input_table/option.h"
+#include "config/tabular_data/input_table_type.h"
+#include "model/table/column_index.h"
+#include "model/transaction/transactional_data.h"
+
+namespace config {
+
+struct TransactionalDataParams {
```
vs9h: I'd prefer to put this structure right to the `TransactionalData` class:
```cpp
class TransactionalData {
public:
    struct Params {
        config::InputTable input_table;
        algos::InputFormat input_format;
        model::ColumnIndex tid_column_index;
        model::ColumnIndex item_column_index;
        bool first_column_tid;
    };
```

And add `CreateFrom` method, which takes this structure as param (existing `TransactionalData::CreateFrom` you can make private to prohibit their usage).

You can also pay attention to the fact that from the point of view of the language, it is strange that the parameters object has a `CreateFrom` method.

src/python_bindings/ar/bind_ar_verification.h
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+#include <pybind11/pybind11.h>
+
+namespace python_bindings {
+void BindArVerification(pybind11::module_& main_module);
```
vs9h: Since we use `AR` in `ARAlgorithm`, let's use everywhere the same naming convention:
```suggestion
void BindARVerification(pybind11::module_& main_module);
```

src/tests/unit/test_apriori.cpp
```diff
@@ -168,7 +168,7 @@ TEST_F(ARAlgorithmTest, PresentationDataset) {
 }
 
 TEST_F(ARAlgorithmTest, SynteticDatasetWithPruning) {
-    auto algorithm = CreateAlgorithmInstance(kRulesSynthetic2, 0.13, 1.00001, 0, 1);
+    auto algorithm = CreateAlgorithmInstance(kRulesSynthetic2, 0.13, 1, 0, 1);
```
vs9h, line 175: Why we must change it?

vs9h (changes requested):
Sorry for the very long review delay. I'll try to provide feedback as quickly as possible after the next round of changes.

If anything is unclear, please don't hesitate to ask.

Also, it would be great to split the history into separate commits (see the comments above from @ol-imorozko)

src/core/algorithms/association_rules/ar_verifier/ar_verifier.cpp
Tydik42: I agree, the options `kARuleLeft`, `kARuleRight` aren't necessary. But why can't we use `kMinimumConfidenceOpt` and `kMinimumSupportOpt`? They're needed for both the validator and the miner, and they contain the value verification logic. The same logic exists in `config/error`, but I thought that semantically this option wasn't quite right for AR case.

src/tests/unit/test_apriori.cpp
Tydik42: Because I added a threshold check in `conifg/ar_data/threshold/option.cpp`:
```c++
extern CommonOption<ArThresholdType> const kMinimumSupportOpt{
        kMinimumSupport, kDMinimumSupport, 0.0, {}, [](ArThresholdType error) {
            if (!(error >= 0 && error <= 1)) {
                throw ConfigurationError("Minimum support value should be between 0 and 1.");
            }
        }};
```
        
By definition, support/confidence values ​​cannot be greater than one. I quote Alexander's words on why it was done this way:

> вроде бы в тест значение больше единицы я впихнул для того, чтобы проверить, что алгоритм не сломается, если пользователь что-то не то введёт. решение временное, но оно так и осталось в коде видимо

src/core/algorithms/ar/ar_verifier/model/rule_coverage.h
Tydik42: The main point of `RuleCoverage` is to store two doubles, and all the methods are just convenient ways to check or evaluate them. So I thought it would be better to call it a structure.

src/tests/unit/test_ar_verifier.cpp
```diff
@@ -0,0 +1,143 @@
...
+    ASSERT_THROW(CreateAndExecute(params), config::ConfigurationError);
+}
+
+TEST_F(TestARVerifierError, TestEmptyRule) {
+    SingularDataset dataset{kRulesBook, 0, 1};
+    RuleConfig config{{}, {"Milk"}, 0.1, 0.1};
+    ARVerifierParams params{dataset, config};
+
+    ASSERT_THROW(CreateAndExecute(params), config::ConfigurationError);
+}
```
vs9h, line 180: Let's use Aggregate initialization here for structures (as we do inside `INSTANTIATE_TEST_SUITE_P`).

For example, let's consider a test:
```cpp
TEST_F(TestARVerifierError, TestInvalidSupport) {
    SingularDataset dataset{kRulesBook, 0, 1};
    RuleConfig config{{"Bread"}, {"Milk"}, 2.0, 0.1};
```
It's difficult to understand what these number actually mean, and where is that incorrect support. I have to go to the top of the file and look for each argument.

Some people also prefer to use comments like
`RuleConfig config{ {"A"} /* lhs */, {"B"} /* rhs */, 0.1 /* ... */, 0.1 /* ... */ };`
but it's hard to maintain it. You can easily mix them up or forget to rename after refactoring.

src/tests/unit/test_ar_verifier.cpp
```diff
@@ -0,0 +1,143 @@
...
+    RuleConfig config{{"Bread"}, {"Milk"}, 2.0, 0.1};
+    ARVerifierParams params(dataset, config);
+
+    ASSERT_THROW(CreateAndExecute(params), config::ConfigurationError);
+}
+
+TEST_F(TestARVerifierError, TestInvalidRule) {
+    SingularDataset dataset{kRulesBook, 0, 1};
+    RuleConfig config{{"42"}, {"Milk"}, 0.1, 0.1};
+    ARVerifierParams params{dataset, config};
```
vs9h: Let's use `kRulesBookSingular`

src/tests/unit/test_ar_verifier.cpp
```diff
@@ -0,0 +1,143 @@
...
+                  {onam::kMinimumConfidence, config.min_conf},
+                  {onam::kLhsRule, config.left},
+                  {onam::kRhsRule, config.right},
+                  {onam::kFirstColumnTId, dataset.first_col_tid}}) {}
+};
+
+SingularDataset const kRulesBookSingular{kRulesBook, 0, 1};
+SingularDataset const kRulesPresentationSingular{kRulesPresentation, 0, 1};
+SingularDataset const kRulesPresentationExtSingular{kRulesPresentationExtended, 0, 1};
+TabularDataset const kRulesKaggleRowsTabular{kRulesKaggleRows, true};
```
vs9h: Let's use Aggregate initialization here too.

src/core/config/ar_data_threshold/option.h
vs9h, line 1: Let's split it into separate files (`config/min_confidence`, `config/min_support`).

And let's avoid using `ArThresholdType`. It's better to define a separate type for each option. And use those types in algorithms and in tests.

Also it's interesting, that we already have a minimum support option in `HyMD` algorithm. But in this case we use a number of pairs (`size_t` instead of `double`). So I suggest using `Ar` prefix for both type and option (`kArMinimum...`).

src/core/algorithms/association_rules/ar_verifier/ar_verifier.h
vs9h: Can we get rid of using these lists? I think it would be better to replace them with a vector both in the existing AR code (if it exists) and in the new code.

src/core/algorithms/association_rules/ar_verifier/model/rule_coverage.cpp
```diff
@@ -0,0 +1,36 @@
+#include "rule_coverage.h"
+
+#include <algorithm>
+#include <set>
+#include <vector>
+
+namespace algos::ar_verifier::util {
+
+RuleCoverage::RuleCoverage(double const left_coverage, double const right_coverage)
```
vs9h: nit: let's remove const here

src/core/algorithms/association_rules/ar_verifier/model/rule_coverage.h
```diff
@@ -0,0 +1,42 @@
+#pragma once
+
+#include <vector>
+
+namespace algos::ar_verifier::util {
```
vs9h: Why do we use `util` here?

I see, that we use such namespace for `nd` primitive, but it's better to use `algos::ar_verifier::model` here

vs9h (changes requested):
Please, rewrite the commit history, this was mentioned above. 

It's better to get into the habit of writing a good commit history right away, and it's much more convenient to review PR in commit order.

I would divide the commits into options refactoring, ar verifier class implementation, examples adding, etc.

It would also be great to have commit descriptions.

src/python_bindings/ar/bind_ar.cpp
```diff
@@ -41,8 +41,8 @@ void BindAr(py::module_& main_module) {
                         auto right_vec = t[1].cast<std::vector<std::string>>();
                         double conf = t[2].cast<double>();
                         double supp = t[3].cast<double>();
-                        std::list<std::string> left_list(left_vec.begin(), left_vec.end());
-                        std::list<std::string> right_list(right_vec.begin(), right_vec.end());
+                        std::vector<std::string> left_list(left_vec.begin(), left_vec.end());
+                        std::vector<std::string> right_list(right_vec.begin(), right_vec.end());
```
vs9h: nit: `left_list` -> `left_vec`

src/core/algorithms/nar/nar_algorithm.cpp
```diff
@@ -29,13 +29,13 @@ void NARAlgorithm::RegisterOptions() {
     DESBORDANTE_OPTION_USING;
 
     RegisterOption(config::kTableOpt(&input_table_));
-    RegisterOption(Option{&minconf_, kMinimumConfidence, kDMinimumConfidence, 0.0});
-    RegisterOption(Option{&minsup_, kMinimumSupport, kDMinimumSupport, 0.0});
+    RegisterOption(Option{&minconf_, kArMinimumConfidence, kDArMinimumConfidence, 0.0});
+    RegisterOption(Option{&minsup_, kArMinimumSupport, kDArMinimumSupport, 0.0});
```
vs9h: Actually, I don't know NAR definition. Can you find out info about these options for this algorithm? Do they match the AR?

If so, then the `kArMinimumSupportOpt` and `kArMinimumConfidenceOpt` should be reused here

src/tests/unit/test_ar_verifier.cpp
```diff
@@ -0,0 +1,191 @@
...
+                  {onam::kArRhsRule, config.right},
+                  {onam::kFirstColumnTId, dataset.first_col_tid}}) {}
+};
+
+// clang-format off
+SingularDataset const kRulesBookSingular{
+    .csv_config = kRulesBook,
+    .tid_col = 0,
+    .item_col = 1
+};
```
vs9h: To keep such formatting you just need to add a comma after the last argument:
```suggestion
SingularDataset const kRulesBookSingular{
        .csv_config = kRulesBook,
        .tid_col = 0,
        .item_col = 1,
};
```

src/python_bindings/py_util/get_py_type.cpp
```diff
@@ -101,7 +103,9 @@ py::tuple GetPyType(std::type_index type_index) {
             PyTypePair<std::vector<std::filesystem::path>, kPyList, kPyStr>,
             PyTypePair<std::unordered_set<size_t>, kPySet, kPyInt>,
             PyTypePair<std::string, kPyStr>,
-    };
+            PyTypePair<std::list<std::string>, kPyList, kPyStr>,
+            PyTypePair<std::unordered_map<std::string, std::vector<unsigned int>>, kPyDict, kPyStr,
+                       kPyList, kPyInt>};
```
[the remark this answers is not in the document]
vs9h: Put here comma at the end (and in the file `py_to_any.cpp‎`)

src/python_bindings/ar/bind_ar.cpp
vs9h: remove those lines, move `left_vec` and `right_vec` to `ARStrings` constructor

src/core/algorithms/association_rules/ar_verifier/ar_verifier.h
```diff
@@ -0,0 +1,84 @@
...
+
+    std::shared_ptr<::model::TransactionalData> transactional_data_;
+    std::vector<std::string> string_rule_left_;
+    std::vector<std::string> string_rule_right_;
+    ::model::ArIDs ar_ids_;
+
+    ARStatsCalculator stats_calculator_;
+
+    double minsup_ = 0.0;
+    double minconf_ = 0.0;
```
vs9h: use type aliases for these options

vs9h (changes requested):
I left a few comments.

Also, it's worth reorganizing the files inside `association_rules` folder. Firstly, I would rename it to `ar` (for consistency). Secondly, it's worth creating a separate folder for `apriori` algorithm and for `ar_verifier`.

desbordante_run.log
vs9h, line 1: There are some files, which should be removed

src/tests/unit/test_ar_verifier.cpp
```diff
@@ -0,0 +1,189 @@
...
+protected:
+    static void CreateAndExecute(ARVerifierParams const& params) {
+        auto verifier =
+                algos::CreateAndLoadAlgorithm<algos::ar_verifier::ARVerifier>(params.params);
+        verifier->Execute();
+    }
+};
+
+TEST_F(TestARVerifierError, TestEmptyTable) {
+    // clang-format off
```
vs9h: Remove these `clang-format off` comments as well

examples/basic/verifying_ar.py
```diff
@@ -0,0 +1,172 @@
...
+          "are different from those requested. Let's examine the clusters that exhibit this violation: "
+          "transaction #2 contains the entire left side of the association rule but is missing the right side, "
+          "indicating a potential error:\n")
+    print(table.iloc[2].to_list())
+    print()
+    print("Assume we are certain that the data is corrupted and that "
+          "the transaction #2 should contain 'Butter'. We will fix this record and re-validate the rule.")
+
+    algo = desbordante.ar_verification.algorithms.Default()
+    algo.load_data(table=(TABLE_FIXED_TABULAR, ',', False), input_format='tabular')
```
vs9h, line 99: It looks like we actually have a header for this table, isn't it?

examples/basic/logging_showcase/custom_handlers.py
vs9h, line 1: I'm not sure about `logging_showcase` example. It looks like it is not related to this PR and should be considered separately. Can you motivate this addition?

vs9h (commented):
There are very few new comments, let's fix them and try to merge this PR

desbordante_run.log
Tydik42: Sorry for this. It was committed accidentally.

examples/basic/logging_showcase/custom_handlers.py
Tydik42: This also was committed accidentally

examples/basic/verifying_ar.py
Tydik42: The CSV file has no header row — the first line is actual data

src/core/algorithms/ucc/ucc_verifier/CMakeLists.txt
```diff
@@ -1,4 +1,6 @@
 set(NAME ucc.verifier)
 desbordante_add_lib(NAME)
 target_sources(${NAME} PRIVATE ucc_verifier.cpp)
-target_link_libraries(${NAME} PRIVATE better-enums Boost::headers)
```
vs9h, line 4: Why do we need to change this line?

src/core/model/transaction/CMakeLists.txt
```diff
@@ -1,3 +1,5 @@
 set(NAME model.transaction)
 desbordante_add_lib(NAME OBJECT)
 target_sources(${NAME} PRIVATE itemset.cpp transactional_data.cpp)
+target_link_libraries(${NAME} PRIVATE better-enums)
+
```
vs9h: Can you remove extra line here and in [‎src/core/config/CMakeLists.txt](https://github.com/Desbordante/desbordante-core/pull/530/changes#diff-1e71f9fb9a6c39d452fc1d333ef204c6f8b821f12534c3a4dfb6ca6c2716f2b9), [src/core/algorithms/ar/CMakeLists.txt](https://github.com/Desbordante/desbordante-core/pull/530/changes#diff-ef5a36701c122470ee850d06eb2f29d8e2b884739bedc6509addaa8583e32c67)?

src/core/algorithms/ar/CMakeLists.txt
```diff
@@ -1,7 +1,10 @@
+add_subdirectory(ar_verifier)
+
 set(NAME ar)
 desbordante_add_lib(NAME)
-target_sources(${NAME} PRIVATE apriori.cpp ar_algorithm.cpp candidate_hash_tree.cpp)
+target_sources(${NAME} PRIVATE apriori/apriori.cpp ar_algorithm.cpp apriori/candidate_hash_tree.cpp)
```
vs9h: It looks like we need add cmakelists file for apriori as well

src/core/algorithms/ucc/ucc_verifier/CMakeLists.txt
Tydik42: It was necessary to fix a linker error caused by ar.verifier depending on model::transaction. Without this, `ColumnLayoutRelationData::CreateFrom`
and `PositionListIndex::Intersect` were unresolved at link time for ucc.verifier

vs9h (approved):

PR #533, opened 2025-02-26
Fix DC bindings

MichaelS239:
Currently, DC objects are depending on the algorithm object: if the algorithm object is destroyed, DC objects become invalidated. For example, the following code results in an error:
```python
import desbordante

TABLE_1 = "examples/datasets/taxes.csv"

def execute():
    algo = desbordante.dc.algorithms.Default()
    algo.load_data(table=(TABLE_1, ',', True))
    algo.execute(evidence_threshold=0, shard_length=0)
    dcs = algo.get_dcs()
    return dcs

dcs = execute()
print("Discovered DCs:")
for dc in dcs:
    print(dc)
```
My solution is similar to the one in #496: add shared pointers so that necessary fields used by DC objects are not destroyed when the algorithm object is destroyed.

### DIFF
--- a/src/core/algorithms/dc/FastADC/fastadc.cpp
+++ b/src/core/algorithms/dc/FastADC/fastadc.cpp
@@ -19,6 +19,7 @@
 namespace algos::dc {
 
 FastADC::FastADC() : Algorithm({}) {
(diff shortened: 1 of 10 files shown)


ol-imorozko (approved):
Whoops :)
Yeah, makes sense

PR #534, opened 2025-02-27
Refactor build.sh, add max-jobs options

Vdaleke:
Add -j option for CMake build
Refactor build.sh using shfmt

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,10 +1,12 @@
 #!/bin/bash
 
+# Formatted with `shfmt -i 4 -ci --w build.sh`
+
 # Stop on error:
 set -e
 
 function print_help() {
-cat << EOF
+    cat <<EOF
 Usage: ./build.sh [options]
 
 Possible options:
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -13,6 +13,7 @@ Possible options:
   -p,         --pybind                Compile python bindings
   -n,         --no-tests              Don't build tests
   -u,         --no-unpack             Don't unpack datasets
+  -j[N],      --parallel [N]          The maximum number of concurrent processes for building
```
vs9h: Maybe it's better to use the option names that people are already used to and have used before?
```
-j[N], --jobs[=N]
```

build.sh
```diff
@@ -23,103 +24,115 @@ Possible options:
...
+        # Use GDB's debug information format
+    -g)
+        GDB_DEBUG=true
+        ;;
+    # Display help
+    -h | --help | *)
+        print_help
+        exit 0
+        ;;
     esac
```
vs9h, line 75: I generally like the idea of ​​fixing the formatting, but I would prefer to keep the original formatting for this block of code.
1) In the original formatting, we use additional indentation for the options themselves inside the case block, which seems more correct to me compared to formatting with the same indentation as the outer block itself.
2) I would prefer to avoid adding indentation between short/long option names. I have never seen such indentation added. But I'm interested to know your motivation.
3) We will completely ruin the history and will not be able to use blame to conveniently jump to specific commits. At the same time, we do not fix anything globally, but simply switch to an alternative formatting option.

Other formatting changes look good.

build.sh
Vdaleke: https://cmake.org/cmake/help/latest/manual/cmake.1.html#cmdoption-cmake-build-j

I made the same as in cmake, because it was easier to do.

if you make the jobs option with equals, you will have to parse the options and substitute number of jobs with `-j`, which will increase the code.

build.sh
Vdaleke: > * In the original formatting, we use additional indentation for the options themselves inside the case block, which seems more correct to me compared to formatting with the same indentation as the outer block itself.
> * I would prefer to avoid adding indentation between short/long option names. I have never seen such indentation added. But I'm interested to know your motivation.

I just used a popular shell formatter for our code https://github.com/mvdan/sh. It's probably not very customizable and I don't think it's possible to add an indent for options here. But overall the idea of ​​using a formatter seems good for me for all our sh scripts and maybe add it to CI later.

> We will completely ruin the history and will not be able to use blame to conveniently jump to specific commits. At the same time, we do not fix anything globally, but simply switch to an alternative formatting option.

It seems to be a fairly common situation that when you blame, you jump over several consecutive commits and we won't break history by having to jump over a commit once.

vs9h: In GitHub it is not so inconvenient, but it will be much more difficult to use the `git blame` command from the terminal.

But to avoid this, I suggest formatting with the following option:
```text
-ci, --case-indent Switch cases will be indented.
```

In this case, we will not affect the lines themselves, where the values ​​for the environment variables are set, thanks to which the history will remain readable.

And I would suggest leaving information in the description about formatter with the command we used for formatting:
```bash
shfmt -i 4 -ci --w build.sh
```

And we really can't remove spaces between options (`sfhmt` doesn't have such option), so let's leave it as you did.

vs9h (commented):
And please split these changes into commits. The formatting change should definitely be in a separate commit.

And in general, it's bad practice when in the commit name we see completely unrelated things that can be easily separated (and which should be separated in a good way). We also have an implicit change, because you fixed the `CMakeCache.txt` removal

Overall, it's a good idea to clean up the formatting. Maybe someday we'll actually add it to CI. But given that the build script may not be needed at all (as I understand, cmake should be enough). Because of this, we won't rush into it.

vs9h (approved):

PR #539, opened 2025-03-07
Enable CTest; allow parallel test running

p-senichenkov:
- Connect GoogleTest to CTest to make test execution with `ctest` command possible.
- Enable parallel test running in CI to speed up `core-tests` task significantly:
  - The whole `core-tests` task:
     Before:  total_before, after: total_after;
  - GCC, Debug:
    Before:
    gcc_debug_before
    After:
    gcc_debug_after
  - Apple Clang, Debug:
    Before:
    apple_clang_before
    After:
    apple_clang_after
- Fix a typo in CMakeLists, in a workaround of Clang Address Sanitizer bug on Ubuntu.

### DIFF
--- a/.github/workflows/core-tests.yml
+++ b/.github/workflows/core-tests.yml
@@ -26,6 +26,11 @@ on:
       #- pyproject.toml
       #- examples/**
   workflow_dispatch:
+    inputs:
+      parallel-tests:
+        type: boolean
+        default: true
+        description: "Run tests in parallel"
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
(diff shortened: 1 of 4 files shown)


.github/workflows/core-tests.yml
```diff
@@ -26,6 +26,11 @@ on:
       #- pyproject.toml
       #- examples/**
   workflow_dispatch:
+    inputs:
+      noParallelTests:
```
Vdaleke, line 30: use kebab-case? we use it install-dependencies 

also may be it's better to _enable_ parallel instead of `disabling`. I thinl it is more recognizable option

README.md
```diff
@@ -356,6 +356,11 @@ cd build/target
 ./Desbordante_test --gtest_filter='*:-*HeavyDatasets*'
 ```
 
+Alternatively, you can run tests with CTest from any directory in `Desbordante` tree:
+```sh
+ctest --test-dir build
```
Vdaleke, line 361: `--exclude-regex ".*HeavyDatasets.*"`?

write about parallel fast-testing?

Vdaleke (changes requested):
LGTM, but finally squash commits into one or two like ctest+parallel but it's harder to separate in git.

Vdaleke (approved):

PR #543, opened 2025-03-12
Created and implemented class DDVerifier

unfadingDawn:
Created class **DDVerifier**, which implementing algorithm of veifying dd. 
Created python bindings for this class. 
Created python example of verifying dd.

### DIFF
--- a/examples/basic/verifying_dd.py
+++ b/examples/basic/verifying_dd.py
@@ -1,7 +1,101 @@
+import csv
+
 import desbordante
-a = desbordante.dd_verification.createDF("Col0", 0, 2)
-b = desbordante.dd_verification.createDF("Col1", 0, 1)
-lhs = [a]
-rhs = [b]
-c = desbordante.dd_verification.createDD(lhs, rhs)
-print(c.__str__())
+import pandas as pd
+
+COLOR_CODES = {
+    "green": "\033[1;32m",
+    "red": "\u001b[1;31m",
+    "yellow": "\033[1;33m",
+    "cyan": "\033[1;36m",
+    "nocolor": "\033[0m"
+}
+
+
+def print_intro(table: str) -> None:
+    print("Let's take the table stores_dd.csv and try to verify the dd in it.\n")
+    data = pd.read_csv(table)
+    print(f"{data}\n")
+
+
+def print_first_dd_verified(algorithm):
+    print(f"At first let`s check the dd: {COLOR_CODES["yellow"]}{dd1.__str__()}{COLOR_CODES["nocolor"]}\n")
+    print_holds(algorithm.dd_holds())
+    print(
+        "It means, that for any pair rows with equal value from column \"product_name\", defined distance from 0 to 20 on column \"stock_quantity\" and from 0 to 60 on column \"price_per_unit\"\n")
+
+
+def print_second_dd_verified(algorithm):
+    print(f"Next dd, which we are going to check: {COLOR_CODES["yellow"]}{dd2.__str__()}{COLOR_CODES["nocolor"]}\n")
+    print_holds(algorithm.dd_holds())
+    print(
+    "It means, that for any pair rows with equal value from columns \"store_name\" and \"category\", defined distance from 0 to 25 on column \"stock_quantity\"\n")
+
+def print_third_dd_verified(algorithm):
+    print(
+        f"Now, let`s check the dd: {COLOR_CODES["yellow"]}{dd3.__str__()}{COLOR_CODES["nocolor"]}, that means, what in one store stock quantity of goods has difference in range from 0 to 25\n")
+    print_holds(algorithm.dd_holds())
+
+def print_about_third_dd(algorithm):
+    print(
+        f"We have {algo.get_num_error_pairs()} pairs of rows, where distance in column \"stock_quantity\" is not holds.\n")
+    highlights = algo.get_highlights()
+    table_copy = table.copy()
+    for highlight in highlights:
+        first_error_row_index = highlight[1][0] + 1
+        second_error_row_index = highlight[1][1] + 1
+        print(f"{first_error_row_index}) ", end="")
+        for i in range(0, 5):
+            if i == highlight[0]:
+                print(f"{COLOR_CODES["red"]}{table_copy[first_error_row_index][i]}{COLOR_CODES["nocolor"]} ", end="")
+            else:
+                print(table_copy[first_error_row_index][i] + " ", end="")
+        print()
+        print(f"{second_error_row_index}) ", end="")
+        for i in range(0, 5):
+            if i == highlight[0]:
+                print(f"{COLOR_CODES["red"]}{table_copy[second_error_row_index][i]}{COLOR_CODES["nocolor"]} ", end="")
+            else:
+                print(table_copy[second_error_row_index][i] + " ", end="")
+        print('\n')
+
+    print(f"Error threshold: {algo.get_error()}")
+
+
+def print_holds(result: bool) -> None:
+    if result:
+        print(f"This {COLOR_CODES["green"]}DD holds{COLOR_CODES["nocolor"]}\n")
+    else:
+        print(f"This {COLOR_CODES["red"]}DD Not holds{COLOR_CODES["nocolor"]}\n")
+
+
+TABLE = 'examples/datasets/stores_dd.csv'
+lhs1 = [desbordante.dd_verification.create_df("product_name", 0, 0)]
+rhs1 = [desbordante.dd_verification.create_df("stock_quantity", 0, 20),
+        desbordante.dd_verification.create_df("price_per_unit", 0, 60)]
+dd1 = desbordante.dd_verification.create_dd(lhs1, rhs1)
+lhs2 = [desbordante.dd_verification.create_df("store_name", 0, 0),
+        desbordante.dd_verification.create_df("category", 0, 0)]
+rhs2 = [desbordante.dd_verification.create_df("stock_quantity", 0, 25)]
+dd2 = desbordante.dd_verification.create_dd(lhs2, rhs2)
+lhs3 = [desbordante.dd_verification.create_df("store_name", 0, 0)]
+rhs3 = [desbordante.dd_verification.create_df("stock_quantity", 0, 25)]
+dd3 = desbordante.dd_verification.create_dd(lhs3, rhs3)
+algo = desbordante.dd_verification.algorithms.DDVerifier()
+algo.load_data(table=(TABLE, ',', True))
+algo.execute(dds=dd1)
+with open(TABLE, newline='') as csvtable:
+    table = list(csv.reader(csvtable))
+print_intro(TABLE)
+print("-"* 100)
+print("Example №1\n")
+print_first_dd_verified(algo)
+print("-"* 100)
+print("Example №2\n")
+algo.execute(dds=dd2)
+print_second_dd_verified(algo)
+algo.execute(dds=dd3)
+print("-"* 100)
+print("Example №3\n")
+print_third_dd_verified(algo)
+print_about_third_dd(algo)
--- a/examples/basic/verifying_dd.py
+++ b/examples/basic/verifying_dd.py
@@ -1,0 +1,7 @@
+import desbordante
+a = desbordante.dd_verification.createDF("Col0", 0, 2)
+b = desbordante.dd_verification.createDF("Col1", 0, 1)
+lhs = [a]
+rhs = [b]
+c = desbordante.dd_verification.createDD(lhs, rhs)
+print(c.__str__())
--- a/examples/basic/verifying_dd.py
+++ b/examples/basic/verifying_dd.py
@@ -29,12 +29,14 @@ def print_second_dd_verified(algorithm):
     print(f"Next dd, which we are going to check: {COLOR_CODES["yellow"]}{dd2.__str__()}{COLOR_CODES["nocolor"]}\n")
     print_holds(algorithm.dd_holds())
     print(
-    "It means, that for any pair rows with equal value from columns \"store_name\" and \"category\", defined distance from 0 to 25 on column \"stock_quantity\"\n")
+        "It means, that for any pair rows with equal value from columns \"store_name\" and \"category\", defined distance from 0 to 25 on column \"stock_quantity\"\n")
+
 
 def print_third_dd_verified(algorithm):
     print(
         f"Now, let`s check the dd: {COLOR_CODES["yellow"]}{dd3.__str__()}{COLOR_CODES["nocolor"]}, that means, what in one store stock quantity of goods has difference in range from 0 to 25\n")
     print_holds(algorithm.dd_holds())
+
 
 def print_about_third_dd(algorithm):
     print(
@@ -87,15 +89,15 @@ algo.execute(dds=dd1)
 with open(TABLE, newline='') as csvtable:
     table = list(csv.reader(csvtable))
 print_intro(TABLE)
-print("-"* 100)
+print("-" * 100)
 print("Example №1\n")
 print_first_dd_verified(algo)
-print("-"* 100)
+print("-" * 100)
 print("Example №2\n")
 algo.execute(dds=dd2)
 print_second_dd_verified(algo)
 algo.execute(dds=dd3)
-print("-"* 100)
+print("-" * 100)
 print("Example №3\n")
 print_third_dd_verified(algo)
 print_about_third_dd(algo)
--- a/src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
+++ b/src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
@@ -20,6 +20,43 @@ namespace algos::dd {
         RegisterOption(Option{&num_rows_, kNumRows, kDNumRows, 0U});
         RegisterOption(Option{&num_columns_, kNumColumns, kDNUmColumns, 0U});
     }
+    //question -- Can I copy this func from Split algorithm
+    double DDVerifier::CalculateDistance(const model::ColumnIndex column_index,
+                                         const std::pair<std::size_t, std::size_t> &tuple_pair) const {
+        model::TypedColumnData const& column = typed_relation_->GetColumnData(column_index);
+        const model::TypeId type_id = column.GetTypeId();
+
+        if (type_id == +model::TypeId::kUndefined) {
+            throw std::invalid_argument("Column with index \"" + std::to_string(column_index) +
+                                        "\" type undefined.");
+        }
+        if (type_id == +model::TypeId::kMixed) {
+            throw std::invalid_argument("Column with index \"" + std::to_string(column_index) +
+                                        "\" contains values of different types.");
+        }
+        if (column.IsNull(tuple_pair.first) || column.IsNull(tuple_pair.second)) {
+            throw std::runtime_error("Some of the value coordinates are nulls.");
+        }
+        if (column.IsEmpty(tuple_pair.first) || column.IsEmpty(tuple_pair.second)) {
+            throw std::runtime_error("Some of the value coordinates are empty.");
+        }
+        double dif = 0;
+        if (column.GetType().IsMetrizable()) {
+            std::byte const* first_value = column.GetValue(tuple_pair.first);
+            std::byte const* second_value = column.GetValue(tuple_pair.second);
+            auto const& type = static_cast<model::IMetrizableType const&>(column.GetType());
+            dif = type.Dist(first_value, second_value);
+        }
+        return dif;
+    }
+
+    std::vector<model::PLI::Cluster> DDVerifier::GetRowsHolds(DF const constraint) {
+        model::ColumnIndex column_index = relation_->GetSchema()->GetColumn(constraint.column_name)->GetIndex();
+        std::shared_ptr<model::PLI const> pli =
+                relation_->GetColumnData(column_index).GetPliOwnership();
+        std::deque<model::PLI::Cluster> const& index = pli->GetIndex();
+    }
+
     void DDVerifier::LoadDataInternal() {
         relation_ = ColumnLayoutRelationData::CreateFrom(*input_table_, false);  // nulls are
         // ignored
@@ -31,6 +68,8 @@ namespace algos::dd {
         using namespace config::names;
         MakeOptionsAvailable({kNumRows, kNumColumns});
     }
+
+
 
     /*VerifyDD(){
      *get tuple of pairs, which holds lhs
--- a/src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
+++ b/src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
@@ -7,6 +7,10 @@
 #include "option_using.h"
 #include "tabular_data/input_table/option.h"
 
+#include <easylogging++.h>
+
+#include "table/vertical.h"
+
 namespace algos::dd {
     DDVerifier::DDVerifier() : Algorithm({}) {
         RegisterOptions();
@@ -17,11 +21,14 @@ namespace algos::dd {
         DESBORDANTE_OPTION_USING;
         const auto default_dd = DDs();
         RegisterOption(config::kTableOpt(&input_table_));
-        RegisterOption(Option{&dd_, kDDString, kDDDString});
+        RegisterOption(Option{&dd_, kDDString, kDDDString, default_dd});
     }
 
+    void DDVerifier::SetLimits() {
+        num_rows_ = typed_relation_->GetNumRows();
+        num_columns_ = typed_relation_->GetNumColumns();
+    }
 
-    //question -- Can I copy this func from Split algorithm
     double DDVerifier::CalculateDistance(const model::ColumnIndex column_index,
                                          const std::pair<std::size_t, std::size_t> &tuple_pair) const {
         model::TypedColumnData const &column = typed_relation_->GetColumnData(column_index);
@@ -58,43 +65,38 @@ namespace algos::dd {
             model::ColumnIndex column_index = relation_->GetSchema()->GetColumn(constraint.column_name)->GetIndex();
             columns.push_back(column_index);
         }
-        auto curr_constraint = constraints.cbegin();
+        auto curr_constraint = constraints.begin();
         for (const auto column_index: columns) {
-            const std::shared_ptr<model::PLI const> pli =
-                    relation_->GetColumnData(column_index).GetPliOwnership();
-            std::deque<model::PLI::Cluster> const &index = pli->GetIndex();
             if (result.empty()) {
-                for (std::size_t i = 0; i < index.size(); i++) {
-                    for (std::size_t j = i; j < index.size(); j++) {
-                        int first_index = index[i][0];
-                        int second_index = index[j][0];
-                        if (const double dif = CalculateDistance(column_index, {first_index, second_index});
-                            dif <= curr_constraint->upper_bound && dif >= curr_constraint->lower_bound) {
-                            result.emplace_back(first_index, second_index);
-                            }
-                        ++curr_constraint;
+                for (size_t i = 0; i < num_rows_; i++) {
+                    for (size_t j = i; j < num_rows_; j++) {
+                        auto dif = CalculateDistance(column_index, {i, j});
+                        LOG(INFO) << "Difference: " << dif;
+                        if (dif <= curr_constraint->upper_bound && dif >= curr_constraint->lower_bound) {
+                            result.emplace_back(i, j);
+                        }
                     }
                 }
-            } else {
-                for (std::size_t i = 0; i < result.size(); i++) {
-                    if (const double dif = CalculateDistance(column_index, result[i]);
-                        dif > curr_constraint->upper_bound || dif < curr_constraint->lower_bound) {
-                        result.erase(result.begin() + i);
-                        i--;
-                        }
-                    ++curr_constraint;
-                }
             }
+            else {
+                for (std::size_t i = 0; i < result.size(); i++) {
+                        if (const double dif = CalculateDistance(column_index, result[i]);
+                            dif > curr_constraint->upper_bound || dif < curr_constraint->lower_bound) {
+                            LOG(INFO) << "Difference: " << dif;
+                            result.erase(result.begin() + i);
+                            --i;
+                            }
+                    }
+            }
+            ++curr_constraint;
         }
         return result;
     }
 
     void DDVerifier::LoadDataInternal() {
-        relation_ = ColumnLayoutRelationData::CreateFrom(*input_table_, false); // nulls are
-        // ignored
+        relation_ = ColumnLayoutRelationData::CreateFrom(*input_table_, false);
         input_table_->Reset();
-        typed_relation_ = model::ColumnLayoutTypedRelationData::CreateFrom(*input_table_,
-                                                                           false); // nulls are ignored
+        typed_relation_ = model::ColumnLayoutTypedRelationData::CreateFrom(*input_table_, false);
     }
 
     void DDVerifier::MakeExecuteOptsAvailable() {
@@ -127,7 +129,9 @@ namespace algos::dd {
 
 
     bool DDVerifier::VerifyDD() const {
+        // LOG(INFO) << dd_.ToString();
         const std::vector<std::pair<int, int>> lhs = GetRowsHolds(dd_.left);
+         // LOG(INFO) << lhs.size();
         if (lhs.empty()) {
             return false;
         }
--- a/src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
+++ b/src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
@@ -14,7 +14,7 @@
 namespace algos::dd {
     DDVerifier::DDVerifier() : Algorithm({}) {
         RegisterOptions();
-        MakeOptionsAvailable({config::kTableOpt.GetName(), config::names::kDDString});
+        MakeOptionsAvailable({config::kTableOpt.GetName()});
     }
 
     void DDVerifier::RegisterOptions() {
@@ -119,6 +119,7 @@ namespace algos::dd {
     }
 
     unsigned long long DDVerifier::ExecuteInternal() {
+        LOG(INFO) << dd_.ToString();
         num_rows_ = typed_relation_->GetNumRows();
         num_columns_ = typed_relation_->GetNumColumns();
 
--- a/src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
+++ b/src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
@@ -1,8 +1,6 @@
 #include "dd_verifier.h"
 
 #include <utility>
-#include <utility>
-#include <boost/mpl/pair.hpp>
 
 #include "descriptions.h"
 #include "names.h"
--- a/src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
+++ b/src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
@@ -165,6 +165,10 @@ namespace algos::dd {
         error_ = static_cast<double>(num_error_rhs_) / static_cast<double>(lhs.size());
     }
 
+    std::vector<std::pair<std::size_t, std::pair<int, int>>> DDVerifier::GetHighlights() const {
+        return highlights_;
+    }
+
     bool DDVerifier::DDHolds() const {
         return !num_error_rhs_;
     }
--- a/src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
+++ b/src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
@@ -106,7 +106,7 @@ namespace algos::dd {
         return 0;
     }
 
-    void DDVerifier::CheckDFOnRhs(const std::vector<std::pair<int, int> > &lhs) const {
+    void DDVerifier::CheckDFOnRhs(const std::vector<std::pair<int, int> > &lhs) {
         std::vector<model::ColumnIndex> columns;
         for (const auto &dd: dd_.right) {
             model::ColumnIndex column_index = relation_->GetSchema()->GetColumn(dd.column_name)->GetIndex();
@@ -133,11 +133,12 @@ namespace algos::dd {
         if (lhs.empty()) {
             return false;
         }
-        const std::vector<std::pair<int, int> > pairs_not_holds = CheckDFOnRhs(lhs);
-        LOG(INFO) << "pairs_not_holds: " << pairs_not_holds.size() << ' ';
-        for (auto pair: pairs_not_holds) {
+        CheckDFOnRhs(lhs);
+        LOG(INFO) << "pairs_not_holds: " << highlights_.size() << ' ';
+        for (auto pair: highlights_) {
             LOG(INFO) << "pair: " << pair.first << ' ' << pair.second << ' ';
         }
-        return pairs_not_holds.empty();
+        return highlights_.empty();
+        //TODO: this return and "LOGS" must be in function PrintStatistics and LOGS is a primitive realization of vizualization highlights
     }
 }
--- a/src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
+++ b/src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
@@ -87,7 +87,7 @@ namespace algos::dd {
                 for (size_t i = 0; i < num_rows_; i++) {
                     for (size_t j = i; j < num_rows_; j++) {
                         if (const auto dif = CalculateDistance(column_index, {i, j});
-                            dif <= curr_constraint->upper_bound && dif >= curr_constraint->lower_bound) {
+                            dif <= curr_constraint->constraint.upper_bound && dif >= curr_constraint->constraint.lower_bound) {
                             result.emplace_back(i, j);
                         }
                     }
@@ -95,7 +95,7 @@ namespace algos::dd {
             } else {
                 for (std::size_t i = 0; i < result.size(); i++) {
                     if (const double dif = CalculateDistance(column_index, result[i]);
-                        dif > curr_constraint->upper_bound || dif < curr_constraint->lower_bound) {
+                        dif > curr_constraint->constraint.upper_bound || dif < curr_constraint->constraint.lower_bound) {
                         result.erase(result.cbegin() + static_cast<int>(i));
                         --i;
                     }
@@ -145,7 +145,7 @@ namespace algos::dd {
             bool is_error = false;
             for (const auto column_index: columns) {
                 if (const double dif = CalculateDistance(column_index, pair); !(
-                    dif >= curr_constraint->lower_bound && dif <= curr_constraint->upper_bound)) {
+                    dif >= curr_constraint->constraint.lower_bound && dif <= curr_constraint->constraint.upper_bound)) {
                     std::pair<std::size_t, std::pair<int, int> > incorrect_pair = {column_index, pair};
                     highlights_.push_back(incorrect_pair);
                     is_error = true;
--- a/src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
+++ b/src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
@@ -7,181 +7,45 @@
 #include "option_using.h"
 #include "tabular_data/input_table/option.h"
 
-#include <easylogging++.h>
-
-#include "table/vertical.h"
-
 namespace algos::dd {
-    DDVerifier::DDVerifier() : Algorithm({}) {
+    DDVerifier::DDVerifier(DD dd) : Algorithm({}), dd_(std::move(dd)) {
         RegisterOptions();
         MakeOptionsAvailable({config::kTableOpt.GetName()});
     }
 
     void DDVerifier::RegisterOptions() {
         DESBORDANTE_OPTION_USING;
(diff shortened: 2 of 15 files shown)


src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
```diff
@@ -0,0 +1,187 @@
...
+    std::vector<std::pair<int, int> > result;
+    std::vector<model::ColumnIndex> columns;
+    for (auto const &constraint : constraints) {
+        model::ColumnIndex column_index =
+                relation_->GetSchema()->GetColumn(constraint.column_name)->GetIndex();
+        columns.push_back(column_index);
+    }
+    auto curr_constraint = constraints.cbegin();
+    for (auto const column_index : columns) {
+        if (result.empty()) {
```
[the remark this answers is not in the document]
unfadingDawn: dd_.left is std::list<T>. It doesn`t have operator[].

src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+double DDVerifier::CalculateDistance(model::ColumnIndex const column_index,
+                                     std::pair<std::size_t, std::size_t> const &tuple_pair) const {
+    double dif = 0;
+    model::TypedColumnData const &column = typed_relation_->GetColumnData(column_index);
+    std::byte const *first_value = column.GetValue(tuple_pair.first);
+    std::byte const *second_value = column.GetValue(tuple_pair.second);
+    auto const &type = static_cast<model::IMetrizableType const &>(column.GetType());
+    dif = type.Dist(first_value, second_value);
+    return dif;
```
ol-imorozko, line 71: We can just return that:
```suggestion
    return type.Dist(first_value, second_value);
```

src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
```diff
@@ -0,0 +1,211 @@
...
+}
+
+std::vector<std::pair<int, int>> DDVerifier::GetRowsWhereLhsHolds() const {
+    std::vector<std::pair<int, int>> result;
+    std::vector<model::ColumnIndex> columns;
+    for (auto const &constraint : dd_.left) {
+        model::ColumnIndex column_index =
+                typed_relation_->GetSchema()->GetColumn(constraint.column_name)->GetIndex();
+        columns.push_back(column_index);
+    }
```
ol-imorozko, line 81: Do `columns.reserve(dd_.left.size());` to avoid reallocations in the loop

src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
```diff
@@ -0,0 +1,211 @@
...
+    ++curr_constraint;
+    for (std::size_t i = 1; i < columns.size(); i++) {
+        std::vector<std::pair<int, int>> new_result;
+        for (std::size_t j = 0; j < result.size(); j++) {
+            if (double const dif = CalculateDistance(columns[i], result[j]);
+                curr_constraint->constraint.Contains(dif)) {
+                new_result.emplace_back(result[j]);
+            }
+        }
```
ol-imorozko, line 99: We don't need value `j`, so let's just use range-based for:
```suggestion
        for (const auto& pair : result) {
            if (double const dif = CalculateDistance(columns[i], pair);
                curr_constraint->constraint.Contains(dif)) {
                new_result.emplace_back(pair);
            }
        }
```

src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
```diff
@@ -0,0 +1,211 @@
...
+        assert(constraint.constraint.upper_bound >= constraint.constraint.lower_bound);
+        assert(constraint.constraint.lower_bound >= 0);
+        assert(constraint.constraint.upper_bound >= 0);
+    }
+    for (auto const &constraint : dd_.right) {
+        assert(constraint.constraint.upper_bound >= constraint.constraint.lower_bound);
+        assert(constraint.constraint.lower_bound >= 0);
+        assert(constraint.constraint.upper_bound >= 0);
+    }
+}
```
ol-imorozko, line 126: Do we have to use assert here?

1. Assertions will be compiled out if `NDEBUG` is set, which is usually the case for release builds.
3. If an assertion fails, it will call `std::abort`, which will abruptly terminate the program. If that's what we want, then go for it, but if not, we might want to consider throwing an exception instead.

src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
```diff
@@ -0,0 +1,211 @@
...
+    return elapsed_milliseconds.count();
+}
+
+void DDVerifier::CheckDFOnRhs(std::vector<std::pair<int, int>> const &lhs) {
+    std::vector<model::ColumnIndex> columns;
+    for (auto const &dd : dd_.right) {
+        model::ColumnIndex column_index =
+                typed_relation_->GetSchema()->GetColumn(dd.column_name)->GetIndex();
+        columns.push_back(column_index);
+    }
```
ol-imorozko, line 165: same with the `columns.reserve()`

ol-imorozko, line 165: nit: you could consider also
```suggestion
    std::ranges::transform(dd_.right, std::back_inserter(columns), [this](const auto& constraint) {
        return typed_relation_->GetSchema()->GetColumn(constraint.column_name)->GetIndex();
    });
```

src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
ol-imorozko: Smth like:
```cpp
void DDVerifier::CheckCorrectnessDd() const {
    auto check_constraint = [](const auto& constraint) {
        if (constraint.constraint.upper_bound < constraint.constraint.lower_bound) {
            throw std::invalid_argument("Invalid constraint bounds for column: " + 
                                       constraint.column_name);
        }
        if (constraint.constraint.lower_bound < 0 || constraint.constraint.upper_bound < 0) {
            throw std::invalid_argument("Constraint bounds cannot be negative for column: " +
                                       constraint.column_name);
        }
    };

    std::ranges::for_each(dd_.left, check_constraint);
    std::ranges::for_each(dd_.right, check_constraint);
}
```

src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
```diff
@@ -0,0 +1,211 @@
...
+}
+
+void DDVerifier::VerifyDD() {
+    std::vector<std::pair<int, int>> const lhs = GetRowsWhereLhsHolds();
+    CheckDFOnRhs(lhs);
+    if (lhs.empty()) {
+        error_ = 0;
+    } else {
+        error_ = static_cast<double>(num_error_rhs_) / static_cast<double>(lhs.size());
+    }
```
ol-imorozko, line 190: We don't have to explicitly cast **both** values to `double`, usually casting the dividend is enough:
```cpp
    error_ = lhs.empty() ? 0 : static_cast<double>(num_error_rhs_) / lhs.size();
```

src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
```diff
@@ -0,0 +1,211 @@
...
+    for (auto const &constraint : dd_.right) {
+        assert(constraint.constraint.upper_bound >= constraint.constraint.lower_bound);
+        assert(constraint.constraint.lower_bound >= 0);
+        assert(constraint.constraint.upper_bound >= 0);
+    }
+}
+
+unsigned long long DDVerifier::ExecuteInternal() {
+    std::vector<model::ColumnIndex> columns;
```
ol-imorozko, line 129: same with `columns.reserve()`

src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
```diff
@@ -0,0 +1,211 @@
...
+    for (auto const &constraint : dd_.left) {
+        model::ColumnIndex column_index =
+                typed_relation_->GetSchema()->GetColumn(constraint.column_name)->GetIndex();
+        columns.push_back(column_index);
+    }
+    for (auto const &constraint : dd_.right) {
+        model::ColumnIndex column_index =
+                typed_relation_->GetSchema()->GetColumn(constraint.column_name)->GetIndex();
+        columns.push_back(column_index);
+    }
```
ol-imorozko, line 139: Nit: You could consider
```suggestion
    std::vector<model::ColumnIndex> columns;
    columns.reserve(dd_.left.size() + dd_.right.size());

    auto add_columns = [this, &columns](const auto& constraints) {
        std::ranges::transform(constraints, std::back_inserter(columns), [this](const auto& c) {
            return typed_relation_->GetSchema()->GetColumn(c.column_name)->GetIndex();
        });
    };

    add_columns(dd_.left);
    add_columns(dd_.right);
```

src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
```diff
@@ -0,0 +1,211 @@
...
+    for (auto const col : columns) {
+        assert(IsColumnMetrizable(col));
+    }
```
ol-imorozko, line 142: same here with the assert, you could consider throwing an `std::invalid_argument` instead

src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+    VerifyDD();
+
+    PrintStatistics();
+
+    auto elapsed_milliseconds = std::chrono::duration_cast<std::chrono::milliseconds>(
+            std::chrono::system_clock::now() - start_time);
+
+    return elapsed_milliseconds.count();
+}
```
ol-imorozko, line 157: nit: you could consider moving all code from `ExecuteInternal` to `VerifyDD`, and do this:
```cpp
unsigned long long DDVerifier::ExecuteInternal() {
    return util::TimedInvoke(&DDVerifier::VerifyDD, this);
}
```

ol-imorozko (changes requested):

ol-imorozko (commented):
> [unfadingDawn](https://github.com/unfadingDawn) wants to merge 60 commits into [Desbordante:main](https://github.com/Desbordante/desbordante-core/tree/main) from [unfadingDawn:main](https://github.com/unfadingDawn/desbordante-core-my/tree/main)

For future contributions, please create a separate branch for each feature instead of pushing from your main. It helps keep PRs cleaner and easier to review.

Also, there are three different authors, is this expected?
```bash
imorozko@laptop:~/Desbordante$ git log origin/main..unfadingDawn/main --pretty=format:"%an" | sort -u
bmwproexpert
yrii
Yuriy Litvinov
```

src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
```diff
@@ -0,0 +1,207 @@
+#include "algorithms/dd/dd_verifier/dd_verifier.h"
+
+#include <chrono>
+#include <cstddef>
+#include <timed_invoke.h>
```
ol-imorozko: Why `#include <timed_invoke.h>`?
It's not a standard library header. We should use `#include "util/timed_invoke.h"` instead

src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
```diff
@@ -0,0 +1,207 @@
+#include "algorithms/dd/dd_verifier/dd_verifier.h"
+
+#include <chrono>
```
ol-imorozko: I think `chrono` is not used in this file, though I might be wrong

src/core/algorithms/dd/dd_verifier/dd_verifier.h
```diff
@@ -0,0 +1,71 @@
...
+
+class DDVerifier : public Algorithm {
+private:
+    DDs dd_;
+    config::InputTable input_table_;
+    std::size_t num_rows_{};
+    std::size_t num_columns_{};
+    std::size_t num_error_rhs_{};
+    double error_ = 0.;
+    std::shared_ptr<model::ColumnLayoutTypedRelationData> typed_relation_;
```
ol-imorozko: _I'm probably wrong_

How do we use `shared_ptr` here? I feel like `unique_ptr` will suffice, but I'm probably missing something

src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+double DDVerifier::GetError() const {
+    return error_;
+}
+
+void DDVerifier::VisualizeHighlights() const {
+    for (auto const &hl : highlights_) {
+        auto const &col_data = typed_relation_->GetColumnData(hl.GetAttributeIndex());
+        auto const col_schema = typed_relation_->GetSchema();
+        auto const &pair = hl.GetPairRows();
```
ol-imorozko: nit: generally we prefer to explicitly specify types when it comes to non-obvious types.
In this partucular example I think it's fine, but you're specifying types right in the next function:
```cpp
bool DDVerifier::IsColumnMetrizable(model::ColumnIndex const column_index) const {
    model::TypedColumnData const &column = typed_relation_->GetColumnData(column_index);
```

So we should specify types here at least for the sake of being consistent

ol-imorozko: Moreover, I see 
```cpp
auto const col_schema = typed_relation_->GetSchema();
````
and it appears to me as if you're forgot to add a reference and you make a copy here.
My LSP tells me that GetSchema returns a pointer so it's actually a perfectly fine code, but it would be easier to notice if it was:
```cpp
RelationalSchema const* col_schema = typed_relation_->GetSchema();
``` 
instead

src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
```diff
@@ -0,0 +1,207 @@
...
+            if (double const dif = CalculateDistance(columns[i], pair);
+                curr_constraint->constraint.Contains(dif)) {
+                new_result.emplace_back(pair);
+            }
+        }
+        result = std::move(new_result);
+        ++curr_constraint;
+    }
+    return result;
+}
```
ol-imorozko: That seems strange. If I get it right, first you find all row pairs that satisfied the first constraint and then you create a temporary vector and do some refiltering?

Why not to check all constraints at once for any given pair of rows `(i, j)` and then add a pair only if it matches all checks?
Something like:
```cpp
std::vector<std::pair<int, int>> DDVerifier::GetRowsWhereLhsHolds() const {
    std::vector<std::pair<int, int>> result;
    for (std::size_t i = 0; i < num_rows_; ++i) {
        for (std::size_t j = i + 1; j < num_rows_; ++j) {
            bool all_lhs_hold = true;
            auto curr_constraint = dd_.left.cbegin();
            for (auto const &column_index : some_lhs_column_indices) { //you can use lhs_column_indices_ from above
                double const diff = CalculateDistance(column_index, {i, j});
                if (!curr_constraint->constraint.Contains(diff)) {
                    all_lhs_hold = false;
                    break;
                }
                ++curr_constraint;
            }

            if (all_lhs_hold) {
                result.emplace_back(i, j);
            }
        }
    }
    return result;
}

```

src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+std::vector<std::pair<int, int>> DDVerifier::GetRowsWhereLhsHolds() const {
+    std::vector<std::pair<int, int>> result;
+    std::vector<model::ColumnIndex> columns;
+    columns.reserve(dd_.left.size());
+    for (auto const &constraint : dd_.left) {
+        model::ColumnIndex column_index =
+                typed_relation_->GetSchema()->GetColumn(constraint.column_name)->GetIndex();
+        columns.push_back(column_index);
+    }
```
ol-imorozko: I noticed that you do a job of finding `dd_.left` or `dd_.right` indices each time in _multiple_ functions. Can we cache those values?

Something like adding the following lines to `dd_verifier.h`:
```cpp
    std::vector<model::ColumnIndex> lhs_column_indices_;
    std::vector<model::ColumnIndex> rhs_column_indices_;
```
do
```cpp
        lhs_column_indices_.clear();
        rhs_column_indices_.clear();
```
in ResetState
and fill them like so
```cpp
    auto get_column_index = [&](auto const &constraint) {
        return typed_relation_->GetSchema()->GetColumn(constraint.column_name)->GetIndex();
    };
 
    std::ranges::transform(dd_.left, std::back_inserter(lhs_column_indices_), get_column_index);
    std::ranges::transform(dd_.right, std::back_inserter(rhs_column_indices_), get_column_index);
```
in `void DDVerifier::VerifyDD()`

src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
```diff
@@ -0,0 +1,207 @@
...
+}
+
+void DDVerifier::CheckDFOnRhs(std::vector<std::pair<int, int>> const &lhs) {
+    std::vector<model::ColumnIndex> columns;
+    columns.reserve(dd_.right.size());
+    for (auto const &dd : dd_.right) {
+        model::ColumnIndex column_index =
+                typed_relation_->GetSchema()->GetColumn(dd.column_name)->GetIndex();
+        columns.push_back(column_index);
+    }
```
ol-imorozko: maybe we can use those cached values as I described above

src/core/algorithms/dd/dd_verifier/highlight.h
```diff
@@ -0,0 +1,32 @@
...
+    model::ColumnIndex attribute_index_;
+    std::pair<std::size_t, std::size_t> pair_rows_;
+    double distance_;
+
+public:
+    Highlight(model::ColumnIndex const& attribute_index,
+              std::pair<std::size_t, std::size_t> const& rows_pair, double const& distance)
+        : attribute_index_(attribute_index),
+          pair_rows_(std::move(rows_pair)),
+          distance_(distance) {}
```
ol-imorozko: Passing fundamental types like `double` by reference is not good as this introduces an unnecessary indirection and could take more space. Pass by value instead.

ol-imorozko: calling `std::move` on const reference does not perform a move.
You could just do:
```suggestion
    Highlight(model::ColumnIndex attribute_index,
              std::pair<std::size_t, std::size_t> rows_pair, double distance)
        : attribute_index_(attribute_index),
          pair_rows_(rows_pair),
          distance_(distance) {}
```
since all types here are basically fundamental.

But IMHO most idiomatic way is to do:
```cpp
    Highlight(model::ColumnIndex attribute_index,
              std::pair<std::size_t, std::size_t> rows_pair, double distance)
        : attribute_index_(std::move(attribute_index)),
          pair_rows_(std::move(rows_pair)),
          distance_(distance) {}
```

src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
```diff
@@ -0,0 +1,207 @@
...
+    for (auto const &pair : lhs) {
+        auto curr_constraint = dd_.right.cbegin();
+        bool is_error = false;
+        for (auto const column_index : columns) {
+            if (double const dif = CalculateDistance(column_index, pair);
+                !curr_constraint->constraint.Contains(dif)) {
+                highlights_.emplace_back(column_index, pair, dif);
```
ol-imorozko: Highlight constructor takes a pair of `std::size_t`:
```cpp
    Highlight(model::ColumnIndex const& attribute_index,
              std::pair<std::size_t, std::size_t> const& rows_pair, double const& distance)
        : attribute_index_(attribute_index),
          pair_rows_(std::move(rows_pair)),
          distance_(distance) {}
```
But here you insert `std::pair<int, int>`. Either use `int` in highlight or `std::size_t` here

src/core/algorithms/dd/dd_verifier/highlight.h
```diff
@@ -0,0 +1,32 @@
+#pragma once
+#include <cstddef>
+#include <memory>
```
ol-imorozko: nit: `<memory>` is not used in this file

src/core/algorithms/dd/dd_verifier/dd_verifier.h
```diff
@@ -0,0 +1,71 @@
...
+public:
+    DDVerifier();
+
+    double GetError() const;
+
+    std::size_t GetNumErrorRhs() const;
+
+    bool DDHolds() const;
+
+    std::vector<Highlight> GetHighlights() const;
```
ol-imorozko: nit: copying vector here is expected? If not, return const ref

src/core/util/create_dd.cpp
```diff
@@ -0,0 +1,20 @@
+#include "create_dd.h"
+
+namespace util {
+model::DFStringConstraint dd::CreateDf(std::string const& col_name, double const lower_bound,
+                                       double const upper_bound) {
```
ol-imorozko: When you call this function with something that isn't already a `std::string`, like a string literal or a `char*`, the compiler must first create a temporary `std::string` object. Then, your function copies from that temporary into its final destination. 
As an example, I think `pybind11` will create a  `std::string` from the Python string, and then your function will copy it.

Instead you could use `std::string_view`.

If I read [this](https://pybind11.readthedocs.io/en/stable/advanced/cast/strings.html#c-17-string-views
) right, `pybind11` can create a `std::string_view` that looks directly into the Python `str` object's underlying buffer, so no extra copies are performed

src/core/util/create_dd.cpp
```diff
@@ -0,0 +1,20 @@
...
+    dd.constraint.lower_bound = lower_bound;
+    dd.constraint.upper_bound = upper_bound;
+    return dd;
+}
+
+model::DDString dd::CreateDd(std::list<model::DFStringConstraint> const& lhs,
+                             std::list<model::DFStringConstraint> const& rhs) {
+    model::DDString dd;
+    dd.left = lhs;
+    dd.right = rhs;
```
ol-imorozko: Here it's kinda the same.

If you did smth like
```python
lhs_list = [df1, df2]
rhs_list = [df3, df4]

my_dd = Desbordante.dd_verification.create_dd(lhs_list, rhs_list)
```

Then `pybind11` will create a `std::list`, pass it to your function and you will copy it.

If you do
```cpp
model::DDString dd::CreateDd(std::vector<model::DFStringConstraint> lhs,
                             std::vector<model::DFStringConstraint> rhs) {
    model::DDString dd;
    dd.left = std::move(lhs);
    dd.right = std::move(rhs);
    return dd;
}
```
instead, we have a chance that `pybind11` will bind a rvalue to `lhs/rhs` and we'll get two moves instead of one copy

src/core/util/create_dd.cpp
```diff
@@ -0,0 +1,20 @@
+#include "create_dd.h"
+
+namespace util {
+model::DFStringConstraint dd::CreateDf(std::string const& col_name, double const lower_bound,
```
MichaelS239: Maybe it would be better to move functions `create_df` and `create_dd` to the constructors of classes `DF` and `DD` in bindings because their purpose is to create objects. The python code would look like this:
```python
lhs1 = [desbordante.dd.DF("product_name", 0, 0)]
rhs1 = [desbordante.dd.DF("stock_quantity", 0, 20),
        desbordante.dd.DF("price_per_unit", 0, 60)]
dd1 = desbordante.dd.DD(lhs1, rhs1)
```
Constructors can be created in `pybind11` using `py::init`.

src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
```diff
@@ -0,0 +1,207 @@
...
+double DDVerifier::GetError() const {
+    return error_;
+}
+
+void DDVerifier::VisualizeHighlights() const {
+    for (auto const &hl : highlights_) {
+        auto const &col_data = typed_relation_->GetColumnData(hl.GetAttributeIndex());
+        auto const col_schema = typed_relation_->GetSchema();
+        auto const &pair = hl.GetPairRows();
+        LOG(DEBUG) << "DD not Holds in " << col_schema->GetColumn(hl.GetAttributeIndex())->GetName()
```
MichaelS239:
```suggestion
        LOG(DEBUG) << "DD does not hold in " << col_schema->GetColumn(hl.GetAttributeIndex())->GetName()
```

src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
```diff
@@ -0,0 +1,207 @@
...
+double DDVerifier::CalculateDistance(model::ColumnIndex const column_index,
+                                     std::pair<std::size_t, std::size_t> const &tuple_pair) const {
+    model::TypedColumnData const &column = typed_relation_->GetColumnData(column_index);
+    std::byte const *first_value = column.GetValue(tuple_pair.first);
+    std::byte const *second_value = column.GetValue(tuple_pair.second);
+    auto const &type = static_cast<model::IMetrizableType const &>(column.GetType());
+    return type.Dist(first_value, second_value);
+}
+
+std::vector<std::pair<int, int>> DDVerifier::GetRowsWhereLhsHolds() const {
```
MichaelS239: I suggest using `std::size_t` or `model::TupleIndex` for row indices.

src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
```diff
@@ -0,0 +1,207 @@
...
+            throw std::invalid_argument("Invalid constraint bounds for column: " +
+                                        constraint.column_name);
+        }
+        if (constraint.constraint.lower_bound < 0 || constraint.constraint.upper_bound < 0) {
+            throw std::invalid_argument("Constraint bounds cannot be negative for column: " +
+                                        constraint.column_name);
+        }
+    };
+
+    std::ranges::for_each(dd_.left, check_constraint);
```
MichaelS239, line 120: It seems that `std::ranges::for_each` is defined in header `<algorithm>`, so add `#include <algorithm>`.

src/core/algorithms/dd/dd_verifier/highlight.h
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    [[nodiscard]] std::pair<std::size_t, std::size_t> GetPairRows() const {
+        return pair_rows_;
+    }
+
+    [[nodiscard]] double GetDistance() const {
+        return distance_;
+    }
+};
```
MichaelS239: Add newline at the end of the file.

src/core/config/descriptions.h
```diff
@@ -191,4 +191,5 @@ constexpr auto kDEvidenceThreshold =
 constexpr auto kDDoCollectViolations = "Flag which tells whether to collect violations or not";
 constexpr auto kDGfdK = "Max size of mined gfds";
 constexpr auto kDGfdSigma = "Min frequency of mined gfds";
+constexpr auto kDDDString = "Difference Dependency that needs to be verified";
```
MichaelS239:
```suggestion
constexpr auto kDDDString = "Differential dependency that needs to be verified";
```

src/core/config/names.h
```diff
@@ -95,6 +95,7 @@ constexpr auto kMinimumSharedValue = "minimum_shared_value";
 constexpr auto kComparableThreshold = "comparable_threshold";
 constexpr auto kEvidenceThreshold = "evidence_threshold";
 constexpr auto kDoCollectViolations = "do_collect_violations";
+constexpr auto kDDString = "dds";
```
MichaelS239:
```suggestion
constexpr auto kDDString = "dd";
```

src/tests/test_dd_verifier.cpp
```diff
@@ -0,0 +1,58 @@
...
+    auto const& p = GetParam();
+    auto const mp = algos::StdParamsMap(p.params);
+    auto const verifier = algos::CreateAndLoadAlgorithm<algos::dd::DDVerifier>(mp);
+    verifier->Execute();
+    EXPECT_EQ(verifier->DDHolds(), p.num_error_pairs == 0);
+    EXPECT_DOUBLE_EQ(verifier->GetError(), p.error);
+    EXPECT_EQ(verifier->GetNumErrorRhs(), p.num_error_pairs);
+}
+
+//clang-format off
```
MichaelS239:
```suggestion
// clang-format off
```

src/tests/test_dd_verifier.cpp
```diff
@@ -0,0 +1,58 @@
...
+                                             {{"Col3", 222, 333}, {"Col4", 111, 550}}},
+                                            1, 1. / 2.),
+                          DDVerifyingParams({{{"Col0", 0, 2}}, {{"Col1", 0, 11}}}, 1, 1. / 7.),
+                          DDVerifyingParams({{{"Col0", 0, 2}}, {{"Col1", 3, 10}}}, 2, 2. / 7.),
+                          DDVerifyingParams({{{"Col0", 2, 3}}, {{"Col1", 9, 10}}}, 4, 4. / 5.),
+                          DDVerifyingParams({{{"Col0", 2, 3}}, {{"Col1", 1, 2}}}, 5, 1.),
+                          DDVerifyingParams({{{"Col1", 2, 16}}, {{"Col3", 112, 333}}}, 5,
+                                            5. / 10.)));
+
+//clang-format off
```
MichaelS239:
```suggestion
// clang-format on
```

examples/datasets/flights_dd_verifier_example.csv
```diff
@@ -0,0 +1,8 @@
+flight_id,date,price
+25,2023-08-19,370
+25,2023-08-22,200
+11,2023-09-01,850
+25,2023-09-02,120
+11,2023-09-07,700
+11,2023-09-12,460
+25,2023-10-11,200
```
MichaelS239: Add newlines at the end of all datasets.

examples/basic/verifying_dd.py
```diff
@@ -0,0 +1,297 @@
...
+
+def print_intro(table: str) -> None:
+    print(
+        f'''This is an example of validating differential dependencies.
+        
+We follow the definition of differential dependencies
+that was introduced by Song, Shaoxu, and Chen, Lei 
+in their 2011 article, "Differential Dependencies: 
+Reasoning and Discovery," published in ACM Transactions
+on Database Systems (Vol. 36, No. 3).
```
MichaelS239:
```suggestion
Differential dependencies were introduced by Song, Shaoxu,
and Chen, Lei in their 2011 article, "Differential Dependencies:
Reasoning and Discovery," published in ACM Transactions
on Database Systems (Vol. 36, No. 3).
```
Actually, the definition of DDs that is used in the implementation is taken from the article "Eﬃcient Discovery of Diﬀerential Dependencies Through Association Rules Mining" by S.Kwashie et al. (2015). I am not sure whether we should mention this article.

examples/basic/verifying_dd.py
```diff
@@ -0,0 +1,297 @@
...
+    
+A differential dependency defines constraints on the
```
MichaelS239:
```suggestion
A differential dependency (DD) defines constraints on the 
```

examples/basic/verifying_dd.py
```diff
@@ -0,0 +1,297 @@
...
+flight_id[0,0]; date[0, 7] -> price[0, 250]. 
+
+This expression is composed of three differential functions:
+flight_id[0,0], date[0, 7], and price[0, 250]. If this 
+dependency is applied to a flight schedule table, it 
+implies that for any two tuples where the difference 
+for identical flights (sharing the same id) in the date 
+attribute is no more than 7 days, the corresponding 
+difference in the price attribute must not exceed 250 units.
+In simpler terms, this means the price difference
```
MichaelS239:
```suggestion
In simpler terms, this means that the price difference 
```

examples/basic/verifying_dd.py
```diff
@@ -0,0 +1,297 @@
...
+To illustrate this further, we will examine the 
+stores_dd.csv dataset and validate a specified 
+differential dependency. 
+
+It is worth noting that this validator implements
+standard distance metrics for numerical data types
+and dates, as well as the Levenshtein distance for 
+strings.
+
+Finally, that an additional example of differential dependency
```
MichaelS239:
```suggestion
Finally, an additional example of differential dependency 
```

examples/basic/verifying_dd.py
```diff
@@ -0,0 +1,297 @@
...
+        print(f"{second_error_row_index}) ", end="")
+        for i in range(0, 5):
+            if i == 0:
+                print(f'{COLOR_CODES["green"]}{table_copy[second_error_row_index][i]}{COLOR_CODES["nocolor"]} ', end="")
+            elif i == highlight.attribute_index:
+                print(f'{COLOR_CODES["red"]}{table_copy[second_error_row_index][i]}{COLOR_CODES["nocolor"]} ', end="")
+            else:
+                print(table_copy[second_error_row_index][i] + " ", end="")
+        print('\n')
+    print('''Clearly, this type of dependency has no practical
```
MichaelS239:
```suggestion
    print('''Clearly, this DD has no practical
```

examples/basic/verifying_dd.py
```diff
@@ -0,0 +1,297 @@
...
+significance, however, it remains useful for
+demonstration purposes. ''')
+
+
+def print_about_fourth_dd():
+    data = pd.read_csv(GRADES_TABLE)
+    print('''Validation of Differential Dependencies (DDs) can
```
MichaelS239:
```suggestion
    print('''Validation of differential dependencies can
```

src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
```diff
@@ -0,0 +1,207 @@
...
+        columns.push_back(column_index);
+    }
+    for (auto const &constraint : dd_.right) {
+        model::ColumnIndex column_index =
+                typed_relation_->GetSchema()->GetColumn(constraint.column_name)->GetIndex();
+        columns.push_back(column_index);
+    }
+    for (auto const col : columns) {
+        if (!IsColumnMetrizable(col)) {
+            throw std::invalid_argument("One of the columns is not metrizable");
```
MichaelS239: Add information about the column in the exception message (at least, its column index).

src/python_bindings/py_util/get_py_type.cpp
```diff
@@ -93,6 +94,8 @@ py::tuple GetPyType(std::type_index type_index) {
                          kPyList,
                          py::type::of<algos::hymd::preprocessing::column_matches::ColumnMatch>());
              }},
+            {typeid(model::DDString),
+             []() { return MakeTypeTuple(kPyList, py::type::of<model::DFStringConstraint>()); }},
```
MichaelS239:
```suggestion
             []() { return MakeTypeTuple(kPyTuple, kPyList, py::type::of<model::DFStringConstraint>()); }},
```
A `DDString` is not one list of `DFStringConstraint` but two lists. As far as I know, there is no python class for pairs, so `kPyTuple` can probably be used instead.

src/core/util/create_dd.cpp
unfadingDawn: The DDString class uses two std::list objects to store lhs and rhs, therefore the arguments of this function are std::list

src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+void DDVerifier::VerifyDD() {
+    auto get_column_index = [&](auto const &constraint) {
+        return typed_relation_->GetSchema()->GetColumn(constraint.column_name)->GetIndex();
+    };
+    std::ranges::transform(dd_.left, std::back_inserter(lhs_column_indices_), get_column_index);
+    std::ranges::transform(dd_.right, std::back_inserter(rhs_column_indices_), get_column_index);
+    for (auto const col : lhs_column_indices_) {
+        if (!IsColumnMetrizable(col)) {
+            std::string message;
```
MichaelS239, line 156: Add `#include <string>`.

src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
```diff
@@ -0,0 +1,195 @@
...
+            message += std::to_string(col);
+            message += " column is not metrizable";
+            throw std::invalid_argument(message);
```
MichaelS239, line 159: Add `#include <stdexcept>`.

src/core/algorithms/dd/dd_verifier/highlight.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+#include <cstddef>
+#include <utility>
+
+#include "model/table/column_index.h"
+
+class Highlight {
+private:
+    model::ColumnIndex attribute_index_;
+    std::pair<int, int> pair_rows_;
```
MichaelS239: Use `std::size_t` for row indices here as well.

src/core/config/descriptions.h
```diff
@@ -211,4 +211,6 @@ constexpr auto kDRatio =
         "cluster's size";
 // UCC verifier
 constexpr auto kDUCCIndices = "column indices for UCC verification";
+// DD verifier
+constexpr auto kDDDString = "Differential dependency that needs to be verified";
```
MichaelS239: I suppose that the options for algorithms are now sorted in the alphabetical order of their names. So put the option after options for DC verifier.

src/core/config/names.h
```diff
@@ -77,6 +77,8 @@ constexpr auto kTimeLimitSeconds = "time_limit";
 // GFD
 constexpr auto kGfdData = "gfd";
 constexpr auto kGraphData = "graph";
+// DD verifier
+constexpr auto kDDString = "dds";
```
MichaelS239: Put the option after the options for DC verifier.

MichaelS239: The name `dds` can be confusing to a user since they can think that several DDs are required. I have already left the suggestion in my previous review (https://github.com/Desbordante/desbordante-core/pull/543#discussion_r2285782549).

examples/basic/verifying_dd.py
```diff
@@ -0,0 +1,320 @@
...
+def print_intro(table: str) -> None:
+    print(
+        f'''This is an example of validating differential dependencies.
+
+Differential dependencies were introduced by Song, Shaoxu,
+and Chen, Lei in their 2011 article, "Differential
+Dependencies: Reasoning and Discovery," published in ACM
+Transactions on Database Systems (Vol. 36, No. 3).
+
+A differential dependency defines constraints on the
```
MichaelS239: Don't forget this small issue (https://github.com/Desbordante/desbordante-core/pull/543#discussion_r2285826639).

src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
```diff
@@ -0,0 +1,195 @@
...
+double DDVerifier::GetError() const {
+    return error_;
+}
+
+void DDVerifier::VisualizeHighlights() const {
+    for (auto const &hl : highlights_) {
+        auto const &col_data = typed_relation_->GetColumnData(hl.GetAttributeIndex());
+        RelationalSchema const *col_schema = typed_relation_->GetSchema();
+        auto const &pair = hl.GetPairRows();
+        LOG(DEBUG) << "DD does not holds in "
```
MichaelS239:
```suggestion
        LOG(DEBUG) << "DD does not hold in "
```

examples/datasets/grades_dd.csv
```diff
@@ -0,0 +1,8 @@
+student_id,student_name,course,exam_date,grade_score
+1,Alice,Math,2025-06-15,95
+1,Akice,Math,2025-06-20,92
+1,Alice,Physics,2025-06-18,80
+2,Alice,Math,2025-06-21,42
+2,Bob,Math,2025-06-16,70
+2,Bob,Math,2025-06-21,68
+3,Charlie,Chemistry,2025-06-17,55
```
MichaelS239: Add newline.

examples/datasets/grades_dd2.csv
```diff
@@ -0,0 +1,8 @@
+student_id,student_name,course,exam_date,grade_score
+1,Alice,Math,2025-06-15,95
+1,Alice,Math,2025-06-20,92
+1,Alice,Physics,2025-06-18,80
+2,Alice,Math,2025-06-21,42
+2,Bob,Math,2025-06-16,70
+2,Bob,Math,2025-06-21,68
+3,Charlie,Chemistry,2025-06-17,55
```
MichaelS239: Add newline.

examples/datasets/grades_dd3.csv
```diff
@@ -0,0 +1,8 @@
+student_id,student_name,course,exam_date,grade_score
+1,Alice,Math,2025-06-15,95
+1,Alice,Math,2025-06-20,92
+1,Alice,Physics,2025-06-18,80
+1,Alice,Math,2025-06-21,42
+2,Bob,Math,2025-06-16,70
+2,Bob,Math,2025-06-21,68
+3,Charlie,Chemistry,2025-06-17,55
```
MichaelS239: Add newline.

examples/basic/verifying_dd.py
```diff
@@ -0,0 +1,320 @@
...
+lhs2 = [desbordante.dd_verification.DF("store_name", 0, 0)]
+rhs2 = [desbordante.dd_verification.DF("stock_quantity", 0, 25)]
+dd2 = desbordante.dd.DD(lhs2, rhs2)
+lhs3 = [desbordante.dd_verification.DF("store_name", 0, 0),
+        desbordante.dd_verification.DF("category", 0, 0)]
+rhs3 = [desbordante.dd_verification.DF("stock_quantity", 0, 25)]
+dd3 = desbordante.dd.DD(lhs3, rhs3)
+algo = desbordante.dd_verification.algorithms.DDVerifier()
+algo.load_data(table=(TABLE, ',', True))
+algo.execute(dds=dd1)
```
MichaelS239:
```suggestion
algo.execute(dd=dd1)
```

src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
```diff
@@ -0,0 +1,195 @@
...
+        }
+    }
+    for (auto const col : rhs_column_indices_) {
+        if (!IsColumnMetrizable(col)) {
+            std::string message;
+            message += std::to_string(col);
+            message += " column is not metrizable";
+            throw std::invalid_argument(message);
+        }
+    }
```
ol-imorozko, line 169: nit: I would do smth like this:
```cpp
    auto ensure_metrizable = [&](auto const& indices) {
        for (auto col : indices) {
            if (!IsColumnMetrizable(col)) {
                throw std::invalid_argument(
                    std::string(col) + " column is not metrizable");
                );
            }
        }
    };

    ensure_metrizable(lhs_column_indices_);
    ensure_metrizable(rhs_column_indices_);
```
but current code is fine too

src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
```diff
@@ -0,0 +1,195 @@
...
+    num_columns_ = typed_relation_->GetNumColumns();
+
+    std::vector<std::pair<std::size_t, std::size_t>> const lhs = GetRowsWhereLhsHolds();
+    CheckDFOnRhs(lhs);
+    error_ = lhs.empty() ? 0 : static_cast<double>(num_error_rhs_) / lhs.size();
+}
+
+std::vector<Highlight> const &DDVerifier::GetHighlights() const {
+    return highlights_;
+}
```
ol-imorozko, line 181: is that what clang-format formats like? I thought it supposed to do ` const& `. Strange

src/core/algorithms/dd/dd_verifier/highlight.h
ol-imorozko: I guess this should be `std::pair<std::size_t, std::size_t>` also?

ol-imorozko (approved):
LGTM

src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
unfadingDawn: Yes, clang-format does this formatting.

src/core/algorithms/dd/dd_verifier/highlight.h
```diff
@@ -0,0 +1,29 @@
...
+public:
+    Highlight(model::ColumnIndex attribute_index, std::pair<std::size_t, std::size_t> rows_pair,
+              double distance)
+        : attribute_index_(attribute_index), pair_rows_(rows_pair), distance_(distance) {}
+
+    [[nodiscard]] model::ColumnIndex GetAttributeIndex() const {
+        return attribute_index_;
+    }
+
+    [[nodiscard]] std::pair<int, int> GetPairRows() const {
```
MichaelS239:
```suggestion
    [[nodiscard]] std::pair<std::size_t, std::size_t> GetPairRows() const {
```

examples/datasets/flights_dd_verifier_example.csv
```diff
@@ -0,0 +1,9 @@
...
+
```
MichaelS239: Screenshot from 2025-09-10 21-16-52
Screenshot from 2025-09-10 21-18-05

It seems that there are two newlines now. Please leave one newline at the end of **all** datasets, like in `stores_dd.csv`. Nothing important, just a matter of style.

Issue #549, opened 2025-04-02
CMake 4.0+ is not supported due to dependency on easylogging


Vdaleke:
The [easylogging](https://github.com/abumq/easyloggingpp) library requires a CMake version compatible with 2.8.7, but compatibility with <3.5 was removed in CMake 4.0.

A possible solution is to switch to another logging library, such as [sdplog](https://github.com/gabime/spdlog).

PR #550, opened 2025-04-02
Explicitly use CMake 3.31, ad-hoc related to #549

Vdaleke:
Set CMake 4.x as unsupported for project.
Use CMake 3.31 in CI explicitly with python venv and CMake from PyPi. 
Fix undocumented commands in datasets/CMakeLists.txt.

Related to #549

### DIFF
--- a/.github/composite-actions/install-dependencies/action.yml
+++ b/.github/composite-actions/install-dependencies/action.yml
@@ -53,6 +53,18 @@ runs:
 
         echo "OS=$(get_short_os_name ${{ inputs.os }})" >> $GITHUB_ENV
 
+    - name: Create python virtual environment
+      run: | 
+          python3 -m venv .venv
+      shell: bash
+
+    # Add-hoc for https://github.com/Desbordante/desbordante-core/issues/549
+    - name: Install cmake 3.31.6 in python venv
+      run: |
+        source .venv/bin/activate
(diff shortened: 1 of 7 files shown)


.github/composite-actions/install-dependencies/action.yml
```diff
@@ -53,6 +53,18 @@ runs:
 
         echo "OS=$(get_short_os_name ${{ inputs.os }})" >> $GITHUB_ENV
 
+    - name: Create python virtual environment
+      run: | 
+          python3 -m venv .venv
```
vs9h, line 58: Why do you want to hide the directory? Usually everyone just leaves venv.
Strictly speaking, there is no difference, so I would prefer to leave it as it was.

vs9h (approved):
Overall it looks good

.github/composite-actions/install-dependencies/action.yml
Vdaleke: >  Usually everyone just leaves venv. 
I don't agree with this, for example, VScode creates venv with a dot by default. Some tools like poetry by default search for .venv names with a dot. As far as I know, this is a generally accepted name.

Vdaleke: Ok, I googled both names are common. I can leave venv, but would merge as is.

PR #554, opened 2025-04-08
Set CMake policy CMP0175 to NEW and fix add_custom_command()

p-senichenkov:
Remove invalid arguments of add_custom_command() and add required ones
Set policy CMP0175 to NEW to forbid invalid arguments

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,8 +1,7 @@
-add_custom_target(symlink-python-examples ALL)
 set(EXAMPLES_DIR "examples")
-add_custom_command(TARGET symlink-python-examples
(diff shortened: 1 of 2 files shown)


examples/CMakeLists.txt
```diff
@@ -5,4 +5,5 @@ add_custom_command(TARGET symlink-python-examples
         ${CMAKE_SOURCE_DIR}/${EXAMPLES_DIR}
```
Vdaleke, line 5: You can combine command with target like it done in datasets/CMakeLists.txt now.

Vdaleke (changes requested):

Vdaleke (approved):
LGTM

PR #555, opened 2025-04-09
Add cli testing in CI

DaniilGoncharov:
This workflow runs test from cli repository


DaniilGoncharov:
This pull request rely on pull request "Add tests for parsing and help pages" №11 from desbordante-cli (https://github.com/Desbordante/desbordante-cli/pull/11)

polyntsov (approved):
I am not 100% sure that we want to run tests from CLI repo as part of CI checks in core repo and not the other way around (it should be possible to trigger CI check in CLI repo from this repo, but might be more complex)

Probably for now this solution is good enough

PR #556, opened 2025-04-14
Reorganize CMake configuration to use CPM.cmake

Vdaleke:
### Refactor CMake configuration:
- Apply cmake-format to standardize the file formatting.
- Add .cmake-format.yaml config to project root.
- Add detailed comments explaining each section.
- Reorganize commands in more logical order.
- Implement proper cleanup for copied files and directories.
- Explicitly removed MSVC support.
- Remove redundant build.bat.

### Use CPM.cmake for managing source dependencies

- Use CPM.cmake for managing source dependencies from GitHub.
- Remove build.sh and actions sections related to source dependencies.
- Remove source installations from GitHub actions.
- Add CPM.cmake cache support for local builds and CI.


.cmake-format.yaml
```diff
@@ -0,0 +1,9 @@
+include: ["cmake/.cmake-format-additional_commands-cpm"]
+format:
+  tab_size: 4
+  line_width: 100
+  # If a statement is wrapped to more than one line, than dangle
```
BUYT-1, line 5:
```suggestion
  # If a statement is wrapped to more than one line, then dangle
```

BUYT-1:
Please create a separate commit for formatting, then another with the reorganization

BUYT-1 (approved):
LGTM

Vdaleke:
Since it is not very clear from the force push, I just swapped the include(cpm-cmake) and the cache variable setting, because the variable is used inside the include

I also bumped the cpm-cmake version from 0.40 to 0.42

PR #557, opened 2025-04-20
Add performance regression tests

p-senichenkov:
Implement performance regression tests for the following algorithms:
- FD: HyFD, Pyro, Tane, EulerFD, AID-FD
- DD: Split
- MD: HyMD
- ADC: FastADC
- NAR: DES
- IND: MIND, Faida
It's able to add more tests easily.

On each test run performance charts are being built, which look like this:
[plots.pdf](https://github.com/user-attachments/files/21232831/plots.pdf)

Also do a little refactoring:
- remove unused variables found by new version of Clang;
- group option names and descriptions by algo/primitive name they used for;
- sort `all_csv_configs` alphabetically.


Vdaleke (changes requested):
Overall it looks reasonable, thanks for the great work. However, it seems like the overall architecture needs some more thought.

Firstly, in the process of discussing the platform for performance testing, we decided to do it on GitHub, because it is free. The decision is clear, but I am still not sure about the stability of the resources provided by GitHub and that the results will not differ greatly, do you have any empirical observations on this matter?

Secondly, checking the correctness of the results remains on manual downloading of the pdf artifact, right? The problem here is that it requires constant monitoring. If it is possible, I would make some threshold of deviation of the results after which a Github Issue with a warning would be automatically created.

Then, it is worth having a solution that can be run on any machine. I understand that this can be done somehow now, but here is what I mean:
- The current main, or probably better the latest release (there should be an option) is run on benchmarks and the results for the current machine are recorded.
- The current local version of the project is run on benchmarks, transferring the results of running the stable solution there and comparing with the current result.
That is, in fact, it should be possible to do the same thing as on GitHub only locally on any machine.
- Here I would think about virtualization running in a Docker container with dedicated resources.
- This should also be described in the README

And I think it's worth using Google benchmarks if possible. It has more advantages than a hand-made implementation.

I hope I haven't forgotten any important ideas. But it seems there's still a lot to work on here.

p-senichenkov:
> Firstly, in the process of discussing the platform for performance testing, we decided to do it on GitHub, because it is free. The decision is clear, but I am still not sure about the stability of the resources provided by GitHub and that the results will not differ greatly, do you have any empirical observations on this matter?

Yes, I've ran these tests several times and 10% threshold for most algorithms look quite ok. You can check threshold for each algo in corresponding `RegisterTest` calls.

> Secondly, checking the correctness of the results remains on manual downloading of the pdf artifact, right? The problem here is that it requires constant monitoring. If it is possible, I would make some threshold of deviation of the results after which a Github Issue with a warning would be automatically created.

Job will fail if tests fail.

> Then, it is worth having a solution that can be run on any machine. I understand that this can be done somehow now, but here is what I mean:
> 
>     * The current main, or probably better the latest release (there should be an option) is run on benchmarks and the results for the current machine are recorded.
> 
>     * The current local version of the project is run on benchmarks, transferring the results of running the stable solution there and comparing with the current result. Actually, it should be possible to do the same thing as on GitHub only locally on any machine.
> 
>     * Here I would think about virtualization running in a Docker container with dedicated resources.
> 
>     * This should also be described in the README

I don't quite understand what do you mean by "run on any machine", but these tests aren't considered to be ran locally -- it's CI-only tool. If one thinks their changes have broken some specific algorithm, it's easier to measure preformance of this algorithm manually.

> And I think it's worth using Google benchmarks if possible. It has more advantages than a hand-made implementation.

I'll look at Google benchmark, but for now it looks like there's no convenient interface suitable for our task, and running performance tests with Google benchmark will require several scripts and json parsers. Also, Google test have some restrictions related to test names, which made Google test-based performance testing very inconvenient.

.github/workflows/performance-tests.yml
```diff
@@ -0,0 +1,114 @@
+name: Performance tests
+on:
+  # Note: schedule only triggers on `main` branch
+  schedule:
+    - cron: 30 3 * * 6
```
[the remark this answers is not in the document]
p-senichenkov: The idea was that there's a general meeting on Sunday, at which bad results of performance tests can be discussed. But it's mostly a random choice.

.github/workflows/performance-tests.yml
```diff
@@ -0,0 +1,114 @@
...
+
+      - name: Get run IDs
+        working-directory: ${{ github.workspace }}
+        run: |
+          curl -L \
+            -o runs.json \
+            -H "Accept: application/vnd.github+json" \
+            -H "X-GitHub-Api-Version: 2022-11-28" \
+            -H "status: success" \
+            -H "exclude_pull_requests: true" \
```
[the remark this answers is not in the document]
p-senichenkov: To reduce response size

.github/workflows/performance-tests.yml
```diff
@@ -0,0 +1,114 @@
...
+            'https://api.github.com/repos/${{ github.repository }}/actions/workflows/performance-tests.yml/runs'
+
+          # [0] is current job
```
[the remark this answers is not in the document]
p-senichenkov: There's a mistake: `success` should be query parameter, not a header. But actually, there's no need in this filter, because normally these jobs fail only when tests fail. In that case artifacts are uploaded anyway.

.github/workflows/performance-tests.yml
```diff
@@ -0,0 +1,114 @@
...
+          toolset: gcc
+      - name: Download datasets
+        uses: ./.github/composite-actions/download-datasets
+      - name: Build
+        run: |
+          export CXX=g++-10
+          ./build.sh --perf-tests --no-tests
+
+      - name: Get run IDs
+        working-directory: ${{ github.workspace }}
```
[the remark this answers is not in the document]
p-senichenkov: I haven't found any mentions that `working-directory` is guaranteed to be `github.workspace` if not specified, so I specify it explicitly just to be sure.

.github/workflows/performance-tests.yml
```diff
@@ -0,0 +1,114 @@
...
+        run: |
+          ./Desbordante_perf_test ${{ github.workspace }}/latest.json \
+            ${{ github.workspace }}/curr-results.json
+      # All next steps should be runned even if this one has failed
+      # Some of them may fail in that case, but it's not a problem
+      - uses: actions/upload-artifact@v4
+        with:
+          name: results.json
+          path: curr-results.json
+        if: ${{ !cancelled() }}
```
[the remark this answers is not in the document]
p-senichenkov: This condition enables job in all cases, except for the case when it was cancelled. See https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#always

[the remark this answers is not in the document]
p-senichenkov: `continue-on-error` isn't suitable here: it disables errors at all. I do want job to fail when tests fail, but some steps must run even in that case.

.github/workflows/performance-tests.yml
```diff
@@ -0,0 +1,114 @@
...
+      - name: Build plots
+        working-directory: ${{ github.workspace }}
+        run: |
+          python3 src/tests/performance_tests/display_performance_tests.py artifacts \
+            curr-results.json plots.pdf
+        if: ${{ !cancelled() }}
+      - uses: actions/upload-artifact@v4
+        with:
+          name: plots.pdf
+          path: ${{ github.workspace }}/plots.pdf
```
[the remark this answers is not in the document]
p-senichenkov: No, `artifacts` folder contains only old results. I'd better rename it

src/python_bindings/py_util/get_py_type.cpp
```diff
@@ -28,7 +28,6 @@ constexpr PyTypeObject* const kPyBool = &PyBool_Type;
 constexpr PyTypeObject* const kPyFloat = &PyFloat_Type;
 constexpr PyTypeObject* const kPyStr = &PyUnicode_Type;
 constexpr PyTypeObject* const kPyList = &PyList_Type;
-constexpr PyTypeObject* const kPyTuple = &PyTuple_Type;
```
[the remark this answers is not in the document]
p-senichenkov: It's an artifact left from HyMD. We've discussed this at one of meetings, and decided to remove it. If someone needed `kPyTuple`, I don't think it would be a problem to define it by analogy with previous lines. Also, as for me, `maybe_unused` on global constant would look strange.

src/tests/common/all_csv_configs.cpp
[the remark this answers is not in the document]
p-senichenkov: It's folder for all common code for tests, not only configs. Maybe one day we'll have some test-specific logger settings, utility functions, or something else. Also, there're some GFD-related configs in `unit tests` forlder, that are not common.

src/tests/performance_tests/display_performance_tests.py
```diff
@@ -0,0 +1,52 @@
+from os import scandir
```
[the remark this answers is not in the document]
p-senichenkov: I'm not sure if such a detailed description is needed for utility script that must be run only from CI. Especially `requirements` part.

datasets/crop_datasets.sh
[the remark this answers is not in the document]
p-senichenkov: This script is needed because I want each test to run about 4 minutes, so that time spent on virtual environment setup and other things doesn't affect test results. And yes, this affects results a lot. Cropped datasets cannot be uploaded into `datasets.zip`, because we've ran out of Git LFS.

Vdaleke:
> Job will fail if tests fail.

I'm not sure that this is a good solution. It's easy to miss a bad result, not to notice it. It turns out that you need to regularly monitor the results of the work every week.

It seems to me that you need to leave an obvious artifact in the form of an automatic issue in which all the plots will be, and also preferably a history of changes to commits from the last test run. If you can do this, it will be much more convenient

> I don't quite understand what do you mean by "run on any machine", but these tests aren't considered to be ran locally -- it's CI-only tool.

I don't want this to be a CI-only tool. By running it only in CI we are forced to test the results on the resources provided by GitHub. The results can vary significantly from system to system and I want to be able to run it manually on any machine with any configuration. This also makes it easier to migrate to other server resources.

> If one thinks their changes have broken some specific algorithm, it's easier to measure preformance of this algorithm manually.

Again, it would be convenient if this one had a ready-made infrastructure for this, so that they would have to do less effort for such a benchmark.

But I'm not only talking about changes in some algorithm. Example: we replaced one logging library with another, we want to check if this slowed down the performance of all algorithms that had logs. Now it is assumed that you will upload your branch to GitHub and run the workflow yourself for your branch (again, it is worth writing about this somewhere), or maybe in a pull request? (I think it would be cool if the reviewer could run performance testing by pressing a button in PR, such an opportunity should be there)

Here you run into the fact that you run all this on GitHub resources, wait for the result for a long time. But if there was a possibility to run it locally, on your machine, where there may be more resources than on GitHub, faster and more efficiently, under your own system and compiler configuration, for example under Apple Clang on MacOS. Then you could understand some special results or understand that everything is bad faster than waiting for it in CI.

> I'll look at Google benchmark,

I'll wait for a comment on this matter, I'd still like to understand all the pros and cons of such a decision.

.github/workflows/performance-tests.yml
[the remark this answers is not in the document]
Vdaleke: I would launch it somewhere on Wednesday, for example. We can discuss it on a call.

p-senichenkov:
I've looked at Google benchmark, and, firstly, there's the same problem as with Google test: it will require users to add benchmark executables manually. With handmade solution, for most algorithms, you can just call
```C++
PerformanceTesting::RegisterSimpleTest<Algo>(dataset_name, algo_options);
```
With Google benchmark you'll need to define executable, benchmark name, etc. Minimal example is
```C++
void AlgoBenchmark() {
    StdParamsMap opts = {
        {config::csv_config, ...},
        ...};
    auto algo = algos::CreateAndLoadAlgorithm<Algo>(opts);
    algo->Execute();
}

BENCHMARK(AlgoBenchmark)->Name("Algo_dataset");
```
I could define a function template for simple tests, so it will look like this:
```C++
BENCHMARK(SimpleBenchmark<Algo>)->Arg(algo_opts)->Name("Algo_dataset");
```
but I'm not sure if it's better than custom solution. Also, Google test allows only alphanumeric characters in test names, and if Google benchamrk does so, even these examples won't compile.

Furhtermore, custom test system can select test name automatically, based on algo and dataset name, so that we have consistent, automatically-updated benchmark names for all algorithms. This can be done with Google benchmark, but due to wide use of macros, the solution will be much more complicated than it's currently is.

Secondly, results format of Google becnhmark is not very convenient: it's JSON with a lot of information that we don't need, so some scripts for parsing these results and comparing them are still needed.
`compare.py` from Google benchmark cannot be used, due to it's not flexible enough. For example, it shows only tests that have been ran in **both** runs (current and previous).

Considering all that, the only thing that Google benchmark will do is
```C++
auto start = std::chrono::steady_clock::now();
Body();
return std::chrono::steady_clock::now() - start;
```
which is now done in custom system. All other things (results parsing, comparison, etc.) I'll have to do by hand.

Overall, Google becnhmark can be used, but the solution won't be simplier then the current one (most likely, it will be more complicated). As a bonus, we get extra dependency.

Vdaleke:
> I could define a function template for simple tests, so it will look like this:
> ```cpp
> BENCHMARK(SimpleBenchmark<Algo>)->Arg(algo_opts)->Name("Algo_dataset");
> ```

It looks good to me. In fact, I would like to see something similar to the existing tests, for example here [here](https://github.com/Desbordante/desbordante-core/blob/6770277a8e1c5e4a48683f0cd256dcb3f6dc3389/src/tests/test_fd_algorithm.cpp#L162)
I looked at the [user guide](https://github.com/google/benchmark/blob/main/docs/user_guide.md#output-formats) and it seems like everything is there and it can all be done quite nicely.

> Also, Google test allows only alphanumeric characters in test names, and if Google benchamrk does so, even these examples won't compile.

I don't really understand what the problem is, do our algorithms and datasets contain non-alphanumeric characters? That's weird.

> Furhtermore, custom test system can select test name automatically, based on algo and dataset name, so that we have consistent, automatically-updated benchmark names for all algorithms. This can be done with Google benchmark, but due to wide use of macros, the solution will be much more complicated than it's currently is.

It seems to me that it has always been accepted to write test names yourself and this is not a problem. But as far as I remember, names in Google tests are generated automatically depending on the transferred template-class, at least that's how we have it now in the same FD.

In general, I don't see a problem here, if you really want an autogen name, then from the example that you attached, you can simply use the function there.

> Secondly, results format of Google becnhmark is not very convenient: it's JSON with a lot of information that we don't need, so some scripts for parsing these results and comparing them are still needed.

I looked at the [output format](https://github.com/google/benchmark/blob/main/docs/user_guide.md#output-formats) in the examples and in my opinion everything is quite useful and even human-readable. And in the python code, which previously collected the necessary fields from the json map, I don’t see a problem here at all.

> `compare.py` from Google benchmark cannot be used, due to it's not flexible enough. For example, it shows only tests that have been ran in both runs (current and previous).

This is not a problem, it is clear that we need our own comparison in any solution

> Overall, Google becnhmark can be used, but the solution won't be simplier then the current one (most likely, it will be more complicated). As a bonus, we get extra dependency.

I don't see the simplicity in using your own tool instead of a commonly used one. Google tests are in fact a standard testing framework that almost any C++ developer is familiar with.

It is a common practice in software development to use third-party libraries instead of writing your own code. If there is a good solution, tested, with the necessary functionality and well supported by the community - it all simplifies life a lot. The popularity of this framework makes it easy for new developers to understand the code. There should be no problems from such a dependence, after all, it is Google and we already use Google tests - this is one of the key factors, the consistency of existing tests, uniformity, which provides easier support.

In general, in my opinion, I did not see any reason not to use this library. It seems to me - this is what we need.

.github/workflows/performance-tests.yml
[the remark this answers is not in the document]
Vdaleke: In general, is there really no obvious way to request the last successful delivery for the required workflow, rather than running through the latest ones?

src/tests/benchmark/benchmark_controller.h
```diff
@@ -0,0 +1,167 @@
...
+    /// @brief Register test to be executed at @c RunAllTests call.
+    /// @tparam Body any callable.
+    /// @param test_body the whole test.
+    /// @param test_name a test name to be displayed. Must be unique. A good test name includes
+    /// algorithm name and dataset name (or size).
+    /// @param threshold execution time threshold, per cent.
+    /// @note If you need more precise control on which part of test is measured, see @c
+    /// RegisterCustomTest.
+    template <typename Body>
+    void RegisterTest(Body&& test_body, std::string const& test_name,
```
Vdaleke, line 109:
```suggestion
    void WrapAndRegisterTest(Body&& test_body, std::string const& test_name,
```
it is not entirely clear what the method does and how it differs from custom

src/tests/benchmark/benchmark_controller.h
```diff
@@ -0,0 +1,167 @@
...
+        prev_results_ = std::move(prev_res);
+    }
+
+    /// @brief Register a complex test to be executed at @c RunAllTests call.
+    /// @tparam Test any callable, which returns @a truthy @a value when test passed,
+    /// @a falsy otherwise. See implementation of @c WrapTest as an example.
+    /// @note Use this method only if you really need precise control on which part of test
+    /// is measured. <b>If you don't know which method to use to register your test, most likely
+    /// it's @c RegisterTest(), not this one.</b>
+    template <typename Test>
```
Vdaleke, line 95: Why not `std::function<bool()>`?

src/tests/benchmark/benchmark_controller.h
```diff
@@ -0,0 +1,167 @@
...
+        auto demangled_name = boost::core::demangle(mangled_name.c_str());
+        auto colon_pos = demangled_name.rfind(':');
+        if (colon_pos == std::string::npos) {
+            return demangled_name;
+        }
+        return demangled_name.substr(colon_pos + 1);
+    }
+
+    /// @note Returned object lives no longer than this BenchmarkController object.
+    template <typename Body>
```
Vdaleke, line 44: Why not `std::function<void()>`?

src/tests/benchmark/benchmark_controller.h
```diff
@@ -0,0 +1,167 @@
...
+                      unsigned char threshold = 10) {
+        RegisterCustomTest(WrapTest(std::forward<Body>(test_body), test_name, threshold));
+    }
+
+    /// @brief Run all registered tests.
+    /// @note It's unlikely that you'll need to call this method.
+    bool RunAllTests() {
+        auto result = true;
+        for (auto const& test : tests_) {
+            result &= test();
```
Vdaleke, line 119:
```suggestion
            result = result && test();
```
I know the current version is correct, but it's a bitwise and, which is a bit confusing to use for logical operations.

src/tests/benchmark/benchmark_controller.h
```diff
@@ -0,0 +1,167 @@
...
+#include "algorithms/algo_factory.h"
+#include "config/names.h"
+#include "parser/csv_parser/csv_parser.h"
+
+namespace benchmark {
+
+/// @brief A singleton that controls execution of benchmarks.
+/// @note Most likely you'll need only these methods: @c Instance, @c RegisterSimpleTest and
+/// @c RegisterTest.
+class BenchmarkController {
```
Vdaleke, line 25: To supplement my comment that I would not like to see a singleton here, this class clearly lacks methods for deregistering tests. For this, it seems worth returning some test id, or perhaps just use its name as an id, but then return a string for the methods that generate it, which is generally logical.

src/tests/benchmark/main.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include "benchmark_results.h"
+#include "dd_benchmark.h"
+#include "fd_benchmark.h"
+#include "ind_benchmark.h"
+#include "md_benchmark.h"
+#include "nar_benchmark.h"
+
+INITIALIZE_EASYLOGGINGPP
+
+int main(int argc, char* argv[]) {
```
Vdaleke, line 19: It would be good to have a comment here about the expected arguments

datasets/CMakeLists.txt
```diff
@@ -15,3 +15,19 @@ add_custom_target(unZip ALL
...
+if (COMPILE_BENCHMARKS)
+	# Please, don't forget to add here names of datasets you want to produce
+	set(cropped_dataset_names
+		adult9attr.csv CIPublicHighway20attr55k.csv iowa550k.csv iowa650k.csv
+		mushroom+3attr1500.csv mushroom+4attr1300.csv neighbors120k.csv)
+
+	add_custom_target(cropDatasets ALL DEPENDS ${cropped_dataset_names})
+	add_custom_command(
+		OUTPUT ${cropped_dataset_names}
+		COMMAND /bin/sh ${CMAKE_SOURCE_DIR}/datasets/crop_datasets.sh
```
Vdaleke, line 28: I'm not thrilled that datasets are cut every time. I think it's not a problem to upload new datasets to the corresponding repository. It's not a problem that we don't have git lfs, we should disable it completely, not that we really need it, it's just a speedup.

It's advisable to put the script in the repository with datasets as an artifact of how these datasets appeared.

.github/workflows/performance-tests.yml
Vdaleke, line 1: In general, I have another concern about artifact versioning. For example, we are merging the current solution, and then we take another one, where there is a different type of Json or other tests. How can we make it so that artifacts from the previous solution do not interfere with the new one? Maybe we should make versioning for artifacts? Is it possible to delete old artifacts from the repository (here we lose information and it seems that versioning would be better)?

Vdaleke (changes requested):
Once again, thank you for your work and cool feature. But my opinion is that the solution still requires some planning and architectural improvements. Yes, there is something that works well now. But is it necessary to spend time on reviewing and refining something that can completely disappear if we take, for example, Google benchmarks or another library? In general, I am ready to approve the current architectural version, but I understand that it can be done better.

I also want to say that I cannot review the selected datasets and settings for specific primitives, because I am not competent in this. For this part of the solution, I would like to see an approval from @chernishev  or someone else.

And the current PR still needs an artifact in the form of a generated pdf, as well as some link to an example of launching an action in a fork or still make the functionality of launching by a button in the PR itself. I would like to see both the appearance and the parameters of the test run time.

src/tests/benchmark/benchmark_controller.h
p-senichenkov: I can't imagine any scenario in which it's necessary to unregister test.

.github/workflows/performance-tests.yml
p-senichenkov: No, there is no generic way to version artifacts -- they are considered to be logs or something like that. The easy solution is to simply name new artifacts differently (e. g. "artifact-2.0"). Artifacts can be deleted from repository manually. Also, they're being deleted automatically after 90 days.

.github/workflows/performance-tests.yml
```diff
@@ -0,0 +1,108 @@
...
+  # ref will be always the same (`main`) for schedule, and can be set manually for workflow-dispatch
+  group: ${{ github.workflow }}-${{ github.ref }}
+  # Cancel in-progress runs when a new workflow with the same group name is triggered
+  cancel-in-progress: true
+
+jobs:
+  performance-tests:
+    runs-on: ubuntu-latest
+    env:
+      # Number of run IDs to select
```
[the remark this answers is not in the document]
p-senichenkov: There's no straightforward way to filter workflow runs by recency. I could filter them manually, but it will result in quite a lot of code, and I don't think it's much better than N last runs.
Consider that this only affects number of entries displayed on plot, comparison is always performed against latest run. Also note that artifacts are being automatically deleted after 90 days (and we cannot increase this number for free).

p-senichenkov:
> In general, is there really no obvious way to request the last successful delivery for the required workflow, rather than running through the latest ones?

No, `actions/download-artifact` is not flexible enough to download artifacts from several workflow runs.

datasets/crop_datasets.sh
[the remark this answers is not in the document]
Vdaleke: It seems worth adding a comment about this to the script that cuts the datasets. Otherwise, this information may get lost, although of course it is intuitively clear why.

.github/workflows/performance-tests.yml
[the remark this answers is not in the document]
Vdaleke: I did some research on the topic and came to the conclusion that such functionality exists. You just need to use the GitHub CLI, not the GitHub REST-API. This is generally more convenient and it seems that this is exactly the approach that GitHub expects when working with workflow.

The solution that is worth using:
```
LAST_SUCCESS_RUN_ID=$(gh run list --workflow  performance-tests.yml --status success --limit 1 --json databaseId -q '.[0].databaseId // ""')
```
[gh run list docs](https://cli.github.com/manual/gh_run_list)
- It is not documented here that the results are displayed in the order of the most recent execution, except that the word recent at the beginning indicates this, but I think we can rely on this, it is quite reasonable behavior and such behavior is also expected from the limit flag. We can leave a comment about this.
- If no result is found, then nothing is written to the variable.

.github/workflows/performance-tests.yml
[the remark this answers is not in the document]
Vdaleke: > In general, is there really no obvious way to request the last successful delivery for the required workflow, rather than running through the latest ones?

See answer on this https://github.com/Desbordante/desbordante-core/pull/557/files#r2096253230

.github/workflows/performance-tests.yml
```diff
@@ -0,0 +1,111 @@
...
+          for (( i = 0; i < ${{ env.num-runs }}; i = $i + 1 )); do
+            ID=$(jq ".[$i]" ids.json)
+
+            curl -L \
+              -o artifacts.json \
+              -H "Accept: application/vnd.github+json" \
+              -H "X-GitHub-Api-Version: 2022-11-28" \
+              "https://api.github.com/repos/${{ github.repository }}/actions/runs/$ID/artifacts"
+
+            ARTIFACT_ID=$(jq '.artifacts.[] | select(.name == "results.json") | .id' artifacts.json)
```
Vdaleke, line 61: It seems that for loading artifacts into CI it is assumed that the download artifact action will be used. https://github.com/actions/download-artifact

Like this
```
      - name: Download artifact from last successful run
        uses: actions/download-artifact@v4
        id: download_previous_artifact
        if: steps.get_last_success_run.outputs.run_id != ''
        with:
          name: 
          run-id: ${{ steps.get_last_success_run.outputs.run_id }}
          path: 
        // continue-on-error: true  ?
```

src/tests/benchmark/benchmark_controller.h
Vdaleke: I think it is acceptable to create a Controller that has an immutable set of tests, but then according to the logic of the language these tests should be passed to the constructor.

Otherwise, it is worth adding test deletion. I can imagine a case in which this is necessary, for example, if we want to run tests in batches in one controller for some reason, for example, to analyze the results in intervals if the testing is very long. It seems that it may be convenient to use one Controller class for this rather than creating a new one each time.

Vdaleke:
> > In general, is there really no obvious way to request the last successful delivery for the required workflow, rather than running through the latest ones?
> 
> No, `actions/download-artifact` is not flexible enough to download artifacts from several workflow runs.

I suggested the option of loading only from the last successful one through it.

Vdaleke (commented):
> No, actions/download-artifact is not flexible enough to download artifacts from several workflow runs.

I suggested the option of loading only from the last successful one through it.

p-senichenkov:
> Yes, that's exactly what I was thinking. I think this option suits us, but maybe it's worth explicitly adding the -v1 suffix to artifact name now and adding a comment that if the json artifact changes incompatiblely, it's worth upgrading the version.

Another option is to add "version" field inside of artifact (like `<xml-version>` tag). It's less obvious, but can be useful if we wanted more fine-grained versions. Also, this seems more widely acceptable way than format version in file name.

.github/workflows/performance-tests.yml
p-senichenkov: As I've already mentioned, `download-artifact` is not flexile enough: it can download artifacts only from a single job and cannot be ran in a loop (there's no loops in YAML -- it's not a programming language). But GitHub CLI can be used here as well as in "`get run IDs`".

Vdaleke: > As I've already mentioned, `download-artifact` is not flexile enough: it can download artifacts only from a single job and cannot be ran in a loop (there's no loops in YAML -- it's not a programming language). But GitHub CLI can be used here as well as in "`get run IDs`".

Yes, I understand this perfectly well, that's why I wrote in the title of the review that this refers to the solution when we unload the artifact for only one run

Vdaleke:
> Another option is to add "version" field inside of artifact (like `<xml-version>` tag). It's less obvious, but can be useful if we wanted more fine-grained versions. Also, this seems more widely acceptable way than format version in file name.

Yes, this is also reasonable, but I think it is not worth doing this, because it is easier to use the original json-a format provided by some testing framework without doing its post-processing. And backward compatibility in theory should break very rarely.

src/tests/performance_tests/display_performance_tests.py
```diff
@@ -0,0 +1,52 @@
...
+    points = []
+    for res in results:
+        points.append(res.get(name, 0) / 1000)
+
+    _, ax = plt.subplots()
+    ax.stairs(points)
+    ax.set_title(name)
+    ax.set_xlabel('Run number')
+    ax.set_ylabel('Time, s')
+    pages.savefig()
```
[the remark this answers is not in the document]
p-senichenkov: Is it necessary to specify `figsize` here? As for me, default size looks fine. Also, one can adjust these values using `matplotlibrc` file if they're not specified here.

.github/workflows/performance-tests.yml
```diff
@@ -0,0 +1,133 @@
...
+            # Don't fail on error
+            echo "$ID"
+            gh run download "$ID" --name 'results-v1.json' --repo "${{ github.repository }}" && \
+              cat curr-results.json && \
+              mv curr-results.json old-results/$i.json || /bin/true
+
+            rm -rf results-v1.json
+          done
+
+      - name: Get latest log
```
Vdaleke, line 61: I don't quite understand why you need to download the last n artifacts to take the latest one? Can't you just use `--limit 1` filter to get it?

Vdaleke (commented):
Hi, I looked at the new iteration and I have a couple of questions and comments in general:

After the last calls and the previous iteration, do I understand correctly that only the logic in the CI part has changed and the python code has been refactored?

At the moment, the solution is ok, it works. We have outlined several points for improvement, but I think we can now merge this solution.

The only thing that is missing now is a guide in the readme or somewhere on how to download plots for a specific workflow - failed or not. If a large script is needed for this, I would put this script somewhere. If it is a couple of commands, I would write them somewhere.

.github/workflows/performance-tests.yml
Vdaleke: Okay, I saw that old results are needed to generate plots, overall this is reasonable, the issue is resolved

src/tests/benchmark/benchmark_runner.h
```diff
@@ -0,0 +1,103 @@
...
+
+        auto bm_body = [other_params] {
+            auto algo = algos::CreateAndLoadAlgorithm<Algo>(other_params);
+            algo->Execute();
+        };
+
+        RegisterBenchmark(name.str(), bm_body, threshold);
+    }
+
+    /// @brief Register a custom benchmark
```
Vdaleke, line 63: Since this is a public method, it is better to specify the descriptions of the parameters in the comment here too.

For all public methods below too.

src/tests/benchmark/benchmark_runner.h
```diff
@@ -0,0 +1,103 @@
...
+#include <vector>
+
+#include <boost/core/demangle.hpp>
+
+#include "algorithms/algo_factory.h"
+#include "config/names.h"
+#include "parser/csv_parser/csv_parser.h"
+
+namespace benchmark {
+struct Benchmark {
```
Vdaleke: This class is only used inside the Runner class, so I would leave it as nested.

Vdaleke: I would call this class BenchmarkData or something like that

src/tests/benchmark/fd_benchmark.h
```diff
@@ -0,0 +1,54 @@
...
+namespace benchmark {
+
+inline void FDBenchmark(BenchmarkRunner& runner) {
+    using namespace config::names;
+
+    runner.RegisterSimpleBenchmark<algos::hyfd::HyFD>(
+            tests::kIowa650k,
+            {{kThreads, static_cast<config::ThreadNumType>(1)},
+             {kMaximumLhs, static_cast<config::MaxLhsType>(2)}},
+            "", 25);
```
Vdaleke: It seems to me that passing the treshhold to the runner class is semantically incorrect.

I would either make it so that each benchmark function also passes a comparer class (optional), in which the treshhold for the benchmark is also registered (in that case the register function should return the name), or move this logic to some controller class, which would be the only one passed to such functions.

Otherwise, it is simply unexpected that the runner contains treshholds and then returns them. The treshhold is an object for the comparer, but not for the runner.

src/tests/benchmark/benchmark_runner.h
```diff
@@ -0,0 +1,103 @@
...
+public:
+    /// @brief Register a benchmark that simply calls and measures Algo.Execute().
+    /// @param csv -- @c CSVConfig to run algo on
+    /// @param other_params -- @c StdParamsMap with other algo options.
+    /// It may or may not contain @c CSVConfig -- @c csv will be used anyway
+    /// @param name_suffix -- suffix that will be added to benchmark name.
+    /// Can be used to distinguish different runs of a single algo
+    /// @param threshold -- threshold (per cent)
+    template <typename Algo>
+    void RegisterSimpleBenchmark(CSVConfig const& csv, algos::StdParamsMap&& other_params = {},
```
Vdaleke: I think it would be useful to return the generated name here

Vdaleke (changes requested):
I propose several changes, mainly I want to rework the class architecture, namely the fact that the treshhold is put and stored in the runner, although the treshhold has nothing to do with the run.

src/tests/benchmark/main.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include "benchmark_runner.h"
+#include "dd_benchmark.h"
+#include "fd_benchmark.h"
+#include "ind_benchmark.h"
+#include "md_benchmark.h"
+#include "nar_benchmark.h"
+
+INITIALIZE_EASYLOGGINGPP
+
+// Expected arguments (both are optional):
```
Vdaleke: It's strange here that previous results are required to save new results?

There should be an option to save and not compare

src/tests/benchmark/README.md
Vdaleke, line 1: need to update readme with comparer class

Vdaleke (changes requested):

PR #558, opened 2025-04-21
easylogging replaced with spdlog

Tydik42:
## Changes
- Removed all EasyLogging++ dependencies and replaced with spdlog
- Added spdlog as header only library
- Rewrote all logging functionality to use spdlog
- Implemented new Python logging module
- Updated README.md documentation to reflect logging changes


src/python_bindings/py_util/logging.cpp
```diff
@@ -0,0 +1,50 @@
...
+    auto logging_module = main_module.def_submodule("logging");
+
+    spdlog::set_pattern("[%Y-%m-%d %H:%M:%S.%e] [%^%l%$] %v");
+    spdlog::set_level(spdlog::level::off);  // Default value
+
+    auto set_log_level = [](std::string const& level) {
+        static std::unordered_map<std::string, spdlog::level::level_enum> const kLevels = {
+                {"trace", spdlog::level::trace}, {"debug", spdlog::level::debug},
+                {"info", spdlog::level::info},   {"warn", spdlog::level::warn},
+                {"error", spdlog::level::err},   {"off", spdlog::level::off}};
```
BUYT-1, line 23: Bind the spdlog enum directly: https://pybind11.readthedocs.io/en/stable/classes.html#enumerations-and-internal-types

src/python_bindings/py_util/logging.cpp
```diff
@@ -0,0 +1,50 @@
...
+        if (auto const it = kLevels.find(level); it != kLevels.end()) {
+            spdlog::set_level(it->second);
+        } else {
+            throw py::value_error(
+                    "Invalid log level. Available options: "
+                    "'trace', 'debug', 'info', 'warn', 'error', 'off'");
+        }
+    };
+
+    logging_module.def("set_level", set_log_level, py::arg("level"),
```
BUYT-1, line 34:
```suggestion
    logging_module.def("set_level", &spdlog::set_level, py::arg("level"),
```

BUYT-1 (changes requested):

datasets/datasets.zip
wildsor, line 1: This shouldn't be here

src/python_bindings/py_util/logging.cpp
```diff
@@ -0,0 +1,228 @@
...
+    py::module_::import("atexit").attr("register")(py::cpp_function(CleanupAtExit));
+
+    py::class_<LoggerProxy>(
+            m, "Logger",
+            "A proxy to a standard Python logger that syncs its level with the C++ core.")
+            .def(py::init([](py::object logger_obj) { return LoggerProxy{std::move(logger_obj)}; }),
+                 py::arg("logger"))
+
+            .def(
+                    "setLevel",
```
wildsor:
```suggestion
                    "set_level",
```

BUYT-1 (changes requested):
Use Python naming conventions for Python

src/core/algorithms/association_rules/ar_algorithm.cpp
```diff
@@ -66,7 +66,13 @@ unsigned long long ARAlgorithm::ExecuteInternal() {
     auto time = FindFrequent();
     time += GenerateAllRules();
 
-    LOG(INFO) << "> Count of association rules: " << ar_collection_.size();
+
+    LOG_TRACE("trace test");
+    LOG_DEBUG("debug test");
+    LOG_INFO("info test");
+    LOG_WARN("warn test");
+    LOG_ERROR("error test");
```
wildsor: extra code

src/core/algorithms/dc/FastADC/model/pli_shard.cpp
```diff
@@ -103,8 +103,8 @@ std::vector<size_t> PliShardBuilder::ColumnToHash(model::TypedColumnData const&
             ColumnToHashTyped<std::string>(hashed_column, column);
             break;
         default:
-            LOG(DEBUG) << "Column type " << column.GetType().ToString()
-                       << " is not supported for FastADC's PLI sharding";
+            LOG_DEBUG("Column type " + column.GetType().ToString() +
+                          " is not supported for FastADC's PLI sharding");
```
wildsor: Use spdlog formatting instead of concatenation

src/core/algorithms/dc/verifier/dc_verifier.cpp
```diff
@@ -71,7 +71,7 @@ unsigned long long int DCVerifier::ExecuteInternal() {
         dc::DCParser parser = dc::DCParser(dc_string_, relation_.get(), data_);
         dc = parser.Parse();
     } catch (std::exception const& e) {
-        LOG(INFO) << e.what();
+        LOG_INFO(e.what());
```
wildsor: In case `e.what()` contains arbitrary text, this is unsafe (suppose the text contains `{}`, then it will try to get an argument, leading to UB), change:
```suggestion
        LOG_INFO("{}", e.what());
```

src/core/algorithms/ucc/hpivalid/hpivalid.cpp
```diff
@@ -77,16 +77,15 @@ void HPIValid::RegisterUCCs(hpiv::ResultCollector const& rc) {
 }
 
 void HPIValid::PrintInfo(hpiv::ResultCollector const& rc) const {
-    LOG(INFO) << "Minimal UCCs: " << rc.UCCs();
-    LOG(DEBUG) << "Mined UCCs:";
+    LOG_INFO("Minimal UCCs: {}", rc.UCCs());
+    LOG_DEBUG("Mined UCCs:");
     for (model::UCC const& ucc : UCCList()) {
-        LOG(DEBUG) << ucc.ToString();
+        LOG_DEBUG(ucc.ToString());
```
wildsor:
```suggestion
        LOG_DEBUG("{}", ucc.ToString());
```

src/core/algorithms/ucc/hyucc/hyucc.cpp
```diff
@@ -55,9 +55,9 @@ unsigned long long HyUCC::ExecuteInternal() {
     auto uccs = ucc_tree->FillUCCs();
     RegisterUCCs(std::move(uccs), og_mapping);
 
-    LOG(DEBUG) << "Mined UCCs:";
+    LOG_DEBUG("Mined UCCs:");
     for (model::UCC const& ucc : UCCList()) {
-        LOG(DEBUG) << ucc.ToString();
+        LOG_DEBUG(ucc.ToString());
```
wildsor:
```suggestion
        LOG_DEBUG("{}", ucc.ToString());
```

src/core/model/table/agree_set_factory.cpp
```diff
@@ -82,11 +82,11 @@ AgreeSetFactory::SetOfAgreeSets AgreeSetFactory::GenAsUsingVectorOfIdSets() cons
...
     auto elapsed_mills_to_gen_id_sets = std::chrono::duration_cast<std::chrono::milliseconds>(
             std::chrono::system_clock::now() - start_time);
-    LOG(INFO) << "TIME TO IDENTIFIER SETS GENERATION: " << elapsed_mills_to_gen_id_sets.count();
+    LOG_INFO("TIME TO IDENTIFIER SETS GENERATION: {}", elapsed_mills_to_gen_id_sets.count());
 
-    LOG(DEBUG) << "Identifier sets:";
+    LOG_DEBUG("Identifier sets:");
     for (auto const& id_set : identifier_sets) {
-        LOG(DEBUG) << id_set.ToString();
+        LOG_DEBUG(id_set.ToString());
```
wildsor:
```suggestion
        LOG_DEBUG("{}", id_set.ToString());
```

src/core/util/auto_join_thread.h
```diff
@@ -28,7 +28,7 @@ class AutoJoinThread {
             t_.join();
         }
     } catch (std::system_error const& e) {
-        LOG(ERROR) << e.what();
+        LOG_ERROR(e.what());
```
wildsor:
```suggestion
        LOG_ERROR("{}", e.what());
```

src/tests/unit/test_data_stats.cpp
```diff
@@ -392,7 +392,7 @@ TEST(TestDataStats, TestShowSample) {
         for (auto const &str : row) {
             result << str << " \t";
         }
-        LOG(INFO) << result.str();
+        LOG_INFO(result.str());
```
wildsor:
```suggestion
        LOG_INFO("{}", result.str());
```

src/tests/unit/test_data_stats.cpp
```diff
@@ -401,7 +401,7 @@ TEST(TestDataStats, TestShowAllStats) {
     auto stats_ptr = MakeStatAlgorithm(kTestDataStats);
     algos::DataStats &stats = *stats_ptr;
     stats.Execute();
-    LOG(INFO) << stats.ToString();
+    LOG_INFO(stats.ToString());
```
wildsor:
```suggestion
    LOG_INFO("{}", stats.ToString());
```

src/core/algorithms/dc/FastADC/misc/typed_column_data_value_differences.cpp
```diff
@@ -68,9 +67,9 @@ double GetSharedPercentage(model::TypedColumnData const& c1, model::TypedColumnD
         case model::TypeId::kString:
             return GetSharedPercentageTyped<std::string>(c1, c2);
         default:
-            LOG(DEBUG) << "Column " << c1.GetColumn()->ToString() << " with type "
-                       << c1.GetType().ToString()
-                       << " is not supported for shared percentage calculation";
+            LOG_DEBUG("Column " + c1.GetColumn()->ToString() + " with type " +
+                      c1.GetType().ToString() +
+                      " is not supported for shared percentage calculation");
```
wildsor:
```suggestion
            LOG_DEBUG("Column {} with type {} is not supported for shared percentage calculation", c1.GetColumn()->ToString(), c1.GetType().ToString());
```

src/python_bindings/py_util/logging.cpp
Tydik42: This naming style is used in [logging module](https://docs.python.org/3/library/logging.html#logging.Logger.setLevel)

wildsor: https://stackoverflow.com/questions/22993667/how-come-the-pythons-logging-module-doesnt-follow-pep8-conventions
We're not Python's `logging` module.

BUYT-1:
` Allow runtime log level control from Python` is the commit that replaces easyloggingpp, rename it to reflect that

src/core/algorithms/fd/fd_verifier/dynamic_stats_calculator.cpp
```diff
@@ -5,7 +5,7 @@
 #include <numeric>
 #include <unordered_map>
 
-#include <easylogging++.h>
+#include "util/logger.h"
```
wildsor: No log statements in this file, might as well remove it, same for other files without log statements

src/core/algorithms/fd/pyro/pyro.cpp
```diff
@@ -85,7 +84,7 @@ unsigned long long Pyro::ExecuteInternal() {
 
     auto const work_on_search_space =
             [this, &progress_step](std::list<std::unique_ptr<SearchSpace>>& search_spaces,
-                                   ProfilingContext* profiling_context, int id) {
+                                   ProfilingContext* profiling_context, [[maybe_unused]] int id) {
```
wildsor:
```suggestion
                                   ProfilingContext* profiling_context, int id) {
```

src/core/algorithms/dc/model/column_operand.h
```diff
@@ -7,12 +7,11 @@
 #include <string>
 #include <vector>
 
-#include <easylogging++.h>
-
 #include "algorithms/dc/model/tuple.h"
 #include "model/table/column.h"
 #include "model/table/relation_data.h"
 #include "model/types/type.h"
+#include "util/logger.h"
```
wildsor, line 14: No logging statements in this file

src/core/algorithms/fd/hyfd/validator.cpp
```diff
@@ -8,11 +8,11 @@
 #include <boost/asio/post.hpp>
 #include <boost/asio/thread_pool.hpp>
 #include <boost/dynamic_bitset.hpp>
-#include <easylogging++.h>
 
 #include "algorithms/fd/hycommon/util/pli_util.h"
 #include "algorithms/fd/hycommon/validator_helpers.h"
 #include "hyfd_config.h"
+#include "util/logger.h"
```
wildsor:
```suggestion
```

src/core/model/table/dynamic_position_list_index.cpp
```diff
@@ -9,7 +9,8 @@
 #include <utility>
 
 #include <boost/dynamic_bitset.hpp>
-#include <easylogging++.h>
+
+#include "util/logger.h"
```
wildsor:
```suggestion
```

src/python_bindings/py_util/py_to_any.cpp
```diff
@@ -2,10 +2,10 @@
 #include <unordered_map>
 
 #include <boost/any.hpp>
-#include <easylogging++.h>
 #include <pybind11/pybind11.h>
 #include <pybind11/stl.h>
 #include <pybind11/stl/filesystem.h>
+#include <spdlog/spdlog.h>
```
wildsor:
```suggestion
```

src/core/util/logger.h
```diff
@@ -0,0 +1,52 @@
...
+    spdlog::register_logger(logger);
+}
+
+inline std::shared_ptr<spdlog::logger> GetLogger(std::string const& logger_name = "desbordante") {
+    return spdlog::get(logger_name);
+}
+
+}  // namespace util::logging
+
+#define GET_CACHED_LOGGER()                                            \
```
wildsor: Does this really belong in the public API? Why not `#undef` it at the end of the file?

src/python_bindings/py_util/logging.cpp
```diff
@@ -0,0 +1,228 @@
...
+        }
+
+    } catch (py::error_already_set const& e) {
+        py::print("ERROR: Error during Python logging setup:", e.what());
+    }
+}
+
+void CleanupAtExit() {
+    g_is_shutting_down = true;
+    g_proxy_instance = py::none();
```
wildsor: `None` is a Python object with its own reference count, so this might be somewhat dangerous. Preferably, use default-constructed `py::object` instead.

src/core/util/logger.h
```diff
@@ -0,0 +1,52 @@
...
+#include <vector>
+
+#include <spdlog/sinks/stdout_color_sinks.h>
+#include <spdlog/spdlog.h>
+
+namespace util::logging {
+
+inline void Initialize(std::string const& logger_name = "desbordante",
+                       std::vector<spdlog::sink_ptr> sinks = {}) {
+    if (spdlog::get(logger_name)) {
```
wildsor, line 14: This part means the function only ensures that the logger is initialized, so name it `EnsureInitialized`.

src/python_bindings/py_util/logging.cpp
```diff
@@ -0,0 +1,228 @@
...
+        }
+
+        spdlog::memory_buf_t formatted;
+        this->formatter_->format(msg, formatted);
+        std::string_view payload(formatted.data(), formatted.size());
+
+        if (!payload.empty() && payload.back() == '\n') {
+            payload.remove_suffix(1);
+        }
+
```
wildsor, line 81: By the way, what will this do in free-threaded Python?
```suggestion

py::gil_scoped_acquire gil
```

src/python_bindings/py_util/logging.cpp
```diff
@@ -0,0 +1,228 @@
...
+}
+
+}  // namespace
+
+namespace python_bindings {
+
+void BindLogging(py::module_& main_module) {
+    auto m = main_module.def_submodule("logging",
+                                       "Logging configuration for the desbordante library.");
+
```
wildsor: Why not do this instead:
1) logging.getLogger("desbordante")
2) Construct PythonSink with this object
3) Initialize spdlog
4) Set the minimum logging level in the appropriate places so all messages go to the sink

In PythonSink's `sink_it_`, check the current level of the Python logger (`.level` attribute), and don't format and call its `log` method if it's too high.

You won't need to actually bind anything in that case and can do it in a separate function in bindings.cpp

BUYT-1 (changes requested):

src/core/util/logger.h
```diff
@@ -0,0 +1,83 @@
...
+    logger->set_level(spdlog::level::trace);
+
+    spdlog::register_logger(logger);
+}
+
+inline std::shared_ptr<spdlog::logger> GetLogger(std::string const& logger_name = "desbordante") {
+    auto logger = spdlog::get(logger_name);
+    if (!logger) {
+        EnsureInitialized(logger_name);
+        logger = spdlog::get(logger_name);
```
wildsor, line 34: nit:
```suggestion
        logger = spdlog::get(logger_name);
        assert(logger);
```

src/core/util/logger.h
```diff
@@ -0,0 +1,83 @@
...
+    auto logger = std::make_shared<spdlog::logger>(logger_name, sinks.begin(), sinks.end());
+    logger->set_pattern("[%Y-%m-%d %H:%M:%S.%e] [%n] [%^%l%$] %v");
+
+    logger->set_level(spdlog::level::trace);
+
+    spdlog::register_logger(logger);
+}
+
+inline std::shared_ptr<spdlog::logger> GetLogger(std::string const& logger_name = "desbordante") {
+    auto logger = spdlog::get(logger_name);
```
wildsor: nit:
```suggestion
    std::shared_ptr<spdlog::logger> logger = spdlog::get(logger_name);
```

src/core/algorithms/od/set_based_verifier/verifier.cpp
```diff
@@ -2,8 +2,6 @@
 
 #include <stdexcept>
 
-#include <easylogging++.h>
```
wildsor, line 5: This works because util/logger.h is transitively included, we don't rely on that, include explicitly

src/python_bindings/py_util/logging.cpp
```diff
@@ -0,0 +1,127 @@
...
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+#include <spdlog/sinks/base_sink.h>
+#include <spdlog/spdlog.h>
+
+#include "util/logger.h"
+
+namespace {
+namespace py = pybind11;
+bool g_is_shutting_down = false;
```
wildsor: Why do we need this check? Add a comment explaining that

src/python_bindings/py_util/logging.cpp
```diff
@@ -0,0 +1,127 @@
...
+        if (!py::hasattr(logging, "TRACE")) {
+            logging.attr("addLevelName")(kPyTraceLevel, "TRACE");
+            logging.add_object("TRACE", py::int_(kPyTraceLevel));
+        }
+
+        py::object py_logger = logging.attr("getLogger")("desbordante");
+
+        py::object root_logger = logging.attr("getLogger")();
+        if (py::len(py_logger.attr("handlers")) == 0 &&
+            py::len(root_logger.attr("handlers")) == 0) {
```
wildsor: Why are we adding a handler in this case?

src/python_bindings/py_util/logging.cpp
```diff
@@ -0,0 +1,127 @@
...
+constexpr int kPyTraceLevel = 5;
+
+template <typename Mutex>
+class PythonLoggerSink : public spdlog::sinks::base_sink<Mutex> {
+public:
+    explicit PythonLoggerSink(py::object py_logger) : py_logger_(std::move(py_logger)) {}
+
+protected:
+    void sink_it_(spdlog::details::log_msg const& msg) override {
+        if (g_is_shutting_down || py_logger_.is_none()) {
```
wildsor: Why are you checking for None? Also, doesn't this check also need the GIL?

src/python_bindings/py_util/logging.cpp
Tydik42: This block provides a default configuration to improve user experience. Without it, C++ logs would be silently discarded if the user hasn't configured Python's logging module, which can be confusing. This conditions ensures we only add this default handler if the user hasn't already configured any handlers themselves (either on our logger or the root logger), so it won't interfere with custom setups. I'll add this in the comments too.

src/python_bindings/py_util/logging.cpp
Tydik42: This is a defensive check to prevent a segfault if the Python logger object becomes invalid for any reason, particularly during interpreter shutdown.  And you are right, the `is_none()` check does require the GIL, thanks!

src/python_bindings/py_util/logging.cpp
wildsor: https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library

> It is strongly advised that you do not add any handlers other than [NullHandler](https://docs.python.org/3/library/logging.handlers.html#logging.NullHandler) to your library’s loggers. This is because the configuration of handlers is the prerogative of the application developer who uses your library. The application developer knows their target audience and what handlers are most appropriate for their application: if you add handlers ‘under the hood’, you might well interfere with their ability to carry out unit tests and deliver logs which suit their requirements.

An advanced user wouldn't appreciate us messing with their configuration, let's remove this section.

src/python_bindings/py_util/logging.cpp
```diff
@@ -0,0 +1,135 @@
...
+                break;
+            case spdlog::level::critical:
+                msg_py_level = 50;
+                break;
+            default:
+                msg_py_level = 0;
+                break;
+        }
+
+        if (msg_py_level < py_effective_level) {
```
wildsor: Use `isEnabledFor`: https://docs.python.org/3/howto/logging.html#optimization

src/python_bindings/py_util/logging.cpp
wildsor: Why would an object become `None`, a different Python value, during shutdown?

src/python_bindings/py_util/logging.cpp
```diff
@@ -0,0 +1,135 @@
...
+            py::object handler = logging.attr("StreamHandler")();
+            py::object formatter = logging.attr("Formatter")(
+                    "%(asctime)s [%(name)s] [%(levelname)-8s] %(message)s", "%Y-%m-%d %H:%M:%S");
+            handler.attr("setFormatter")(formatter);
+            py_logger.attr("addHandler")(handler);
+            py_logger.attr("setLevel")(logging.attr("INFO"));
+        }
+
+        auto python_sink = std::make_shared<PythonSink>(py_logger);
+        ::util::logging::EnsureInitialized("desbordante", {python_sink});
```
wildsor:
```suggestion
        auto python_sink = std::make_shared<PythonSink>(std::move(py_logger));
        ::util::logging::EnsureInitialized("desbordante", {std::move(python_sink)});
```

src/python_bindings/py_util/logging.cpp
Tydik42: You are right, the object wouldn't transform into None during shutdown. My `is_none()` check was intended as a general guard against the handle pointing to a None value (which would cause an AttributeError), though this is admittedly an edge case.

So a more critical danger is a null C++ handle, the current code doesn't check for that.  So I'll add a check for `!py_logger_` to make sure that doesn't happen.

wildsor: Why would any of this happen? Is there a situation where this check would help? From what I see, something would have to go catastrophically wrong for a situation like that to arise, so I say just let it crash or error out immediately.

src/python_bindings/py_util/logging.cpp
```diff
@@ -0,0 +1,122 @@
...
+        std::string_view payload(msg.payload.data(), msg.payload.size());
+        (void)py_logger_.attr("log")(msg_py_level, payload);
+    }
+
+    void flush_() override {
+        py::gil_scoped_acquire gil;
+        if (g_is_shutting_down || !py_logger_ || py_logger_.is_none()) {
+            return;
+        }
+        (void)py_logger_.attr("handlers").attr("flush")();
```
BUYT-1: `.handlers` is a `list`, this doesn't work. Why would we do that in the first place? The handlers are Python's domain, I don't see a reason to mess with them.

src/python_bindings/py_util/logging.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+}  // namespace
+
+namespace python_bindings {
+
+void BindLogging(py::module_&) {
+    static std::once_flag once;
+    std::call_once(once, SetupLoggingBridge);
+
+    py::module_::import("atexit").attr("register")(py::cpp_function(CleanupAtExit));
```
BUYT-1: You might want to put this under `std::call_once` too. I think the function can be called another time if [importlib.reload](https://docs.python.org/dev/library/importlib.html#importlib.reload) is called

wildsor: This is not going to work with subinterpreters.

From https://docs.python.org/3/library/atexit.html:
> When used with C-API subinterpreters, registered functions are local to the interpreter they were registered in.

src/python_bindings/py_util/logging.cpp
Tydik42: Fair point. I'll leave only `g_is_shutting_down`.

src/python_bindings/py_util/logging.cpp
wildsor: This function is called every time a module is imported. This means it needs to work with subinterpreters too. Let's assume it's only ever called once per subinterpreter.

https://pybind11.readthedocs.io/en/stable/advanced/misc.html#concurrency-and-parallelism-in-python-with-pybind11

I propose:
* Create a custom sink that stores objects that call `.log`, assign one to each subinterpreter (i.e. every time this function is called)
* On each logging call, find the current subinterpreter's object (`py::subinterpreter::current.id()`), then execute the call on it
* I think we can get away with still caching the Python's logger object
* In the cleanup function, we remove the object for the current subinterpreter, acquiring exclusive lock on the sink

I guess storing in `std::unordered_map` with a `std::shared_mutex` works well enough. But maybe spdlog does its own locking, have a look at the source code there. You get the "it's only done once" check for free in this setup, just use `try_emplace`.

Also, you can pass objects, like the sink, to Python's atexit if you need to (I don't think you do in this case).

src/python_bindings/py_util/logging.cpp
wildsor: I think the situation you want to prevent is impossible. The objects in the library are only ever going to log when their methods are called, but if their methods are called, the interpreter is still active.

src/python_bindings/py_util/logging.cpp
```diff
@@ -0,0 +1,128 @@
...
+    void flush_() override {
+        py::gil_scoped_acquire gil;
+        if (g_is_shutting_down) {
+            return;
+        }
+        for (auto handler : py_logger_.attr("handlers")) {
+            if (py::hasattr(handler, "flush")) {
+                handler.attr("flush")();
+            }
+        }
```
wildsor: I don't think this is needed.

wildsor: Whoops, forgot to finish this one. Handlers are Python's domain and we don't interact with them in any other place in the code, so it doesn't make sense to do that here.

"Flush" methods are meant to push through the lines that are stuck buffered and would be gone if the logger just stopped. That is not the case here as we push the relevant lines through to Python immediately. It would have been useful to implement this method if we instead held up some lines and only called `.log` on several in a row to, for example, avoid calling Python too much. We need to make 2 hash-table accesses + actually format the message on each call, so it is something to consider in the future.

src/python_bindings/py_util/logging.cpp
Tydik42: Thanks! I'll change this. But to use `py::subinterpreter::current.id()` we will need to move to pybind11 [3.0.0](https://github.com/pybind/pybind11/releases/tag/v3.0.0) version. There won't be any problems with this?

src/python_bindings/py_util/logging.cpp
Tydik42: Agreed, I'll leave the implementation empty.

src/python_bindings/py_util/logging.cpp
wildsor: This part will need a workaround for versions before 3.12. Wrap that call in a function and if `PYBIND11_HAS_SUBINTERPRETER_SUPPORT` is not defined, just return 0 in it and don't include the subinterpreter.h header.

wildsor: Everything should be fine in other aspects, you can just bump the version in a separate commit.

src/python_bindings/py_util/logging.cpp
Tydik42: I forgot to reply to this, sorry. This is true for synchronous calls, but the library might define background threads or global objects with destructors that attempt to log after the Python interpreter has started finalizing

src/python_bindings/py_util/logging.cpp
```diff
@@ -0,0 +1,185 @@
+#include "logging.h"
```
wildsor: nit:
```suggestion
#include "py_util/logging.h"
```

src/python_bindings/py_util/logging.cpp
wildsor: Maybe. I think that scenario is far-fetched for a library that is, most broadly, supposed to take some data and spit out something built using it and some parameters, but perhaps in some far away future.

src/python_bindings/py_util/logging.cpp
```diff
@@ -0,0 +1,185 @@
...
+    }
+}
+
+void CleanupAtExit() {
+#ifdef PYBIND11_HAS_SUBINTERPRETER_SUPPORT
+    if (py::subinterpreter::current().id() == py::subinterpreter::main().id()) {
+        g_main_interpreter_shutting_down.store(true, std::memory_order_relaxed);
+    }
+#else
+    g_main_interpreter_shutting_down.store(true, std::memory_order_relaxed);
```
wildsor: Relaxed memory order pretty much only guarantees no UB, the order of events between threads can be whatever it happens to be. I think you want release-acquire, but I say don't bother and just use the default order.

src/python_bindings/py_util/logging.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+#include "util/logger.h"
+
+namespace {
+namespace py = pybind11;
+
+// A global flag set by the Python atexit handler to signal that the
+// interpreter is shutting down. The logging sink checks this flag to prevent
+// unsafe calls back into the Python C-API during this critical phase.
+std::atomic<bool> g_main_interpreter_shutting_down{false};
```
wildsor: What's the idea with the "g_" prefix?

src/python_bindings/py_util/logging.cpp
```diff
@@ -0,0 +1,185 @@
...
+        if (!py::hasattr(logging, "TRACE")) {
+            logging.attr("addLevelName")(kPyTraceLevel, "TRACE");
+            logging.add_object("TRACE", py::int_(kPyTraceLevel));
+        }
+
+        py::object py_logger = logging.attr("getLogger")("desbordante");
+        if (py::len(py_logger.attr("handlers")) == 0) {
+            py::object handler = logging.attr("NullHandler")();
+            py_logger.attr("addHandler")(handler);
+        }
```
wildsor: The default handler outputs messages with levels WARNING and above to stderr. I checked the messages in the library, and they do belong in warnings, so let's not silence them.

src/python_bindings/py_util/logging.cpp
```diff
@@ -0,0 +1,185 @@
...
+class MultiTenantPythonSink : public spdlog::sinks::base_sink<Mutex> {
+public:
+    MultiTenantPythonSink() = default;
+
+    void RegisterLogger(py::object logger) {
+        if (g_main_interpreter_shutting_down.load(std::memory_order_relaxed)) {
+            return;
+        }
+
+        py::gil_scoped_acquire gil;
```
wildsor: You already hold the lock if you're in this function. You can put an assert if you want

src/python_bindings/py_util/logging.cpp
```diff
@@ -0,0 +1,185 @@
...
+        InterpreterId id = GetCurrentInterpreterId();
+        if (id < 0) return;
+
+        std::unique_lock lock(registry_mutex_);
+        interpreters_loggers_.try_emplace(id, std::move(logger));
+    }
+
+    void UnregisterLogger() {
+        py::gil_scoped_acquire gil;
```
wildsor: same here, you already have the lock

src/python_bindings/py_util/logging.cpp
```diff
@@ -0,0 +1,185 @@
...
+        if (py::len(py_logger.attr("handlers")) == 0) {
+            py::object handler = logging.attr("NullHandler")();
+            py_logger.attr("addHandler")(handler);
+        }
+
+        auto python_sink = GetGlobalPythonSink();
+        python_sink->RegisterLogger(std::move(py_logger));
+
+        ::util::logging::EnsureInitialized("desbordante", {std::move(python_sink)});
+
```
wildsor: nit:
```suggestion
```

src/python_bindings/py_util/logging.cpp
Tydik42: Just to mark the variable as global and separate it from others

wildsor: Why? You don't need to mark global variables.

src/python_bindings/py_util/logging.cpp
wildsor: Okay, I've [read](https://peps.python.org/pep-0788/) a bit. It looks like you're trying to solve a problem that is unsolvable with the current Python API. In addition to that, as the library is right now (and likely will keep being in the future), it really shouldn't start background threads or store globals with logging destructors, as its behavior is literally "take data, give result, then lay dormant". Therefore, it's pointless to account for this case, just remove it.

src/python_bindings/py_util/logging.cpp
```diff
@@ -0,0 +1,178 @@
...
+            case spdlog::level::err:
+                msg_py_level = 40;
+                break;
+            case spdlog::level::critical:
+                msg_py_level = 50;
+                break;
+            default:
+                msg_py_level = 0;
+                break;
+        }
```
wildsor: nit: move into a map

CMakeLists.txt
```diff
@@ -324,13 +321,7 @@ include_directories(
...
 add_subdirectory("src/core")
 
-# Configure logs
-add_subdirectory("cfg")
-
-# --- Optional Modules Setup ---
-
-# Configure tests
-if(COMPILE_TESTS)
+if (COMPILE_TESTS)
```
Vdaleke: The last time CMake formatting was done using the cmake-format tool. We don't have it in CI right now. There are no spaces between the if and the parenthesis, just like in other parts of the cmake code.

src/python_bindings/py_util/logging.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+        py::object target_logger;
+        {
+            std::shared_lock lock(registry_mutex_);
+            auto it = interpreters_loggers_.find(id);
+            if (it == interpreters_loggers_.end()) {
+                return;
+            }
+            target_logger = it->second;
+        }
```
wildsor: nit: move into `GetCurrentSubinterpreterLogger`
tip: py::object{} is an invalid object and can indicate failure

src/python_bindings/py_util/logging.cpp
```diff
@@ -0,0 +1,178 @@
...
+                break;
+            case spdlog::level::critical:
+                msg_py_level = 50;
+                break;
+            default:
+                msg_py_level = 0;
+                break;
+        }
+
+        if (!target_logger.attr("isEnabledFor")(msg_py_level).cast<bool>()) {
```
wildsor: nit:
for maximum clarity, give it a name
```suggestion
        if (!ShouldLog(target_logger)) {
```

src/python_bindings/py_util/logging.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+private:
+    std::shared_mutex registry_mutex_;
+    std::unordered_map<InterpreterId, py::object> interpreters_loggers_;
+};
+
+using PythonSink = MultiTenantPythonSink<spdlog::details::null_mutex>;
+
+std::shared_ptr<PythonSink> GetGlobalPythonSink() {
+    static std::shared_ptr<PythonSink> sink = std::make_shared<PythonSink>();
```
wildsor: (obvious type, see the styleguide)
```suggestion
    static auto sink = std::make_shared<PythonSink>();
```

src/python_bindings/py_util/logging.cpp
```diff
@@ -0,0 +1,178 @@
...
+        }
+
+        py::object py_logger = logging.attr("getLogger")("desbordante");
+
+        auto python_sink = GetGlobalPythonSink();
+        python_sink->RegisterLogger(std::move(py_logger));
+
+        ::util::logging::EnsureInitialized("desbordante", {std::move(python_sink)});
+    } catch (py::error_already_set const& e) {
+        py::print("ERROR: Error during Python logging setup:", e.what());
```
wildsor: While it may make sense to work without logging in theory, I'm not sure that would ever be wanted. Let's reraise
```suggestion
        py::print("ERROR: Error during Python logging setup:", e.what());
        throw;
```

wildsor: And also, print to sys.stderr

CMakeLists.txt
```diff
@@ -99,12 +99,6 @@ set(CMAKE_COLOR_DIAGNOSTICS ON)
 set(CMAKE_CXX_STANDARD 20)
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/target")
 
-# -DELPP_THREAD_SAFE -- for easylogging++ thread safety
-set(BUILD_OPTS "-DELPP_THREAD_SAFE")
-if(BUILD_NATIVE)
-    string(JOIN ";" BUILD_OPTS "${BUILD_OPTS}" "-march=native")
```
wildsor, line 105: Why is this gone?

src/python_bindings/py_util/logging.cpp
Tydik42: Thanks! I'll remove this logic

CMakeLists.txt
Tydik42: It was a mistake, sorry. I'll get it back

CMakeLists.txt
Vdaleke: Since this PR removes the limitation, the maximum CMake version should be removed here and above.

If possible, I would also remove the forced use of CMake 3.x in GitHub Actions in this PR

In the PR description add "Resolves #549".

Tydik42: Thank you, I'll change the versions now

.github/composite-actions/install-dependencies/action.yml
```diff
@@ -67,11 +67,10 @@ runs:
           python3 -m venv .venv
       shell: bash
 
-    # Add-hoc for https://github.com/Desbordante/desbordante-core/issues/549
-    - name: Install cmake 3.31.6 in python venv
+    - name: Install cmake in python venv
```
Vdaleke: It's not that simple, actually.

I'd like to see a partial revert of two commits:
https://github.com/Desbordante/desbordante-core/commit/a3d05dad2e96af79af7d7ccacf1220545bb17bb9
and
https://github.com/Desbordante/desbordante-core/commit/0c66f28b4fd5be71ca5aae6961ea30b1b53be1f2

Install CMake through package managers, remove restrictions in pyproject.toml, remove the virtual environment creation in install-dependecies, and add its creation to the binding tests (keep the name ".venv").

Tydik42: Thanks! I'll fix it now

BUYT-1 (approved):
nit:
> The old CMake version requirement was previously imposed by easylogging++.

Max line length in commit body is 72 characters

PR #559, opened 2025-04-23
Cinderella

rpuwa:


ol-imorozko:
If this PR is not ready for review, I suggest marking it as "Draft".

src/core/algorithms/cind/condition.h
```diff
@@ -0,0 +1,86 @@
...
+};
+}  // namespace algos::cind
+
+template <>
+struct std::hash<algos::cind::Condition> {
+    size_t operator()(algos::cind::Condition const& cond) const noexcept {
+        size_t result = 0;
+        boost::hash_combine(result, boost::hash_value(cond.validity));
+        boost::hash_combine(result, boost::hash_value(cond.completeness));
+        boost::hash_combine(result, boost::hash_value(cond.completeness));
```
Oddin60F: Do we really need to hash the same thing twice?

src/core/algorithms/cind/cind.h
```diff
@@ -0,0 +1,49 @@
...
+    model::IND const& ind;
+    std::vector<Condition> conditions;
+    std::vector<std::string> conditional_attributes;
+
+    std::string ToString() const {
+        std::string result = ind.ToLongString();
+        result.append("\nPossible conditions number: ");
+        result.append(std::to_string(conditions.size()));
+        result.append("\n");
+        if (conditions.size()) {
```
Oddin60F, line 18: Using the `empty()` function will be more correct.
```suggestion
        if (!conditions.empty()) {
```

src/core/algorithms/cind/condition_miners/itemset_node.h
```diff
@@ -0,0 +1,134 @@
...
+    std::shared_ptr<ItemsetNode> CreateChild(Item value, std::list<BasketInfo> baskets_info,
+                                             size_t included_baskets_cnt, double min_completeness) {
+        double included_contained_baskets_cnt = 0;
+        for (auto const& [_, real_id, is_included] : baskets_info) {
+            included_contained_baskets_cnt += is_included;
+        }
+        double completeness = included_contained_baskets_cnt / included_baskets_cnt;
+        if (completeness >= min_completeness) {
+            double validity = -1;
+            if (baskets_info.size()) {
```
Oddin60F: It's also better to use `empty()`
```suggestion
            if (!baskets_info.empty()) {
```

src/core/algorithms/cind/condition_miners/itemset_node.h
```diff
@@ -0,0 +1,134 @@
...
+                validity = included_contained_baskets_cnt / baskets_info.size();
+            }
+            std::shared_ptr<ItemsetNode> new_element =
+                    std::make_shared<ItemsetNode>(shared_from_this(), std::move(value),
+                                                  std::move(baskets_info), validity, completeness);
```
Oddin60F, line 40: It would be better to use `auto` here.
```suggestion
            auto new_element =
                    std::make_shared<ItemsetNode>(shared_from_this(), std::move(value),
                                                  std::move(baskets_info), validity, completeness);
```

src/core/algorithms/cind/condition.h
```diff
@@ -0,0 +1,86 @@
...
+    std::vector<std::string> condition_attrs_values;
+    double validity;
+    double completeness;
+
+    Condition(std::shared_ptr<ItemsetNode> itemset,
+              std::vector<model::EncodedColumnData const*> const& condition_attrs)
+        : validity(itemset->GetValidity()), completeness(itemset->GetCompleteness()) {
+        condition_attrs_values.resize(condition_attrs.size(), kAnyValue);
+        ItemsetNode const* item_ptr = itemset.get();
+        for (size_t column_id = condition_attrs.size() - 1; column_id >= 0; --column_id) {
```
Oddin60F: `column_id >= 0` is always true, because `size_t` is unsigned type. It doesn't seem to matter in what order we view them, so it's easier to use a direct traversal.
```suggestion
        for (size_t column_id = 0; column_id < condition_attrs.size() ; ++column_id) {
```

src/core/algorithms/cind/condition_miners/cinderella.cpp
```diff
@@ -0,0 +1,156 @@
...
+            }
+        }
+    }
+
+    // result stores all frequent and valid itemsets
+    std::vector<Condition> result;
+
+    Itemset itemset(std::move(first_level_items), included_baskets_cnt, min_completeness_);
+    while (!itemset.GetItems().empty()) {
+        for (auto candidate : itemset.GetItems()) {
```
Oddin60F, line 122: Here, it's unnecessary to copy each `shared_ptr`; it's better to get it by reference
```suggestion
        for (auto const& candidate : itemset.GetItems()) {
```

src/core/algorithms/cind/condition_miners/pli_cind.cpp
```diff
@@ -0,0 +1,153 @@
...
+             ++next_attr_idx) {
+            auto conditions = Analyze(next_attr_idx, new_curr_attrs, new_curr_pls, cond_attrs,
+                                      row_to_group, included_pos);
+            for (auto& cond : conditions) {
+                result.push_back(std::move(cond));
+            }
+        }
+    }
+    return result;
+}
```
Oddin60F, line 147: I can suggest some improvements that can be accepted if they are really justified.

When entering the function from `getConditions()` at the first level of recursion, `curr_attrs` is always empty, so we always skip the `Intersect()` call at the first level.  This means that at the first level, `curr_comb_pls` is always equal to the pointer from `attr_idx_to_pls_`. `attr_idx_to_pls_` is an object that is used at all levels of recursion, so we cannot change it. At the first level, when we traverse the clusters from `curr_comb_pls->GetClusters()`, we copy the `cluster` when we add it to `good_clusters`. This is fine. However, starting from the second level, we start using the intersection from `Intersect()` in `curr_comb_pls`, which has a limited lifetime due to the `for (auto& [cluster_value, cluster] : curr_comb_pls->GetClusters())` loop, as it is not used after this loop. This means that starting from the second level of recursion in the `for` loop, we can move `cluster` from `GetClusters()` when adding them to `good_clusters`.

This is the first solution that came to my mind, but there are probably better ways to do it. You can also come up with something to move cluster_value, but you need to remember that the unordered_map key is constant.
```suggestion
std::vector<Condition> PliCind::Analyze(size_t attr_idx, std::vector<int> const& curr_attrs,
                                        PLSetShared const& curr_pls, AttrsType const& cond_attrs,
                                        std::vector<int> const& row_to_group,
                                        std::vector<int> const& included_pos) {
    std::vector<int> new_curr_attrs = curr_attrs;
    new_curr_attrs.push_back(attr_idx);
    PLSetShared curr_comb_pls = attr_idx_to_pls_.at(attr_idx);
    bool is_first_level = curr_attrs.empty();
    if (!is_first_level) {
        curr_comb_pls = curr_pls->Intersect(attr_idx_to_pls_.at(attr_idx).get());
    }

    std::vector<Condition> result;
    std::vector<PLSet::ClusterCollection::value_type> good_clusters;
    for (auto& [cluster_value, cluster] : curr_comb_pls->GetClusters()) {
        std::vector<int> included_cluster;
        size_t cluster_size = cluster.size();
        if (condition_type_._value == CondType::group) {
            std::set<int> group_cluster;
            for (int row_id : cluster) {
                group_cluster.insert(row_to_group.at(row_id));
            }
            std::set_intersection(included_pos.begin(), included_pos.end(), group_cluster.begin(),
                                  group_cluster.end(), std::back_inserter(included_cluster));
            cluster_size = group_cluster.size();
        } else {
            // both included_pos and cluster are sorted
            std::set_intersection(included_pos.begin(), included_pos.end(), cluster.begin(),
                                  cluster.end(), std::back_inserter(included_cluster));
        }
        if (double completeness = static_cast<double>(included_cluster.size()) / included_pos.size();
            completeness >= min_completeness_) {
            if (double validity = static_cast<double>(included_cluster.size()) / cluster_size;
                validity >= min_validity_) {
                result.emplace_back(new_curr_attrs, cluster_value, cond_attrs, validity,
                                    completeness);
            }
            if (is_first_level) {
                good_clusters.emplace_back(cluster_value, cluster);
            ] else { 
                good_clusters.emplace_back(cluster_value, std::move(cluster));
            }
        }
    }
    if (!good_clusters.empty()) {
        PLSetShared new_curr_pls = PLSet::CreateFor(std::move(good_clusters), relation_size_);
        for (size_t next_attr_idx = attr_idx + 1; next_attr_idx < cond_attrs.size();
             ++next_attr_idx) {
            auto conditions = Analyze(next_attr_idx, new_curr_attrs, new_curr_pls, cond_attrs,
                                      row_to_group, included_pos);
            for (auto& cond : conditions) {
                result.push_back(std::move(cond));
            }
        }
    }
    return result;
}
```

examples/basic/sandbox.py
Oddin60F, line 1: It is worth deleting this file.

src/core/algorithms/cind/condition_miners/pli_cind.cpp
Oddin60F: And another thing. You also need to use `static_cast` here

src/core/algorithms/cind/condition_miners/position_lists_set.h
```diff
@@ -0,0 +1,81 @@
+#pragma once
+#include <memory>
+#include <unordered_map>
+#include <vector>
+
+#include <boost/container_hash/hash.hpp>
+
+class ColumnLayoutRelationData;
```
Oddin60F, line 8: This class is not used anywhere.

src/core/algorithms/cind/condition_miners/cinderella.cpp
```diff
@@ -0,0 +1,156 @@
...
+        if (index1 < index2) {
+            ++it1;
+        } else if (index1 > index2) {
+            ++it2;
+        } else {
+            std::list<size_t> positions_intersection;
+            std::set_intersection(positions1.begin(), positions1.end(), positions2.begin(),
+                                  positions2.end(), std::back_inserter(positions_intersection));
+            if (!positions_intersection.empty()) {
+                result.emplace_back(index1, positions_intersection, is_included1);
```
Oddin60F, line 33: Here you can use std::move()
```suggestion
                result.emplace_back(index1, std::move(positions_intersection), is_included1);
```

src/core/algorithms/cind/condition_miners/cinderella.cpp
```diff
@@ -0,0 +1,156 @@
...
+    }
+    return result;
+}
+
+void Cinderella::CreateNewItemsets(Itemset& itemset) const {
+    std::vector<std::tuple<std::shared_ptr<ItemsetNode>, Item, std::list<BasketInfo>>>
+            new_items_info;
+    for (auto const& itemset_prefix : itemset.GetPrevItems()) {
+        std::vector<Item> candidate = itemset_prefix->GetContents();
+        for (std::shared_ptr<ItemsetNode> parent : itemset_prefix->GetChildNodes()) {
```
Oddin60F, line 137: It is also better to avoid unnecessary copying here.
```suggestion
        for (auto const& parent : itemset_prefix->GetChildNodes()) {
```

src/core/algorithms/cind/condition_miners/itemset_node.h
```diff
@@ -0,0 +1,134 @@
...
+        double included_contained_baskets_cnt = 0;
+        for (auto const& [_, real_id, is_included] : baskets_info) {
+            included_contained_baskets_cnt += is_included;
+        }
+        double completeness = included_contained_baskets_cnt / included_baskets_cnt;
+        if (completeness >= min_completeness) {
+            double validity = -1;
+            if (baskets_info.size()) {
+                validity = included_contained_baskets_cnt / baskets_info.size();
+            }
```
Oddin60F, line 37: Imho, it's more readable this way.
```suggestion
            double validity = baskets_info.empty() ? -1 : included_contained_baskets_cnt / baskets_info.size();
```

examples/basic/mining_cind.py
```diff
@@ -0,0 +1,64 @@
...
+COMPLETENESS_THRESHOLD = 0.25
+
+algo.execute(error=ERROR_THRESHOLD, validity=VALIDITY_THRESHOLD, completeness=COMPLETENESS_THRESHOLD, condition_type=CONDITION_TYPE)
+
+print('Found condotoinal inclusion dependencies (-> means "is included in"):')
+for cind in algo.get_cinds():
+    print(cind)
+
+print("CIND's and their conditions, that found by algorithm, are also a valid python objects")
+cind = algo.get_cinds()[0]
```
ALanovaya, line 44: Can be IndexError

examples/basic/mining_cind.py
```diff
@@ -0,0 +1,64 @@
...
+print()
+print("Example CIND object:")
+print(cind)
+print("CIND object has methods:")
+print("    get_condition_attributes: ", cind.get_condition_attributes())
+print("    conditions_number: ", cind.conditions_number())
+print("    get_conditions (get all found conditions as Condition objects)")
+
+condition = cind.get_conditions()[0]
```
ALanovaya, line 53: Here is too
maybe write smth like this:
```c++
if not conditions:
    print("CIND has no conditions")
    return
```

examples/basic/sandbox.py
```diff
@@ -0,0 +1,69 @@
...
+
+# for cond_type in [
+#     "row",
+#     # "group"
+#     ]:
+#     algo1.execute(error=0.5, validity=VALIDITY, completeness=COMPLETENESS, condition_type=cond_type)
+#     algo2.execute(error=0.5, validity=VALIDITY, completeness=COMPLETENESS, condition_type=cond_type)
+#     check(algo1, algo2, cond_type)
+
+import desbordante
```
ALanovaya, line 58: why is it here again?

src/core/algorithms/cind/cind_algorithm.h
```diff
@@ -0,0 +1,56 @@
...
+        size_t load;    /**< time taken for the data loading */
+        size_t compute; /**< time taken for the inds computing */
+        size_t total;   /**< total time taken for all stages */
+    };
+
+public:
+    explicit CindAlgorithm(std::vector<std::string_view> phase_names = {});
+
+    unsigned long long TimeTaken() const {
+        return timings_.total;
```
ALanovaya, line 22: `total` has type of `size_t` which is implementation-defined, but returns `unsigned long long` 
for timings use `std::chrono` or smth like `std::uint64`

src/core/algorithms/cind/condition.h
ALanovaya, line 26: `size_t` is unsigned type so `column_id >= 0` is always true 

If `condition_attrs.size() == 0`, then` size() - 1` immediately results very large number

You can use `std::ssize` or this option 
```c++
for (size_t column_id = condition_attrs.size(); column_id-- > 0; ) {
    // use of column_id
}
```

src/core/algorithms/cind/condition.h
```diff
@@ -0,0 +1,86 @@
...
+    Condition(std::vector<int> const& cluster_attrs_ids,
+              model::PLSet::ClusterValue const& cluster_value,
+              std::vector<model::EncodedColumnData const*> const& condition_attrs, double _validity,
+              double _completeness)
+        : validity(_validity), completeness(_completeness) {
+        condition_attrs_values.reserve(condition_attrs.size());
+        size_t item_id = 0;
+        for (size_t column_id = 0; column_id < condition_attrs.size(); ++column_id) {
+            if (item_id < cluster_attrs_ids.size() &&
+                column_id == (size_t)cluster_attrs_ids[item_id]) {
```
ALanovaya, line 48: use c++ casts pls

src/core/algorithms/cind/condition.h
```diff
@@ -0,0 +1,86 @@
...
+        result.append("validity = ")
+                .append(std::to_string(validity))
+                .append(", completeness = ")
+                .append(std::to_string(completeness))
+                .append(")");
+        return result;
+    }
+
+    bool operator==(Condition const& that) const {
+        return this->completeness == that.completeness && this->validity == that.validity &&
```
ALanovaya, line 71: `validity` and `completeness` are doubles, so check for equality is incorrect 
Make it using some `eps`

src/core/algorithms/cind/condition.h
ALanovaya, line 83: duplicate for `completeness`

src/core/algorithms/cind/condition.h
```diff
@@ -0,0 +1,86 @@
...
+               this->condition_attrs_values == that.condition_attrs_values;
+    }
+};
+}  // namespace algos::cind
+
+template <>
+struct std::hash<algos::cind::Condition> {
+    size_t operator()(algos::cind::Condition const& cond) const noexcept {
```
ALanovaya, line 79: Violation of the invariant for unordered containers:

+ `operator==` takes into account `condition_attrs_values`, but they do not participate in the hash at all.

+ This violates the contract: `if a == b`, their hashes must match. It's easy to get now `a != b`, but `hash(a) == hash(b)`, it can lead to lots of collisions.

src/core/algorithms/cind/condition_miners/basket.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+#include <list>
+#include <unordered_map>
+
+#include "item.h"
+
+namespace algos::cind {
+using ItemsInfo = std::unordered_map<Item, std::list<size_t>>;
```
ALanovaya, line 9: `std::list` is soo heavy data structure with a lot of allocations
Can it be replaced with vector?

src/core/algorithms/cind/condition_miners/basket.h
```diff
@@ -0,0 +1,19 @@
...
+
+struct Basket {
+    bool is_included;
```
ALanovaya, line 12: Not initialized

src/core/algorithms/cind/cind_algorithm.h
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    std::list<model::IND> const& AINDList() const noexcept {
```
ALanovaya, line 25: Can returned type be replaced with vector?

src/core/algorithms/cind/cind_algorithm.h
```diff
@@ -0,0 +1,56 @@
...
+        return spider_algo_->INDList();
+    }
+
+    std::list<CIND> const& CINDList() const noexcept {
```
ALanovaya, line 29: Can returned type be replaced with vector?

src/core/algorithms/cind/condition_miners/cind_miner.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+namespace algos::cind {
+CindMiner::CindMiner(config::InputTables& input_tables)
+    : tables_(input_tables), condition_type_(CondType::group) {}
+
+CindMiner::Attributes CindMiner::ClassifyAttributes(model::IND const& aind) const {
+    Attributes result;
+    for (auto const& column : tables_.GetTable(aind.GetLhs().GetTableIndex()).GetColumnData()) {
+        auto const& ind_columns = aind.GetLhs().GetColumnIndices();
+        auto const& ind_columns_rhs = aind.GetRhs().GetColumnIndices();
```
ALanovaya, line 15: 14-15 lines: Is it cycle invariant ? If yes, you can take it out of the loop

src/core/algorithms/cind/condition_miners/cind_miner.cpp
```diff
@@ -0,0 +1,60 @@
...
+        AttrsType const& condition_attrs) {
+    if (condition_attrs.empty()) {
+        return {};
+    }
+    std::vector<std::string> result(condition_attrs.size(),
+                                    condition_attrs.back()->GetColumn()->GetSchema()->GetName());
+    for (size_t i = 0; i < result.size(); ++i) {
+        result[i].append(".").append(condition_attrs[i]->GetColumn()->GetName());
+    }
+    return result;
```
ALanovaya, line 58: 53-58 lines: it is hard to read
Maybe make smth like this?
```c++
std::vector<std::string> result;
result.reserve(condition_attrs.size());
for (auto* col_data : condition_attrs) {
    auto* col = col_data->GetColumn();
    result.emplace_back(col->GetSchema()->GetName() + "." + col->GetName());
}
```

src/core/algorithms/cind/condition_miners/cinderella.cpp
```diff
@@ -0,0 +1,156 @@
...
+#include <tuple>
+#include <unordered_map>
+
+#include "cind/condition.h"
+#include "cind/condition_miners/basket.h"
+#include "cind/condition_type.h"
+
+namespace algos::cind {
+namespace {
+std::list<BasketInfo> MergeBaskets(std::list<BasketInfo> const& baskets1,
```
ALanovaya, line 16: Maybe use vector :(
intersection also works with a vector, and it will be much cheaper for cache and allocations

src/core/algorithms/cind/condition_miners/cinderella.cpp
```diff
@@ -0,0 +1,156 @@
...
+    std::map<std::vector<int>, int> basket_id_by_value;
+    for (size_t index = 0; index < attributes.lhs_inclusion.front()->GetNumRows(); ++index) {
+        std::vector<int> row;
+        for (auto& attr : attributes.lhs_inclusion) {
+            row.push_back(attr->GetValue(index));
+        }
+        if (condition_type_._value == CondType::group) {
+            if (auto const& it = basket_id_by_value.find(row); it == basket_id_by_value.cend()) {
+                result.emplace_back(rhs_values.contains(row), ItemsInfo{});
+                basket_id_by_value[row] = basket_id_by_value.size();
```
ALanovaya, line 75: The order of calculation of subexpressions is not defined. That is, `size()` can be calculated first, and then `operator[]`, or vice versa. This gives:

+ potentially UB (non-sequential access with modification and reading of the same object),

+ and even with a specific procedure, the indexes are incorrect.

ALanovaya: The correct `basket_id` must match the index of the newly added basket in result, i.e. `result.size() - 1` after `emplace_back`.

`basket_id_by_value.size()` is not tied to anything here at all.

src/core/algorithms/cind/condition_miners/cinderella.cpp
```diff
@@ -0,0 +1,156 @@
...
+            }
```
ALanovaya, line 76: Maybe do smth like this?
```c++
if (auto const& it = basket_id_by_value.find(row); it == basket_id_by_value.cend()) {
    bool included = rhs_values.contains(row);
    std::size_t id = result.size();
    result.emplace_back(included, ItemsInfo{});
    basket_id_by_value.emplace(row, static_cast<int>(id));
}
auto basket_id = basket_id_by_value.at(row);

```

src/core/algorithms/cind/condition_miners/cinderella.cpp
```diff
@@ -0,0 +1,156 @@
...
+    auto attributes = ClassifyAttributes(aind);
+    CIND cind{.ind = aind,
+              .conditions = GetConditions(GetBaskets(attributes), attributes.conditional),
+              .conditional_attributes = GetConditionalAttributesNames(attributes.conditional)};
+    return cind;
+}
+
+std::vector<Basket> Cinderella::GetBaskets(Attributes const& attributes) {
+    // algorithm uses modified left-outer join representation to build the baskets
+    std::set<std::vector<int>> rhs_values;
```
ALanovaya, line 55: Can it be replaced with `std::unordered_set<std::vector<int>, VectorHash>`?
Set on a tree can be too heavy here because the key is `std::vector<int>` (O(len) comparison)

src/core/algorithms/cind/condition_miners/cinderella.cpp
```diff
@@ -0,0 +1,156 @@
...
+            }
+            result.emplace_back(rhs_values.contains(row), std::move(basket_items));
+        }
+    }
+    return result;
+}
+
+std::vector<Condition> Cinderella::GetConditions(std::vector<Basket> const& baskets,
+                                                 AttrsType const& condition_attrs) const {
+    std::unordered_map<Item, std::list<BasketInfo>> first_level_items;
```
ALanovaya, line 96: `vector<BasketInfo>`

src/core/algorithms/cind/condition_miners/cinderella.cpp
ALanovaya: The same with map

src/core/algorithms/cind/condition_miners/itemset_node.h
```diff
@@ -0,0 +1,134 @@
...
+          completeness_(completeness),
+          parent_node_(std::move(parent)) {}
+
+    std::shared_ptr<ItemsetNode> CreateChild(Item value, std::list<BasketInfo> baskets_info,
+                                             size_t included_baskets_cnt, double min_completeness) {
+        double included_contained_baskets_cnt = 0;
+        for (auto const& [_, real_id, is_included] : baskets_info) {
+            included_contained_baskets_cnt += is_included;
+        }
+        double completeness = included_contained_baskets_cnt / included_baskets_cnt;
```
ALanovaya, line 32: So if `included_baskets_cnt == 0` (it can be, there is no checks in `Cinderella::GetConditions`) it would be +-inf or nan in `completeness`, is it correct logic?

src/core/algorithms/cind/condition_miners/itemset_node.h
```diff
@@ -0,0 +1,134 @@
...
+
+    bool CheckSubsetItem(std::vector<Item> const& candidate, size_t curr_idx, size_t offset,
+                         bool do_branch) const {
+        if (curr_idx + offset < candidate.size()) {
+            auto const& it = std::find_if(child_nodes_.begin(), child_nodes_.end(),
+                                          [&](std::shared_ptr<ItemsetNode> const& node) {
+                                              return node->GetValue() == candidate.at(curr_idx + 1);
+                                          });
+            return it != child_nodes_.end() &&
+                   it->get()->CheckSubsets(candidate, curr_idx + 1, do_branch);
```
ALanovaya, line 100: why here is `curr_idx + 1` not `curr_idx + offset` ?

src/core/algorithms/cind/condition_miners/itemset_node.h
```diff
@@ -0,0 +1,134 @@
...
+            parent_node_->Erase(shared_from_this());
+        }
+    }
+
+private:
+    Item value_;
+    std::list<BasketInfo> baskets_info_;
+    double validity_;
+    double completeness_;
+    std::unordered_set<std::shared_ptr<ItemsetNode>> child_nodes_;
```
ALanovaya, line 123: why not `std::unordered_map<Item, std::shared_ptr<ItemsetNode>> child_nodes_`?
It is faster then `std::find_if`

src/core/algorithms/cind/condition_miners/itemset_node.h
```diff
@@ -0,0 +1,134 @@
...
+};
+}  // namespace algos::cind
+
+template <>
+struct std::hash<algos::cind::ItemsetNode> {
+    size_t operator()(algos::cind::ItemsetNode const& node) const {
+        std::hash<algos::cind::Item> item_hasher;
+        return item_hasher(node.GetValue());
+    }
+};
```
ALanovaya, line 134: this hash is unused now. Can be deleted

src/core/algorithms/cind/condition_miners/pli_cind.cpp
```diff
@@ -0,0 +1,153 @@
...
+    for (size_t index = 0; index < attrs.rhs_inclusion.front()->GetNumRows(); ++index) {
+        std::vector<std::string> row;
+        for (auto& attr : attrs.rhs_inclusion) {
+            row.push_back(attr->GetStringValue(index));
+        }
+        rhs_values.insert(std::move(row));
+    }
+    //  included row_id for rows, group_id for groups
+    std::vector<int> included_pos;
+    std::map<std::vector<std::string>, int> group_idx;
```
ALanovaya, line 42: `rhs_values` and `group_idx`:
You can use `std::unordered_set`/`std::unordered_map` with a hash of a combination of values, or even encode strings into whole IDs (dictionary encoding).

src/core/algorithms/cind/condition_miners/pli_cind.cpp
```diff
@@ -0,0 +1,153 @@
...
+        curr_comb_pls = curr_pls->Intersect(attr_idx_to_pls_.at(attr_idx).get());
+    }
+
+    std::vector<Condition> result;
+    std::vector<PLSet::ClusterCollection::value_type> good_clusters;
+    for (auto& [cluster_value, cluster] : curr_comb_pls->GetClusters()) {
+        std::vector<int> included_cluster;
+        size_t cluster_size = cluster.size();
+        if (condition_type_._value == CondType::group) {
+            std::set<int> group_cluster;
```
ALanovaya, line 113: can be replaced with vector + sort+unique

src/core/algorithms/cind/condition_miners/pli_cind.cpp
```diff
@@ -0,0 +1,153 @@
...
+    }
+    if (!good_clusters.empty()) {
+        PLSetShared new_curr_pls = PLSet::CreateFor(std::move(good_clusters), relation_size_);
+        for (size_t next_attr_idx = attr_idx + 1; next_attr_idx < cond_attrs.size();
+             ++next_attr_idx) {
+            auto conditions = Analyze(next_attr_idx, new_curr_attrs, new_curr_pls, cond_attrs,
+                                      row_to_group, included_pos);
+            for (auto& cond : conditions) {
+                result.push_back(std::move(cond));
+            }
```
ALanovaya, line 143: Can be rewritten with `insert` with move-iterators

src/core/model/table/encoded_column_data.h
```diff
@@ -0,0 +1,68 @@
...
+    model::ColumnIndex GetColumnId() const noexcept {
+        return column_->GetIndex();
+    }
+
+    std::set<std::string> const& GetUniqueValues() const noexcept {
+        return unique_values_;
+    }
+
+    std::string DecodeValue(int value) const {
+        return value_dictionary_->ToString(value);
```
ALanovaya, line 59: Maybe add check value for `kNullValueId`? If `value `would be 0 it would be assertion failure
It is actual for `Condition `:
```c++
condition_attrs_values[column_id] = attribute->DecodeValue(item.value);
```
Here `item.value` is from `cond_attr->GetValue()` and can be 0

src/core/model/table/encoded_column_data.h
```diff
@@ -0,0 +1,68 @@
...
+          unique_values_(std::move(unique_values)),
+          value_dictionary_(std::move(value_dictionary)) {}
+
+    std::string ToString() const final {
+        return "Data for " + column_->ToString();
+    }
+
+    std::string const& GetStringValue(size_t index) const {
+        assert(index < column_data_.size());
+        return value_dictionary_->ToString(column_data_.at(index));
```
ALanovaya, line 30: `column_data_[index]` because there is `assert` after

src/core/model/table/encoded_tables.h
```diff
@@ -0,0 +1,34 @@
...
+                                                                    table_id, value_dictionary_));
+        }
+    }
+
+    ColumnEncodedRelationData& GetTable(size_t index) const {
+        assert(index < tables_.size());
+        return *tables_[index];
+    }
+
+    const std::vector<std::unique_ptr<ColumnEncodedRelationData>>& GetTables() const noexcept {
```
ALanovaya, line 26: it turns out that through `const EncodedTables`, you can modify the tables themselves. Is it correct?

ALanovaya (changes requested):
Overall the CIND miner design is sound and the high-level logic of both the Cinderella and PliCind algorithms matches the intended “conditional IND as AIND + conditions” model, but there are several important issues and optimization opportunities.

src/core/algorithms/cind/cind_algorithm.h
```diff
@@ -0,0 +1,56 @@
...
+namespace algos::cind {
+class CindAlgorithm final : public Algorithm {
+    /// timing information for algorithm stages
+    struct StageTimings {
+        size_t load;    /**< time taken for the data loading */
+        size_t compute; /**< time taken for the inds computing */
+        size_t total;   /**< total time taken for all stages */
+    };
+
+public:
```
Oddin60F, line 18: We have an exception from styleguide about the order of fields in a class (see Wiki/Development‐General/C++ Style Guide).

src/core/algorithms/cind/condition_miners/item.h
```diff
@@ -0,0 +1,33 @@
...
+    bool operator==(Item const& that) const {
+        return this->column_id == that.column_id && this->value == that.value;
+    }
+
+    bool operator<(Item const& that) const {
+        if (this->column_id != that.column_id) {
+            return this->column_id < that.column_id;
+        }
+        return this->value < that.value;
+    }
```
Oddin60F, line 21: maybe:
```suggestion
    bool operator<(Item const& that) const {
        return std::tie(column_id, value) < std::tie(that.column_id, that.value);
    }
```

src/core/algorithms/cind/condition_miners/pli_cind.cpp
```diff
@@ -0,0 +1,153 @@
...
+#include <vector>
+
+#include "cind/condition.h"
+#include "cind/condition_miners/position_lists_set.h"
+#include "cind/condition_type.h"
+#include "table/encoded_column_data.h"
+#include "table/table_index.h"
+
+namespace algos::cind {
+using model::TableIndex;
```
Oddin60F, line 18: It seems like `TableIndex` is not used anywhere here.

src/core/algorithms/cind/condition_miners/cind_miner.cpp
```diff
@@ -0,0 +1,60 @@
...
+    }
+    for (auto const& column : tables_.GetTable(aind.GetRhs().GetTableIndex()).GetColumnData()) {
+        auto const& ind_columns = aind.GetRhs().GetColumnIndices();
+        if (std::find(ind_columns.cbegin(), ind_columns.cend(), column.GetColumn()->GetIndex()) !=
+            ind_columns.cend()) {
+            result.rhs_inclusion.push_back(&column);
+        }
+    }
+    return result;
+}
```
Oddin60F, line 36: In my opinion, the function is very hard to read, I would rewrite it somehow using `ranges`
```suggestion
CindMiner::Attributes CindMiner::ClassifyAttributes(model::IND const& aind) const {
    Attributes result;
    auto const& lhs = aind.GetLhs();
    auto const& rhs = aind.GetRhs();
    auto const& lhs_indices = lhs.GetColumnIndices();
    auto const& rhs_indices = rhs.GetColumnIndices();

    for (auto const& column : tables_.GetTable(lhs.GetTableIndex()).GetColumnData()) {
        auto col_index = column.GetColumn()->GetIndex();

        if (std::ranges::find(lhs_indices, col_index) != lhs_indices.end()) {
            result.lhs_inclusion.push_back(&column);
        } else if (lhs.GetTableIndex() == rhs.GetTableIndex()) {
            if (std::ranges::find(rhs_indices, col_index) == rhs_indices.end()) {
                result.conditional.push_back(&column);
            }
        } else {
            result.conditional.push_back(&column);
        }
    }

    for (auto const& column : tables_.GetTable(rhs.GetTableIndex()).GetColumnData()) {
        auto col_index = column.GetColumn()->GetIndex();

        if (std::ranges::find(rhs_indices, col_index) != rhs_indices.end()) {
            result.rhs_inclusion.push_back(&column);
        }
    }
    return result;
}
```

src/core/algorithms/cind/condition_type.h
Oddin60F, line 1: This is going to sound picky, but I would rename this file. It defines not only the types of conditions, but also the types of algorithms.

src/python_bindings/cind/bind_cind.cpp
```diff
@@ -0,0 +1,76 @@
...
+            })
+            .def("precision", [](Condition const& cond) {
+                return cond.validity;
+            })
+            .def("completeness", [](Condition const& cond) {
+                return cond.completeness;
+            })
+            .def("recall", [](Condition const& cond) {
+                return cond.completeness;
+            });
```
Oddin60F, line 54: As I understand it, `validity` is an analog of precision from machine learning theory, and `completeness` is the same analog of recall. Do we really need to get the same value under different names?

chernishev:
This PR has been completed and merged here [664](https://github.com/Desbordante/desbordante-core/pull/664).

PR #561, opened 2025-05-03
Fix includes part 1

PodushkaPIR:
- Refactoring tests files includes
 - Refactoring includes in core modules: (model, util, parser)

All files that do not fit into the category above were either obtained using clang-format-18, or were slightly modified to build project correctly.

About include comments: it's temporary and they don't make sense. 

This PR is one of the 3 parts of the larger #531. And it's also Draft, because I need to clean up the comments and run IWYU for full improvement.


PodushkaPIR:
Forget text about "src/core/confg" directory. I will review the changes made to this directory. They will be added to this PR later.

PR #573, opened 2025-06-21
[DRAFT] DynFD algorithm

m4xxx1m:


src/core/algorithms/fd/dynfd/dynfd.h
```diff
@@ -0,0 +1,40 @@
+#pragma once
+#include <FDTrees/fd_tree.h>
```
ol-imorozko, line 3: nit: Separate pragma with an empty line

src/core/algorithms/fd/dynfd/dynfd.h
```diff
@@ -0,0 +1,40 @@
...
+class DynFD final : public FDAlgorithm {
+    config::InputTable input_table_;
+    config::InputTable insert_statements_table_ = nullptr;
+    config::InputTable update_statements_table_ = nullptr;
+    std::unordered_set<size_t> delete_statement_indices_;
+    std::shared_ptr<DynamicRelationData> relation_ = nullptr;
+    std::shared_ptr<model::FDTree> positive_cover_tree_ = nullptr;
+    std::shared_ptr<NonFDTree> negative_cover_tree_ = nullptr;
+    std::shared_ptr<Validator> validator_ = nullptr;
+    config::ThreadNumType threads_num_ = 1;
```
ol-imorozko: You init all shared pointers with `nullptr` except `input_table_`. Is this intentional? 
I feel like we should init it too

src/core/algorithms/fd/dynfd/dynfd.cpp
```diff
@@ -0,0 +1,239 @@
...
+    if (is_fd_validation_needed) {
+        validator_->ValidateFds(first_insert_batch_id);
+    }
+
+    SetProgress(kTotalProgressPercent);
+    RegisterFDs(positive_cover_tree_->FillFDs());
+    auto const elapsed_milliseconds = std::chrono::duration_cast<std::chrono::milliseconds>(
+            std::chrono::system_clock::now() - start_time);
+    return elapsed_milliseconds.count();
+}
```
ol-imorozko: nit: I suggest to exract mining functionality to a separate method and use `TimedInvoke` here.
For example, in `src/core/algorithms/ind/mind/mind.cpp`:
```cpp
unsigned long long Mind::ExecuteInternal() {
    timings_.compute_uinds = util::TimedInvoke(&Mind::MineUnaryINDs, this);
    timings_.compute_ninds = util::TimedInvoke(&Mind::MineNaryINDs, this);
    return timings_.compute_uinds + timings_.compute_ninds;
}
```

src/core/algorithms/fd/dynfd/dynfd.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+    if (!relation_->Empty()) {
+        ExecuteHyFD();
+    }
+
+    negative_cover_tree_ = std::make_shared<NonFDTree>(GetRelation().GetNumColumns());
+
+    // Cover inversion
+    for (size_t i = 0; i < relation_->GetNumColumns(); i++) {
+        boost::dynamic_bitset<> lhs(relation_->GetNumColumns());
```
ol-imorozko: Sometimes you use `GetRelation().GetNumColumns()`, sometimes `relation_->GetNumColumns()`.
It's better to stay consistent. I suggest to just define a variable like `num_columns = relation_->GetNumColumns()`.

src/core/algorithms/fd/dynfd/dynfd.cpp
```diff
@@ -0,0 +1,239 @@
...
+#include "tabular_data/crud_operations/operations.h"
+#include "tabular_data/crud_operations/update/option.h"
+#include "tabular_data/input_table/option.h"
+
+namespace algos::dynfd {
+
+void DynFD::ExecuteHyFD() {
+    std::shared_ptr hy_fd_relation = ColumnLayoutRelationData::CreateFrom(*input_table_, true);
+
+    auto [plis, pli_records, og_mapping] = hy::Preprocess(hy_fd_relation.get());
```
ol-imorozko: Here you create new variables and since `hy::Preprocess(hy_fd_relation.get())` is a prvalue expression, you move-initialize them.
A better alternative would be to use `auto &&` to create references that bind directly to the members of the temporary, thus removing the additional move.
```suggestion
    auto&& [plis, pli_records, og_mapping] = hy::Preprocess(hy_fd_relation.get());
```

src/core/algorithms/fd/dynfd/dynfd.cpp
```diff
@@ -0,0 +1,239 @@
...
+    hyfd::Inductor inductor(positive_cover_tree);
+    hyfd::Validator validator(positive_cover_tree, plis_shared, pli_records_shared, threads_num_);
+
+    hy::IdPairs comparison_suggestions;
+
+    while (true) {
+        auto non_fds = sampler.GetNonFDs(comparison_suggestions);
+
+        inductor.UpdateFdTree(std::move(non_fds));
+
```
ol-imorozko: `UpdateFdTree` does not use the fact that it accepts only rvalue references. It only calls some methods of `NonFDList` and that's it:
```cpp
void Inductor::UpdateFdTree(NonFDList&& non_fds) {
    unsigned const max_level = non_fds.GetDepth();

    for (unsigned level = max_level; level != 0; level--) {
        for (auto const& lhs_bits : non_fds.GetLevel(level)) {
```

So I suggest to just  use regular const reference, and remove `std::move()` there

src/core/algorithms/fd/dynfd/dynfd.cpp
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    for (size_t rhs = 0; rhs < hy_fd_relation->GetNumColumns(); ++rhs) {
+        positive_cover_tree_->Remove(boost::dynamic_bitset(hy_fd_relation->GetNumColumns()), rhs);
+    }
+    for (auto fd : positive_cover_tree->FillFDs()) {
+        fd.lhs_ = hy::RestoreAgreeSet(fd.lhs_, og_mapping, hy_fd_relation->GetNumColumns());
+        fd.rhs_ = og_mapping[fd.rhs_];
+        positive_cover_tree_->AddFD(fd.lhs_, fd.rhs_);
+    }
```
ol-imorozko: That's a bit confusing, since you declare a local variable `positive_cover_tree` which shadows the class member `positive_cover_tree_`. I suggest to rename that local variable to smth like `hyfd_positive_cover`

src/core/algorithms/fd/dynfd/dynfd.cpp
```diff
@@ -0,0 +1,239 @@
...
+            break;
+        }
+
+        LOG(TRACE) << "Cycle done";
+    }
+
+    for (size_t rhs = 0; rhs < hy_fd_relation->GetNumColumns(); ++rhs) {
+        positive_cover_tree_->Remove(boost::dynamic_bitset(hy_fd_relation->GetNumColumns()), rhs);
+    }
+    for (auto fd : positive_cover_tree->FillFDs()) {
```
ol-imorozko: You create a copy here, which is probably not intentional
```suggestion
    for (RawFD const& fd : positive_cover_tree->FillFDs()) {
```

src/core/algorithms/fd/dynfd/dynfd.cpp
```diff
@@ -0,0 +1,239 @@
...
+            for (size_t bit = lhs.find_first(); bit != boost::dynamic_bitset<>::npos;
+                 bit = lhs.find_next(bit)) {
+                boost::dynamic_bitset<> new_lhs = non_fd;
+                new_lhs.reset(bit);
+                if (!negative_cover_tree_->ContainsNonFdOrSpecial(new_lhs, rhs)) {
+                    negative_cover_tree_->AddNonFD(new_lhs, rhs, std::nullopt);
+                }
+            }
+        }
+    }
```
ol-imorozko: That's the Algorithm 1 from the paper. It doubt that needs to be in `LoadDataInternal`, as it seems strange that we mix data loading with complex algorithmic logic. At the very least, consider extracting that to the separate method/static function and just call it `CoverInversion` instead of having to write a comment

src/core/algorithms/fd/dynfd/dynfd.cpp
```diff
@@ -0,0 +1,239 @@
...
+        if (comparison_suggestions.empty()) {
+            break;
+        }
+
+        LOG(TRACE) << "Cycle done";
+    }
+
+    for (size_t rhs = 0; rhs < hy_fd_relation->GetNumColumns(); ++rhs) {
+        positive_cover_tree_->Remove(boost::dynamic_bitset(hy_fd_relation->GetNumColumns()), rhs);
+    }
```
ol-imorozko, line 51: `positive_cover_tree_` is created fresh in `LoadDataInternal` just before `ExecuteHyFD` function is called, so isn't it should already be empty? Correct me if I'm wrong, please

src/core/algorithms/fd/dynfd/fd_inductor.h
```diff
@@ -0,0 +1,24 @@
+#pragma once
+
+#include <vector>
+
```
ol-imorozko: It's unused in this file

src/core/algorithms/fd/dynfd/fd_inductor.cpp
```diff
@@ -0,0 +1,40 @@
...
+            for (auto rhs = rhss.find_first(); rhs != boost::dynamic_bitset<>::npos;
+                 rhs = rhss.find_next(rhs)) {
+                DeduceDependencies({lhs, rhs});
+            }
+        }
+    }
+}
+
+void algos::dynfd::FDInductor::DeduceDependencies(RawFD non_fd) {
+    auto rhs = non_fd.rhs_;
```
ol-imorozko:
```suggestion
    size_t rhs = non_fd.rhs_;
```
Generally, prefer using explicit types to `auto`, unless the type is obvious like an iterator, `std::shared_ptr` etc.

src/core/algorithms/fd/dynfd/fd_inductor.cpp
```diff
@@ -0,0 +1,40 @@
...
+void algos::dynfd::FDInductor::UpdateCovers(algos::hy::ColumnCombinationList const& agree_sets) {
+    for (int level = agree_sets.GetDepth(); level > 0; --level) {
+        for (auto const& lhs : agree_sets.GetLevel(level)) {
+            auto rhss = lhs;
+            rhss.flip();
+
+            for (auto rhs = rhss.find_first(); rhs != boost::dynamic_bitset<>::npos;
+                 rhs = rhss.find_next(rhs)) {
+                DeduceDependencies({lhs, rhs});
+            }
```
ol-imorozko: You create a new bitset and flip it's bits, which is O(N) already:
```cpp
template <typename Block, typename Allocator>
dynamic_bitset<Block, Allocator>&
dynamic_bitset<Block, Allocator>::flip()
{
    for (size_type i = 0; i < num_blocks(); ++i)
        m_bits[i] = ~m_bits[i];
    m_zero_unused_bits();
    return *this;
}
```

I think it's simpler and easier to just do a regular loop over all values, since you're doing it anyway in `.flip()`:
```suggestion
            for (size_t rhs = 0; rhs < lhs.size(); ++rhs) {
                if (!lhs[rhs]) {
                    DeduceDependencies({lhs, rhs});
                }
            }
```

src/core/algorithms/fd/dynfd/fd_inductor.cpp
```diff
@@ -0,0 +1,40 @@
+#include "fd_inductor.h"
+
+void algos::dynfd::FDInductor::UpdateCovers(algos::hy::ColumnCombinationList const& agree_sets) {
+    for (int level = agree_sets.GetDepth(); level > 0; --level) {
+        for (auto const& lhs : agree_sets.GetLevel(level)) {
```
ol-imorozko, line 5:
```suggestion
        for (boost::dynamic_bitset<> const& lhs : agree_sets.GetLevel(level)) {
```
Generally, prefer using explicit types to auto, unless the type is obvious like an iterator, std::shared_ptr etc.

src/core/algorithms/fd/dynfd/model/cluster_ids_array.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+    return ClusterIdsArray(std::move(cluster_ids));
+}
+
+std::size_t std::hash<ClusterIdsArray>::operator()(ClusterIdsArray const& key) const {
+    std::size_t hash = 1;
+    for (int const elem : key.GetCluster()) {
+        hash = 31 * hash + elem;
+    }
+    return hash;
```
ol-imorozko: you can use `boost`:
```suggestion
    return boost::hash_range(key.GetCluster().begin(), key.GetCluster().end());
```

src/core/algorithms/fd/dynfd/model/cluster_ids_array.h
```diff
@@ -0,0 +1,38 @@
...
+
+using algos::dynfd::CompressedRecord;
+
+class ClusterIdsArray {
+public:
+    ClusterIdsArray(std::vector<int> cluster_ids) : cluster_ids_(std::move(cluster_ids)) {}
+
+    std::vector<int> const& GetCluster() const;
+
+    bool operator==(ClusterIdsArray const& other) const;
```
ol-imorozko, line 13: It is expected that if some class has `operator==`, it also has `operator!=` implemented as `!(*this == other);`

src/core/algorithms/fd/dynfd/model/cluster_ids_array.h
```diff
@@ -0,0 +1,38 @@
...
+
+template <>
+struct std::hash<ClusterIdsArray> {
+    std::size_t operator()(ClusterIdsArray const& key) const;
+};
+
+class ClusterIdsArrayWithRecord : public ClusterIdsArray {
+public:
+    ClusterIdsArrayWithRecord(std::vector<int> cluster_ids, size_t record_id)
+        : ClusterIdsArray(cluster_ids), record_id_(record_id) {}
```
ol-imorozko:
```suggestion
    ClusterIdsArrayWithRecord(std::vector<int> cluster_ids, size_t record_id)
        : ClusterIdsArray(std::move(cluster_ids)), record_id_(record_id) {}
```
Otherwise we do a copy

src/core/algorithms/fd/dynfd/model/dynamic_position_list_index.h
```diff
@@ -0,0 +1,119 @@
+#pragma once
+
```
ol-imorozko, line 2: We have `src/core/model/table/dynamic_position_list_index.cpp`. Is this implementation an ad hoc for dynfd algorithm?

src/core/algorithms/fd/dynfd/model/compressed_column_data.h
```diff
@@ -0,0 +1,35 @@
...
+        : AbstractColumnData(column), position_list_index_(std::move(position_list_index)) {}
+
+    [[nodiscard]] size_t GetNumRows() const {
+        return position_list_index_->GetSize();
+    }
+
+    [[nodiscard]] std::string ToString() const final {
+        return "Data for " + column_->ToString();
+    }
+
```
ol-imorozko: Why is only `ToString` method is `final`?
I don't see any usage of `CompressedColumnData` as a parent class, so I thinkg we should mark the whole class as `final` to help compiler with devirtualization

src/core/algorithms/fd/dynfd/model/dynamic_position_list_index.cpp
```diff
@@ -0,0 +1,172 @@
+#include "dynamic_position_list_index.h"
+
```
ol-imorozko, line 2: nit: It would be a nice touch to add unit tests for `DynamicPositionListIndex` class as it's big, complex, uses a lot of iterators, adding, erasing, loops etc. Kind algorithmically heavy.

src/core/algorithms/fd/dynfd/dynfd.cpp
m4xxx1m: Constructor of model::FDTree sets all FDs in its root, which is correct for empty dataset. But for construction DynFD's positive cover we need to remove these FDs.

src/core/algorithms/fd/dynfd/fd_inductor.cpp
m4xxx1m: I think it's obvious from the context that lhs is boost::dynamic_bitset<>. But I checked every auto I used and replaced with explicit types where needed.

src/core/algorithms/fd/dynfd/model/dynamic_position_list_index.h
m4xxx1m: DPLI in `src/core/model/table/dynamic_position_list_index.cpp` is meant for DynFD Verifier.  We decided not to merge these versions for now.

src/core/algorithms/fd/dynfd/model/dynamic_relation_data.cpp
```diff
@@ -0,0 +1,265 @@
...
+    std::vector<size_t> all_ids(next_record_id);
+    std::iota(all_ids.begin(), all_ids.end(), 0);
+
+    input_table->Reset();
+
+    return std::make_unique<DynamicRelationData>(std::move(schema), std::move(column_data),
+                                                 std::unordered_set(all_ids.begin(), all_ids.end()),
+                                                 std::move(value_dictionary), next_value_id,
+                                                 next_record_id, std::move(compressed_records));
+}
```
ol-imorozko: On the left we have `CreateFrom`, on the right `InsertBatch`:
image
Code looks virutally the same inside. Same for the `InsertRecordsFromUpdateBatch`. Extract common logic into a helper method, please

src/core/algorithms/fd/dynfd/model/dynamic_relation_data.cpp
```diff
@@ -0,0 +1,265 @@
...
+                value_id = location->second;
+
+                if (value_id < 0) {
+                    int column_ind = -value_id - 1;
+                    value_id = -compressed_records[column_ind][index] - 1;
+                    location->second = value_id;
+                    compressed_records[column_ind][index] = value_id;
+                }
+
+                compressed_records.back()[index] = value_id;
```
ol-imorozko: The dictionary encoding uses negative values as temporary back-references for newly seen strings?
It's hard to understand, IMHO this really needs some sort of comment. Preferably, each method should come with some documentation explaining what it is intended to achieve and how.

src/core/algorithms/fd/dynfd/model/non_fd_tree.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+void NonFDTree::RemoveGenerals(boost::dynamic_bitset<> const& lhs, size_t rhs) {
+    root_->RemoveGeneralsRecursive(lhs, rhs, lhs.find_first(), false);
+}
+
+bool NonFDTree::ContainsNonFdOrSpecial(boost::dynamic_bitset<> const& lhs, size_t rhs) const {
+    return root_->ContainsNonFdOrSpecialRecursive(lhs, rhs, 0);
+}
+
+std::vector<LhsPair> NonFDTree::GetLevel(unsigned int target_level) {
```
ol-imorozko: In the header that is declared as `std::vector<LhsPair> GetLevel(unsigned target_level);`
I guess remove `int` here

src/core/algorithms/fd/dynfd/model/non_fd_tree.cpp
```diff
@@ -0,0 +1,97 @@
...
+#include <memory>
+#include <vector>
+
+#include <boost/dynamic_bitset.hpp>
+
+namespace algos::dynfd {
+
+std::shared_ptr<NonFDTreeVertex> NonFDTree::AddNonFD(
+        boost::dynamic_bitset<> const& lhs, size_t rhs,
+        ViolatingRecordPair violationPair = std::nullopt) {
```
ol-imorozko: Default argument makes no effect here. You should define it in the header file

src/core/algorithms/fd/dynfd/model/non_fd_tree.cpp
```diff
@@ -0,0 +1,97 @@
...
+    NonFDTreeVertex* cur_node = root_.get();
+    cur_node->SetAttribute(rhs);
```
ol-imorozko, line 13: why do we use raw pointer here and shared pointer in `FindNonFdVertex`? I think using shared pointer to traverse is fine

src/core/algorithms/fd/dynfd/model/non_fd_tree.cpp
```diff
@@ -0,0 +1,97 @@
...
+            added_node->SetNonFd(rhs, violationPair);
+            return added_node;
+        }
+
+        cur_node = cur_node->GetChild(bit);
+        cur_node->SetAttribute(rhs);
+    }
+    cur_node->SetNonFd(rhs, violationPair);
+    return nullptr;
+}
```
ol-imorozko: it returns a pointer to a newly created terminal node *only* in a specific branch of an `if` statement, and `nullptr` otherwise. Maybe it's better to return the vertex corresponding to the `lhs`, regardless of whether it was newly created. The caller can then inspect the node if needed.

Also I failed to find any usage of the return value of this method. Maybe we should just make it `void` and return nothing?

ol-imorozko: Also why do we need the second if?
I think this is more clear:
```cpp
    auto current_node = root_;
    current_node->SetAttribute(rhs);

    for (size_t bit = lhs.find_first(); bit != boost::dynamic_bitset<>::npos;
         bit = lhs.find_next(bit)) {
        current_node->AddChild(bit);
        current_node = current_node->GetChildShared(bit);
        current_node->SetAttribute(rhs);
    }

    current_node->SetNonFd(rhs, violationPair);
```

src/core/algorithms/fd/dynfd/model/non_fd_tree_vertex.h
```diff
@@ -0,0 +1,181 @@
...
+
+#include <boost/dynamic_bitset.hpp>
+
+#include "algorithms/fd/raw_fd.h"
+#include "dynamic_relation_data.h"
+
+namespace algos::dynfd {
+
+/**
+ * Pair of pointer ot FD tree node and the corresponding LHS.
```
ol-imorozko:
```suggestion
 * Pair of pointer of FD tree node and the corresponding LHS.
```

src/core/algorithms/fd/dynfd/model/non_fd_tree_vertex.h
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    std::shared_ptr<NonFDTreeVertex> GetChildShared(size_t const pos) const {
+        return children_.at(pos);
+    }
+
+    std::shared_ptr<NonFDTreeVertex> GetChildIfExists(size_t pos) const;
+
+    bool ContainsChildAt(size_t const pos) const {
+        return children_.at(pos) != nullptr;
```
ol-imorozko:
```suggestion
        return !children_.empty() && children_.at(pos) != nullptr;
```

src/core/algorithms/fd/dynfd/model/non_fd_tree_vertex.h
```diff
@@ -0,0 +1,181 @@
...
+          attributes_(numAttributes),
+          num_attributes_(numAttributes) {}
+
+    size_t GetNumAttributes() const noexcept {
+        return num_attributes_;
+    }
+
+    boost::dynamic_bitset<> GetNonFDs() const noexcept {
+        return non_fds_;
+    }
```
ol-imorozko: do we need to copy? If not, return const reference

src/core/algorithms/fd/dynfd/model/non_fd_tree_vertex.h
```diff
@@ -0,0 +1,181 @@
...
+        return children_.at(pos).get();
+    }
+
+    std::shared_ptr<NonFDTreeVertex> GetChildShared(size_t pos) {
+        return children_.at(pos);
+    }
+
+    boost::dynamic_bitset<> GetAttributes() const noexcept {
+        return attributes_;
+    }
```
ol-imorozko: do we need to copy? If not, return const reference

src/core/algorithms/fd/dynfd/model/non_fd_tree_vertex.h
```diff
@@ -0,0 +1,181 @@
...
+
+    friend class NonFDTree;
+
+    NonFDTreeVertex* GetChild(size_t pos) {
+        return children_.at(pos).get();
+    }
+
+    std::shared_ptr<NonFDTreeVertex> GetChildShared(size_t pos) {
+        return children_.at(pos);
+    }
```
ol-imorozko: why do we need all those
```cpp
// NonFDTreeVertex* GetChild(size_t pos)
// std::shared_ptr<NonFDTreeVertex> GetChildShared(size_t pos)
// std::shared_ptr<NonFDTreeVertex> GetChildPtr(size_t const pos)
// NonFDTreeVertex const* GetChild(size_t const pos) const
// std::shared_ptr<NonFDTreeVertex> GetChildShared(size_t const pos) const
```
Can't we just use
```cpp
    std::shared_ptr<NonFDTreeVertex> GetChild(size_t pos) {
        return children_.at(pos);
    }

    std::shared_ptr<const NonFDTreeVertex> GetChild(size_t pos) const {
        return children_.at(pos);
    }
```
and that's it?

src/core/algorithms/fd/dynfd/model/non_fd_tree_vertex.h
```diff
@@ -0,0 +1,181 @@
...
+    /**
+     * Constructs empty child node at the given position. Does nothing if the child already exists.
+     *
+     * @param pos child position
+     * @return whether a child was constructed
+     */
+    bool AddChild(size_t pos);
+
+    void GetLevelRecursive(unsigned target_level, unsigned cur_level, boost::dynamic_bitset<> lhs,
+                           std::vector<LhsPair>& vertices);
```
ol-imorozko: uses `unsigned` for levels, while the rest of the code uses `size_t`

ol-imorozko (changes requested):

src/core/algorithms/fd/dynfd/model/non_fd_tree.cpp
m4xxx1m: `FindNonFdVertex` returns shared_ptr that I use to traverse.
I use raw pointers here to avoid overhead of shared_ptr and because I don't need to share ownership of this pointer here.

PR #574, opened 2025-06-22
Pattern FD verifier algorithm

itisalisas:


src/core/algorithms/fd/pattern_fd_verifier/highlight.h
```diff
@@ -0,0 +1,24 @@
...
+namespace algos::pattern_fd {
+
+class Highlight {
+private:
+    model::PLI::Cluster cluster_;
+    std::vector<size_t> violating_rows_;
+
+public:
+    Highlight(model::PLI::Cluster cluster, std::vector<size_t> violating_rows)
+        : cluster_(cluster), violating_rows_(violating_rows) {}
```
ol-imorozko:
```suggestion
    Highlight(model::PLI::Cluster cluster, std::vector<size_t> violating_rows)
        : cluster_(std::move(cluster)), violating_rows_(std::move(violating_rows)) {}
```

src/core/algorithms/fd/pattern_fd_verifier/highlight.h
```diff
@@ -0,0 +1,24 @@
...
+
+    model::PLI::Cluster const& GetCluster() const {
+        return cluster_;
+    }
+
+    std::vector<size_t> GetViolatingRows() const {
+        return violating_rows_;
+    }
```
ol-imorozko: is no const ref expected here?

src/core/algorithms/fd/pattern_fd_verifier/model/pattern_info.h
```diff
@@ -0,0 +1,109 @@
...
+public:
+    TokenPatternInfo(TokenNGram const& token, Pos position) : token_(token), position_(position) {}
+
+    std::string Type() const override {
+        return "TokenPatternInfo";
+    }
+
+    bool operator==(TokenPatternInfo const& other) const {
+        return token_ == other.token_ && position_ == other.position_;
+    }
```
ol-imorozko, line 43: compiler can handle it `bool operator==(TokenPatternInfo const& other) const = default`. And also don't forget the `operator!=(..) = default` too

src/core/algorithms/fd/pattern_fd_verifier/model/pattern_info.h
```diff
@@ -0,0 +1,109 @@
...
+    std::string Type() const override {
+        return "RegexPatternInfo";
+    }
+
+    std::vector<std::pair<std::string, size_t>> GetLiterals() const {
+        return literals_;
+    }
+
+    std::vector<std::string> ExtractConstrainedParts(std::string const& value) const;
+
```
ol-imorozko, line 81: this does
```cpp
std::regex re(this->regex_);
```

That recompiles the regex every time. Store a compiled regex as a member.

src/core/algorithms/fd/pattern_fd_verifier/model/pattern_info.h
```diff
@@ -0,0 +1,109 @@
...
+        return regex_;
+    }
+
+    std::string Type() const override {
+        return "RegexPatternInfo";
+    }
+
+    std::vector<std::pair<std::string, size_t>> GetLiterals() const {
+        return literals_;
+    }
```
ol-imorozko: I suggest to return const reference

src/core/algorithms/fd/pattern_fd_verifier/model/pattern_info.h
```diff
@@ -0,0 +1,109 @@
...
+};
+
+class TokenPatternInfo : public PatternInfo {
+private:
+    TokenNGram token_;
+    Pos position_;
+
+public:
+    TokenPatternInfo(TokenNGram const& token, Pos position) : token_(token), position_(position) {}
```
ol-imorozko: nit: doing this 
```suggestion
    TokenPatternInfo(TokenNGram token, Pos position) : token_(std::move(token)), position_(position) {}
```
will allow for better usage when passing rvalues

src/core/algorithms/fd/pattern_fd_verifier/model/pattern_info.h
```diff
@@ -0,0 +1,109 @@
...
+namespace std {
+template <>
+struct hash<algos::pattern_fd::TokenPatternInfo> {
+    size_t operator()(algos::pattern_fd::TokenPatternInfo const& t) const noexcept {
+        size_t h1 = std::hash<std::string_view>{}(t.Token());
+        size_t h2 = std::hash<algos::pattern_fd::Pos>{}(t.Position());
+        return h1 ^ (h2 << 1);
+    }
+};
+}  // namespace std
```
ol-imorozko: It's better not to open `namespace std`, but to add specialization directly to `std::hash`:
```suggestion
template <>
struct std::hash<algos::pattern_fd::TokenPatternInfo> {
    size_t operator()(algos::pattern_fd::TokenPatternInfo const& t) const noexcept {
        size_t h1 = std::hash<std::string_view>{}(t.Token());
        size_t h2 = std::hash<algos::pattern_fd::Pos>{}(t.Position());
        return h1 ^ (h2 << 1);
    }
};
```

ol-imorozko: Also we have `boost::hash_combine`, it's better to use it instead of `h1 ^ (h2 << 1)`

src/core/algorithms/fd/pattern_fd_verifier/model/pattern_info.h
ol-imorozko: pattern parsing is duplicated across `ToRegex()`, `CountConstrainedGroups()` and `ExtractLiterals()`.
This is easy to desync, code duplication should generally be avoided.
I would suggest to do a single parse pass that produces smth like {regex_str,vector of literals, constrained_groups}.

src/core/algorithms/fd/pattern_fd_verifier/model/pattern_info.cpp
```diff
@@ -0,0 +1,155 @@
+#include "pattern_info.h"
+
+#include <regex>
+
+namespace algos::pattern_fd {
+
```
ol-imorozko: I assume that you've added `std::regex compiled_regex_;` as a member.

src/core/algorithms/fd/pattern_fd_verifier/model/pattern_info.cpp
```diff
@@ -0,0 +1,155 @@
...
+            size_t end_brace = p.find('}', i);
+            if (end_brace != std::string::npos) {
+                std::string num_str = p.substr(i + 1, end_brace - i - 1);
+                int count = std::stoi(num_str);
+                current_pos_in_regex += count;
+                i = end_brace;
+            }
+        } else if (c == '.' || c == '^' || c == '$' || c == '|' || c == '?' || c == '+' ||
+                   c == '*' || c == '[' || c == ']' || c == '(' || c == ')' || c == '<' ||
+                   c == '>') {
```
ol-imorozko: Better to extract it:
```cpp
constexpr bool IsSpecialRegexChar(char c) {
    return c == '.' || c == '^' || c == '$' || c == '|' || c == '?' || c == '+' || c == '*' ||
           c == '[' || c == ']' || c == '(' || c == ')' || c == '<' || c == '>';
```

src/core/algorithms/fd/pattern_fd_verifier/model/pattern_info.cpp
```diff
@@ -0,0 +1,155 @@
...
+    size_t i = 0;
+    while (i < pattern_regex.size()) {
+        if (pattern_regex[i] == '<') {
+            this->num_constrained_groups_++;
+            i++;
+        } else {
+            i++;
+        }
+    }
+}
```
ol-imorozko: Those functions should no longer be needed (and ExtractLiterals). We can do all the work inside the constructor to avoid code duplication. If we need `ToRegex` method, it can just do `return compiled_regex_;`

ol-imorozko: Something like this:
```cpp
RegexPatternInfo::RegexPatternInfo(std::string const& pattern) {
    // 1. Reserve memory to avoid repeated allocations
    regex_.reserve(pattern.size() * 2);
    literals_.reserve(pattern.size() / 2);

    std::string current_literal;
    size_t current_pos_in_match = 0;

    // Helper to flush accumulated literals
    auto flush_literal = [&](size_t offset_adjustment = 0) {
        if (!current_literal.empty()) {
            // calculated position subtracts the size of the literal we just built
            literals_.emplace_back(current_literal,
                                   current_pos_in_match - current_literal.size() - offset_adjustment);
            current_literal.clear();
        }
    };

    // 2. Single pass parsing using string_view
    std::string_view p(pattern);
    size_t i = 0;

    while (i < p.size()) {
        const char c = p[i];

        if (c == '<') {
            flush_literal();
            regex_ += '(';
            num_constrained_groups_++;
            i++;
        } else if (c == '>') {
            flush_literal();
            regex_ += ')';
            i++;
        } else if (c == '\\') {
            flush_literal();

            if (i + 1 >= p.size()) {
                // Trailing backslash is invalid
                throw std::invalid_argument("Invalid escape sequence at end of pattern");
            }

            const char next = p[i + 1];
            switch (next) {
                case 'A': regex_ += "."; break;
                case 'D': regex_ += "\\d"; break;
                case 'U': regex_ += "[A-Z]"; break;
                case 'L': regex_ += "[a-z]"; break;
                case 'S': regex_ += "[^\\w]"; break;
                default:
                    throw std::invalid_argument(std::string("Invalid escape sequence: \\") + next);
            }

            // All supported escapes seem to advance logical match position by 1
            current_pos_in_match += 1;
            i += 2;
        } else if (c == '{') {
            flush_literal();

            size_t end = p.find('}', i);
            if (end == std::string_view::npos) {
                throw std::invalid_argument("Unclosed quantifier at position " + std::to_string(i));
            }

            // High-performance integer parsing
            size_t count = 0;
            auto result = std::from_chars(p.data() + i + 1, p.data() + end, count);

            if (result.ec != std::errc()) {
                 throw std::invalid_argument("Invalid number in quantifier");
            }

            // Append {N} to regex
            regex_ += '{';
            regex_.append(p.substr(i + 1, end - i - 1));
            regex_ += '}';

            // Update logical position based on quantifier
            // Note: Preserving original logic where {N} adds N to position
            current_pos_in_match += count;
            i = end + 1;
        } else if (IsSpecialRegexChar(c)) {
            flush_literal();
            regex_ += c;
            current_pos_in_match += 1;
            i++;
        } else {
            // Standard character, part of a literal
            current_literal += c;
            regex_ += c;
            current_pos_in_match += 1;
            i++;
        }
    }

    flush_literal();

    // 3. Compile Regex Once
    // Using optimize flag for better matching performance at cost of construction time
    compiled_regex_.assign(regex_, std::regex::optimize);
}
```

Please note that you should write it yourself, this code is mostly generated by LLM and I'm not sure whether it's correct or not

src/core/algorithms/fd/pattern_fd_verifier/model/pattern_info.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+std::vector<std::string> RegexPatternInfo::ExtractConstrainedParts(std::string const& value) const {
+    std::vector<std::string> parts;
+    if (num_constrained_groups_ == 0) {
+        return parts;
+    }
+
+    std::smatch match_results;
+    std::regex re(this->regex_);
```
ol-imorozko: won't be needed

src/core/algorithms/fd/pattern_fd_verifier/model/tokenizer.h
```diff
@@ -0,0 +1,41 @@
...
+    std::vector<TokenPatternInfo> GetTokenPatternInfo(std::string const& value);
+
+    std::vector<TokenPatternInfo> GetNGramPatternInfo(std::string_view value, double avg_len);
+
+public:
+    Tokenizer(std::shared_ptr<model::ColumnLayoutTypedRelationData> typed_relation)
+        : typed_relation_(std::move(typed_relation)), tokens_map_() {}
+
+    Tokenizer() : tokens_map_() {}
+
```
ol-imorozko: Is it required to have a default constructor here? It leaves `typed_relation_` as null which is dangerous. Maybe we should remove it?

src/core/algorithms/fd/pattern_fd_verifier/model/tokenizer.h
```diff
@@ -0,0 +1,41 @@
...
+namespace algos::pattern_fd {
+
+enum class TokenNGramType { kToken, kNGram };
+
+class Tokenizer {
+private:
+    std::shared_ptr<model::ColumnLayoutTypedRelationData> typed_relation_;
+
+    std::unordered_map<config::IndexType, std::unordered_map<TokenPatternInfo, std::vector<size_t>>>
```
ol-imorozko: nit:
smth like this will be nice for readability:
```cpp
using TokenToRows = std::unordered_map<TokenPatternInfo, std::vector<size_t>>;
using TokensMap = std::unordered_map<config::IndexType, TokenToRows>;
```

ol-imorozko: Do you mix `size_t`  and `config::IndexType`, or is this intentional?

ol-imorozko (changes requested):

PR #576, opened 2025-07-30
Rewrite AC example

Sneper-Breeze:
This PR extends the Python example for Algebraic Constraints (AC) to make it more understandable and extensive. The improvements include better explanations, additional usage examples, and edge-case handling to help users work with AC more effectively.

### DIFF
--- a/examples/test_examples/snapshots/snap_test_examples_pytest.py
+++ b/examples/test_examples/snapshots/snap_test_examples_pytest.py
@@ -929,25 +929,146 @@ Num distinct rhs values: 2
 
 '''
 
-snapshots['test_example[basic/mining_ac.py-None-mining_ac_output] mining_ac_output'] = '''Discovered ranges for (Delivery date - Dispatch date) are:
+snapshots['test_example[basic/mining_ac.py-None-mining_ac_output] mining_ac_output'] = '''This example is dedicated to Fuzzy Algebraic Constraints (AC). The definition and algorithm 
+are based on article "B-HUNT: Automatic Discovery of Fuzzy Algebraic Constraints in Relational 
+Data" by Paul G. Brown & Peter J. Haas presented at VLDB in 2003.
+
+First of all, let's figure out what AC is. However, to avoid going too deep, we will give you 
+a simple definition without formalization. AC represents the results of applying binary 
+operations between two table columns, with values grouped into meaningful intervals. 
+Let's illustrate this with an example.
+
+We have a table examples/datasets/player_stats.csv with the following data:
+|   id |   Strength |   Agility |
+|-----:|-----------:|----------:|
+|    0 |          3 |         1 |
+|    1 |          4 |         1 |
+|    2 |          1 |         3 |
+|    3 |          2 |         2 |
+|    4 |          1 |         4 |
+|    5 |         10 |        12 |
+|    6 |         14 |        10 |
+|    7 |          1 |        23 |
+|    8 |          6 |        16 |
+
+Let's apply binary operation "+" to the Strength and Agility columns and observe the results.
+
+\x1b[1;42mDiscovered ranges\x1b[0m for (Strength + Agility) are:
+[(4.0, 5.0), (22.0, 24.0)]
+
+Rows in which the result of the chosen operation (+) is \x1b[1;41moutside\x1b[1;49m of discovered ranges:
+\x1b[1;46mNone\x1b[1;49m
+
+As shown, the sum of Strength and Agility falls within either the (4, 5) or (22, 24) ranges. 
+This pattern may emerge because player characters with similar combined attribute values 
+likely belong to the same tier.
+
+
+To run the algorithm, you must configure the parameters below:
+For \x1b[1;34mbinary arithmetic operations\x1b[0m, you can use four options:\x1b[1;49m
+"+"
+"-"
+"*"
+"/"
+\x1b[0m
+Furthermore, AC mining algorithm provides five \x1b[1;34mparameters for setting up execution\x1b[0m:
+\x1b[1;49mweight
+fuzziness
+p_fuzz
+bumps_limit
+iterations_limit
+ac_seed
+\x1b[0m
+\x1b[1;42mWeight\x1b[0m accepts values in the range (0, 1]. 
+Values closer to 1 force the algorithm to produce fewer larger intervals (up to a single 
+interval covering all values).
+Values closer to 0 force the algorithm to produce smaller intervals.
+
+\x1b[1;42mFuzziness\x1b[0m belongs to (0,1) range while \x1b[1;42mp_fuzz\x1b[0m belongs to [0,1] range. These parameters 
+control precision and the number of considered rows.
+Fuzziness values closer to 0 and p_fuzz values closer to 1 force the algorithm to include 
+more rows (higher accuracy).
+Fuzziness values closer to 1 and p_fuzz values closer to 0 force the algorithm to include 
+fewer rows (higher chance of skipping rows, lower precision, but faster execution).
+
+\x1b[1;42mBumps_limit\x1b[0m accepts only natural numbers from the range [1, inf) and limits the number of 
+intervals for all column pairs. To set bumps_limit to inf you should use the 0 value.
+
+\x1b[1;42mIterations_limit\x1b[0m accepts only natural numbers. 
+Lower values (close to 1) reduce accuracy due to algorithm performing fewer iterations.
+
+\x1b[1;42mAC_seed\x1b[0m accepts only natural numbers. 
+B-HUNT is a randomized algorithm that accepts the seed parameter (AC_seed). Fixing this 
+parameter ensures reproducible results, which are necessary for verifying results during 
+testing of the algorithm. Furthermore, we need to fix it in this example for demonstration 
+purposes; otherwise, we may obtain a different number of intervals with different boundaries 
+that will not correspond to the text we wrote for our output.
+
+Let's proceed to a visual example. We will use dataset from this path: 
+examples/datasets/cargo_march.csv.
+For default parameters we will use those values:
+binary operation is "-", weight - 0.1, fuzziness - 0.2, p_fuzz - 0.85, bumps_limit - 0, 
+iterations_limit - 4, AC_seed - 11.
+
+Let's see the result of the algorithm with these parameters.
+
+\x1b[1;42mDiscovered ranges\x1b[0m for (Delivery date - Dispatch date) are:
 [(2.0, 7.0), (15.0, 22.0)]
 
-Rows in which the result of the chosen operation (-) is outside of discovered ranges:
-id: 7
+Rows in which the result of the chosen operation (-) is \x1b[1;41moutside\x1b[1;49m of discovered ranges:
+\x1b[1;46mid: 7\x1b[1;49m
 Dispatch date: 1
 Delivery date: 30
 Difference: 29
 
-id: 26
+\x1b[1;46mid: 26\x1b[1;49m
 Dispatch date: 7
 Delivery date: 18
 Difference: 11
 
-id: 30
+\x1b[1;46mid: 30\x1b[1;49m
 Dispatch date: 11
 Delivery date: 22
 Difference: 11
 
+You can see that the algorithm creates two intervals for the binary operation "-": (2-7) and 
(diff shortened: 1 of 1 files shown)


examples/basic/mining_ac.py
```diff
@@ -1,6 +1,22 @@
 import desbordante
 import pandas
 import operator
+from tabulate import tabulate
+
+
+#\x1b = \033
```
m4xxx1m: Комментарий кажется лишним

examples/basic/mining_ac.py
```diff
@@ -18,33 +34,160 @@
...
-    print(f'Dispatch date: {dispatch_date}')
-    print(f'Delivery date: {delivery_date}')
-    print(f'{operation_name}: {operation(delivery_date, dispatch_date)}')
+
+
+def print_table(table_, headers_ = None, title_ = None):
+    if title_ is not None:
+        print(title_)
+
+    table_ = pandas.read_csv(table_, sep=SEPARATOR, header=HEADER)
```
m4xxx1m: Датасет из файла считываем сначала здесь, а потом еще раз в методе result. Хотелось бы лишний раз не считывать

examples/basic/mining_ac.py
```diff
@@ -18,33 +34,160 @@
...
-print(f'Rows in which the result of the chosen operation ({BIN_OPERATION}) is outside of discovered ranges:')
-for ac_exception in ac_exceptions:
-    id, delivery_date, dispatch_date = df.iloc[ac_exception.row_index]
-    print(f'id: {id}')
-    print(f'Dispatch date: {dispatch_date}')
-    print(f'Delivery date: {delivery_date}')
-    print(f'{operation_name}: {operation(delivery_date, dispatch_date)}')
+
+
+def print_table(table_, headers_ = None, title_ = None):
```
m4xxx1m, line 41: Нижние подчеркивания в конце слова кажутся лишними (здесь и далее)

examples/basic/mining_ac.py
```diff
@@ -18,33 +34,160 @@
...
+
+    table_ = pandas.read_csv(table_, sep=SEPARATOR, header=HEADER)
+
+    print(tabulate(table_, headers=headers_, tablefmt='pipe', showindex=False, stralign='center'))
     print()
 
+
+def result(columns, p_fuzz_=P_FUZZ, fuzziness_=FUZZINESS, bumps_limit_=BUMPS_LIMIT, 
+         weight_=WEIGHT, bin_operation_=BIN_OPERATION, ac_seed_=AC_SEED, 
+         iterations_limit_=ITERATIONS_LIMIT, table_=TABLE):
```
m4xxx1m: `result`, на мой взгляд, - не очень хорошее название для данного метода, оно совсем не отражает то, что метод делает. Я бы предложил что-то вроде `run_ac_mining`

examples/basic/mining_ac.py
```diff
@@ -18,33 +34,160 @@
...
+
+    ac_exceptions = algo.get_ac_exceptions()
+    print()
+    print(f'Rows in which the result of the chosen operation ({bin_operation_}) is ' +
+          f'{COLOR_CODES["red_bg"]}outside{COLOR_CODES["default_bg"]} of discovered ranges:')
+    for ac_exception in ac_exceptions:
+        id, delivery_date, dispatch_date = df.iloc[ac_exception.row_index]
+        print(f'{COLOR_CODES["blue_bg"]}id: {id}{COLOR_CODES["default_bg"]}')
+        print(f'Dispatch date: {dispatch_date}')
+        print(f'Delivery date: {delivery_date}')
```
m4xxx1m: Данный метод принимает таблицу как параметр и работает с двумя разными таблицами. Даже если по факту для одной из таблиц `ac_exceptions` пустое, не очень хорошо хардкодить здесь поля второй таблицы.

examples/basic/mining_ac.py
```diff
@@ -18,33 +34,160 @@
...
+        print(f'{operation_name}: {operation(delivery_date, dispatch_date)}')
+        print()
+    if len(ac_exceptions) == 0:
+        print(f'{COLOR_CODES["blue_bg"]}None{COLOR_CODES["default_bg"]}\n')
+
+
+print(f'This example is dedicated to Fuzzy Algebraic Constraints (AC). ' +
+      'The definition and algorithm \nare based on article "B-HUNT: Automatic ' +
+      'Discovery of Fuzzy Algebraic Constraints in Relational \nData" by Paul G. ' +
+      'Brown & Peter J. Haas presented at VLDB in 2003.\n')
```
m4xxx1m: Вместо конкатенации нескольких строчек и использования `\n` лучше будет использовать многострочные комментарии (здесь и далее)

examples/basic/mining_ac.py
```diff
@@ -18,33 +34,160 @@
...
+      '\nlook into the underlying data. We can imagine several reasons for this result, such as: 1) ' +
+      '\nnearby addresses versus far addresses; 2) air shipping versus regular shipping.\n')
+print('There are three parcels that fall outside of these delivery intervals. Why? This is a point \nfor ' +
+      'further investigation, which requires additional context. There are many possible reasons \nfor ' +
+      'this: 1) on these dates there was a workers\' strike in some regions; or 2) an incorrect \naddress ' +
+      'was specified, which increased the delivery time; or 3) it is just a typo in the table.\n')
+
+print('Now we reduce the value of the parameter weight to 0.05.')
+
+result(['Delivery date', 'Dispatch date'], weight_=0.05)
```
m4xxx1m: То, что предыдущее значение `weight` вынесено в константу, а это захордкожено, вызывает некоторую несогласованность

examples/basic/mining_ac.py
```diff
@@ -18,33 +34,160 @@
...
+print(f'{COLOR_CODES["green_bg"]}AC_seed{COLOR_CODES["default"]} accepts only natural numbers. \n' +
+      'B-HUNT is a randomized algorithm that accepts the seed parameter (AC_seed). Fixing this \n' +
+      'parameter ensures reproducible results, which are necessary for verifying results during \ntesting ' +
+      'of the algorithm. Furthermore, we need to fix it in this example for demonstration \npurposes; ' +
+      'otherwise, we may obtain a different number of intervals with different boundaries \nthat will ' +
+      'not correspond to the text we wrote for our output.\n')
+
+print(f'Let\'s proceed to a visual example. We will use dataset from this path: \n{TABLE}. '+
+      '\nFor default parameters we will use those values:')
+print(f'binaty operation is "{BIN_OPERATION}", weight - {WEIGHT}, ' +
```
m4xxx1m:
```suggestion
print(f'binary operation is "{BIN_OPERATION}", weight - {WEIGHT}, ' +
```

examples/basic/mining_ac.py
```diff
@@ -18,33 +34,160 @@
...
+
+print('Now we reduce the value of the parameter weight to 0.05.')
+
+result(['Delivery date', 'Dispatch date'], weight_=0.05)
+
+print('You can see that the number of intervals increases, and there ' +
+      'is no longer any data outside of \nthe discovered ranges.')
+
+
+print('However, with this amount of intervals, it is difficult to draw immediate conclusions ' +
```
m4xxx1m:
```suggestion
print('However, with this number of intervals, it is difficult to draw immediate conclusions ' +
```

examples/basic/mining_ac.py
```diff
@@ -18,33 +34,160 @@
...
+      'about the \ndelivery date. In this case, a detailed analysis of other attributes might ' +
+      'enable more meaningful \npredictions for delivery times. For example, it may be a good ' +
+      'idea to partition data by the region \nattribute (or by month/quarter) and consider each ' +
+      'partition individually. ')
+
+print('\nAnother option is to try to find a parameter combination that will result in a fewer ' +
+      'number of \nintervals. Next, remember that the algorithm is randomized (unless you run it ' +
```
m4xxx1m:
```suggestion
print('\nAnother option is to try to find a parameter combination that will result in a fewer' +
      '\nintervals. Next, remember that the algorithm is randomized (unless you run it ' +
```
Или smaller number of intervals

examples/basic/mining_ac.py
Sneper-Breeze: Removing table reading outside the print_table() function would require removing it from the result() function as well. This would introduce several issues:

1. Increased function dependency on external data, making the code less modular
2. Greater difficulty for users to extract these functions for their own needs
3. Minimal benefit - the current implementation is only used once to display a small table with negligible performance impact

The current approach keeps the functions more self-contained and reusable while maintaining clean code structure.

m4xxx1m: In my opinion taking pandas dataframe as `print_table` and `result` parameters is much more reusable than a path to dataset in csv format

examples/basic/mining_ac.py
Sneper-Breeze: This is needed when demonstrating how to modify parameters

examples/basic/mining_ac.py
Sneper-Breeze: Then the names of the variables will match the names of the arguments to the some methods.

examples/basic/mining_ac.py
m4xxx1m: I just meant that defining 0.05 as a constant (REDUCED_WEIGHT, for example) would be more consistent (considering previous WEIGHT)

examples/basic/mining_ac.py
m4xxx1m: I don't see a problem here. There's no variable shadowing and meaning of each name occurrence remain quite clear. But it doesn't really matter, so it's up to you.

examples/basic/mining_ac.py
```diff
@@ -18,33 +34,211 @@
...
+'''First of all, let's figure out what AC is. However, to avoid going too deep, we will give you 
+a simple definition without formalization. AC represents the results of applying binary 
+operations between two table columns, with values grouped into meaningful intervals. 
+Let's illustrate this with an example.
+'''
+)
+
+print('We have a table examples/datasets/player_stats.csv with the following data:')
+
+table_for_example = pandas.read_csv('examples/datasets/player_stats.csv',
```
m4xxx1m: This hardcoded table name is inconsistent with the following constant TABLE (it's same as the comment about weight)

examples/basic/mining_ac.py
```diff
@@ -18,33 +34,211 @@
...
 
+
+def run_ac_mining(table_, columns_, p_fuzz_=P_FUZZ, fuzziness_=FUZZINESS,
+         bumps_limit_=BUMPS_LIMIT, weight_=WEIGHT, bin_operation_=BIN_OPERATION,
+         ac_seed_=AC_SEED, iterations_limit_=ITERATIONS_LIMIT):
+    operation, operation_name = OPERATIONS[bin_operation_]
+
+    algo = desbordante.ac.algorithms.Default()
+
+    df_without_id = table_[columns_]
```
m4xxx1m: It is not obvious that `columns_` argument contains all table columns except for `id`.
Maybe `columns_` argument isn't even needed here.
Otherwise I would change name of either `columns_` or `df_without_id`

examples/basic/mining_ac.py
```diff
@@ -18,33 +34,211 @@
...
+
+    ac_exceptions = algo.get_ac_exceptions()
+    print()
+    print(f'Rows in which the result of the chosen operation ({bin_operation_}) is ' +
+          f'{COLOR_CODES["red_bg"]}outside{COLOR_CODES["default_bg"]} of discovered ranges:')
+    for ac_exception in ac_exceptions:
+        id_range, column1, column2 = table_.iloc[ac_exception.row_index]
+        print(f'{COLOR_CODES["blue_bg"]}id: {id_range}{COLOR_CODES["default_bg"]}')
+        print(f'{columns_[1]}: {column2}')
+        print(f'{columns_[0]}: {column1}')
```
m4xxx1m: It assumes that table always contains only 3 columns.

Sneper-Breeze: No, here it is assumed that the table has at least 2 columns, since otherwise it is impossible to calculate the primitive. It's probably worth changing the name `column1` and `column2` to `column0` and `column1`. I just left the old names of the variables that were before me. But now it's probably just confusing.

chernishev (approved):
Approving on behalf of Maxim Ivanov

PR #578, opened 2025-08-01
Remove frozen from install-dependencies

wildsor:


Vdaleke (changes requested):
LGTM, but I would add to the commit description that it was missed in #556, to make the story clearer

Vdaleke (approved):

PR #580, opened 2025-08-06
Fix IsLastNodeOf method of HyFD's FD Tree

m4xxx1m:


src/core/algorithms/fd/hyfd/model/fd_tree_vertex.cpp
```diff
@@ -35,6 +35,7 @@ void FDTreeVertex::GetFdAndGeneralsRecursive(boost::dynamic_bitset<> const& lhs,
                                              std::vector<boost::dynamic_bitset<>>& result) const {
     if (IsFd(rhs)) {
         result.push_back(cur_lhs);
+        return;
```
wildsor: comment on why this is correct

src/core/algorithms/fd/hyfd/model/fd_tree_vertex.cpp
```diff
@@ -9,16 +9,16 @@ namespace algos::hyfd::fd_tree {
 
 void FDTreeVertex::GetLevelRecursive(unsigned target_level, unsigned cur_level,
                                      boost::dynamic_bitset<> lhs, std::vector<LhsPair>& vertices) {
-    if (cur_level == target_level) {
+    if (cur_level == target_level && fds_.any()) {
```
wildsor:
```suggestion
    if (cur_level == target_level) {
```
Move to line below: `if (fds_.any()) vertices. ...`

src/core/algorithms/fd/hyfd/model/fd_tree_vertex.cpp
```diff
@@ -84,13 +85,11 @@ bool FDTreeVertex::RemoveRecursive(boost::dynamic_bitset<> const& lhs, size_t rh
         }
 
         if (!children_[current_lhs_attr]->GetAttributes().any()) {
-            children_[current_lhs_attr].reset();
             children_[current_lhs_attr] = nullptr;
         }
     }
 
     if (IsLastNodeOf(rhs)) {
-        contains_children_ = false;
```
wildsor, line 93: Looks like this flag does not get set to false anymore

BUYT-1 (changes requested):

src/core/algorithms/fd/hyfd/model/fd_tree_vertex.cpp
m4xxx1m: If this node is valid minimal FD, than all its specializations will be not minimal FDs

wildsor: I meant adding a comment in the code.

src/core/algorithms/fd/hyfd/model/fd_tree_vertex.h
```diff
@@ -44,7 +44,7 @@ class FDTreeVertex : public std::enable_shared_from_this<FDTreeVertex> {
     /**
      * Flag for optimizing child existence check. Is true iff any children_ is set
      */
-    bool contains_children_ = false;
+    size_t children_count_ = 0;
```
wildsor, line 47: Comment above says "flag" but this is not a flag anymore.

BUYT-1 (changes requested):

src/core/algorithms/fd/hyfd/model/fd_tree_vertex.cpp
```diff
@@ -35,6 +37,8 @@ void FDTreeVertex::GetFdAndGeneralsRecursive(boost::dynamic_bitset<> const& lhs,
                                              std::vector<boost::dynamic_bitset<>>& result) const {
     if (IsFd(rhs)) {
         result.push_back(cur_lhs);
+        return;  // If this vertex is valid minimal FD, than all its specializations will be not
+                 // minimal FDs
```
wildsor:
```suggestion
        return;  // If this vertex has the RHS bit set, then none of its children will have this bit set.
```
A vertex is not an FD.

Here is how I understand what's happening here.
Each vertex in this trie represents an FD's LHS. FD LHSs are partially ordered by inclusion (`X` <= `Y` iff `X` \subseteq `Y`).
We can order the vertices themselves on the LHSs they represent. This structure's invariant is then this: none of the vertices following the one with bit `n` set have the bit `n` also set.
In particular, the child vertices always follow the parent in that order (they represent the parent's LHSs with added elements, so the parent's LHS is a subset). That's the fact that is used here.

PR #591, opened 2025-09-19
Fix bitset tests

p-senichenkov:
Fix custom bitset implementation tests:
* Replace `#if defined(__GNUG__) || !defined(__clang__)` with `#if defined(__GLIBCXX__)` to check standard library implementation instead of checking compilers
* Remove CheckImplementation test, because it can fail without breaking critical behaviour
* Rename FindFirst and FindNext tests to FindFirst_libstdcxxExtensions and FindNext_libstdcxxExtensions, because they are being activated only on libstdc++

### DIFF
--- a/src/tests/unit/test_bitset.cpp
+++ b/src/tests/unit/test_bitset.cpp
@@ -425,8 +425,8 @@ TYPED_TEST_P(BitsetTest, FromStream) {
 
 // SGI Extensions:
 // We don't test _Unchecked_XXX methods, as they're used inside of standard ones
-TYPED_TEST_P(BitsetTest, FindFirst) {
-#if defined(__GNUG__) && !defined(__clang__)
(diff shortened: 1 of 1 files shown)


src/tests/unit/test_bitset.cpp
```diff
@@ -426,7 +426,7 @@ TYPED_TEST_P(BitsetTest, FromStream) {
 // SGI Extensions:
 // We don't test _Unchecked_XXX methods, as they're used inside of standard ones
 TYPED_TEST_P(BitsetTest, FindFirst) {
-#if defined(__GNUG__) && !defined(__clang__)
```
ol-imorozko: nit: Maybe it's better to change it's name from `FindFirst` to `FindFirst_libstdcxxExtensions`
I think test reports will be cleaner this way

ol-imorozko (approved):

PR #593, opened 2025-09-22
CFD Python bindings

Godrik0:
This pull request refactors the way Conditional Functional Dependency (CFD) rules are represented and handled, moving from string-based attribute/value pairs to `RawCFD` model. This change affects the core algorithm, configuration, Python bindings, and tests.

### Core CFD Rule Representation and Algorithm Changes

* Replaced the use of string-based attribute/value pairs for CFD rules in `CFDVerifier` with the new `RawCFD` structure, which uses attribute indices and optional values for both LHS and RHS. 
* Updated configuration option names to use a single `cfd_rule` option instead of separate left/right rule options, and updated related descriptions. 

### RawCFD Model Enhancements

* Added methods, constructors, and operators to `RawCFD` and `RawCFD::RawItem`, including `ToString`, equality comparison, and default constructors.

### Python Bindings Improvements

* Refactored Python bindings for CFD rules to use the new `RawCFD` model, including custom constructors, string representations, equality, and hash support for both `RawCFD` and `RawItem`. Added helper functions for conversion between Python and C++. 
* Added necessary includes and type converters for `RawCFD` in Python utility files. 

### Test Suite Updates

* Updated all CFD verifier tests to use the new `RawCFD` rule representation. 

### Python Examples

* Python examples for mining and verification of CFD rules have been added and revised (guides with code samples).
* The examples show how to build `RawCFD`, run mining, and run verification on the results.


src/core/algorithms/cfd/model/raw_cfd.h
```diff
@@ -19,8 +19,15 @@ class RawCFD {
...
         }
+
+        std::string ToString() const;
+
         AttributeIndex attribute;         /* attribute column index */
         std::optional<std::string> value; /* pattern value is optional */
+
+        bool operator==(RawItem const& other) const {
+            return attribute == other.attribute && value == other.value;
+        }
```
MichaelS239:
```suggestion
        bool operator==(RawItem const& other) const = default;
```
Standard equality operators can be defaulted since C++20.

src/core/algorithms/cfd/model/raw_cfd.h
```diff
@@ -43,6 +52,10 @@ class RawCFD {
     RawItem const& GetRhs() const {
         return rhs_;
     }
+
+    bool operator==(RawCFD const& other) const {
+        return lhs_ == other.lhs_ && rhs_ == other.rhs_;
+    }
```
MichaelS239:
```suggestion
    bool operator==(RawCFD const& other) const = default;
```

src/core/algorithms/cfd/cfd_verifier/cfd_verifier.cpp
```diff
@@ -19,38 +19,27 @@ CFDVerifier::CFDVerifier() : Algorithm({}) {
...
 
-    auto validate_rule_part = [check_item_ids](std::vector<CFDAttributeValuePair> const& part) {
-        check_item_ids(part);
-    };
-
-    auto validate_single_pair = [check_item_ids](CFDAttributeValuePair const& pair) {
-        check_item_ids({pair});
+    auto validate_rule = [check_items](cfd::RawCFD const& rule) {
+        check_items(rule.GetLhs());
+        check_items({rule.GetRhs()});
```
MichaelS239: Creating temporary vector results in constructing unnecessary copies. Add separate `check_item` lambda for checking one item and call it from `check_items`.

src/core/algorithms/cfd/cfd_verifier/cfd_verifier.cpp
```diff
@@ -88,7 +75,8 @@ unsigned long long CFDVerifier::ExecuteInternal() {
         return item_ids;
     };
 
-    cfd_ = {build_item_ids(string_rule_left_), build_item_ids({string_rule_right_}).front()};
+    cfd_ = {build_item_ids(raw_cfd_rule_.GetLhs()),
+            build_item_ids({raw_cfd_rule_.GetRhs()}).front()};
```
MichaelS239: Add `build_item_id` lambda for building item_id for single `RawItem` and call it from `build_item_ids`.

src/core/algorithms/cfd/model/raw_cfd.h
```diff
@@ -32,6 +39,8 @@ class RawCFD {
 public:
     explicit RawCFD(RawItems lhs, RawItem rhs) : lhs_(std::move(lhs)), rhs_(std::move(rhs)) {}
```
MichaelS239, line 39: `std::move` is defined in header `<utility>`. So add `#include <utility>`.

src/python_bindings/cfd/bind_cfd.cpp
```diff
@@ -1,13 +1,55 @@
...
+        lhs.emplace_back(RawCFD::RawItem{attr, value});
+    }
+
+    if (rhs_tuple.size() != 2) throw py::value_error("RHS must be an (attribute, value) tuple");
+
+    int rhs_attr = py::cast<int>(rhs_tuple[0]);
+    std::optional<std::string> rhs_value;
+    if (!rhs_tuple[1].is_none()) rhs_value = py::cast<std::string>(rhs_tuple[1]);
+
+    return RawCFD(lhs, RawCFD::RawItem{rhs_attr, rhs_value});
```
MichaelS239:
```suggestion
    return RawCFD(std::move(lhs), RawCFD::RawItem{rhs_attr, rhs_value});
```

src/python_bindings/cfd/bind_cfd.cpp
```diff
@@ -1,13 +1,55 @@
...
+}
+
+template <typename ElementType>
+py::tuple VectorToTuple(std::vector<ElementType> vec) {
+    std::size_t const size = vec.size();
```
MichaelS239: Add `#include <cstddef>` for `std::size_t`.

examples/basic/mining_cfd.py
MichaelS239, line 1: As far as I understand from the article, there are two types of CFDs:
1. Constant (e.g. **{(0, Los Angeles)} -> (4, high)**) - basically, this is an AR
2. Variable (wildcard on the RHS)
    - All wildcards (e.g. **{(1, _)} -> (3, _)**) - basically, this is an AFD, with confidence serving as its `g3` measure
    - Mixed type (e.g. **{(0, Los Angeles), (1, _)} -> (3, _)**) - this case is CFD-specific, it is not covered by other primitives

Feel free to correct me, if I understood something wrongly.
If I'm right, then I have several suggestions:
1. All CFDs presented in the mining example are essentially AFDs, and all CFDs presented in the verification example are essentially ARs. The mixed type, which is specific to CFDs, is not covered by these examples. So it would be nice to show an example of such CFD in at least one of the examples, if it doesn't take a lot of your time.
2. Connection between CFDs, ARs and AFDs is not explained in the examples. I guess a proper explanation of this connection with references to AR and AFD examples may be useful.
3. The meaning of wildcard is not fully explained. The wildcard presence in the RHS determines the type of CFD (constant or variable).
4. As fas as I understand, CFD is considered to hold, if its support and confidence are greater than defined minimum ones. Also confidence being equal to 1 means exact holding (no violations), while a lower confidence means greater number of violations. These two insights are not clear from the examples. I suggest adding these facts to the examples and explaining a little bit more about support and confidence.

examples/basic/verifying_cfd.py
```diff
@@ -55,55 +148,250 @@ def print_table_with_highlight(table, indices_to_highlight):
...
+    print(f"Violating rows: {Colors.RED_FG}{sorted(violating_rows)}{Colors.RESET}")
+
+    lhs_items = cfd.lhs
+    rhs_item = cfd.rhs
+
+    for j, highlight in enumerate(highlights[:2], start=1):
+        cluster_violating = highlight.get_violating_rows
+        if cluster_violating:
+            print(f"  Cluster #{j} violations:")
+            for row_idx in list(cluster_violating)[:3]:
```
MichaelS239:
```suggestion
            for row_idx in list(cluster_violating)[:5]:
```
There is an example with four violations that is not fully printed because of this limit.

src/python_bindings/cfd/bind_cfd_verification.cpp
```diff
@@ -12,9 +15,11 @@ namespace py = pybind11;
...
 void BindCFDVerification(py::module_& main_module) {
     using namespace algos;
+    using namespace algos::cfd;
     using namespace algos::cfd_verifier;
 
     auto cfd_verification_module = main_module.def_submodule("cfd_verification");
+
     py::class_<Highlight>(cfd_verification_module, "Highlight")
             .def_property_readonly("cluster", &Highlight::GetCluster)
             .def_property_readonly("get_violating_rows", &Highlight::GetViolatingRows);
```
MichaelS239:
```suggestion
            .def_property_readonly("violating_rows", &Highlight::GetViolatingRows);
```

src/python_bindings/cfd/bind_cfd.cpp
```diff
@@ -1,13 +1,55 @@
...
+    for (std::size_t i = 0; i < size; ++i) {
+        tuple[i] = std::move(vec[i]);
+    }
+    return tuple;
+}
+
+py::tuple MakeCfdTuple(algos::cfd::RawCFD const& cfd) {
+    auto const& lhs = cfd.GetLhs();
+    auto const& rhs = cfd.GetRhs();
+    return py::make_tuple(VectorToTuple(std::move(lhs)), py::cast(rhs));
```
MichaelS239:
```suggestion
    RawItems lhs = cfd.GetLhs();
    RawItem rhs = cfd.GetRhs();
    return py::make_tuple(VectorToTuple(std::move(lhs)), py::cast(rhs));
```
`std::move` being called on a `const&` object doesn't actually move the object; it performs a copy instead.

src/core/algorithms/cfd/cfd_verifier/cfd_verifier.cpp
```diff
@@ -18,38 +18,31 @@ CFDVerifier::CFDVerifier() : Algorithm({}) {
...
+    auto check_items = [check_item](cfd::RawCFD::RawItems const& items) {
+        for (auto const& item : items) {
+            check_item(item);
+        }
     };
 
-    auto validate_single_pair = [check_item_ids](CFDAttributeValuePair const& pair) {
-        check_item_ids({pair});
+    auto validate_rule = [check_item, check_items](cfd::RawCFD const& rule) {
+        check_items(rule.GetLhs());
```
MichaelS239:
```suggestion
        std::ranges::for_each(rule.GetLhs(), check_item);
```
We can do something like this instead of writing `check_items` lambda.

src/core/algorithms/cfd/cfd_verifier/cfd_verifier.cpp
```diff
@@ -18,38 +18,31 @@ CFDVerifier::CFDVerifier() : Algorithm({}) {
...
-        check_item_ids(part);
+    auto check_items = [check_item](cfd::RawCFD::RawItems const& items) {
+        for (auto const& item : items) {
+            check_item(item);
+        }
     };
 
-    auto validate_single_pair = [check_item_ids](CFDAttributeValuePair const& pair) {
-        check_item_ids({pair});
+    auto validate_rule = [check_item, check_items](cfd::RawCFD const& rule) {
```
MichaelS239:
```suggestion
    auto validate_rule = [&check_item](cfd::RawCFD const& rule) {
```

src/python_bindings/cfd/bind_cfd.cpp
```diff
@@ -1,13 +1,57 @@
 #include "python_bindings/cfd/bind_cfd.h"
 
+#include <cstddef>
```
MichaelS239:
```suggestion
#include <cstddef>
#include <optional>
#include <string>
#include <utility>
#include <vector>
```

src/tests/unit/test_cfd_verifier.cpp
```diff
@@ -1,26 +1,29 @@
 #include <iostream>
+#include <optional>
 #include <string>
 #include <vector>
```
MichaelS239, line 3:
```suggestion
#include <utility>
```

src/python_bindings/cfd/bind_cfd.cpp
```diff
@@ -1,13 +1,57 @@
...
 #include "core/algorithms/cfd/model/raw_cfd.h"
 #include "python_bindings/py_util/bind_primitive.h"
 
+namespace {
+
 namespace py = pybind11;
+using algos::cfd::RawCFD;
+
+RawCFD CreateRawCFDFromTuples(py::list lhs_tuples, py::tuple rhs_tuple) {
+    RawCFD::RawItems lhs;
```
MichaelS239, line 23:
```suggestion
    RawCFD::RawItems lhs;
    lhs.reserve(lhs_tuples.size());
```
Add `reserve` to avoid reallocations.

src/tests/unit/test_cfd_verifier.cpp
```diff
@@ -1,26 +1,29 @@
...
```
MichaelS239:
```suggestion
#include <optional>
```
I guess that these includes are no longer needed.

src/core/algorithms/cfd/cfd_verifier/cfd_verifier.cpp
MichaelS239: Also `#include <algorithm>` is needed for `std::ranges::for_each`.

src/core/algorithms/cfd/cfd_verifier/cfd_verifier.cpp
```diff
@@ -64,30 +57,33 @@ void CFDVerifier::LoadDataInternal() {
...
-            if (item_name != "_") {
-                item_ids.push_back(relation_->GetItem(attr_id, item_name));
-            } else {
-                // Negative values (-1 - attr_id) are used to represent wildcards,
-                // indicating that the rule applies to any value in this column
-                item_ids.push_back(-1 - attr_id);
-            }
+        for (auto const& item : rule_part) {
+            item_ids.push_back(build_item_id(item));
         }
```
MichaelS239:
```suggestion
        std::ranges::transform(rule_part, std::back_inserter(item_ids), build_item_id);
```
This is probably a nicer way to do it. Note that `#include <iterator>` is needed for `std::back_inserter`.

MichaelS239 (commented):
Only [these](https://github.com/Desbordante/desbordante-core/pull/593/files#r2417855415) suggestions to examples are left, everything else is fine.

examples/basic/mining_cfd.py
Godrik0: Hi! I've added a few lines about it, could you take a [look](https://github.com/Desbordante/desbordante-core/compare/dad8041228051b989b987e28340f077b12c76f06..4bd2501419e1df6098785f89487e9ed28cfb27ef#diff-6a3eece31db97605ea5c16a2bec8ddb61932217f4e5ef96c31eb2ef650efe433)?

examples/basic/mining_cfd.py
```diff
@@ -1,108 +1,330 @@
...
+    df = pd.read_csv(TABLE_PATH)
+
+    print("Let’s consider an example dataset containing employee data.")
+    print()
+    print(df.to_string())
+    print()
+
+    print("Expected patterns in this dataset:")
+    print("1. We can see that each position has the same salary level. This can be expressed in the")
+    print("   following CFD: Position -> Salary. All managers have high salaries, all developers ")
```
MichaelS239:
```suggestion
    print("   following CFD: (Position=_) -> (Salary=_). All managers have high salaries, all developers ")
```

examples/basic/mining_cfd.py
```diff
@@ -1,108 +1,330 @@
...
+    print("   have average salaries, etc.")
+    print("2. We might expect certain positions to be department-specific, but examining the data reveals")
+    print("   that Managers work across multiple departments (IT, HR, Finance). This violates the potential")
+    print("   CFD: Position -> Department, showing that management roles span organizational boundaries.")
```
MichaelS239:
```suggestion
    print("   CFD: (Position=_) -> (Department=_), showing that management roles span organizational boundaries.")
```

examples/basic/mining_cfd.py
```diff
@@ -1,108 +1,330 @@
...
+    print("3. Looking at the data, one can note that all HR specialists reside in LA, except for one. This")
+    print("   fact can be expressed by the following CFD: Department=HR -> Location=LA. However, an HR")
```
MichaelS239:
```suggestion
    print("   fact can be expressed by the following CFD: (Department=HR) -> (Location=LA). However, an HR")
```

examples/basic/verifying_cfd.py
MichaelS239, line 1: I see that the content of the mining and verifying examples overlaps. So, I guess we need to copy the newly added information to the verifying example.

examples/basic/verifying_cfd.py
```diff
@@ -55,55 +148,250 @@ def print_table_with_highlight(table, indices_to_highlight):
...
+    explain_cfd_concept()
+
+    demonstrate_algorithm_parameters()
+    demonstrate_cfd_object_operations()
+
+    df = create_and_explain_dataset()
+    scenario_find_rhs_violations(df)
+
+    df = pd.read_csv(TABLE_PATH)
+    scenario_find_lhs_violations(df)
```
MichaelS239: Maybe it would be better to run these scenarios on `df_fixed` obtained from the first scenario: it looks natural that we fix the errors first ans then check our hypotheses on a cleaned dataset.

examples/basic/verifying_cfd.py
```diff
@@ -1,51 +1,144 @@
...
-    confidence = cfd_verifier.get_real_confidence()
-    print(f"{COLOR_CODES['default_bg']}CFD: {lhs} -> {rhs}")
-    if holds:
-        print(f"CFD holds: {COLOR_CODES['bold_green']}{holds}{COLOR_CODES['default']}")
-    else:
-        print(f"CFD holds: {COLOR_CODES['bold_red']}{holds}{COLOR_CODES['default']}")
+import pandas as pd
+
+TABLE_PATH = 'examples/datasets/cfd_datasets/city.csv'
+TABLE_PATH_FIXED = 'examples/datasets/cfd_datasets/city_fixed.csv'
```
MichaelS239: As far as I can see, this variable is not used anymore. So we can delete `city_fixed.csv` along with this variable.

examples/basic/verifying_cfd.py
```diff
@@ -55,55 +148,250 @@ def print_table_with_highlight(table, indices_to_highlight):
...
 
-    lhs, rhs = [("City", "Los Angeles"), ("BuildingType", "_")], ("BuildingCost", "high")
-    print(f'This hypothesis will be expressed as a rule: [("City", "Los Angeles"), ("BuildingType", "_")] -> ("BuildingCost", "high")\n')
-    algo.execute(cfd_rule_left=lhs, cfd_rule_right=rhs, cfd_minconf = 1)
+def print_cfd_metrics(df, cfd, verifier):
+    """Print CFD verification metrics and violations."""
 
-    print_results(table, algo, lhs, rhs)
-    print("\nWe can see that the rule is violated in line 6, which may indicate incorrect data entry. Let's fix them.\n") #
+    verifier.execute(cfd_rule=cfd, cfd_minconf=0)
```
MichaelS239:
```suggestion
    verifier.execute(cfd_rule=cfd, cfd_minconf=0.9)
```
We can set a reasonable confidence threshold, so that the message `CFD holds: False` appears when there are violations. We should also print the value of this execution parameter in the example for user's convenience.

examples/basic/verifying_cfd.py
MichaelS239: Also, it would be really nice to showcase variable and mixed CFDs in the verifying example. I have come up with the following examples:
- `(Street='_') -> (BuildingCost='_')` - buildings on the same street have relatively close costs (doesn't hold)
- `(City='New York' AND Street='_') -> (BuildingCost='_')` - the above hypothesis is true for buildings in New York

If you can think of better examples, feel free to suggest them.

src/python_bindings/py_util/py_to_any.cpp
```diff
@@ -204,6 +205,7 @@ std::unordered_map<std::type_index, ConvFunc> const kConverters{
         kNormalConvPair<std::pair<std::string, std::string>>,
         kNormalConvPair<std::vector<std::string>>,
         kNormalConvPair<std::unordered_map<std::string, std::vector<unsigned int>>>,
+        kNormalConvPair<algos::cfd::RawCFD>,
```
MichaelS239, line 270: We should also add info about `RawCFD` to `get_py_type.h` and `opt_to_py.h`.

MichaelS239 (commented):
Sorry, previously I said that there were no other issues, but I've found a few more.

src/python_bindings/cfd/bind_cfd.cpp
```diff
@@ -2,13 +2,62 @@
...
+
+template <typename ElementType>
+py::tuple VectorToTuple(std::vector<ElementType> vec) {
+    std::size_t const size = vec.size();
+    py::tuple tuple(size);
+    for (std::size_t i = 0; i < size; ++i) {
+        tuple[i] = std::move(vec[i]);
+    }
+    return tuple;
+}
```
Oddin60F: This method is included in the file `py_util/vector_to_tuple.h`. Use it.

examples/basic/verifying_cfd.py
```diff
@@ -1,51 +1,165 @@
...
+    print("For example, consider employee data (a table), containing “Department”, “Head”, “Position”,")
+    print("and “Location” attributes. In this case:")
+    print()
+    print("* An FD Department -> Head means that the head is unambiguously determined by the ")
+    print("  department, indicating that each department (mentioned in the table) has exactly one head.")
+    print()
+    print("* On the other hand, a CFD: (Department=IT) -> (Head=Smith) means that the IT department")
+    print("  has Smith as its head. Other departments may or may not follow suit.")
+    print()
+    print("* Finally, there is CFD with a wildcard: (Department=IT, Position=_) -> (Location=NYC). The")
```
MichaelS239:
```suggestion
    print("* Finally, there is CFD with a wildcard: (Department=IT, Position=_) -> (Location=_). The")
```
As far as I can see, CFD `(Department=IT, Position=_) -> (Location=NYC)` is not valid according to the definition (the article says that for constant CFDs, LHS consists of constants only), so a wildcard on the RHS should be added. The meaning also changes: you can write something like "the location is unambiguously determined by the position for the IT department" or "all employees of the IT department with the same position are located in the same city".

Same applies to `mining_cfd.py`.

MichaelS239 (commented):
Only one minor issue left.

examples/basic/verifying_cfd.py
Godrik0: Thanks! I fixed it.

MichaelS239 (approved):

PR #598, opened 2025-10-04
[DRAFT] build: simplify compiler check with regex

metrapoliten:
The previous approach was too verbose. Using regex makes the code more concise and easier to mantain.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,11 +40,7 @@ if(PROJECT_SOURCE_DIR STREQUAL PROJECT_BINARY_DIR)
 endif()
 
 # Untested compiler notification
-if(NOT
-   ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU"
-    OR "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang"
-    OR "${CMAKE_CXX_COMPILER_ID}" STREQUAL "AppleClang")
-)
+if(NOT "${CMAKE_CXX_COMPILER_ID}" MATCHES "(^GNU|AppleClang|Clang)$")
     message(WARNING "This project has primarily been tested with GCC, LLVM Clang, and Apple Clang."
                     "Other compilers may not be fully supported."
     )


CMakeLists.txt
```diff
@@ -40,11 +40,7 @@ if(PROJECT_SOURCE_DIR STREQUAL PROJECT_BINARY_DIR)
 endif()
 
 # Untested compiler notification
-if(NOT
-   ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU"
-    OR "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang"
-    OR "${CMAKE_CXX_COMPILER_ID}" STREQUAL "AppleClang")
-)
+if(NOT "${CMAKE_CXX_COMPILER_ID}" MATCHES "(^GNU|AppleClang|Clang)$")
```
wildsor: Should it not be this instead?
```suggestion
if(NOT "${CMAKE_CXX_COMPILER_ID}" MATCHES "^(GNU|AppleClang|Clang)$")
```

p-senichenkov: Why don't just

```suggestion
if(NOT "${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU|AppleClang|Clang")
```

metrapoliten: Ok, I'll check it again.

metrapoliten: Without anchors it won't catch smth like `1GNU`

metrapoliten: Thank you, the brackets are really unnecessary in the case ([docs on CMake regex](https://cmake.org/cmake/help/latest/command/string.html#regex-specification)). About `$` and `^` I've replied in the message below

p-senichenkov: Ok, I thought it matches the entire string, like "match" usually does. But what is the purpose of grouping?

metrapoliten: What do you mean by "grouping"?

p-senichenkov: By grouping I mean parens: `"^GNU|AppleClang|Clang$"`

metrapoliten: Oh, sorry, I've answered about `()`  to BUYT-1 above and already have pushed the commit that removes `()`.

BUYT-1 (approved):

PR #600, opened 2025-10-05
Remove TypoMiner

wildsor:
This pipeline has been recreated in `examples/expert/mine_typos.py`.
The original class is not reachable from Python and is no longer
accessed in the Web version, so we can remove it.
`ColumnLayoutRelationDataManager` was implemented to avoid creating
PLIs multiple times in this pipeline and is no longer needed, so was
removed.


polyntsov (approved):
It's `examples/expert/mine_typos.**py**`, not `examples/expert/mine_typos.h` in the commit descripton

PR #601, opened 2025-10-05
Add CFDFinder algorithm

Oddin60F:
Add CFDFinder, a mining algorithm for discovering Conditional Functional Dependencies (CFDs). It uses a novel, multi-stage approach involving candidate generation, pruning, and post-processing to find interesting CFDs in a dataset. The algorithm first discovers a set of minimal FDs, then generates and generalizes CFD candidates based on non-FDs, and finally constructs near-optimal tableaux for them. For more details on the approach, refer to the thesis "Discovering Interesting Conditional Functional Dependencies" by Maximilian Grundke.


src/core/algorithms/cfd/cfdfinder/model/expansion/constant_expansion_strategy.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+        auto new_entries = entries;
+        new_entries[i].entry = std::make_shared<ConstantEntry>(value);
+        results.emplace_back(std::move(new_entries));
+    }
+
+    return results;
+}
+
+}  // namespace algos::cfdfinder
```
p-senichenkov: All files should end with newline.

src/core/algorithms/cfd/cfdfinder/model/expansion/constant_expansion_strategy.cpp
```diff
@@ -0,0 +1,43 @@
+#include "constant_expansion_strategy.h"
+
+#include "algorithms/cfd/cfdfinder/model/pattern/constant_entry.h"
+#include "algorithms/cfd/cfdfinder/model/pattern/variable_entry.h"
```
p-senichenkov: Don't forget to include standard headers that you use. For example, here it's `boost/dynamic_bitset` and `cstddef`.

This applies to all files.

src/core/algorithms/cfd/cfdfinder/model/expansion/constant_expansion_strategy.h
```diff
@@ -0,0 +1,24 @@
+#pragma once
+
+#include <memory>
+#include <vector>
+
+#include "expansion_strategy.h"
```
p-senichenkov: We use filenames relative to 'src/core'

src/core/algorithms/cfd/cfdfinder/model/expansion/constant_expansion_strategy.h
```diff
@@ -0,0 +1,24 @@
...
+
+namespace algos::cfdfinder {
+
+class ConstantExpansion : public ExpansionStrategy {
+protected:
+    hy::RowsPtr compressed_records_;
```
p-senichenkov: You took `RowsPtr` definition form HyFD, aren't you?
I don't think it's a good idea, as we try to avoid dependencies between algorithms. As for me, it's better to define `RowPtr = std::shared_ptr<std::vector<std::vector<model::ColumnIndex>>>` somewhere in CFD tree or move `RowsPtr` definition to `model`.

Anyway, include header in which `RowsPtr` is defined.

src/core/algorithms/cfd/cfdfinder/model/expansion/constant_expansion_strategy.h
```diff
@@ -0,0 +1,24 @@
...
+
+    virtual std::list<Pattern> GetChildPatterns(Pattern const& pattern, Cluster const& cluster);
+
+public:
+    explicit ConstantExpansion(hy::RowsPtr compressed_records)
```
p-senichenkov: I haven't seen all usages of this constructor yet, but looks like `shared_ptr` can be taken by rvalue reference here.

src/core/algorithms/cfd/cfdfinder/model/expansion/positive_negative_constant_strategy.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+std::list<Pattern> PositiveNegativeConstantExpansion::GetChildPatterns(Pattern const& pattern,
+                                                                       Cluster const& cluster) {
+    std::list<Pattern> results;
+    auto const& entries = pattern.GetEntries();
+    for (size_t i = 0; i < entries.size(); ++i) {
+        if (entries[i].entry->GetType() != EntryType::kVariable) {
+            continue;
+        }
+        int value = compressed_records_->at(cluster[0])[entries[i].id];
```
p-senichenkov: If you use `at` to keep `compressed_records_` constant, mark this method as `const`.
Otherwise, prefer subscript operator, as it doesn't have overhead for key presence check:

```suggestion
        int value = (*compressed_records_)[cluster[0]][entries[i].id];
```

src/core/algorithms/cfd/cfdfinder/model/expansion/range_pattern_expansion.cpp
```diff
@@ -0,0 +1,75 @@
...
+        std::sort(mappings.begin(), mappings.end(), [](auto const& a, auto const& b) {
+            if (a.second.empty() && b.second.empty()) {
+                return false;
+            }
+            if (a.second.empty()) {
+                return true;
+            }
+            if (b.second.empty()) {
+                return false;
+            }
```
p-senichenkov: This gives the same result:
```suggestion
            if (b.second.empty()) {
                return false;
            }
            if (a.second.empty()) {
                return true;
            }
```

src/core/algorithms/cfd/cfdfinder/model/expansion/range_pattern_expansion.cpp
```diff
@@ -0,0 +1,75 @@
...
+            return a.second < b.second;
+        });
+
+        std::vector<size_t> sorted_cluster_ids;
+        for (auto const& entry : mappings) {
+            sorted_cluster_ids.push_back(entry.first);
+        }
```
p-senichenkov:
```suggestion
        std::vector<size_t> sorted_cluster_ids(mappings.size());
        std::transform(mappings.begin(), mappings.end(), sorted_cluster_ids.begin(), [](auto const& entry) { return entry.first; });
```

Or, even more clear:

```suggestion
        std::vector<size_t> sorted_cluster_ids(mappings.size());
        std::ranges::transform(mappings, sorted_cluster_ids.begin(), [](auto const& entry) { return entry.first; });
```

src/core/algorithms/cfd/cfdfinder/model/pattern/constant_entry.h
```diff
@@ -0,0 +1,38 @@
...
+public:
+    explicit ConstantEntry(size_t constant) : constant_(constant) {}
+
+    inline bool Matches(size_t value) const override final {
+        return constant_ == value;
+    }
+
+    bool operator==(Entry const& other) const override final {
+        if (other.GetType() != EntryType::kConstant) return false;
+        return constant_ == static_cast<ConstantEntry const&>(other).constant_;
```
p-senichenkov: It's not recommended to use such "home-brew RTTI" in C++ (see, for example, [styleguide](https://google.github.io/styleguide/cppguide.html#Run-Time_Type_Information__RTTI_), or "C++ Gotchas" by Stephen Dewhrust). Standard RTTI is also not encouraged, but here it doesn't look so bad.
I'd prefer `typeid` or `dynamic_cast` here.

src/core/algorithms/cfd/cfdfinder/model/hyfd/inductor.cpp
```diff
@@ -0,0 +1,50 @@
+#include "inductor.h"
+
+namespace algos::cfdfinder {
+
+void Inductor::UpdateFdTree(NonFDList&& non_fds) {
```
p-senichenkov, line 7: Why not `NonFDList const& non_fds`?

src/core/algorithms/cfd/cfdfinder/model/hyfd/preprocessor.cpp
```diff
@@ -0,0 +1,29 @@
...
+#include "algorithms/fd/hycommon/preprocessor.h"
+
+namespace {
+using PLIs = algos::hy::PLIs;
+
+PLIs BuildPLIs(algos::cfdfinder::CFDFinderRelationData* relation) {
+    PLIs plis;
+    std::transform(relation->GetColumnData().begin(), relation->GetColumnData().end(),
+                   std::back_inserter(plis),
+                   [](auto& column_data) { return column_data.GetPositionListIndex(); });
```
p-senichenkov: 1. `plis` size is known in advance
2. `std::mem_fn` can be used

```suggestion
    auto const& col_data = relation.GetColumnData();
    PLIs plis(col_data.size());
    std::transform(col_data.begin(), col_data.end(),
                   plis.begin(),
                   std::mem_fn(&ColumnData::GetPositionalListIndex));
```

Also, `ranges` can be used for brevity:

```suggestion
    auto const& col_data = relation.GetColumnData();
    PLIs plis(col_data.size());
    std::ranges::transform(col_data,
                   plis.begin(),
                   std::mem_fn(&ColumnData::GetPositionalListIndex));
```

src/core/algorithms/cfd/cfdfinder/model/hyfd/preprocessor.cpp
```diff
@@ -0,0 +1,29 @@
...
+}  // namespace
+
+namespace algos::cfdfinder {
+
+std::tuple<hy::PLIsPtr, ColumnsPtr, hy::RowsPtr> Preprocess(CFDFinderRelationData* relation) {
+    hy::PLIs plis = BuildPLIs(relation);
+    auto inverted_plis = hy::util::BuildInvertedPlis(plis);
+    auto compressed_records = hy::util::BuildRecordRepresentation(inverted_plis);
+
+    return std::make_tuple(std::make_shared<hy::PLIs>(std::move(plis)),
```
p-senichenkov: Building PLIs and then moving it into `shared_ptr` looks strange. `BuildPlots` can work directly with `shared_ptr`

src/core/algorithms/cfd/cfdfinder/model/hyfd/validator.cpp
```diff
@@ -0,0 +1,323 @@
...
+#include "algorithms/fd/hycommon/validator_helpers.h"
+#include "algorithms/fd/hyfd/hyfd_config.h"
+#include "types/bitset.h"
+
+namespace {
+using BitSet = boost::dynamic_bitset<>;
+
+std::unordered_set<size_t> AsSet(BitSet const& bitset) {
+    std::unordered_set<size_t> valid_rhss(bitset.count());
+    for (size_t attr = bitset.find_first(); attr != BitSet::npos; attr = bitset.find_next(attr)) {
```
p-senichenkov: I've seen such a loop over set bits multiple times. Maybe it's better to define a function like `std::for_each`?
```C++
void for_each_set_bit(boost::dynamic_bitset<> const& bs, std::function<void(std::size_t)>&& fun) {
    for (std::size_t attr = bs.find_first(); attr != boost::dynamic_bitset<>::npos; attr = bs.find_next(attr)) {
        fun(attr);
    }
}
```

src/core/algorithms/cfd/cfdfinder/model/hyfd/validator.cpp
```diff
@@ -0,0 +1,323 @@
...
+#include "algorithms/fd/hycommon/util/pli_util.h"
+#include "algorithms/fd/hycommon/validator_helpers.h"
+#include "algorithms/fd/hyfd/hyfd_config.h"
+#include "types/bitset.h"
+
+namespace {
+using BitSet = boost::dynamic_bitset<>;
+
+std::unordered_set<size_t> AsSet(BitSet const& bitset) {
+    std::unordered_set<size_t> valid_rhss(bitset.count());
```
p-senichenkov: Do you really need a set with `bitset.count()` buckets, or you just want to `reserve` this number of elements?

src/core/algorithms/cfd/cfdfinder/model/hyfd/validator.cpp
```diff
@@ -0,0 +1,323 @@
...
+        size_t const rhs_column = *it;
+        size_t const value = compressed_records[row][rhs_column];
+
+        if (algos::hy::PLIUtil::IsSingletonCluster(value) ||
+            value != rhs_record.first[rhs_ranks[rhs_column]]) {
+            comparison_suggestions.emplace_back(row, rhs_record.second);
+            it = valid_rhs_ids.erase(it);
+            if (valid_rhs_ids.empty()) {
+                return;
+            }
```
p-senichenkov, line 60: `it` here is iterator after removed element. It will be `valid_rhs_ids.end()` on empty set, so this check is redundant.

src/core/algorithms/cfd/cfdfinder/model/hyfd/validator.cpp
```diff
@@ -0,0 +1,323 @@
...
+}
+
+RhsRowId BuildRhsRowId(algos::hy::Rows const& compressed_records, BitSet const& rhs,
+                       std::vector<size_t> const& rhs_column_ids, size_t row) {
+    std::vector<size_t> rhs_sub_cluster(rhs.count());
+    for (size_t i = 0; i < rhs.count(); ++i) {
+        rhs_sub_cluster[i] = compressed_records[row][rhs_column_ids[i]];
+    }
+
+    return std::make_pair(std::move(rhs_sub_cluster), row);
```
p-senichenkov: nit:

```suggestion
    return {std::move(rhs_sub_cluster), row};
```

src/core/algorithms/cfd/cfdfinder/model/hyfd/validator.cpp
```diff
@@ -0,0 +1,323 @@
...
+#include <boost/asio/thread_pool.hpp>
+#include <easylogging++.h>
+
+#include "algorithms/fd/hycommon/util/pli_util.h"
+#include "algorithms/fd/hycommon/validator_helpers.h"
+#include "algorithms/fd/hyfd/hyfd_config.h"
+#include "types/bitset.h"
+
+namespace {
+using BitSet = boost::dynamic_bitset<>;
```
p-senichenkov: You've included `model/types/bitset.h` (btw, I can't see its usages in this file), which defines `Bitset` class, and defined `BitSet` here.
It's very likely that these types will be misused. I suggest renaming this to

```suggestion
using DynBitSet = boost::dynamic_bitset<>;
```

src/core/algorithms/cfd/cfdfinder/model/hyfd/validator.cpp
```diff
@@ -0,0 +1,323 @@
...
+    auto vertex = lhsPair.first;
+    auto const lhs = lhsPair.second;
+    auto const rhs = vertex->GetFDs();
+    size_t const rhs_count = rhs.count();
+
+    result.SetCountValidations(rhs_count);
+    result.SetCountIntersections(rhs_count);
+
+    for (size_t attr = rhs.find_first(); attr != BitSet::npos; attr = rhs.find_next(attr)) {
+        if (!(*plis_)[attr]->IsConstant()) {
```
p-senichenkov: It's matter of taste, but I'd rather used `at` here and marked this method as `const`.

src/core/algorithms/cfd/cfdfinder/model/hyfd/validator.cpp
```diff
@@ -0,0 +1,323 @@
...
+        size_t candidates = AddExtendedCandidatesFromInvalid(
+                next_level, *fds_, result.InvalidInstances(), num_attributes, max_non_fds_);
+        algos::hy::LogLevel(cur_level_vertices, result, candidates, current_level_number_, "FD");
+
+        size_t const num_invalid_fds = result.InvalidInstances().size();
+        size_t const num_valid_fds = result.CountValidations() - num_invalid_fds;
+        cur_level_vertices = std::move(next_level);
+        current_level_number_++;
+
+        if (num_invalid_fds > (long double)hyfd::HyFDConfig::kEfficiencyThreshold * num_valid_fds &&
```
p-senichenkov: The [styleguide](https://google.github.io/styleguide/cppguide.html#Floating-Point_Types) doesn't ecourage using `long double`, because it's not portable.
It's not a must, but I'd prefer division here (`num_invalid_fds / num_valid_fds > threshold`).

Also, according to [cppreference](https://en.cppreference.com/w/cpp/language/types.html), even a single `float` has greater range than `size_t`, so I don't understand why cast is needed (even if it's false, I think that maximum number of FDs is about 10^6).

Anyway, don't use C-style cast.

src/core/algorithms/cfd/cfdfinder/model/pattern/pattern.cpp
```diff
@@ -0,0 +1,68 @@
+#include "pattern.h"
+
+#include <numeric>
+
+namespace algos::cfdfinder {
+bool Pattern::Matches(algos::hy::Row const& tuple) const {
+    for (auto const& [id, entry] : entries_) {
+        if (!entry->Matches(tuple[id])) return false;
```
p-senichenkov: Our styleguide forbids two statements on one line, even in such simple cases.

src/core/algorithms/cfd/cfdfinder/model/pattern/pattern.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+bool Pattern::operator<(Pattern const& other) const noexcept {
+    if (other.support_ != support_) {
+        return other.support_ > support_;
+    }
+    if (PatternDebugController::IsDebugEnabled()) {
+        if (other.num_keepers_ != num_keepers_) {
+            return other.num_keepers_ > num_keepers_;
+        }
+        return other.number_ > number_;  // debug
```
p-senichenkov: Looks like this comment is not needed.

src/core/algorithms/cfd/cfdfinder/model/pattern/pattern.cpp
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    return other.num_keepers_ > num_keepers_;
```
p-senichenkov: As for me, tuple comparison here will be more readable:

```suggestion
    if (PatternDebugController::IsDebugEnabled()) {
        return std::tie(support, num_keepers_, number_,) < std::tie(other.support_, other.num_keepers_, other.number_);
    else {
        return std::tie(support_, num_keepers_) < std::tie(other.support_, other.num_keepers_);
    }
```

**Important:** if you plan to use this operator in `set`, this can result in undefined behaviour in Release mode, because patterns that differ only in number will be considered equal.
I recommend either to always compare patterns in Debug mode (by three keys) or to add this notion in operator's documentation.

Also, it's strongly recommended not to use different behavoiur in Debug and Release modes, because in this case it's actually needed to test and debug two different programs, and in future, someone may, for example, change Debug behaviour and don't change the Release one.

src/core/algorithms/cfd/cfdfinder/model/pattern/range_entry.h
```diff
@@ -0,0 +1,71 @@
...
+
+public:
+    RangeEntry(std::shared_ptr<std::vector<size_t>> sorted_clusters, size_t min_cluster,
+               size_t max_cluster)
+        : sorted_cluster_ids_(std::move(sorted_clusters)),
+          min_cluster_(min_cluster),
+          max_cluster_(max_cluster) {}
+
+    bool IncreaseLowerBound() {
+        min_cluster_++;
```
p-senichenkov: Previous value is not needed here, so pre-increment should be used. See [styleguide](https://google.github.io/styleguide/cppguide.html#Preincrement_and_Predecrement).

src/core/algorithms/cfd/cfdfinder/model/pattern/range_entry.h
```diff
@@ -0,0 +1,71 @@
...
+        return min_cluster_ <= max_cluster_;
+    }
+
+    bool DecreaseUpperBound() {
+        if (max_cluster_ == 0) {
+            return max_cluster_ >= (min_cluster_ + 1);
```
p-senichenkov: `min_cluster_` is `size_t`. How `0` can be greater than `min_cluster_ + 1)`?

If you meant `min_cluster == numeric_limits<size_t>::max()`, it's not the best way of expressing this. For example, new Clang will optimize the whole expression away, so it becomes `return false`.

src/core/algorithms/cfd/cfdfinder/model/pattern/range_entry.h
```diff
@@ -0,0 +1,71 @@
...
+        if (other.GetType() != EntryType::kRange) return false;
+        auto const* entry = static_cast<RangeEntry const*>(&other);
+
+        return min_cluster_ == entry->min_cluster_ && max_cluster_ == entry->max_cluster_ &&
+               sorted_cluster_ids_ == entry->sorted_cluster_ids_;
+    }
+
+    size_t Hash() const override {
+        size_t hash = boost::hash_range(sorted_cluster_ids_->begin(), sorted_cluster_ids_->end());
+        hash += std::hash<size_t>{}(min_cluster_);
```
p-senichenkov: nit: Usually, XOR is used for combining hashes.

src/core/algorithms/cfd/cfdfinder/model/pruning/legacy_pruning_strategy.cpp
```diff
@@ -0,0 +1,71 @@
...
+void LegacyPruning::AddPattern(Pattern const& pattern) {
+    cumulative_support_ += pattern.GetSupport();
+}
+
+void LegacyPruning::ExpandPattern(Pattern const& pattern) {
+    visited_.insert(pattern);
+}
+
+bool LegacyPruning::HasEnoughPatterns([[maybe_unused]] std::vector<Pattern> const& tableau) {
+    return cumulative_support_ >= min_support_ * static_cast<double>(num_tuples_);
```
p-senichenkov: `min_support` is `double`, so `num_tuples` will be implicitly converted to `double`, won't it?

src/core/algorithms/cfd/cfdfinder/model/pruning/legacy_pruning_strategy.cpp
```diff
@@ -0,0 +1,71 @@
...
+                entries[i].entry->GetType() == EntryType::kNegativeConstant) {
+                auto new_entries = entries;
+                new_entries[i].entry = std::make_shared<VariableEntry>();
+                if (!visited_.contains(Pattern(std::move(new_entries)))) {
+                    return false;
+                }
+            }
+        }
+        return true;
+    }
```
p-senichenkov: Again, I don't like the idea of two different methods for debug and release modes. It's twice harder to test, debug and keep them coherent.

src/core/algorithms/cfd/cfdfinder/model/pruning/legacy_pruning_strategy.cpp
```diff
@@ -0,0 +1,71 @@
...
+                auto new_entries = entries;
+                new_entries[i].entry = std::make_shared<VariableEntry>();
+                parent_patterns.emplace_back(std::move(new_entries));
+            }
+        }
+        for (auto const& parent : parent_patterns) {
+            if (!visited_.contains(parent)) {
+                return false;
+            }
+        }
```
p-senichenkov: I don't quite understand, what is the purpose of this modification (compared to release varitant)?

src/core/algorithms/cfd/cfdfinder/model/pruning/partial_fd_pruning.h
```diff
@@ -0,0 +1,87 @@
...
+            for (auto const& entry : value_counts) {
+                non_violation_pairs += entry.second * entry.second;
+            }
+
+            size_t cluster_violations = total_pairs - non_violation_pairs;
+            violations += cluster_violations;
+        }
+
+        return violations;
+    }
```
p-senichenkov: nit: This method is quite long, so it should be in `.cpp` file

src/core/algorithms/cfd/cfdfinder/model/result/result_lattice.h
```diff
@@ -0,0 +1,58 @@
...
+#include "algorithms/cfd/cfdfinder/util/lhs_utils.h"
+#include "raw_cfd.h"
+
+namespace algos::cfdfinder {
+class ResultLattice {
+private:
+    size_t rhs_;
+    std::map<size_t, std::list<RawCFD>, std::greater<size_t>> layers_;
+
+    bool HasChildrens(RawCFD const& node) const {
```
p-senichenkov: nit: "children" is already in plural form

src/core/algorithms/cfd/cfdfinder/model/result/result_lattice.h
```diff
@@ -0,0 +1,58 @@
...
+        size_t level_num = node.embedded_fd_.lhs_.count() - 1;
+        if (!layers_.contains(level_num)) {
+            return false;
+        }
```
p-senichenkov, line 21: You can save a lookup by using `find` here.

src/core/algorithms/cfd/cfdfinder/model/cfdfinder_relation_data.cpp
```diff
@@ -0,0 +1,86 @@
...
+    for (auto& iter : cluster_map) {
+        size += iter.second.size();
+        if (iter.second.size() == 1) {
+            continue;
+        }
+        clusters.push_back(iter.second);
+    }
+    relation_size += size;
+    std::sort(clusters.begin(), clusters.end(),
+              [](std::vector<int> const& a, std::vector<int> const& b) { return a[0] < b[0]; });
```
p-senichenkov: FYI: `ranges` can be used for brevity:
```suggestion
    std::ranges::sort(clusters, {}, [](std::vector<int> const& a) { return a[0]; });
```

src/core/algorithms/cfd/cfdfinder/model/cfdfinder_relation_data.cpp
```diff
@@ -0,0 +1,86 @@
...
+        row = data_stream.GetNextRow();
+
+        if (row.size() != num_columns) {
+            LOG(WARNING) << "Unexpected number of columns for a row, skipping (expected "
+                         << num_columns << ", got " << row.size() << ")";
+            continue;
+        }
+
+        int attribute_id = 0;
+        for (auto field : row) {
```
p-senichenkov: `field` is `std::string`, so it's better to capture it by reference

src/core/algorithms/cfd/cfdfinder/types/bitset.h
```diff
@@ -0,0 +1,8 @@
+#pragma once
+
+#include "boost/dynamic_bitset.hpp"
+
+namespace algos::cfdfinder {
+using BitSet = boost::dynamic_bitset<>;
```
p-senichenkov: I don't think that such a shorthand needs a separate header file.

Also, this file is named `types/bitset.h` and defines `BitSet`, and there is a file `model/types/bitset.h`, which defines `Bitset`. This may lead to misuse of these types and hard-to-track problems.
Also note that we have `include_directory("model")` in CMakeLists, so when you say `#include "types/bitset.h"` it's not obvious which of them will be included.

src/core/algorithms/cfd/cfdfinder/util/lhs_utils.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+        if (subset.any()) {
+            subsets.push_back(std::move(subset));
+        }
+    }
+
+    return subsets;
+}
+
+void AddLhsSubsets(Candidate const& candidate, std::set<Candidate>& level) {
```
p-senichenkov: It's matter of taste, but looks like usually such "C-style methods" take their "hand-made this" as first argument.

src/core/algorithms/cfd/cfdfinder/util/violations_util.cpp
```diff
@@ -0,0 +1,23 @@
...
+size_t CalculateViolations(Cluster const& cluster, hy::Row const& inverted_rhs_pli) {
+    size_t max_cluster_size = 0;
+    std::unordered_map<size_t, size_t> rhs_cluster_counts;
+
+    for (auto tuple : cluster) {
+        auto cluster_id = inverted_rhs_pli[tuple];
+        if (hy::PLIUtil::IsSingletonCluster(cluster_id)) continue;
+
+        rhs_cluster_counts[cluster_id] += 1;
+        max_cluster_size = std::max(max_cluster_size, rhs_cluster_counts[cluster_id]);
```
p-senichenkov:
```suggestion
        max_cluster_size = std::max(max_cluster_size, ++rhs_cluster_counts[cluster_id]);
```

src/core/algorithms/cfd/cfdfinder/util/violations_util.cpp
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    return (max_cluster_size > 0) ? (cluster.size() - max_cluster_size) : cluster.size() - 1;
```
p-senichenkov: Add parens for consistency:

```suggestion
    return (max_cluster_size > 0) ? (cluster.size() - max_cluster_size) : (cluster.size() - 1);
```

src/core/algorithms/cfd/cfdfinder/cfd.cpp
```diff
@@ -0,0 +1,99 @@
...
+Condition GetEntriesString(Pattern const& pattern,
+                           InvertedClusterMaps const& inverted_cluster_maps) {
+    Condition result;
+    static std::string const kNullRepresentation = "null";
+    static std::string const kNegationSign = "¬";
+    static std::string const kWildCard = "_";
+
+    for (auto const& [id, entry] : pattern.GetEntries()) {
+        auto const& inverted_cluster_map = inverted_cluster_maps[id];
+        switch (entry->GetType()) {
```
p-senichenkov: Again, about "self-made RTTI". Its main disadvantage is that it leads to such 'switch`es, which are hard to keep consistent with type system. I highly recommend using Visitor pattern here.

src/core/algorithms/cfd/cfdfinder/cfd.cpp
```diff
@@ -0,0 +1,99 @@
...
+            case EntryType::kVariable:
+                result.push_back(kWildCard);
+                break;
+            case EntryType::kConstant: {
+                auto const* constant_entry = static_cast<ConstantEntry const*>(entry.get());
+                std::string value =
+                        inverted_cluster_map.find(constant_entry->GetConstant())->second;
```
p-senichenkov: Why not `inverted_cluster_map.at(...)`?

src/core/algorithms/cfd/cfdfinder/cfd.cpp
```diff
@@ -0,0 +1,99 @@
...
+                break;
+            }
+            case EntryType::kNegativeConstant: {
+                auto const* neg_constant_entry =
+                        static_cast<NegativeConstantEntry const*>(entry.get());
+                std::string value =
+                        inverted_cluster_map.find(neg_constant_entry->GetConstant())->second;
+
+                value = (!value.empty()) ? kNegationSign + value
+                                         : kNegationSign + kNullRepresentation;
```
p-senichenkov:
```suggestion
                value = kNegationSign + (!value.empty() ? value
                                         : kNullRepresentation);
```

src/core/algorithms/cfd/cfdfinder/cfd.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+                std::string lower_bound;
+                std::string upper_bound;
+
+                lower_bound = inverted_cluster_map.find(range_entry->GetLowerBound())->second;
+                upper_bound = inverted_cluster_map.find(range_entry->GetUpperBound())->second;
+                // if (range_entry->GetLowerBound() == range_entry->GetUpperBound()) {
+                //     result.push_back("[" + lower_bound + "]");
+                //     break;
+                // }
```
p-senichenkov: Remove commented code

src/core/algorithms/cfd/cfdfinder/cfdfinder.cpp
```diff
@@ -0,0 +1,496 @@
...
+                                         {support_independent_eq,
+                                          {kPatternTreshold, kMinSupportGain, kMaxLevelSupportDrop,
+                                           kCfdMinimumConfidence}},
+                                         {partial_fd_eq, {kMaximumG1}},
+                                         {rhs_filter_eq,
+                                          {kPatternTreshold, kMinSupportGain, kMaxLevelSupportDrop,
+                                           kCfdMinimumConfidence, kRhsIndices}}}));
+}
+
+unsigned long long CFDFinder::ExecuteInternal() {
```
p-senichenkov: This function is very long. It seems to need to be split into stages.

src/core/algorithms/cfd/cfdfinder/cfdfinder.cpp
```diff
@@ -0,0 +1,496 @@
...
+            }
+        }
+
+        enriched_plis.push_back(std::move(enriched_pli));
+        inverted_cluster_maps.push_back(std::move(inverted_cluster_map));
+    }
+
+    // если хранить единичные кластеры в model::PositionListIndex,
+    // хотя их интерфейс не предполагает этого, то можно
+    // обойтись без этого метода
```
p-senichenkov: Don't write comments in russian

src/core/algorithms/cfd/cfdfinder/cfdfinder.cpp
```diff
@@ -0,0 +1,496 @@
...
+
+    auto pruning_stategy = InitPruningStrategy(inverted_plis_shared);
+    auto result_receiver = InitResultStrategy();
+    auto expansion_stategy =
+            InitExpansionStrategy(compressed_records_shared, inverted_cluster_maps);
+
+    PLICache pli_cache(limit_pli_cache_, relation_->GetSchema());
+
+    size_t num_results = 0;
+    int height = levels.size() - 1;
```
p-senichenkov: In case of empty `levels`, `levels.size() - 1` will be first calculated as `numeric_limits<size_t>::max()` (due to unsigned underflow) and then **on some machines** become `-1` (due to signed underflow, **which is UB**). The right approach here will be (despite it being a little ugly):

```suggestion
    int height = levels.size();
    --height;
```

src/core/algorithms/cfd/cfdfinder/cfdfinder.cpp
```diff
@@ -0,0 +1,496 @@
...
+                util::AddLhsSubsets(candidate, levels[height - 1]);
+            }
+            num_results++;
+            result_receiver->ReceiveResult(std::move(candidate), std::move(pattern_tableau));
+        }
+        auto elapsed_seconds = std::chrono::duration_cast<std::chrono::seconds>(
+                std::chrono::system_clock::now() - start_level_time);
+        LOG(INFO) << "Finished level " << height << "(" << elapsed_seconds.count() << ", "
+                  << num_results << " results)";
+        height--;
```
p-senichenkov:
```suggestion
        --height;
```

src/core/algorithms/cfd/cfdfinder/cfdfinder.cpp
```diff
@@ -0,0 +1,496 @@
...
+    LOG(INFO) << "Total PLI computations: "
+              << pli_cache.GetTotalMisses() + pli_cache.GetFullHits() + pli_cache.GetPartialHits();
+    LOG(INFO) << "Total PLI cache misses: " << pli_cache.GetTotalMisses();
+    LOG(INFO) << "Total PLI cache hits: " << pli_cache.GetFullHits() + pli_cache.GetPartialHits();
+    LOG(INFO) << "Total full PLI cache hits: " << pli_cache.GetFullHits();
+    LOG(INFO) << "Total partial PLI cache hits: " << pli_cache.GetPartialHits();
+
+    auto elapsed_milliseconds = std::chrono::duration_cast<std::chrono::milliseconds>(
+            std::chrono::system_clock::now() - start_time);
+    unsigned long long apriori_millis = elapsed_milliseconds.count();
```
p-senichenkov: Apriori here is a name of algorithm, from which you've copied these lines, yeah? Here it's either `cfdfinder_millis` or `millis_count`.

src/core/algorithms/cfd/cfdfinder/cfdfinder.cpp
```diff
@@ -0,0 +1,496 @@
...
+        } else {
+            pli_cache.AddMiss();
+
+            auto intersected = result->Intersect(pli);
+            auto shared_intersected =
+                    std::shared_ptr<model::PositionListIndex>(std::move(intersected));
+
+            pli_cache.Put(current_lhs, shared_intersected);
+
+            result = pli_cache.Get(current_lhs) ? pli_cache.Get(current_lhs) : shared_intersected;
```
p-senichenkov: Double lookup doesn't look good.

```suggestion
            result = pli_cache.Get(current_lhs);
            if (!result) {
                result = shared_intersected;
            }
```

src/core/algorithms/cfd/cfdfinder/cfdfinder.cpp
```diff
@@ -0,0 +1,496 @@
...
+        if (comparison_suggestions.empty()) {
+            break;
+        }
+    }
+
+    auto fds = positive_cover_tree->FillFDs();
+
+    std::sort(fds.begin(), fds.end(), [](RawFD const& first, RawFD const& second) {
+        return first.lhs_.count() > second.lhs_.count();
+    });
```
p-senichenkov: Again, you can use `ranges` to save a few clicks and make it more readable, if you'd like to:

```suggestion
    std::ranges::sort(fds, std::greater<>{}, [](RawFD const& raw_fd) {
        return raw_fd.lhs_.count();
    });
```

src/core/algorithms/cfd/cfdfinder/cfdfinder.cpp
```diff
@@ -0,0 +1,496 @@
...
+                candidates.emplace_back(std::move(subset), fd.rhs_);
+            }
+        }
+    }
+
+    LOG(INFO) << candidates.size() << " maximal non-FDs (initial candidates).";
+    Lattice levels(relation_->GetNumColumns() - 1);
+
+    for (auto&& candidate : candidates) {
+        size_t level = candidate.lhs_.count() - 1;
```
p-senichenkov, line 379: Are you sure `lhs` cannot be empty here?

src/core/algorithms/cfd/cfdfinder/cfdfinder.cpp
```diff
@@ -0,0 +1,496 @@
...
+    while (!frontier.Empty() && !pruning_strategy->HasEnoughPatterns(tableau)) {
+        auto current_pattern = frontier.Poll();
+
+        if (pruning_strategy->IsPatternWorthAdding(current_pattern)) {
+            pruning_strategy->AddPattern(current_pattern);
+
+            Frontier new_frontier;
+
+            while (!frontier.Empty()) {
+                auto pattern(frontier.Poll());
```
p-senichenkov: nit: Paren-initialization looks strange with `auto`. I'd prefer `=`.

src/core/config/descriptions.h
```diff
@@ -79,6 +87,14 @@ auto const kDCfdSubstrategy = details::kDCfdSubstrategyString.c_str();
...
         "Number of tuples in the part of the dataset if you "
         "want to use algo not on the full dataset, but on its part";
+constexpr auto kDMaximumG1 = "maximum G1 value for the algorithm";
+constexpr auto kDPatternTreshold = "minimum number of patterns in the tableau";
+constexpr auto kDMinSupportGain = "minimum support gain for the algorithm.";
+constexpr auto kDMaxLevelSupportDrop = "maximum support drop for the algorithm.";
+constexpr auto kDLimitPliCache = "limit for the number of PLIs in the cache";
+auto const kDCfdExpansionStrategy = details::kDCfdExpansionStrategyString.c_str();
+auto const kDCfdPruningStrategy = details::kDCfdPruningStrategyString.c_str();
+auto const kDCfdResultStrategy = details::kDCfdResultStrategyString.c_str();
```
p-senichenkov: Names and descriptions were sorted alphabetically once...

p-senichenkov (changes requested):

src/core/algorithms/cfd/cfdfinder/model/expansion/range_pattern_expansion.cpp
```diff
@@ -0,0 +1,70 @@
...
+            new_entries[i].entry = lentry;
+            result.emplace_back(std::move(new_entries));
+        }
+
+        auto rentry = std::static_pointer_cast<RangeEntry>(range_entry->Clone());
+        if (rentry->DecreaseUpperBound()) {
+            auto new_entries = entries;
+            new_entries[i].entry = rentry;
+            result.emplace_back(std::move(new_entries));
+        }
```
Tydik42: The logic for creating child patterns is nearly identical. We could refactor this into a small helper lambda to avoid code duplication and make the intent clearer.

Also, `auto new_entries = entries` might become a performance hotspot. Is this copy necessary for each child pattern, or could there be a more efficient way to construct them, perhaps by sharing the unchanged parts?

src/core/algorithms/cfd/cfdfinder/model/expansion/range_pattern_expansion.cpp
```diff
@@ -0,0 +1,70 @@
...
+    }
+    return Pattern(std::move(entries));
+}
+
+std::list<Pattern> RangePatternExpansion::GetChildPatterns(Pattern const& current_pattern) {
+    std::list<Pattern> result;
+    auto const& entries = current_pattern.GetEntries();
+
+    for (size_t i = 0; i < entries.size(); ++i) {
+        auto range_entry = static_cast<RangeEntry const*>(entries[i].entry.get());
```
Tydik42: This `static_cast` assumes that the `entry` here will always be a `RangeEntry`. Is this invariant guaranteed by the design?

src/core/algorithms/cfd/cfdfinder/model/expansion/range_pattern_expansion.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+    for (size_t i = attributes.find_first(); i != BitSet::npos; i = attributes.find_next(i)) {
+        auto const& clusters = sorted_clusters_ids_.at(i);
+        entries.emplace_back(i, std::make_shared<RangeEntry>(clusters, 0, clusters->size() - 1));
+    }
+    return Pattern(std::move(entries));
+}
+
+std::list<Pattern> RangePatternExpansion::GetChildPatterns(Pattern const& current_pattern) {
+    std::list<Pattern> result;
```
Tydik42: Is there a specific reason for using `std::list` here?

src/core/algorithms/cfd/cfdfinder/model/expansion/range_pattern_expansion.cpp
```diff
@@ -0,0 +1,70 @@
...
+                return false;
+            }
+            if (a.second.empty()) {
+                return true;
+            }
+            if (b.second.empty()) {
+                return false;
+            }
+            return a.second < b.second;
+        });
```
Tydik42: nit: maybe it would be better to use `std::tie` here:
```suggestion
        std::sort(mappings.begin(), mappings.end(), [](auto const& a, auto const& b) {
            return std::tie(!a.second.empty(), a.second) < 
                   std::tie(!b.second.empty(), b.second);
});
```

Tydik42 (commented):
Just a few thoughts and suggestions here. Not blockers, just some ideas for potential improvements. Feel free to ignore them if they don't fit. Great work overall!

src/core/algorithms/cfd/cfdfinder/candidate.h
```diff
@@ -0,0 +1,37 @@
...
+    Candidate(boost::dynamic_bitset<> lhs, size_t rhs) noexcept : lhs_(std::move(lhs)), rhs_(rhs) {}
+
+    bool operator<(Candidate const& other) const {
+        if (rhs_ != other.rhs_) {
+            return rhs_ < other.rhs_;
+        }
+        return lhs_ < other.lhs_;
+    }
+
+    bool operator==(Candidate const& other) const {
```
xJoskiy: nit: `noexcept`

src/core/algorithms/cfd/cfdfinder/candidate.h
```diff
@@ -0,0 +1,37 @@
...
+namespace algos::cfdfinder {
+struct Candidate {
+    boost::dynamic_bitset<> lhs_;
+    size_t rhs_;
+
+    Candidate() = default;
+
+    Candidate(boost::dynamic_bitset<> lhs, size_t rhs) noexcept : lhs_(std::move(lhs)), rhs_(rhs) {}
+
+    bool operator<(Candidate const& other) const {
```
xJoskiy: nit: `noexcept`

src/core/algorithms/cfd/cfdfinder/cfd.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+namespace {
+using namespace algos::cfdfinder;
+using Condition = std::vector<std::string>;
+
+Condition GetEntriesString(Pattern const& pattern,
+                           InvertedClusterMaps const& inverted_cluster_maps) {
+    Condition result;
+    for (auto const& [id, entry] : pattern.GetEntries()) {
+        auto const& inverted_cluster_map = inverted_cluster_maps[id];
```
xJoskiy, line 19: `auto` type is not obvious. If types don't have very long name, it would be more clear for others if you wrote types explicitly

src/core/algorithms/cfd/cfdfinder/cfd.cpp
```diff
@@ -0,0 +1,49 @@
+#include "algorithms/cfd/cfdfinder/cfd.h"
+
+#include <boost/algorithm/string/join.hpp>
+
+#include "algorithms/cfd/cfdfinder/model/entries.h"
+
+namespace {
+using namespace algos::cfdfinder;
+using Condition = std::vector<std::string>;
```
xJoskiy, line 10: iwu for vector, string

src/core/algorithms/cfd/cfdfinder/cfd.h
```diff
@@ -0,0 +1,46 @@
...
+    double confidence_;
+
+public:
+    CFD(Vertical lhs, Column rhs, PatternTableau const& tableau,
+        std::shared_ptr<RelationalSchema const> schema,
+        InvertedClusterMaps const& inverted_cluster_maps);
+
+    std::string ToString() const;
+
+    double GetSupport() const {
```
xJoskiy: Also can be marked as `noexcept`

src/core/algorithms/cfd/cfdfinder/cfdfinder.cpp
```diff
@@ -0,0 +1,514 @@
...
+
+    for (auto const& fd : fds) {
+        for (auto&& subset : util::GenerateLhsSubsets(fd.lhs_)) {
+            if (!std::any_of(candidates.begin(), candidates.end(),
+                             [&subset, rhs = fd.rhs_](auto const& candidate) {
+                                 return rhs == candidate.rhs_ &&
+                                        subset.is_subset_of(candidate.lhs_);
+                             })) {
+                candidates.emplace_back(std::move(subset), fd.rhs_);
+            }
```
xJoskiy: You might want to split this predicate in parts for better readability, and use `std::ranges::any_of` to avoid boilerplates.

src/core/algorithms/cfd/cfdfinder/cfdfinder.cpp
```diff
@@ -0,0 +1,513 @@
...
+                for (auto&& subset : utils::GenerateLhsSubsets(candidate.lhs_)) {
+                    target_level.emplace(std::move(subset), candidate.rhs_);
+                }
+            }
+            ++num_results;
+            result_receiver->ReceiveResult(std::move(candidate), std::move(pattern_tableau));
+        }
+        auto elapsed_seconds = std::chrono::duration_cast<std::chrono::seconds>(
+                std::chrono::system_clock::now() - start_level_time);
+        LOG_INFO("Finished level {} ({}), {} results.", height, elapsed_seconds.count(),
```
p-senichenkov: nit:
```suggestion
        LOG_INFO("Finished level {} ({}ms), {} results.", height, elapsed_seconds.count(),
```

src/tests/unit/test_cfdfinder.cpp
```diff
@@ -0,0 +1,280 @@
...
+    template <typename... Args>
+    static std::unique_ptr<algos::cfdfinder::CFDFinder> CreateAlgorithmInstance(
+            algos::cfdfinder::Expansion expansion, algos::cfdfinder::Result result,
+            unsigned int max_lhs, Args&&... args) {
+        return algos::CreateAndLoadAlgorithm<algos::cfdfinder::CFDFinder>(
+                GetParamMap(expansion, result, max_lhs, std::forward<Args>(args)...));
+    }
+};
+
+TEST_F(CFDFinderAlgorithmTest, TennisConstantExpansionLegacyPruning) {
```
p-senichenkov: Maybe it's better to use paramethrized tests (`TEST_P`)?

p-senichenkov (approved):

src/core/algorithms/cfd/cfdfinder/model/expansion/range_pattern_expansion.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+namespace algos::cfdfinder {
+
+RangePatternExpansion::RangePatternExpansion(InvertedClusterMaps const& inverted_cluster_maps) {
+    sorted_clusters_ids_.reserve(inverted_cluster_maps.size());
+    for (size_t i = 0; i < inverted_cluster_maps.size(); ++i) {
+        std::vector<std::pair<ClusterId, AttributeValue>> mappings(inverted_cluster_maps[i].begin(),
+                                                                   inverted_cluster_maps[i].end());
+
+        std::sort(mappings.begin(), mappings.end(), [](auto const& a, auto const& b) {
```
xJoskiy: For consistency std::ranges::sort

src/core/algorithms/cfd/cfdfinder/model/hyfd/validator.cpp
```diff
@@ -0,0 +1,323 @@
...
+    result.SetCountIntersections(rhs_count);
+    result.SetCountValidations(rhs_count);
+
+    util::ForEachIndex(rhs, [&](size_t attr) {
+        for (auto const& cluster : plis_->at(lhs_attr)->GetIndex()) {
+            size_t const cluster_id = (*compressed_records_)[cluster[0]][attr];
+            if (algos::hy::PLIUtil::IsSingletonCluster(cluster_id) ||
+                std::any_of(cluster.cbegin(), cluster.cend(), [this, attr, cluster_id](int id) {
+                    return (*compressed_records_)[id][attr] != cluster_id;
+                })) {
```
xJoskiy: `std::ranges::any_of`

src/core/algorithms/cfd/cfdfinder/model/pattern/entry.h
```diff
@@ -0,0 +1,23 @@
...
+
+#include <cstddef>
+#include <string>
+
+#include "algorithms/cfd/cfdfinder/types/inverted_cluster_maps.h"
+
+namespace algos::cfdfinder {
+class Entry {
+protected:
+    inline static std::string const kNullRepresentation = "null";
```
xJoskiy: Note that static `std::string` is not constexpr and it may fail on memory allocation before main call, you may use `static std::string_view` so the data is guaranteed stored in .rodata segment. Applies everywhere else

src/core/algorithms/cfd/cfdfinder/model/pattern/pattern.cpp
```diff
@@ -0,0 +1,70 @@
...
+    support_ = static_cast<double>(GetNumCover());
+}
+
+void Pattern::UpdateKeepers(Row const& inverted_pli_rhs) {
+    size_t child_violations =
+            std::accumulate(cover_.begin(), cover_.end(), 0u,
+                            [&inverted_pli_rhs](size_t sum, Cluster const& cluster) {
+                                return sum + algos::cfdfinder::utils::CalculateViolations(
+                                                     cluster, inverted_pli_rhs);
+                            });
```
xJoskiy: Split it for better readability

src/core/algorithms/cfd/cfdfinder/model/pattern/pattern.h
```diff
@@ -0,0 +1,113 @@
...
+
+namespace algos::cfdfinder {
+
+class PatternDebugController {
+private:
+    inline static bool debug_enabled_ = false;
+    inline static unsigned long long counter_ = 0;
+
+public:
+    static bool IsDebugEnabled() {
```
xJoskiy: nit: `noexcept`

src/core/algorithms/cfd/cfdfinder/model/pattern/pattern.h
```diff
@@ -0,0 +1,113 @@
...
+    Pattern(Pattern&& other) noexcept = default;
+    Pattern(Pattern const& other) = default;
+    Pattern& operator=(Pattern&& other) noexcept = default;
+    Pattern& operator=(Pattern const& other) = default;
+
+    bool operator<(Pattern const& other) const noexcept;
+
+    bool operator==(Pattern const& other) const {
+        return entries_ == other.entries_;
+    };
```
xJoskiy: If you define < and == operators explicitly, you would also like to use other operators like !=, >, >=, but defining such operators by yourself is not neccessary. If you define spaceship operator <=> they will be available

src/core/algorithms/cfd/cfdfinder/model/pattern/pattern.h
```diff
@@ -0,0 +1,113 @@
...
+
+    bool Matches(Row const& tuple) const;
+    void UpdateCover(Pattern const& pattern);
+    void UpdateKeepers(Row const& inverted_pli_rhs);
+    size_t GetNumCover() const;
+
+    Entries const& GetEntries() const {
```
xJoskiy: nit: `noexcept`

src/core/algorithms/cfd/cfdfinder/model/pattern/pattern_item.h
```diff
@@ -0,0 +1,36 @@
...
+namespace algos::cfdfinder {
+struct PatternItem {
+    size_t id;
+    std::shared_ptr<Entry> entry;
+
+    PatternItem(size_t id, std::shared_ptr<Entry> entry) : id(id), entry(std::move(entry)) {}
+
+    bool operator==(PatternItem const& other) const {
+        return id == other.id && *entry == *(other.entry);
+    }
```
xJoskiy, line 18: `bool operator!=(PatternItem const& other) const = default`

src/core/algorithms/cfd/cfdfinder/model/pattern/pattern_item.h
```diff
@@ -0,0 +1,36 @@
+#pragma once
+#include <memory>
+#include <vector>
```
xJoskiy, line 3: `vector` is not used as I see

src/core/algorithms/cfd/cfdfinder/util/pli_cache.h
```diff
@@ -0,0 +1,71 @@
...
+
+    void Put(boost::dynamic_bitset<> columns, std::shared_ptr<model::PLI> pli) {
+        if (size_ >= limit_) {
+            return;
+        }
+        index_->Put(Vertical(schema_, std::move(columns)), std::move(pli));
+        ++size_;
+    }
+
+    void AddFullHit() {
```
xJoskiy: `noexcept`

src/core/algorithms/cfd/cfdfinder/model/result/result_lattice.h
```diff
@@ -0,0 +1,61 @@
...
+#include "raw_cfd.h"
+
+namespace algos::cfdfinder {
+class ResultLattice {
+private:
+    size_t rhs_;
+    std::map<size_t, std::list<RawCFD>, std::greater<size_t>> layers_;
+
+    bool HasChildren(RawCFD const& node) const {
+        size_t level_num = node.embedded_fd_.lhs_.count() - 1;
```
xJoskiy, line 18: Set an assertion that `lhs_.count()` is always positive, as somewhere above

src/core/algorithms/cfd/cfdfinder/model/pruning/support_independent_strategy.cpp
```diff
@@ -0,0 +1,74 @@
...
+                                                     double min_support_gain,
+                                                     double max_level_support_drop,
+                                                     double min_confidence)
+    : max_patterns_(pattern_threshold),
+      min_support_gain_(min_support_gain),
+      min_confidence_(min_confidence),
+      max_level_support_drop_(max_level_support_drop),
+      insufficient_support_gain_(false) {}
+
+void SupportIndependentPruning::StartNewTableau(Candidate const& candidate) {
```
xJoskiy, line 11: `Candidate`, `Pattern`, `PatterTableau` are not included

xJoskiy (changes requested):

src/core/algorithms/cfd/cfdfinder/cfdfinder.cpp
```diff
@@ -0,0 +1,513 @@
...
+    PLICache pli_cache(limit_pli_cache_, relation_->GetSchema());
+
+    size_t num_results = 0;
+    int height = levels.size();
+    --height;
+    while (height >= 0) {
+        auto start_level_time = std::chrono::system_clock::now();
+        auto& current_level = levels[height];
+        while (!current_level.empty()) {
+            auto candidate = current_level.extract(current_level.begin()).value();
```
xJoskiy: Why do you keep extracting values from `std::set` instead of iterating over it? Also should better replace `auto` for implicit types. It is not obvious what are the `candidate` and `current_level` types for example.

src/core/algorithms/cfd/cfdfinder/model/hyfd/inductor.h
```diff
@@ -0,0 +1,35 @@
+#pragma once
```
xJoskiy, line 1: Why did you copy all of hyfd files here? Why can't you just include hyfd?

src/core/algorithms/cfd/cfdfinder/model/pattern/constant_entry.h
```diff
@@ -0,0 +1,46 @@
...
+    explicit ConstantEntry(size_t constant) : constant_(constant) {}
+
+    inline bool Matches(size_t value) const override final {
+        return constant_ == value;
+    }
+
+    bool operator==(Entry const& other) const override final {
+        auto const* other_constant = dynamic_cast<ConstantEntry const*>(&other);
+        return other_constant != nullptr && constant_ == other_constant->constant_;
+    }
```
xJoskiy, line 23: if you add == also add `operator!=`

src/core/algorithms/cfd/cfdfinder/cfdfinder.cpp
```diff
@@ -0,0 +1,513 @@
...
+        cfd_collection_.emplace_back(std::move(lhs_v), std::move(rhs_c), result.patterns_,
+                                     relation_->GetSharedPtrSchema(), inverted_cluster_maps);
+    }
+}
+
+std::vector<Cluster> CFDFinder::EnrichPLI(model::PLI const* pli, int num_tuples) const {
+    auto const& original_clusters = pli->GetIndex();
+    std::vector<Cluster> enriched_clusters;
+
+    std::unordered_set<int> existing_elements;
```
xJoskiy: Move it closer to a place where it is used

src/core/algorithms/cfd/cfdfinder/cfdfinder.cpp
```diff
@@ -0,0 +1,513 @@
...
+
+    size_t total_elements = 0;
+    for (auto const& cluster : original_clusters) {
+        total_elements += cluster.size();
+    }
+    existing_elements.reserve(total_elements);
```
xJoskiy: Isn't total number of elements stored in PLI? If not it would be more efficient to cache total size inside PLI class

src/core/algorithms/cfd/cfdfinder/model/pattern/pattern_item.h
Oddin60F: `vector` is used below to define Entries:
`using Entries = std::vector<PatternItem>;`

src/core/algorithms/cfd/cfdfinder/model/hyfd/inductor.h
Oddin60F: The algorithm uses HyFD components with some changes that may affect the speed of the main algorithm if these changes are made to their code. It was decided to duplicate the necessary files.

src/core/algorithms/cfd/cfdfinder/cfdfinder.cpp
```diff
@@ -0,0 +1,501 @@
...
+
+    auto const& original_clusters = pli->GetIndex();
+    for (auto const& cluster : original_clusters) {
+        existing_elements.insert(cluster.begin(), cluster.end());
+    }
+    size_t missing_count = num_tuples - existing_elements.size();
+
+    std::vector<Cluster> enriched_clusters;
+    enriched_clusters.reserve(original_clusters.size() + missing_count);
+    enriched_clusters.assign(original_clusters.begin(), original_clusters.end());
```
p-senichenkov:
```suggestion
    std::vector<Cluster> enriched_clusters{original_clusters.begin(), original_clusters.end()};
    enriched_clusters.reserve(enriched_clusters.size() + missing_count);
```

PR #602, opened 2025-10-05
Fix CPM cache location set logic

BUYT-1:
The previous code was written incorrectly: `CACHE` was missing in `set`s. This lead to `CPM_CMAKE_CACHE` being set to `OFF` in the CMake cache. As `cmake/CPM.cmake` downloads `CPM.cmake` to the `CPM_CMAKE_CACHE` directory if said variable is set and to `build/cmake/` otherwise, the `CPM.cmake` files that were included in the end ended up being different. This was detected by the second `CPM.cmake` in `build/cmake` using its internal `CPM_DIRECTORY` and `CPM_CURRENT_DIRECTORY` variables and it returned before defining `CPMAddPackage`, resulting in failures if CMake was run without removing `CMakeCache.txt` prior.

Also, this set is now only performed if desbordante is built as a top-level project to avoid unrelated packages ending up in the source tree.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -213,13 +213,19 @@ endif()
 
 # --- Dependencies installation ---
 
-# Set CPM cache location
-if(NOT DEFINED CPM_SOURCE_CACHE)
-    if(DEFINED ENV{CPM_SOURCE_CACHE})
-        set(CPM_SOURCE_CACHE "$ENV{CPM_SOURCE_CACHE}")
-    else()
-        set(CPM_SOURCE_CACHE "${CMAKE_CURRENT_SOURCE_DIR}/cmake/cpm-cache")
-    endif()
+# Set CPM cache location. By default CPM does not enable cache. Here we set the source cache
+# location to a place in the source tree. This is only done if we are the top-level project, i.e.
+# this CMakeLists is used directly by `cmake ..` or similar to avoid the situation where an
+# unrelated project starts caching its dependencies in Desbordante's source tree.
+if(CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR
+   AND NOT DEFINED CPM_SOURCE_CACHE
+   AND NOT DEFINED ENV{CPM_SOURCE_CACHE}
+)
+    set(CPM_SOURCE_CACHE
+        "${CMAKE_CURRENT_SOURCE_DIR}/cmake/cpm-cache"
+        CACHE PATH "Directory to download CPM dependencies"
+    )
+    message(STATUS "Set CPM source cache location to ${CPM_SOURCE_CACHE}")
 endif()
 
 include(cmake/CPM.cmake)


Vdaleke (changes requested):
You also need to remove default in pyproject.toml for CPM_SOURCE_CACHE or remove the entire line to set default variable to nothing there (I'm not sure about how scikit-build and cmake work in both cases, need to check)
https://github.com/Desbordante/desbordante-core/blob/7898607c107b14080e0e3522a516cf6f15a06646/pyproject.toml#L36

As far as I remember, I've set default to this path for similar reasons, to have the one path for both cmake and toml. Maybe `pip install .` command doesn't create `build` directory that we use by default and in this case install dependencies twice and this is was a workaround for both install interfaces? (need to be check also)

Vdaleke:
You can also remove .gitignore inside cmake directory if we don't plan to store the cache locally in this location anyway.

Vdaleke:
Could you also explain where dependencies are stored if the variable isn't defined? If it's in build/_deps, as I recall, that's not very good.

CMake assumes that different configurations can be stored in different directories, for example, build/release or build/test.
I suspect that in this case, it will download a new configuration each time, although that doesn't make sense. I expect we'll see a similar folder separation after the introduction of CMakePresets; it makes sense.

If it's confirmed that pip doesn't use already cached dependencies, I'm more opposed to not specifying a single path for caching, as was previously the case, because downloading multiple dependencies doesn't make sense. If the CACHE variable was missing in the cmake set, you can add it.

BUYT-1:
> CMake assumes that different configurations can be stored in different directories, for example, build/release or build/test. I suspect that in this case, it will download dependencies for new configuration each time, although that doesn't make sense. I expect we'll see a similar folder separation after moving to CMakePresets.json, it makes sense for this case.

On one hand, I want to say that whoever is building should set the cache variable on their own, but on the other, I don't know who is actually going to do that, especially given that most contributors here are novices.
However, specifying the variable on our own has the potential to break things if desbordante is included in another CMakeLists.

Vdaleke:
> However, specifying the variable on our own has the potential to break things if desbordante is included in another CMakeLists.

Since I exposed the variable without the CACHE or PARENT_SCOPE identifier, it doesn't propagate to the parent cmake, and now we have an if-guard that doesn't override the variable in this case. I just see a case where only the ENV variable was used, and we need to change that to `if(NOT DEFINED CPM_SOURCE_CACHE OR DEFINED ENV{CPM_SOURCE_CACHE})`

But I agree there may be nuances here. Generally, it's safer to set this variable in build.sh or CMakePresets.json and stay with current pyproject.toml if we're leaving the cache in that path and in this case we can add CACHE to set() function.

Regarding the idea of ​​not enabling the cache by default and leaving it to the developer: a lot of applications, including package managers, enable their own cache by default, and I don't understand why it's a bad thing in this case. Few people will consider enabling the cache until they encounter consistently slow first builds, so why not help them right away?

Overall, I haven't yet heard any arguments why caching dependencies for all configurations in a single place by default in a project would be bad. This solution could be refined, taking into account the risk of breaking other projects that depend on desbordante-core, but the task of getting our CMake fully up to date for this case is assigned to someone else and part of a different, larger task.

build.sh
```diff
@@ -129,5 +129,4 @@ if [[ -n $SANITIZER ]]; then
     CMAKE_OPTS="$CMAKE_OPTS -D SANITIZER=${SANITIZER}"
 fi
 
-rm -f build/CMakeCache.txt
```
Vdaleke, line 132: This is done because there's a problem with caching variables across different configurations. This refers to debug/sanitizers and other things. If you don't remove them before building, the configurations won't change, apparently due to variable caching. This is supposedly the reason why it's a good idea to create separate build directories for different configurations, like build/debug and the like.

We removed them recently by mistake, and someone started complaining about glitches, but we quickly revert it https://github.com/Desbordante/desbordante-core/commit/b26a05d5a0257fcf38dca5722a521d16a5452fea

Vdaleke: In fact, to avoid deleting this file, you need always to explicitly define all the variables used in different configurations in build.sh so that they are updated in CMakeCache. You can do this if you want.

wildsor: Okay, that can be done in a separate PR

Vdaleke (approved):
Pls rename PR before merging

PR #604, opened 2025-10-07
Update Boost version requirements in README.md

perfotoran:
Exclude Boost 1.87.0 from supported versions due to test failures experienced by multiple contributors. Specify compatible versions as 1.85-1.86, 1.88+.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -293,7 +293,7 @@ if (COMPILE_BENCHMARKS)
 	list(APPEND boost_libraries program_options)
 endif()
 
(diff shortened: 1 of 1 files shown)


p-senichenkov (approved):

p-senichenkov:
If we really support only boost 1.85+, `CMakeLists.txt` should be updated. It's said that minimum required version is 1.81 there (on [line 296](https://github.com/Desbordante/desbordante-core/blob/main/CMakeLists.txt#L296C1-L296C66))

PR #606, opened 2025-10-08
Workflow for IWYU

PodushkaPIR:
# Include-What-You-Use (IWYU) Analysis Workflow

## Overview

This project provides a streamlined workflow for running Include-What-You-Use (IWYU) analysis on Desbordante configurable filtering and reporting. The tools help identify unnecessary includes and suggest missing ones.

### IWYU Installation

Choose one of the following installation methods:

#### Method 1: Package Installation
```bash
# Ubuntu/Debian
sudo apt install iwyu
```

#### Method 2: Build from Source

```bash
git clone https://github.com/include-what-you-use/include-what-you-use.git
cd include-what-you-use
mkdir build && cd build
cmake -G "Unix Makefiles" -DCMAKE_PREFIX_PATH=/usr/lib/llvm-19 ..
make
sudo make install
```

#### Required Versions:
- IWYU 0.23+
- Clang 19+

#### Additional Dependencies

- jq: JSON processor for configuration parsing
 
- Python 3: For output filtering script
 
- CMake: For generating build files

## Install dependencies on Ubuntu
```bash
sudo apt install jq python3 cmake
```

#### Core Scripts

- start_iwyu.sh - Main analysis script with command-line flags support

- output_cleaner.py - Filters and cleans IWYU output, removing noise and organizing results

#### Configuration Files

- iwyu_config.json - Central configuration for analysis parameters and filtering rules

- *.imp - IWYU mapping files [Format specification](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUMappings.md)


iwyu/start_iwyu.sh
p-senichenkov, line 1: Is it custom script or some part of IWYU tooling?

If it's custom, I'll suggest some changes to improve readability.
Anyway, looks like bash isn't a perfect language for such tasks.

iwyu/start_iwyu.sh
```diff
@@ -0,0 +1,184 @@
+#!/bin/sh
+
```
p-senichenkov:
```suggestion
set -e
```

This will abort script execution on first error, so `|| {...; exit 1}` on every line won't be needed.

p-senichenkov (commented):
I'll run these scripts, read IWYU docs and give more comments later.

iwyu/start_iwyu.sh
PodushkaPIR: Yes, this script is custom. I would be glad for help and suggestions.

iwyu/start_iwyu.sh
```diff
@@ -0,0 +1,184 @@
...
+BUILD_DIR="$TARGET_PATH/build"
+COMPILE_COMMANDS_FILE="$BUILD_DIR/compile_commands.json"
+
+if [ -d "$BUILD_DIR" ]; then
+    echo "Build directory exists"
+    if [ -f "$COMPILE_COMMANDS_FILE" ]; then
+        echo "compile_commands.json found, skipping rebuild"
+    else
+        echo "compile_commands.json not found, rebuilding"
+        rm -rf "$BUILD_DIR" || { echo "Error: Cannot remove $BUILD_DIR"; exit 1; }
```
p-senichenkov: FYI: You will never get this message, because `rm -f` always succeeds

iwyu/start_iwyu.sh
```diff
@@ -0,0 +1,184 @@
+#!/bin/sh
+set -e
```
p-senichenkov: Now you can remove all `|| {echo "Error...`. Error messages given by commands are clearer than custom ones.

iwyu/start_iwyu.sh
```diff
@@ -0,0 +1,184 @@
...
+$flags
+EOF
+    fi
+    
+    # Add keep files
+    local keep_files=$(jq -r '.keep_files[]?' "$CONFIG_FILE" 2>/dev/null)
+    if [ -n "$keep_files" ]; then
+        while read -r keep_file; do
+            keep_file=$(resolve_path "$keep_file")
+            args="$args -Xiwyu --keep=$keep_file"
```
p-senichenkov: I haven't read IWYU docs thoroughly yet, but looks like `-Xiwyu` is needed only when you pass options as `CXXFLAGS`.

PodushkaPIR: Unfortunately, this is necessary flag. Otherwise, I couldn't do without it.

iwyu/run_iwyu.py
```diff
@@ -0,0 +1,254 @@
...
+        "-p", str(build_dir),
+        "-j", str(jobs),
+        "-e", str(project_root / "cmake" / "cpm-cache"),
+        "--",
+        "-Xiwyu", "--no_fwd_decls",
+    ]
+    
+    # Add --no_internal_mappings to disable IWYU's mappings
+    # We use our own mapping files instead
+    cmd.extend(["-Xiwyu", "--no_internal_mappings"])
```
polyntsov, line 158: Why do we want to disable IWYU mappings? This requires us to copy standard lib mappings to the project, which will inevitably diverge as IWYU gets updated

polyntsov (commented):
I am not entirely convinced that we need support for the third party libraries, they require explicit mappings which might be quite painful to support. Although of course with them the coverage is considerably higher, hm.

Have you considered `clang-tidy`'s [misc-include-cleaner](https://clang.llvm.org/extra/clang-tidy/checks/misc/include-cleaner.html)? It doesn't support custom mappings, so third-party includes will have to be disabled, but it seems like it's a 5-line configuration in the existing `.clang-tidy`. It should work for free in CI as a part of the existing clang-tidy check and also on diff of a PR only, so we won't have to fix all the errors we have in the current code.

PR #637, opened 2025-12-03
Allow multiple Python interpreter use in bindings

wildsor:
Depends on #744, #768, #769, #773, last 2 commits belong to this one

### DIFF
--- a/src/python_bindings/CMakeLists.txt
+++ b/src/python_bindings/CMakeLists.txt
@@ -6,7 +6,7 @@ project(
     DESCRIPTION "Python bindings to the interface of the Desbordante C++ core library"
 )
 
-CPMAddPackage("gh:pybind/pybind11@2.13.4")
+CPMAddPackage("gh:pybind/pybind11@3.0.4")
 
 # ---- Python Bindings Configuration ----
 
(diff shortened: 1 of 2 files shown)


wildsor:
Reproducibility:
[find_globals.sh](https://github.com/user-attachments/files/28235073/find_globals.sh)
Basically says to run an AST matcher for non-const globals that are not in an included header, excluding those from googletest and pybind11 macros, on all headers and implementation files, using C++20 and all the libraries (except for one that was causing issues, emhash packages with boost for some reason) because compile_commands.json does not handle headers.

Parsing compile_commands.json to get compiler arguments for using libraries. One-off REPL stuff but putting it here so you don't have to think again:
```python
t = eval(open('build/compile_commands.json').read())
system = set()
incl = set()
for info in t:
    words = info['command'].split()
    for word in words:
        if word.startswith('-I'):
            incl.add(word)
    sys_include = False
    for word in words:
        if sys_include:
            system.add(word)
            sys_include = False
        if word == '-isystem':
            sys_include = True
print(' '.join(f'--extra-arg="-isystem {path}"' for path in system))
print(' '.join(f"--extra-arg='{i}'" for i in incl))
```

wildsor:
Needs more consideration

PR #642, opened 2025-12-09
Some additions to bindings according to current bindings requirements

artem-smola:
The requirements [here](https://github.com/Desbordante/desbordante-core/wiki/Development‐Features#что-значит-разработать-байндинги).

- Add __eq__ and __hash__ methods to ind, md, nar, sfd
- Add miner-validator compatibility to ind, ucc


src/python_bindings/dc/bind_fastadc.cpp
```diff
@@ -39,6 +39,26 @@ PredicatePtr DeserializePredicate(py::tuple t, std::shared_ptr<RelationalSchema
...
+
+py::tuple GetSortedPredicatesAsTuple(DC const& dc) {
+    std::vector<PredicatePtr> predicate_ptrs;
+    for (PredicatePtr predicate : dc.GetPredicateSet()) {
+        predicate_ptrs.push_back(predicate);
+    }
+    std::sort(predicate_ptrs.begin(), predicate_ptrs.end(),
+              [](PredicatePtr const& p1, PredicatePtr const& p2) {
+                  return p1->ToString() < p2->ToString();
+              });
```
Oddin60F: Use `std::ranges` for better readability
```suggestion
    std::ranges::sort(predicate_ptrs,
              [](auto const& p1, auto const& p2) {
                  return p1->ToString() < p2->ToString();
              });
```

src/python_bindings/ind/bind_ind.cpp
```diff
@@ -1,50 +1,111 @@
...
+        std::string s_name = schema->GetName();
+        std::vector<std::string> s_col_names;
+        for (std::unique_ptr<Column> const& col_ptr : schema->GetColumns()) {
+            s_col_names.push_back(col_ptr->GetName());
+        }
+        schemas_state.push_back(py::make_tuple(std::move(s_name), std::move(s_col_names)));
+    }
+    double error = ind.GetError();
+    return py::make_tuple(std::move(lhs_state), std::move(rhs_state), std::move(schemas_state),
+                          error);
```
Oddin60F: I think the function names are enough to understand what we're putting in the tuple, so these variables are unnecessary.

```suggestion
    std::vector<py::tuple> schemas_state;
    for (auto const& schema : schemas_vec) {
        std::vector<std::string> col_names;
        for (std::unique_ptr<Column> const& col_ptr : schema->GetColumns()) {
            col_names.push_back(col_ptr->GetName());
        }
        schemas_state.push_back(py::make_tuple(schema->GetName(), std::move(col_names)));
    }
    return py::make_tuple(std::move(lhs_state), std::move(rhs_state), std::move(schemas_state),
                          ind.GetError());
```

src/python_bindings/dc/bind_fastadc.cpp
```diff
@@ -48,6 +68,31 @@ void BindFastADC(py::module_& main_module) {
...
+            .def("__hash__",
+                 [](DC const& dc) {
+                     py::tuple dc_predicates_tuple = GetSortedPredicatesAsTuple(dc);
+                     py::tuple schema_tuple = table_serialization::ConvertSchemaToImmutableTuple(
+                             dc.GetSchema().get());
+
+                     py::tuple state_tuple = py::make_tuple(std::move(dc_predicates_tuple),
+                                                            std::move(schema_tuple));
+
+                     return py::hash(state_tuple);
```
Oddin60F: It's better to check, but it seems to me that `py::hash` takes a copy of the hashed value, so it's better to use a move here.

```suggestion
                     py::tuple state_tuple = py::make_tuple(std::move(dc_predicates_tuple),
                                                            std::move(schema_tuple));

                     return py::hash(std::move(state_tuple));
```

src/python_bindings/ind/bind_ind.cpp
```diff
@@ -1,50 +1,111 @@
...
+                     py::tuple ind1_state_tuple =
+                             ind_serialization::ConvertIndToImmutableTuple(ind1);
+                     py::tuple ind2_state_tuple =
+                             ind_serialization::ConvertIndToImmutableTuple(ind2);
+                     return ind1_state_tuple.equal(ind2_state_tuple);
+                 })
+            .def("__hash__",
+                 [](IND const& ind) {
+                     py::tuple ind_state_tuple = ind_serialization::ConvertIndToImmutableTuple(ind);
+                     return py::hash(ind_state_tuple);
```
Oddin60F: Similar to the previous remark:

```suggestion
                     return py::hash(ind_serialization::ConvertIndToImmutableTuple(ind));
```

src/python_bindings/ucc/bind_ucc_verification.cpp
Oddin60F, line 20: Why do we need this empty namespace?

src/python_bindings/dc/bind_fastadc.cpp
```diff
@@ -39,6 +39,26 @@ PredicatePtr DeserializePredicate(py::tuple t, std::shared_ptr<RelationalSchema
...
+        predicate_ptrs.push_back(predicate);
+    }
+    std::sort(predicate_ptrs.begin(), predicate_ptrs.end(),
+              [](PredicatePtr const& p1, PredicatePtr const& p2) {
+                  return p1->ToString() < p2->ToString();
+              });
+    py::tuple predicates(predicate_ptrs.size());
+    for (size_t i = 0; i < predicate_ptrs.size(); i++) {
+        predicates[i] = fastadc_serialization::SerializePredicate(predicate_ptrs[i]);
+    }
```
Oddin60F: Here and further in all the binding code, there is one similar pattern: there is some vector of elements, for each vector object, we create a py::tuple of some object fields and add it to one common tuple.
There is also a slightly different pattern: we add each vector object to one tuple.

 Usually it looks like this:
```
std::vector<T> vec = ...;
py::tuple tup(vec.size());
for (std::size_t i = 0; i < vec.size(); ++i) {
 tup[i] = vec[i];
 //or
 tup[i] = py::make_tuple(vec[i].name, vec[i].age, foo());

```
This code is duplicated in many places, and an attempt has been made to reduce code repetition in the CFD, FD, and ND bindings _(see the corresponding bindings and the VectorToTuple() function)_.

I think it's a good idea to make a utility function somewhere in the py_util folder, similar to VectorToTuple() and use it throughout the code. For example, the method might look like this: 
```
template <typename ElementType>
py::tuple VectorToTuple(std::vector<ElementType> vec, std::function<py::tuple(ElementType)> build_tuple_for_elem) {
 py::tuple tup(vec.size());
for (std::size_t i = 0; i < vec.size(); ++i) {
 tup[i] = build_tuple_for_elem(vec[i]);
}
return tup;
```
As I see it, this solution will improve readability, make the code easier to understand, and significantly reduce the size of the code.

src/python_bindings/dc/bind_fastadc.cpp
```diff
@@ -48,6 +68,31 @@ void BindFastADC(py::module_& main_module) {
     py::class_<DC>(dc_module, "DC")
             .def("__str__", &DC::ToString)
             .def("__repr__", &DC::ToString)
+            .def("__eq__",
+                 [](DC const& dc1, DC const& dc2) {
+                     if (&dc1 == &dc2) {
+                         return true;
+                     }
```
Oddin60F, line 117: This is a good idea, but it should probably be done for all primitives _(for example, it is not done in MD and SFD)_.

src/python_bindings/ind/bind_ind.cpp
```diff
@@ -1,50 +1,111 @@
 #include "python_bindings/ind/bind_ind.h"
 
+#include <cmath>
```
Oddin60F, line 5: If you're using any standard structures and functions, add them using include. This applies to the entire code _(for example, there's a missing_ `#include <vector>`_)_.

examples/basic/mining_sfd.py
Oddin60F, line 1: I think this is unnecessary in PR.

src/python_bindings/md/bind_md.cpp
```diff
@@ -65,6 +65,66 @@ void BindColumnMatchWithConstructor(Args&&... args) {
                          .none(false),
                  "column_functions"_a = ColumnFunctions{});
 }
+
+py::tuple SerializeColumnMatch(MD const& md_obj) {
+    std::vector<py::tuple> match_tuples;
+    std::shared_ptr<std::vector<model::md::ColumnMatch> const> matches_ptr =
+            md_obj.GetColumnMatches();
+    if (matches_ptr) {
```
Oddin60F: Maybe we should change the condition to the opposite for better clarity, and then use `VectorToTuple()`?

src/python_bindings/py_util/table_serialization.cpp
```diff
@@ -32,6 +32,18 @@ std::shared_ptr<RelationalSchema const> DeserializeRelationalSchema(py::tuple t)
     return schema;
 }
 
+// relational schema conversion suitable for py::hash
+py::tuple ConvertSchemaToImmutableTuple(RelationalSchema const* schema) {
```
Oddin60F, line 38: As I understand it, you added your `ConvertSchemaToImmutableTuple()` method because the base `SerializeRelationalSchema()` method creates a mutable tuple that cannot be hashed. Maybe you should use `SerializeRelationalSchema()` everywhere instead? Or do you need a mutable schema as well?

src/python_bindings/ind/bind_ind.cpp
```diff
@@ -1,50 +1,111 @@
...
-                            }
-                            schemas_state.push_back(
-                                    py::make_tuple(std::move(s_name), std::move(s_col_names)));
-                        }
-                        double error = ind.GetError();
-                        return py::make_tuple(std::move(lhs_state), std::move(rhs_state),
-                                              std::move(schemas_state), error);
-                    },
+                    [](IND const& ind) { return ind_serialization::SerializeInd(ind); },
                     // __setstate__
```
Oddin60F, line 138: In all primitives, the serialization logic is wrapped in a separate method. Maybe it's worth doing the same for the deserialization logic?

src/python_bindings/dc/bind_fastadc.cpp
```diff
@@ -48,6 +68,31 @@ void BindFastADC(py::module_& main_module) {
...
+                     py::tuple schema_tuple = table_serialization::ConvertSchemaToImmutableTuple(
+                             dc.GetSchema().get());
+
+                     py::tuple state_tuple = py::make_tuple(std::move(dc_predicates_tuple),
+                                                            std::move(schema_tuple));
+
+                     return py::hash(state_tuple);
+                 })
             .def(py::pickle(
                     //__getstate__
```
Oddin60F: Maybe we should move serialization to a separate method for easier reading, just like we do here?

Oddin60F, line 130: Maybe we should move the serialization to a separate method, so it's easier to read how we do it further in the code?

src/python_bindings/nar/bind_nar.cpp
```diff
@@ -79,6 +86,70 @@ std::unordered_map<size_t, std::shared_ptr<model::ValueRange>> DeserializeRangeM
...
+    return ranges_tuple;
+}
+
+py::tuple ConvertNarToImmutableTuple(model::NAR const& nar) {
+    py::tuple ante_tuple = ConvertRangeMapToImmutableTuple(nar.GetAnte());
+    py::tuple cons_tuple = ConvertRangeMapToImmutableTuple(nar.GetCons());
+
+    bool is_consistent = nar.IsQualitiesConsistent();
+
+    if (is_consistent) {
```
Oddin60F: For readability, it is better to invert the condition.

src/python_bindings/bind_main_classes.cpp
```diff
@@ -14,7 +14,7 @@
 #include "python_bindings/py_util/opt_to_py.h"
 #include "python_bindings/py_util/py_to_any.h"
 
-namespace {
+namespace configure_algorithm_bind_main_classes {
```
Oddin60F: Maybe you should shorten the namespace name to `configure_algorithm`?

src/python_bindings/nar/bind_nar.cpp
```diff
@@ -79,6 +86,70 @@ std::unordered_map<size_t, std::shared_ptr<model::ValueRange>> DeserializeRangeM
...
+}
+
+py::tuple ConvertRangeMapToImmutableTuple(
+        std::unordered_map<size_t, std::shared_ptr<model::ValueRange>> const& map) {
+    std::vector<std::pair<size_t, std::shared_ptr<model::ValueRange>>> sorted_ranges;
+    sorted_ranges.reserve(map.size());
+    for (auto const& [key, value] : map) {
+        sorted_ranges.emplace_back(key, value);
+    }
+    std::sort(sorted_ranges.begin(), sorted_ranges.end(),
```
Oddin60F: Use std::ranges::sort

src/python_bindings/nar/bind_nar.cpp
```diff
@@ -79,6 +86,70 @@ std::unordered_map<size_t, std::shared_ptr<model::ValueRange>> DeserializeRangeM
     }
     return map;
 }
+
+py::tuple ConvertValueRangeToImmutableTuple(std::shared_ptr<model::ValueRange> const& vr) {
```
Oddin60F, line 135: This question is similar to the immutable schema. Why do we use the immutable and mutable versions of the `ValueRange` tuple for serialization and hashing?

src/python_bindings/nar/bind_nar.cpp
```diff
@@ -126,6 +197,31 @@ void BindNar(py::module_& main_module) {
...
+                 })
+            .def("__hash__",
+                 [](NAR const& nar) {
+                     try {
+                         py::tuple state_tuple = nar_serialization::ConvertNarToImmutableTuple(nar);
+                         return py::hash(state_tuple);
+                     } catch (std::exception const& e) {
+                         return static_cast<long>(0);
+                     }
+                 })
```
Oddin60F, line 254: I think that silently shadowing the error is not a good idea. The exception that is caught here indicates that the received `NAR` object is in an inconsistent state, as one of its fields has an unsupported type, so it is better to report this.
It is possible that we will never encounter a situation where we pass a `NAR` with an unsupported type for serialization/deserialization (this is just a guess, and it is better to consider it more carefully).

Oddin60F (changes requested):

Oddin60F (approved):

PR #643, opened 2025-12-10
Pep8 action (ruff)

PodushkaPIR:
- Add base ruff action in CI to catch pep8 suggestions (can be extended)
- Add ruff-tool in the pyproject.toml (can be extend)

This PR will change depending on the offers. I'm not sure which Errors and Warnings are better to introduce and which are better to ignore. I am ready for suggestions.

Also ruff is a great tool and and some features from it can be transferred to the project.


wildsor (changes requested):
the ruff-check rename should be squashed with the initial commit

pyproject.toml
```diff
@@ -49,3 +49,19 @@ exclude = [
...
     "examples/*",
 ]
+
+[tool.ruff]
+select = [
+    "E", "W", "F",
+]
+
+ignore = [
+    "W291",
```
wildsor: Are there places where we want trailing spaces? Exclude them individually, this should not be disabled for all Python files.

pyproject.toml
```diff
@@ -49,3 +49,19 @@ exclude = [
...
+    "E501",
```
wildsor: Are there overlong lines that can't be fixed? This should definitely not be disabled for the whole project.

pyproject.toml
```diff
@@ -49,3 +49,19 @@ exclude = [
...
+]
+
+exclude = [
+    ".git",
+    "build/",
+    "examples/notebooks/*.ipynb",
```
wildsor: I think we should lint and format notebooks.

.github/workflows/check-codestyle.yml
```diff
@@ -53,6 +53,7 @@ jobs:
           echo "### $formatterOutput :heavy_check_mark:" >> $GITHUB_STEP_SUMMARY
 
   clang-tidy-check:
+    continue-on-error: true
```
wildsor: Why was this added?

pyproject.toml
```diff
@@ -49,3 +49,19 @@ exclude = [
     "src/tests/*",
     "examples/*",
 ]
+
+[tool.ruff]
+select = [
+    "E", "W", "F",
```
wildsor: Why not enable ALL and exclude stuff that doesn't make sense?

wildsor (changes requested):
I know that's not how they are written now, but I'd like CI config files to be written so that anyone looking at them for the first time can at least somewhat understand what's going on with minimal disruptions.
At least, if you're disabling or enabling something, there should be a rationale: if I just look at `[tool.ruff] ignore = [ "W291", ...]`, there is no way I can guess what it is doing without going into ruff's docs.

pyproject.toml
PodushkaPIR: Enabling ALL creates many of false positives for code. Then you need a massive ignore list with 50+ rules. It's easier for a developer to understand what's being checked by looking at an explicit `select` list than trying to  create a ignore list.

.github/workflows/check-codestyle.yml
PodushkaPIR: Sorry. Realized that this is not necessary. Fixed it

pyproject.toml
PodushkaPIR: Fixed it

pyproject.toml
PodushkaPIR: Fixed it

pyproject.toml
```diff
@@ -131,3 +131,70 @@ extend-ignore-re = [
...
+    "I",   # sort imports properly
+    "UP",  # modernize Python syntax (safe for 3.8+)
+    "ISC", # good use of string concatenation
+    "ICN", # use common import conventions
+]
+# Ignore rules that are too strict for examples/notebooks:
+# - Q000: requires double quotes (examples use single quotes for consistency)
+# - Q001: single quote multiline strings (used for output formatting)
+# - Q002: move single-quoted strings to double (related to Q001)
+# - Q003: change outer quotes to avoid escaping (examples use escaped quotes)
```
wildsor, line 161: These ones prevent inconsistencies. Let's enable them.

pyproject.toml
```diff
@@ -131,3 +131,70 @@ extend-ignore-re = [
...
+# - T201: warns about print() (examples use print for output demonstration)
+# - D100: missing docstring in public module (examples are scripts, not libraries)
+# - D101: missing docstring in public class (simple example classes)
+# - D102: missing docstring in public method (example methods prioritize clarity)
+# - D103: missing docstring in public function (simple helper functions)
+# - D104: missing docstring in public nested (helper functions in examples)
+# - D107: missing docstring in __init__ (simple init methods)
+# - D203: one blank line required before class docstring
+#         (conflicts with D211: no blank line required before class)
+# - D210: one-line docstring should fit on first line (formatting preference)
```
wildsor, line 171: The summary is incorrect

```suggestion
# - D210: No whitespaces allowed surrounding docstring text
```

I think we should enable this

pyproject.toml
```diff
@@ -131,3 +131,70 @@ extend-ignore-re = [
...
+# - D212: summary must start at first line of multiline docstring
+#         (conflicts with D213: summary must start at second line)
+# - D300: use triple double quotes (benchmark uses single quotes)
```
wildsor, line 174: Let's fix it, PEP257 recommends double quotes for docstrings (consistency)

pyproject.toml
```diff
@@ -131,3 +131,70 @@ extend-ignore-re = [
...
+# - D400: one-line docstring does not match summary (formatting preference)
```
wildsor, line 175:
```suggestion
# - D400: First line should end with a period (formatting preference)
```
Let's fix this one

pyproject.toml
```diff
@@ -131,3 +131,70 @@ extend-ignore-re = [
...
+# - D413: bad multiline summary (formatting preference)
+# - D415: section underline (formatting preference)
+# - S101: use of assert detected (examples use assert for test demonstrations)
+# - S603: use of subprocess call (not a security issue in examples)
```
wildsor, line 179: https://docs.astral.sh/ruff/rules/ has short descriptions for the names, you can use them

pyproject.toml
```diff
@@ -131,3 +131,70 @@ extend-ignore-re = [
...
+# - D104: missing docstring in public nested (helper functions in examples)
+# - D107: missing docstring in __init__ (simple init methods)
+# - D203: one blank line required before class docstring
+#         (conflicts with D211: no blank line required before class)
+# - D210: one-line docstring should fit on first line (formatting preference)
+# - D212: summary must start at first line of multiline docstring
+#         (conflicts with D213: summary must start at second line)
+# - D300: use triple double quotes (benchmark uses single quotes)
+# - D400: one-line docstring does not match summary (formatting preference)
+# - D413: bad multiline summary (formatting preference)
```
wildsor, line 176:
```suggestion
# - D413: Missing blank line after last section (formatting preference)
```

pyproject.toml
```diff
@@ -131,3 +131,70 @@ extend-ignore-re = [
...
+# - D415: section underline (formatting preference)
```
wildsor, line 177:
```suggestion
# - D415: First line should end with a period, question mark, or exclamation point
```
Let's fix this

pyproject.toml
```diff
@@ -131,3 +131,70 @@ extend-ignore-re = [
     "ser\\.iloc\\[pos\\]", # "ser" is short for pandas Series
     "flights_dd_dif_tabl", # "dif" / "tabl" are short for difference / table
 ]
+
+[tool.ruff]
+# line-length: 120 instead of default 88
+# Reason: Examples often have long f-strings with column names, URLs, table data
```
wildsor, line 137: https://docs.astral.sh/ruff/rules/line-too-long/:
> In the interest of pragmatism, this rule makes a few exceptions when determining whether a line is overlong. Namely, it:
> * Ignores lines that consist of a single "word" (i.e., without any whitespace between its characters).
> * Ignores lines that end with a URL, as long as the URL starts before the line-length threshold.
> * Ignores line that end with a pragma comment (e.g., # type: ignore or # noqa), as long as the pragma comment starts before the line-length threshold. That is, a line will not be flagged as overlong if a pragma comment causes it to exceed the line length. (This behavior aligns with that of the Ruff formatter.)
> * Ignores SPDX license identifiers and copyright notices (e.g., # SPDX-License-Identifier: MIT), which are machine-readable and should not wrap over multiple lines.

Where are the false positives? I wonder if it would be better to explicitly ignore exceptions.

pyproject.toml
```diff
@@ -131,3 +131,70 @@ extend-ignore-re = [
...
+]
+
+exclude = [
+    ".git",
+    "build/",
+]
+
+[tool.ruff.lint.per-file-ignores]
+# E702: Multiple statements on one line (e.g., G.add_node(); G.add_edge())
+"examples/**/*.py" = ["E501", "E702"]
```
wildsor, line 195: Let's fix E702, I think it's better to leave awkward code in, it lets us see awkward interfaces

wildsor (changes requested):

pyproject.toml
```diff
@@ -131,3 +131,70 @@ extend-ignore-re = [
...
+# Notebooks have generated cell code with unique scoping
+# E501: Long lines in cell outputs
+# E731: lambda assigned to variable (cell-specific pattern)
+# F401, F811, F821: Unused imports/redefinitions (cell scoping)
+"examples/notebooks/*.ipynb" = ["E501", "E731", "F401", "F811", "F821"]
```
wildsor, line 200: How does it break? Does it actually check every cell individually? That's strange, I wonder if there is a setting for it

pyproject.toml
PodushkaPIR: thanks

PR #649, opened 2025-12-15
Improvment of class DDVerifier

unfadingDawn:
Implemented support of user-defined metrics for validation. 
Implemented Python bindings for these metrics and added missing capabilities for accessing DD and DF instances from python.
Added code to the DD validation example using user-defined metrics.


src/core/algorithms/dd/dd.h
```diff
@@ -122,11 +122,26 @@ struct DFStringConstraint {
 
     auto operator<=>(DFStringConstraint const& other) const = default;
 
+    auto operator==(DFStringConstraint const& other) const {
+        return column_name == other.column_name && constraint == other.constraint;
+    }
+
```
ol-imorozko: that is redundant, we already have `operator<=>` that implicitly declares `operator==` with the same semantics

src/core/algorithms/dd/dd.h
```diff
@@ -122,11 +122,26 @@ struct DFStringConstraint {
...
+    std::string to_JSON() const {
+        std::stringstream s;
+        s << "{ column_name: " << column_name << ", lower_bound: " << constraint.lower_bound
+          << ", upper_bound: " << constraint.upper_bound << "}";
+        return s.str();
+    }
+
+    int hash() const {
+        return static_cast<int>(constraint.lower_bound + constraint.upper_bound);
+    }
```
ol-imorozko: Is it expected that we ignore `std::string column_name;` field?
Also, the current usage of `__hash__` in repo is this:
```bash
imorozko@i112809612:~/work/Desbordante(main)$ rg "__hash__" -A 1
src/python_bindings/ucc/bind_ucc.cpp
42:            .def("__hash__", [](UCC const& ucc) { return py::hash(MakeUCCNameTuple(ucc)); })
43-            .def(py::pickle(

src/python_bindings/od/bind_od.cpp
32:            .def("__hash__",
33-                 [](AscCanonicalOD const& od) { return py::hash(py::str(od.ToString())); })
--
60:            .def("__hash__",
61-                 [](DescCanonicalOD const& od) { return py::hash(py::str(od.ToString())); })
--
88:            .def("__hash__",
89-                 [](SimpleCanonicalOD const& od) { return py::hash(py::str(od.ToString())); })

src/python_bindings/nd/bind_nd.cpp
42:            .def("__hash__", [](ND const& nd) { return py::hash(MakeNdNameTuple(nd)); });
43-}

src/python_bindings/fd/bind_fd.cpp
51:            .def("__hash__", [](FD const& fd) { return py::hash(MakeFdNameTuple(fd)); })
52-            .def_property_readonly("lhs_indices", &FD::GetLhsIndices)
```

So I guess the correct ways is to define some sort of tuple or use `ToString`

src/core/algorithms/dd/dd.h
```diff
@@ -137,6 +152,37 @@ struct DDString {
...
+                    }
+                }
+                if (!is_match) {
+                    equal = false;
+                    break;
+                }
+            }
+        }
+        return equal;
+    }
```
ol-imorozko, line 150: Is it supposed to be non-symmetric? Your check is "every element of `left` exists in `other.left`", so you check only injection, but not surjection. For equality you want bijection, which is both.

Right now `A == B` can be true while `B == A` is false if B has extra constraints.

I guess what you actually need is
```cpp
bool operator==(DDString const& other) const {
    return std::ranges::is_permutation(left, other.left) &&
           std::ranges::is_permutation(right, other.right);
}
```

src/core/algorithms/dd/dd.h
```diff
@@ -149,6 +195,31 @@ struct DDString {
...
+                s << ", ";
+            }
+        }
+        s << " ] }";
+        return s.str();
+    }
+
+    int hash() const {
+        return left.size() + right.size();
+    }
```
ol-imorozko: same thing here, maybe it's better to use `DFToString` like other alogirhtms do

src/core/algorithms/dd/dd.h
```diff
@@ -122,11 +122,26 @@ struct DFStringConstraint {
...
         s << column_name << " " << constraint.ToString();
         return s.str();
     }
+
+    std::string to_JSON() const {
+        std::stringstream s;
+        s << "{ column_name: " << column_name << ", lower_bound: " << constraint.lower_bound
+          << ", upper_bound: " << constraint.upper_bound << "}";
+        return s.str();
+    }
```
ol-imorozko: Is it valid JSON? If I remember correctly JSON keys must always be enclosed in double quotes.

src/core/algorithms/dd/dd.h
```diff
@@ -149,6 +195,31 @@ struct DDString {
...
+        s << " ], right: [ ";
+        for (auto it = right.begin(); it != right.end(); ++it) {
+            s << it->to_JSON();
+            if (it != --right.end()) {
+                s << ", ";
+            }
+        }
+        s << " ] }";
+        return s.str();
+    }
```
ol-imorozko: Is it valid JSON? If I remember correctly JSON keys must always be enclosed in double quotes.

Also `--left.end()` recomputes `--end()` every iteration and a bit unreadable. I suggest smth like this:
```cpp
    std::stringstream s;

    auto DumpList = [&s](std::string_view name,
                        std::list<DFStringConstraint> const& xs) {
        s << "\"" << name << "\":[";
        bool first = true;
        for (DFStringConstraintconst& x : xs) {
            if (!first) s << ",";
            first = false;
            s << x.ToJson();   // must also be valid JSON
        }
        s << "]";
    };

    s << "{";
    DumpList("left", left);
    s << ",";
    DumpList("right", right);
    s << "}";

    return s.str();
```

src/core/algorithms/dd/dd_verifier/FuncMetric.h
```diff
@@ -0,0 +1,19 @@
...
+
+#include "Metric.h"
+
+template <typename T>
+class FuncMetric : public Metric {
+private:
+    std::function<double(T, T)> metric_function_;
+
+public:
+    explicit FuncMetric(std::function<double(T, T)> func) : metric_function_(func) {}
```
ol-imorozko: we will do two copies of `std::string` per call this way. Maybe `T const&`?

src/core/algorithms/dd/dd_verifier/FuncMetric.h
```diff
@@ -0,0 +1,19 @@
...
+
+    double Dist(std::byte const* first, std::byte const* second) const override {
+        return metric_function_(*reinterpret_cast<T const*>(first),
+                                *reinterpret_cast<T const*>(second));
```
ol-imorozko: We can use
```cpp
    template <typename T>
    [[nodiscard]] static T const& GetValue(std::byte const* buf) {
        return *reinterpret_cast<T const*>(buf);
    }
```
from `src/core/model/types/type.h`

src/core/algorithms/dd/dd_verifier/dd_verifier.cpp
```diff
@@ -64,9 +67,16 @@ bool DDVerifier::IsColumnMetrizable(model::ColumnIndex const column_index) const
...
                                      std::pair<std::size_t, std::size_t> const &tuple_pair) const {
     model::TypedColumnData const &column = typed_relation_->GetColumnData(column_index);
+    auto const column_name = column.GetColumn()->GetName();
     std::byte const *first_value = column.GetValue(tuple_pair.first);
     std::byte const *second_value = column.GetValue(tuple_pair.second);
     auto const &type = static_cast<model::IMetrizableType const &>(column.GetType());
+    for (auto const &it : metrics_) {
+        if (column_name == it.first) {
+            return it.second->Dist(first_value, second_value);
+            ;
```
ol-imorozko: extra semicolon

src/tests/unit/test_dd_verifier.cpp
```diff
@@ -52,7 +59,8 @@ INSTANTIATE_TEST_SUITE_P(
                           DDVerifyingParams({{{"Col0", 2, 3}}, {{"Col1", 9, 10}}}, 4, 4. / 5.),
                           DDVerifyingParams({{{"Col0", 2, 3}}, {{"Col1", 1, 2}}}, 5, 1.),
                           DDVerifyingParams({{{"Col1", 2, 16}}, {{"Col3", 112, 333}}}, 5,
-                                            5. / 10.)));
+                                            5. / 10.),
+			  DDVerifyingParams({{{"Col0", 0, 0}}, {{"Col2", 0, 1},{"Col1", 0, 20}}}, 0, 0., {{"Col0", (new FuncMetric<int>([](int first, int second){return first - first + second - second;}))},{"Col2", new FuncMetric<int>([](int first, int second){return first - first + second - second + 1;})}})));
```
ol-imorozko: Is that a leak, like, who deletes that `new FuncMetric`?

src/python_bindings/dd/bind_dd_verification.cpp
```diff
@@ -12,17 +22,71 @@
...
+                            return new FuncMetric<std::string>(
+                                    std::move(metric_tuple[0]
+                                                      .cast<std::function<double(std::string,
+                                                                                 std::string)>>()));
+                        } else {
+                            throw std::runtime_error(
+                                    "Unsupported type tag. Use 'int', 'double', or 'string'.");
+                        }
+                    },
+                    py::return_value_policy::take_ownership)
```
ol-imorozko: You pass ownership to python here with `take_ownership`, so Python will delete it when that Python object dies.

`DDVerifier` stores Metrics as raw pointers `std::unordered_map<std::string, Metric const*> metrics_;` and uses them in `CalculateDistance`. Can we be sure that those pointers remain valid as long as DDVerifier may call CalculateDistance?

If not, I think we need to use `std::shared_ptr` instead of raw pointers. If I remember correctly, `pybind11` knows how to hold `shared_ptr` objects and keep reference counts correct.

ol-imorozko (changes requested):

src/core/algorithms/dd/dd_verifier/FuncMetric.h
```diff
@@ -0,0 +1,20 @@
...
+#include "core/model/types/type.h"
+
+template <typename T>
+class FuncMetric : public Metric {
+private:
+    std::function<double(T const &, T const &)> metric_function_;
+
+public:
+    explicit FuncMetric(std::function<double(T const &, T const &)> func)
+        : metric_function_(func) {}
```
ol-imorozko: `std::move(func)`

src/core/algorithms/dd/dd.h
ol-imorozko: > unfadingDawn marked this comment as resolved

You forgot to remove this code :)

ol-imorozko (approved):
don't forget to fix merge conflicts and those two minor comments, LTGM

PR #662, opened 2026-01-02
Dev/new stats

VinogradovDmitri:
Summary
---
This PR adds 6 new statistics to the DataStats class. These statistics help users understand their data better.

New Statistics
---
1. Interquartile Range — shows the spread of the middle 50% of numeric data
2. Coefficient of Variation — shows how much data varies compared to the average
3. Monotonicity — tells if data always goes up, always goes down, always the same, or neither
4. Jarque-Bera Statistic — tests if numeric data has a normal distribution
5. Entropy — measures how mixed categorical data is
6. Gini Coefficient — measures inequality in categorical data

Changes
---

1. ### Core
    **Location**: `src/core/algorithms/statistics/`
    **Change**: extended DataStats class with six new statistical methods. Enhanced ColumnStats data structure to store new statistics.
2. ### Python Bindings
    **Location**: `src/python_bindings/statistics/`
    **Change**: added pybind methods for all new statistics.
3. ### Tests
    **Location**: `src/tests/unit/`
    **Change**: implemented google-testing for new statistical methods.

    **Location**: `src/python_bindings/`
    **Change**: implemented  Python integration tests for new statistical methods.
4. ### Examples
    **Location**: `examples/basic/`
    **Change**: Updated demonstration case to show usage of all new statistics.


src/core/algorithms/statistics/data_stats.cpp
```diff
@@ -986,4 +1004,193 @@ void DataStats::LoadDataInternal() {
...
+    std::byte* q1_val = mo::DoubleType::MakeFrom(q1.GetData(), *q1.GetType());
+    std::byte* q3_val = mo::DoubleType::MakeFrom(q3.GetData(), *q3.GetType());
+    std::byte* iqr_val = double_type.Allocate();
+
+    double_type.Sub(q3_val, q1_val, iqr_val);
+
+    double_type.Free(q1_val);
+    double_type.Free(q3_val);
+
+    return Statistic(iqr_val, &double_type, false);
```
ol-imorozko, line 1169: `GetInterquartileRange` allocates `iqr_val` and constructs `Statistic` object with it. Will `Statistic` destructor free it?

src/core/algorithms/statistics/data_stats.cpp
```diff
@@ -986,4 +1004,193 @@ void DataStats::LoadDataInternal() {
...
+            }
+
+            if (!increasing && !decreasing) break;
+        } else {
+            has_prev = true;
+        }
+        prev = current;
+    }
+
+    if (!has_prev) return {};
```
ol-imorozko: Since `has_data` is set to `true` on the first non-null element and never changes after that, I think we can just just check if `prev` is `nullptr`:
```suggestion
    bool increasing = true;
    bool decreasing = true;
    std::byte const* prev = nullptr;

    for (size_t i = 0; i < col.GetNumRows(); ++i) {
        if (col.IsNullOrEmpty(i)) continue;

        std::byte const* current = col.GetData()[i];

        if (prev != nullptr) {
            mo::CompareResult cmp = col.GetType().Compare(prev, current);

            if (cmp == mo::CompareResult::kLess) {
                decreasing = false;
            } else if (cmp == mo::CompareResult::kGreater) {
                increasing = false;
            }

            if (!increasing && !decreasing) break;
        }
        prev = current;
    }

    if (prev == nullptr) return {};
```

src/core/algorithms/statistics/data_stats.cpp
```diff
@@ -986,4 +1004,193 @@ void DataStats::LoadDataInternal() {
...
+    std::string result;
+    if (increasing && !decreasing) {
+        result = "ascending";
+    } else if (!increasing && decreasing) {
+        result = "descending";
+    } else if (increasing && decreasing) {
+        result = "equal";
+    } else {
+        result = "none";
+    }
```
ol-imorozko: IMO this is cleaner, but feel free to ignore:

```suggestion
    std::string_view result = (increasing && decreasing) ? "equal" :
                              increasing ? "ascending" :
                              decreasing ? "descending" : "none";
```

ol-imorozko (changes requested):

src/core/algorithms/statistics/data_stats.h
```diff
@@ -167,6 +167,13 @@ class DataStats : public Algorithm {
     // Returns the amount of entirely lowercase words in a string column.
     Statistic GetNumberOfEntirelyLowercaseWords(size_t index) const;
 
+    Statistic GetInterquartileRange(size_t index) const;
+    Statistic GetCoefficientOfVariation(size_t index) const;
+    Statistic GetMonotonicity(size_t index) const;
+    Statistic GetJarqueBeraStatistic(size_t index) const;
+    Statistic GetEntropy(size_t index) const;
+    Statistic GetGiniCoefficient(size_t index) const;
```
Firsov62121, line 186: Add a description as for statics above

src/core/algorithms/statistics/data_stats.cpp
VinogradovDmitri: Yes, because we forbade to clone data in constructor

src/core/algorithms/statistics/data_stats.cpp
VinogradovDmitri: `MakeValue` take string value, so why `std::string_view`? I'll leave `std::srting`

ol-imorozko: Yeah, you're right

src/core/algorithms/statistics/data_stats.h
VinogradovDmitri: Done

src/core/algorithms/statistics/data_stats.cpp
VinogradovDmitri: Done

ol-imorozko (approved):

Firsov62121 (approved):

PR #664, opened 2026-01-04
Add Cinderella & PLICind, CIND Verifier, Py Bindings and Example

ALanovaya:
This PR adds support for **Conditional Inclusion Dependencies (CINDs)** to Desbordante — both mining and verification. CINDs extend standard INDs by restricting the inclusion check to tuples that satisfy a condition on non-inclusion attributes, enabling more precise data quality rules.

### Formal Definition

A **CIND** is defined over a pair of relations R₁ and R₂ as:

$$\phi = (R_1(X; X_p) \subseteq R_2(Y; Y_p), T_p)$$

where:
- $R_1[X] \subseteq R_2[Y]$ is a standard IND (the **embedded IND**)
- $X_p, Y_p$ are **pattern attributes** from R₁ and R₂ respectively
- $T_p$ is a **pattern tableau** — a set of tuples with constants from the domains of $X_p$ and $Y_p$

An instance $(D_1, D_2)$ satisfies the CIND $\phi$ iff for each tuple $t_p$ in $T_p$ and for each $t_1$ in $D_1$ such that $t_1[X_p] = t_p[X_p]$, there must exist $t_2$ in $D_2$ such that $t_1[X] = t_2[Y]$ and $t_2[Y_p] = t_p[Y_p]$.

### Approximate CINDs — Validity and Completeness

In practice, CINDs rarely hold exactly. We use two quality metrics:

**Validity** measures what fraction of condition-matching LHS groups are included in the RHS:

$$validity = \frac{|included\ support|}{|supporting\ baskets|}$$

where *supporting_baskets* are the distinct LHS groups matching the condition, and *included_support* are those whose inclusion key also appears in the RHS.

**Completeness** measures what fraction of all RHS-included groups also match the condition:

$$completeness = \frac{|included\ support|}{|included\ baskets\ total|}$$

A CIND **holds** when both `validity >= min_validity` and `completeness >= min_completeness`.

### Condition Modes

- **`group`** mode: each distinct value combination of inclusion attributes is one basket. Validity counts distinct groups.
- **`row`** mode: each row is treated individually. Validity counts individual matching rows.

### References

- L. Bravo, W. Fan, S. Ma. **"Extending Dependencies with Conditions"**. VLDB 2007, pp. 243–254. — introduces CINDs (formal definition).
- J. Bauckmann, Z. Abedjan, U. Leser, H. Müller, F. Naumann. **"Discovering Conditional Inclusion Dependencies"**. CIKM 2012, pp. 2094–2098. — proposes the CINDERELLA and PLI-CIND mining algorithms used in this PR.
- O. Curé. "Improving the Data Quality of Drug Databases using Conditional Dependencies and Ontologies". ACM JDIQ 4(1):20, 2012. — earlier work on CIND violation detection (SPARQL over ontologies); referenced as motivation for the verifier.

### 1. CIND Data Model
- **`CIND`** class (`core/algorithms/cind/cind.h`) — represents a discovered CIND: an IND with a list of `Condition` objects, each carrying validity and completeness metrics.
- **`Condition`** class (`core/algorithms/cind/condition.h`) — stores condition attribute values, validity, and completeness for a single pattern.
- **`CondType` / `AlgoType`** enums (`core/algorithms/cind/types.h`) — condition evaluation mode (`row`/`group`) and mining algorithm selection (`cinderella`/`pli_cind`).

### 2. CIND Mining (`CindAlgorithm`)
- **`CindAlgorithm`** (`core/algorithms/cind/cind_algorithm.{h,cpp}`) — the top-level mining algorithm. It first runs **Spider** to discover approximate INDs, then passes them to a condition miner.
- **`CindMiner`** (`condition_miners/cind_miner.{h,cpp}`) — abstract base class for condition mining strategies. Processes each IND independently via `ExecuteSingle()`.
- **`Cinderella`** (`condition_miners/cinderella.{h,cpp}`) — itemset-based condition miner. Builds baskets from inclusion/conditional attributes and mines frequent conditions using an itemset lattice.
- **`PliCind`** (`condition_miners/pli_cind.{h,cpp}`) — PLI-based condition miner. Uses position list intersections to classify rows and discover conditions. Generally faster than Cinderella.

Supporting data structures: `Basket`, `Item`, `Itemset`, `ItemsetNode`, `PositionListsSet`.

### 3. CIND Verification (`CINDVerifier`)
- **`CINDVerifier`** (`cind_verifier/cind_verifier.{h,cpp}`) — verifies whether a given CIND holds on input data. Supports:
  - Two-table and single-table verification
  - Wildcard and specific condition values
  - `group` and `row` condition modes
  - Returns: `real_validity`, `real_completeness`, `holds()`, violating clusters with basket/violating row indices

### 4. Encoded Table Infrastructure
New model classes to support encoded multi-table operations needed by the verifier:
- **`EncodedTables`** (`model/table/encoded_tables.h`) — encodes multiple input tables with a shared `ValueDictionary`.
- **`EncodedColumnData`** (`model/table/encoded_column_data.h`) — integer-encoded column with decode support.
- **`ColumnEncodedRelationData`** (`model/table/column_encoded_relation_data.{h,cpp}`) — builds encoded relation from a dataset stream.

### 5. Config Options
New options in `core/config/conditions/`:
- `validity` — minimum validity threshold
- `completeness` — minimum completeness threshold
- `condition_type` — `row` or `group` mode
- `algo_type` — `cinderella` or `pli_cind`

### 6. Python Bindings
- `desbordante.cind` — exposes `CIND`, `Condition` classes and the `Cind` mining algorithm
- `desbordante.cind_verification` — exposes `CINDVerifier` with all result getters (`holds`, `get_real_validity`, `get_real_completeness`, `get_violating_clusters`, etc.)

### 7. Build System Integration
- CMakeLists for `cind`, `cind.miners`, `cind.verifier` libraries
- Registration in `algorithm_types.h`, `algorithms.h`, `bindings.cpp`
- Config conditions sources added via glob to `config/CMakeLists.txt`
- `column_encoded_relation_data.cpp` added to `model.table`

### 8. Example
- `examples/basic/verifying_cind.py` — demonstrates 4 scenarios:
  1. Empty condition (CIND behaves like IND)
  2. Condition that makes a failing IND hold
  3. Partial validity with a fix by adding data
  4. Error handling for wrong condition values count

### 9. Datasets
- `examples/datasets/ind_datasets/cind_test_en.csv` and `cind_test_de.csv` — test datasets placed alongside IND datasets, based on the multilingual person entity example from CIND literature.


src/core/algorithms/cind/cind_verifier/cind_verifier.cpp
```diff
@@ -0,0 +1,468 @@
...
+void CINDVerifier::LoadDataInternal() {}
+
+unsigned long long CINDVerifier::ExecuteInternal() {
+    return util::TimedInvoke(&CINDVerifier::VerifyCIND, this);
+}
+
+void CINDVerifier::VerifyCIND() {
+    if (input_tables_.empty()) {
+        throw std::runtime_error("No input tables configured");
+    }
```
Oddin60F: This is an unnecessary check. When loading the InputTables option, it is already checked that the parameter is not empty.

src/core/algorithms/cind/cind_verifier/cind_verifier.cpp
```diff
@@ -0,0 +1,468 @@
...
+    real_completeness_ = 0.0;
+}
+
+void CINDVerifier::LoadDataInternal() {}
+
+unsigned long long CINDVerifier::ExecuteInternal() {
+    return util::TimedInvoke(&CINDVerifier::VerifyCIND, this);
+}
+
+void CINDVerifier::VerifyCIND() {
```
Oddin60F, line 158: This function is very large, and it should be divided into phases.

src/core/algorithms/cind/cind_verifier/cind_verifier.cpp
```diff
@@ -0,0 +1,468 @@
...
+    if (input_tables_.empty()) {
+        throw std::runtime_error("No input tables configured");
+    }
+    if (ind_.lhs.empty() || ind_.rhs.empty()) {
+        throw std::runtime_error("CIND verification requires non-empty LHS and RHS indices");
+    }
+
```
Oddin60F: Similarly, when loading parameters for execution, it checks whether they are empty inside the options.

examples/datasets/ind_datasets/cind_test_de.csv
```diff
@@ -0,0 +1,11 @@
...
+Cecil Kellaway,18,Kap,LA,Schauspieler
+Cecil Kellaway,18,Kap,Cal,Schauspieler
+Cecil Kellaway,18,Kap,USA,Schauspieler
+Cecil Kellaway,18,Sud,LA,Schauspieler
+Cecil Kellaway,18,Sud,Cal,Schauspieler
+Cecil Kellaway,18,Sud,USA,Schauspieler
+Sam Sheppard,19,-,-,Mediziner
+Mel Sheppard,18,Almonesson Lake,Queens,Leichtathlet
+Isobel Elsom,18,Cambridge,LA,Schauspielerin
+Isobel Elsom,18,Cambridge,Cal,Schauspielerin
```
Oddin60F: All files should end with newline.

src/core/algorithms/cind/cind_verifier/cind_verifier.cpp
```diff
@@ -0,0 +1,468 @@
...
+    std::vector<std::string> cond_vals = condition_values_;
+    if (cond_vals.empty()) {
+        cond_vals.assign(conditional_indices.size(), kAnyValue);
+    } else {
+        if (cond_vals.size() != conditional_indices.size()) {
+            throw std::runtime_error(
+                    "cind_condition_values size must equal number of conditional attributes");
+        }
+    }
+    for (auto& s : cond_vals) s = StripQuotes(s);
```
Oddin60F: Our styleguide forbids two statements on one line, even in such simple cases.

src/core/algorithms/cind/cind_verifier/cind_verifier.cpp
```diff
@@ -0,0 +1,468 @@
...
+
+    supporting_baskets_ = 0;
+    included_support_ = 0;
+    included_baskets_total_ = 0;
+
+    real_validity_ = -1.0;
+    real_completeness_ = 0.0;
+}
+
+void CINDVerifier::LoadDataInternal() {}
```
Oddin60F: When working with the verifier, the user is supposed to first load data into the algorithm using the `algo.load_data(tables=[...])` function, and then check one or more instances of the primitive for this data. Now, to check several different CIND for the same tables, you need to create different algorithm objects. At the same time, for each validation, you will have to re-parse the input tables, because all the logic of processing tables is performed when the algorithm is running. It will be much better if we can work with the algorithm object, as it assumes an interface.

As I understood from the following code, we parse tables in different ways depending on the input pattern. You can transfer the table parsing logic to `LoadDataInternal()` using the `model::EncodedTables` (which is used further anyway) and  if you definitely need to divide the logic into two cases, then for the `all_wildcard` case, for example, you can transpose the internal representation of the table columns to get a row representation and use them for validation.

src/core/algorithms/cind/cind_verifier/cind_verifier.cpp
```diff
@@ -0,0 +1,468 @@
...
+std::vector<std::string> MakeKeyDecoded(std::size_t row,
+                                        std::vector<model::EncodedColumnData const*> const& cols) {
+    std::vector<std::string> key;
+    key.reserve(cols.size());
+    for (auto const* c : cols) {
+        int code = c->GetValue(row);
+        key.push_back(c->DecodeValue(code));
+    }
+    return key;
+}
```
Oddin60F: Considering how this function is used, you can do without translating codes into strings. When encoding tables in `EncodedTables`, one common `value_dictionary` is created for any tables (see the `EncodedTables` constructor). Then if there is a row in two tables ("Sun", "Apple", "Bob"), which for the first table is encoded as (1,2,3), then for the second table this row will also be encoded as (1,2,3).
Therefore, it is better to work with numeric values to reduce memory usage.
```suggestion
std::vector<int> MakeKeyDecoded(std::size_t row,
                                        std::vector<model::EncodedColumnData const*> const& cols) {
    std::vector<int> key;
    key.reserve(cols.size());
    for (auto const* c : cols) {
        key.push_back(c->GetValue(row));
    }
    return key;
}
```

src/core/algorithms/cind/cind_verifier/cind_verifier.cpp
Oddin60F: Our styleguide forbids two statements on one line, even in such simple cases.

```suggestion
    for (auto& s : cond_vals) {
    s = StripQuotes(s);
    }
```

src/core/algorithms/cind/cind_verifier/cind_verifier.cpp
```diff
@@ -0,0 +1,468 @@
...
+        if (lhs_set.contains(col_index)) continue;
+
+        if (!same_table_enc) {
+            conditional.push_back(&column);
+            continue;
+        }
+
+        if (!rhs_set.contains(col_index)) {
+            conditional.push_back(&column);
+        }
```
Oddin60F:
```suggestion
       if (lhs_set.contains(col_index) || (same_table_enc && rhs_set.contains(col_index)) ) { 
           continue;
       }
       conditional.push_back(&column);
```

src/core/algorithms/cind/cind_verifier/cind_verifier.cpp
```diff
@@ -0,0 +1,468 @@
...
+                ++current_row_id;
+            }
+        }
+
+        for (auto& [row, cluster] : violating_map) {
+            ViolatingCluster vc;
+            vc.violating_rows = std::move(cluster);
+            vc.basket_rows = vc.violating_rows;
+            violating_clusters_.push_back(std::move(vc));
+        }
```
Oddin60F:
```suggestion
        for (auto& [_, cluster] : violating_map) {
        ViolatingCluster vc{.basket_rows = cluster, .violating_rows = std::move(cluster)};
            violating_clusters_.push_back(std::move(vc));
        }
```

src/core/algorithms/cind/cind_verifier/cind_verifier.cpp
```diff
@@ -0,0 +1,468 @@
...
+    model::TableIndex const rhs_table_idx =
+            (input_tables_.size() == 1) ? 0
+                                        : static_cast<model::TableIndex>(input_tables_.size() - 1);
+
+    auto const& lhs_enc = tables.GetTable(lhs_table_idx);
+    auto const& rhs_enc = tables.GetTable(rhs_table_idx);
+
+    if (lhs_enc.GetColumnData().empty() || lhs_enc.GetColumnData().front().GetNumRows() == 0) {
+        throw std::runtime_error("Got an empty LHS table: CIND verification is meaningless.");
+    }
```
Oddin60F: This can be checked in `LoadDataInternal()`.

src/core/algorithms/cind/cind_verifier/cind_verifier.cpp
```diff
@@ -0,0 +1,468 @@
...
+
+    std::vector<std::string> cond_vals2 = condition_values_;
+    if (cond_vals2.empty()) {
+        cond_vals2.assign(conditional.size(), kAnyValue);
+    }
+    if (cond_vals2.size() != conditional.size()) {
+        throw std::runtime_error(
+                "cind_condition_values size must equal number of conditional attributes");
+    }
+    for (auto& s : cond_vals2) s = StripQuotes(s);
```
Oddin60F: How does this code differ from the one we executed in lines 196--211? It feels like we're doing the same thing twice.

src/core/algorithms/cind/cind_verifier/cind_verifier.cpp
```diff
@@ -0,0 +1,468 @@
...
+
+                if (!rhs_rows.contains(row)) {
+                    violating_map[row].push_back(current_row_id);
+                    ++violating_rows_;
+                }
+
+                ++current_row_id;
+            }
+        }
+
```
Oddin60F, line 229: I think it's still worth moving the logic of parsing tables to `LoadDataInternal()` using `EncodedTables`. It is assumed that the `ExecuteInternal()` method will only perform validation logic, and not input data processing. 

If you create `EncodedTables` in `LoadDataInternal()`, then
1) The user will be able to verify several different instances of the primitive for two tables using a single algorithm object, he will not need to recreate the algorithm for each new validation.
2) In `ExecuteInternal()`, you will not need to parse the tables separately for the `all_wildcards` case, the original full rows can be restored by analogy with the modified `MakeKeyDecoded()` method, where `cols` are all the columns of the table.
3) The logic will become much clearer.
4) Many points below can be simplified.

src/core/algorithms/cind/cind_verifier/cind_verifier.cpp
```diff
@@ -0,0 +1,468 @@
...
+        auto it = decode_to_code.find(want);
+        if (it == decode_to_code.end()) {
+            real_validity_ = -1.0;
+            real_completeness_ = 0.0;
+            return;
+        }
+
+        expected_codes.push_back(it->second);
```
Oddin60F: `EncodedColumnData` already has a container storing string values that occur in this column. If we need to check that the value from the pattern is found in the column, then it's easier to do this:
```suggestion
        auto const* col = conditional[i];
         if (!col->GetUniqueValues().contains(want)) {
            real_validity_ = -1.0;
            real_completeness_ = 0.0;
            return;
         }

        expected_codes.push_back(col->GetValueDict()->ToInt(want));
```
P.S. To use this code, the `GetValueDict()` dictionary method must be added to the `EncodedColumnData` class, as it is currently missing.

src/core/algorithms/cind/cind_verifier/cind_verifier.cpp
```diff
@@ -0,0 +1,468 @@
...
+    RegisterOption(config::kRhsRawIndicesOpt(
+            &ind_.rhs, [this]() { return input_tables_.back()->GetNumberOfColumns(); },
+            [&lhs = ind_.lhs, &check_uniqueness](config::IndicesType const& indices) {
+                check_uniqueness(indices);
+                if (!lhs.empty() && lhs.size() != indices.size()) {
+                    throw config::ConfigurationError{
+                            "Invalid input: LHS and RHS indices must have the same size"};
+                }
+            }));
+
```
Oddin60F: It seems to me that it is better to use the options `kRhsIndicesOpt` and `kLhsIndicesOpt` here, then it will be possible to rewrite it like this.

```suggestion
    RegisterOption(config::kLhsIndicesOpt(
            &ind_.lhs, [this]() { return input_tables_.front()->GetNumberOfColumns(); },
            [&rhs = ind_.rhs](config::IndicesType const& indices) {
                if (rhs.size() != indices.size()) {
                throw config::ConfigurationError{"Invalid input: all indices should be unique"};
                }
            }));

    RegisterOption(config::kRhsIndicesOpt(
            &ind_.rhs, [this]() { return input_tables_.back()->GetNumberOfColumns(); },
            [&lhs = ind_.lhs](config::IndicesType const& indices) {
                if (lhs.size() != indices.size()) {
                    throw config::ConfigurationError{"Invalid input: all indices should be unique"};
                }
            }));
```

src/core/algorithms/cind/cind_verifier/cind_verifier.cpp
```diff
@@ -0,0 +1,468 @@
...
+    out.reserve(indices.size());
+    for (auto idx : indices) {
+        auto it = by_index.find(static_cast<IndexT>(idx));
+        if (it == by_index.end()) {
+            throw std::runtime_error("Internal error: inclusion index not found in encoded table");
+        }
+        out.push_back(it->second);
+    }
+    return out;
+}
```
Oddin60F: Why does the method return Ordered columns? It seems that right now we are passing `ind.lhs` and `ind.rhs`, which are not necessarily sorted. At the same time, if we had used the `kLhsIndicesOpt` option (as noted below), this would have been guaranteed, since the values would have been sorted when the options were loaded.
In any case, it seems that templates can be dispensed with here, in the code we use this function twice for the same types.

src/python_bindings/cind/bind_cind_verification.cpp
```diff
@@ -0,0 +1,40 @@
...
+            .def("get_real_completeness", &CINDVerifier::GetRealCompleteness)
+            .def("get_violating_rows_count", &CINDVerifier::GetViolatingRowsCount)
+            .def("get_violating_clusters_count", &CINDVerifier::GetViolatingClustersCount)
+            .def("get_violating_clusters", &CINDVerifier::GetViolatingClusters)
+            .def("get_supporting_baskets", &CINDVerifier::GetSupportingBaskets)
+            .def("get_included_supporting_baskets", &CINDVerifier::GetIncludedSupportingBaskets)
+            .def("get_included_baskets_total", &CINDVerifier::GetIncludedBasketsTotal);
+
+    py::module_ alg = m.def_submodule("algorithms");
+    alg.attr("Default") = py::reinterpret_borrow<py::object>(cls);
```
Oddin60F: You can remove it, `BindPrimitiveNoBase` will do it itself.

src/python_bindings/cind/bind_cind_verification.cpp
```diff
@@ -0,0 +1,40 @@
...
+void BindCindVerification(py::module_& main_module) {
+    using algos::cind::CINDVerifier;
+
+    py::module_ m = main_module.def_submodule("cind_verification");
+
+    using ViolatingCluster = CINDVerifier::ViolatingCluster;
+
+    py::class_<ViolatingCluster>(m, "ViolatingCluster")
+            .def_readonly("basket_rows", &ViolatingCluster::basket_rows)
+            .def_readonly("violating_rows", &ViolatingCluster::violating_rows);
```
Oddin60F: I'm not sure, but it seems that serialization and deserialization methods need to be added for this type (see the binding requirements in _Wiki/Development‐Features_ and an example of the `__getstate__` and `__setstate__` methods, for example in `bind_md.cpp` )

src/core/algorithms/cind/cind_verifier/cind_verifier.cpp
```diff
@@ -0,0 +1,468 @@
...
+        expected_codes.push_back(it->second);
+    }
+
+    auto row_matches = [&](std::size_t row) -> bool {
+        for (std::size_t i = 0; i < conditional.size(); ++i) {
+            if (!expected_codes[i].has_value()) continue;
+            if (conditional[i]->GetValue(row) != *expected_codes[i]) return false;
+        }
+        return true;
+    };
```
Oddin60F, line 282: Imho, I would move it closer to the place of use.

src/core/algorithms/cind/cind_verifier/cind_verifier.cpp
```diff
@@ -0,0 +1,468 @@
...
+
+    using Key = std::vector<std::string>;
```
Oddin60F: If we had used the `MakeKeyDecoded()` version without decoding to a string, we could have worked with numbers.

```suggestion
    using Key = std::vector<int>;
```

src/core/algorithms/cind/cind_verifier/cind_verifier.cpp
```diff
@@ -0,0 +1,468 @@
...
+    included_baskets_total_ = 0;
+    if (condition_type_._value == CondType::group) {
+        for (auto const& [_, acc] : acc_by_key) {
+            if (acc.included) ++included_baskets_total_;
+        }
+    } else {
+        for (auto const& [_, acc] : acc_by_key) {
+            if (acc.included) included_baskets_total_ += acc.basket_rows.size();
+        }
+    }
```
Oddin60F: It seems that it will be more readable this way.
```suggestion
for (auto const& [_, acc] : acc_by_key) {
    if (acc.included) {
        included_baskets_total_ += (condition_type_._value == CondType::group) 
                                    ? 1 
                                    : acc.basket_rows.size();
    }
}
```

src/core/algorithms/cind/cind_verifier/cind_verifier.cpp
```diff
@@ -0,0 +1,468 @@
...
+                included_support_ += acc.matching_rows.size();
+            } else {
+                violating_rows_ += acc.matching_rows.size();
+                violating_clusters_.push_back(ViolatingCluster{
+                        .basket_rows = std::move(acc.basket_rows),
+                        .violating_rows = std::move(acc.matching_rows),
+                });
+            }
+        }
+    }
```
Oddin60F, line 344:
```suggestion
    const bool is_group = (condition_type_._value == CondType::group);
    for (auto& [_, acc] : acc_by_key) {
    if (acc.matching_rows.empty()) continue;

    const size_t contribution = is_group ? 1 : acc.matching_rows.size();

    supporting_baskets_ += contribution;

    if (acc.included) {
        included_support_ += contribution;
        continue;
    }
    violating_rows_ += acc.matching_rows.size();
        violating_clusters_.push_back(ViolatingCluster{
            .basket_rows = std::move(acc.basket_rows),
            .violating_rows = std::move(acc.matching_rows),
        });
}
```

src/core/algorithms/cind/cind_verifier/cind_verifier.cpp
```diff
@@ -0,0 +1,468 @@
...
+#include "core/util/timed_invoke.h"
+
+namespace algos::cind {
+namespace {
+
+static constexpr char kCindCondValues[] = "cind_condition_values";
+static constexpr char kDCindCondValues[] =
+        "Condition values aligned with conditional attributes order from "
+        "CindMiner::ClassifyAttributes. "
+        "Use '-' or '_' as wildcard. If empty => all wildcards.";
```
Oddin60F: This needs to be moved to `core/config/names `and `core/config/descriptions`.

Oddin60F (changes requested):

src/core/algorithms/cind/condition_miners/pli_cind.cpp
```diff
@@ -0,0 +1,196 @@
...
+std::pair<std::vector<int>, std::vector<int>> PliCind::ClassifyRows(Attributes const& attrs) {
+    std::unordered_set<std::vector<std::string>, VectorStringHash> rhs_values;
+    for (size_t index = 0; index < attrs.rhs_inclusion.front()->GetNumRows(); ++index) {
+        std::vector<std::string> row;
+        row.reserve(attrs.rhs_inclusion.size());
+        for (auto& attr : attrs.rhs_inclusion) {
+            row.push_back(attr->GetStringValue(index));
+        }
+        rhs_values.insert(std::move(row));
+    }
```
Oddin60F: You can also use codes instead of strings here. Moreover, it is strange that strings are used here, and in a similar logic `Cinderella::GetBaskets()` codes.

src/core/algorithms/cind/condition_miners/pli_cind.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+    // included row_id for rows, group_id for groups
+    std::vector<int> included_pos;
+
+    std::unordered_map<std::vector<std::string>, int, VectorStringHash> group_idx;
+    std::vector<int> row_to_group;
+    row_to_group.reserve(attrs.lhs_inclusion.front()->GetNumRows());
```
Oddin60F: It seems to me that it is impractical to always allocate memory for `row_to_group`, even if a type of condition "Row" is selected. We need to do something to make reservations happen only if we have a type of condition "Group". Otherwise, we will have a large container (the size of the number of rows in the lhs table) in our memory that we don't even use.

src/core/algorithms/cind/condition_miners/itemset_node.h
```diff
@@ -0,0 +1,109 @@
...
+          validity_(validity),
+          completeness_(completeness),
+          parent_node_(std::move(parent)) {}
+
+    std::shared_ptr<ItemsetNode> CreateChild(Item value, std::vector<BasketInfo> baskets_info,
+                                             size_t included_baskets_cnt, double min_completeness) {
+        double included_contained_baskets_cnt = 0.0;
+        for (auto const& [_, real_id, is_included] : baskets_info) {
+            included_contained_baskets_cnt += is_included;
+        }
```
Oddin60F:
```suggestion
        for (auto const& [_, _, is_included] : baskets_info) {
            included_contained_baskets_cnt += is_included ? 1 : 0;
        }
```

src/core/algorithms/cind/condition_miners/itemset_node.h
```diff
@@ -0,0 +1,109 @@
...
+                                                      static_cast<double>(baskets_info.size());
+
+            auto new_element =
+                    std::make_shared<ItemsetNode>(shared_from_this(), std::move(value),
+                                                  std::move(baskets_info), validity, completeness);
+
+            child_nodes_.emplace(new_element->GetValue(), new_element);
+            return new_element;
+        }
+        return nullptr;
```
Oddin60F: imho, it will be better to read this way.

```suggestion
        if (completeness < min_completeness) {
            return nullptr;
        }
        double const validity = baskets_info.empty()
                                            ? -1.0
                                            : included_contained_baskets_cnt /
                                                      static_cast<double>(baskets_info.size());

            auto new_element =
                    std::make_shared<ItemsetNode>(shared_from_this(), std::move(value),
                                                  std::move(baskets_info), validity, completeness);

            child_nodes_.emplace(new_element->GetValue(), new_element);
            return new_element;
```

src/core/algorithms/cind/condition_miners/itemset_node.h
```diff
@@ -0,0 +1,109 @@
...
+        }
+    }
+
+private:
+    Item value_;
+    std::vector<BasketInfo> baskets_info_;
+    double validity_;
+    double completeness_;
+    std::unordered_map<Item, std::shared_ptr<ItemsetNode>> child_nodes_;
+    std::weak_ptr<ItemsetNode> parent_node_;
```
Oddin60F: We have an exception from Google's styling guide for the order of modifiers in a class (see _Wiki/Development‐General_)

src/core/algorithms/cind/cind_verifier/cind_verifier.cpp
Oddin60F: I mistakenly said about the ordering of `ind.lhs` and `ind.rhs`, but still, `by_index` is also not sorted.

src/core/algorithms/cind/condition_miners/cind_miner.cpp
```diff
@@ -0,0 +1,76 @@
...
+        }
+    }
+
+    for (auto const& column : tables_.GetTable(rhs_table).GetColumnData()) {
+        auto const col_index = column.GetColumn()->GetIndex();
+        if (std::ranges::find(rhs_indices, col_index) != rhs_indices.end()) {
+            result.rhs_inclusion.push_back(&column);
+        }
+    }
+
```
Oddin60F, line 47: imho, I would rewrite the method by analogy with the same logic in the verifier, where the logic looks more obvious and acceptable. You can also load indexes into `unordered_set` and search in it and change the conditions for readability.

src/core/algorithms/cind/condition_miners/cinderella.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+                    candidate.pop_back();
+                }
+            }
+
+            candidate.pop_back();
+        }
+    }
+
+    itemset.CreateNewLayer(new_items_info);
```
Oddin60F:
```suggestion
    itemset.CreateNewLayer(std::move(new_items_info));
```

src/core/algorithms/cind/cind.h
```diff
@@ -0,0 +1,48 @@
...
+                result.append(attr).append(", ");
+            }
+            result.resize(result.size() - 2);
+            result.append(");\n");
+            for (auto const& condition : conditions) {
+                result.append("\t").append(condition.ToString()).append(";\n");
+            }
+        }
+        return result;
+    }
```
Oddin60F, line 30: imho, it will be more readable if you use `std::ostringstream`.
```suggestion
    std::string ToString() const {
    std::ostringstream oss;
    oss << ind.ToLongString()
        << "\nPossible conditions number: " << conditions.size() << "\n";
    if (conditions.empty()) {
        return oss.str();
    }
    oss << "Possible conditions:\n\t(";
        for (size_t i = 0; i < conditional_attributes.size(); ++i) {
            if (i != 0) oss << ", ";
            oss << conditional_attributes[i];
        }
        oss << ");\n";
        for (auto const& condition : conditions) {
            oss << "\t" << condition.ToString() << ";\n";
        }
    return oss.str();
}
    }
```

src/core/algorithms/cind/condition.h
```diff
@@ -0,0 +1,106 @@
...
+                }
+            }
+        }
+    }
+
+    Condition(std::vector<int> const& cluster_attrs_ids,
+              model::PLSet::ClusterValue const& cluster_value,
+              std::vector<model::EncodedColumnData const*> const& condition_attrs, double _validity,
+              double _completeness)
+        : validity(_validity), completeness(_completeness) {
```
Oddin60F: Some kind of weird naming style.

```suggestion
    Condition(std::vector<int> const& cluster_attrs_ids,
              model::PLSet::ClusterValue const& cluster_value,
              std::vector<model::EncodedColumnData const*> const& condition_attrs, double validity,
              double completeness)
        : validity(validity), completeness(completeness) {
```

src/core/algorithms/cind/cind_verifier/cind_verifier.cpp
```diff
@@ -0,0 +1,468 @@
...
+};
+
+struct VecStrHash {
+    std::size_t operator()(std::vector<std::string> const& v) const noexcept {
+        return boost::hash_value(v);
+    }
+};
+
+inline bool IsWildcard(std::string const& s) {
+    return s == "-" || s == "_" || s == kAnyValue;
```
Oddin60F: It seems that kAnyValue is "-".

src/core/algorithms/cind/condition_miners/cind_miner.h
```diff
@@ -0,0 +1,47 @@
...
+#include "core/algorithms/cind/types.h"
+#include "core/algorithms/ind/ind.h"
+#include "core/config/tabular_data/input_tables_type.h"
+#include "core/model/table/encoded_tables.h"
+#include "core/util/primitive_collection.h"
+
+namespace algos::cind {
+class CindAlgorithm;
+using model::EncodedColumnData;
+using model::EncodedTables;
```
Oddin60F, line 12: It seems that forward declaration is not needed here, since we have included the header file.

src/core/algorithms/cind/condition_miners/itemset.h
```diff
@@ -0,0 +1,102 @@
...
+            new_items.push_back(child);
+
+            if (seen_parents.insert(parent.get()).second) {
+                new_prev_items.push_back(parent);
+            }
+        }
+
+        items_ = std::move(new_items);
+        prev_items_ = std::move(new_prev_items);
+    }
```
Oddin60F, line 89: imho
```suggestion
    void CreateNewLayer(
            std::vector<std::tuple<std::shared_ptr<ItemsetNode>, Item, std::vector<BasketInfo>>>
                    new_items_info) {
        items_.clear();
        prev_items_.clear();
        items_.reserve(new_items_info.size());

        std::unordered_set<ItemsetNode*> seen_parents;
        seen_parents.reserve(new_items_info.size());

        for (auto& [parent, value, baskets] : new_items_info) {
            auto child = parent->CreateChild(std::move(value), std::move(baskets),
                                             included_baskets_cnt_, min_completeness_);
            if (!child) {
                continue;
            }

            new_items.push_back(child);

            if (seen_parents.insert(parent.get()).second) {
                new_prev_items.push_back(parent);
            }
        }
    }
```

src/core/model/table/column_encoded_relation_data.cpp
```diff
@@ -0,0 +1,46 @@
...
+        }
+        for (size_t index = 0; index < row.size(); ++index) {
+            std::string const& field = row[index];
+            unique_values[index].insert(field);
+            if (field.empty()) {
+                column_vectors[index].push_back(kNullValueId);
+            } else {
+                column_vectors[index].push_back(value_dictionary->ToInt(field));
+            }
+        }
```
Oddin60F:
```suggestion
        size_t index = 0;
        for (auto&& field: row) {
            if (field.empty()) {
                column_vectors[index].push_back(kNullValueId);
            } else {
                column_vectors[index].push_back(value_dictionary->ToInt(field));
            }
            unique_values[index].insert(std::move(field));
            ++index;
        }
```

src/python_bindings/cind/bind_cind.cpp
```diff
@@ -0,0 +1,70 @@
...
+            .def("__hash__",
+                 [](Condition const& cond) {
+                     return py::hash(py::int_(std::hash<Condition>{}(cond)));
+                 })
+            .def("data",
+                 [](Condition const& cond) { return VectorToTuple(cond.condition_attrs_values); })
+            .def("validity", [](Condition const& cond) { return cond.validity; })
+            .def("completeness", [](Condition const& cond) { return cond.completeness; });
+
+    py::class_<CIND>(cind_module, "CIND")
```
Oddin60F, line 65: There are definitely not enough serialization and deserialization methods here.

src/python_bindings/cind/bind_cind.cpp
```diff
@@ -0,0 +1,70 @@
...
+                     }
+                     return result;
+                 })
+            .def("get_condition_attributes",
+                 [](CIND const& cind) { return VectorToTuple(cind.conditional_attributes); });
+
+    BindPrimitiveNoBase<CindAlgorithm>(cind_module, "Cind")
+            .def("get_cinds", &CindAlgorithm::CINDList)
+            .def("time_taken", &CindAlgorithm::TimeTaken);
+}
```
Oddin60F: It seems worth giving the name `Cinderella` instead of `Cind` and `time_taken` looks superfluous.

.gitignore
```diff
@@ -12,6 +12,7 @@ datasets/*.zip
 .cache/
 .idea/
 .vscode/
+**/compile_commands.json
```
Oddin60F: It doesn't seem to need to be added to PR.

Oddin60F (changes requested):

src/core/config/descriptions.h
```diff
@@ -243,4 +243,15 @@ constexpr auto kDOcRightIndex = "Index of the right attribute of the OC to verif
...
 constexpr auto kDOFDContext = "Context of the OFD to verify";
 constexpr auto kDOFDRightIndex = "Right index of the OFD to verify";
+
+// CIND
+constexpr auto kDValidity =
+        "Percentage of rows/groups that form IND when the pattern tuple is satisfied";
+constexpr auto kDCompleteness =
+        "Percentage of rows/groups from IND that satisfies the pattern tuple";
+constexpr auto kDConditionType = "Type of condition units: can be either rows or groups";
+constexpr auto kDAlgoType = "cinderella or pli_cind";
```
Oddin60F: It is better to define a function for this :

```suggestion
// in the namespace `detail` above
`std::string const kDConditionType =
        "CIND condition types to use\n" +
        util::EnumToAvailableValues<algos::cind::CondType>();`
`std::string const kDAlgoType =
        "CIND algoritm types to use\n" +
        util::EnumToAvailableValues<algos::cind::AlgoType>();`
// ...
auto kDConditionType = detail::kDConditionType.c_str();
auto kDAlgoType = detail::kDAlgoType.c_str();
// ...
```

src/core/algorithms/cind/condition.h
```diff
@@ -0,0 +1,113 @@
...
+        std::string result = "(";
+        for (auto const& value : condition_attrs_values) {
+            result.append("\"").append(value).append("\", ");
+        }
+        result.append("validity = ")
+                .append(std::to_string(validity))
+                .append(", completeness = ")
+                .append(std::to_string(completeness))
+                .append(")");
+        return result;
```
Oddin60F:
```suggestion
        std::ostringstream oss;
        oss << "(";
        for (auto const& value : condition_attrs_values) {
              oss << "\"" << value << "\", ";
        }
        oss << "validity = " << validity
        << ", completeness = " << completeness << ")";
        return oss.str();
```

src/core/algorithms/cind/condition_miners/cind_miner.h
```diff
@@ -0,0 +1,46 @@
...
+    model::EncodedTables tables_;
+    double min_validity_;
+    double min_completeness_;
+    CondType condition_type_;
+
+    util::PrimitiveCollection<CIND> cind_collection_;
+
+private:
+    friend class CindAlgorithm;
+};
```
Oddin60F, line 43: The remark about the order of field modifiers in the class applies to all files.

src/core/algorithms/cind/cind_verifier/cind_verifier.cpp
```diff
@@ -0,0 +1,350 @@
...
+                                 ? -1.0
+                                 : static_cast<double>(included_support_) /
+                                           static_cast<double>(supporting_baskets_);
+
+        real_completeness_ = (included_baskets_total_ == 0)
+                                     ? 0.0
+                                     : static_cast<double>(included_support_) /
+                                               static_cast<double>(included_baskets_total_);
+        return;
+    }
```
Oddin60F: I think it's better to put this logic in a separate function for readability. It also seems that `lhs_key_set` is superfluous here.

```suggestion
///
void CINDVerifier::ValidateAIND(){
// Simple inclusion check: every LHS row key must appear in RHS keys
        std::unordered_map<std::vector<int>, Cluster, VecIntHash> violating_map;
        std::size_t lhs_rows = lhs_inclusion.front()->GetNumRows();
        
        for (std::size_t l = 0; l < lhs_rows; ++l) {
            auto key = MakeKey(l, lhs_inclusion);

            if (!rhs_keys.contains(key)) {
                violating_map[std::move(key)].push_back(static_cast<model::TupleIndex>(l));
                ++violating_rows_;
            }
        }
        // Don't forget to add the necessary headers
        std::transform(violating_map.begin(), violating_map.end(), 
        std::back_inserter(violating_clusters_),
        [](auto& kv) {
        auto& cluster = kv.second;
        return ViolatingCluster{
            .basket_rows = cluster,
            .violating_rows = std::move(cluster)
            };
        });

        supporting_baskets_ = lhs_rows;
        included_support_ = lhs_rows - violating_clusters_.size();
        included_baskets_total_ = included_support_;

        real_validity_ = (supporting_baskets_ == 0)
                                 ? -1.0
                                 : static_cast<double>(included_support_) /
                                           supporting_baskets_;

        real_completeness_ = (included_baskets_total_ == 0)
                                     ? 0.0
                                     : static_cast<double>(included_support_) /
                                               included_baskets_total_;
}
///
    if (all_wildcards) {
        ValidateAIND();
        return;
    }
```

src/core/algorithms/cind/cind_verifier/cind_verifier.cpp
```diff
@@ -0,0 +1,350 @@
...
+    // Count included baskets total
+    for (auto const& [_, acc] : acc_by_key) {
+        if (acc.included) {
+            included_baskets_total_ +=
+                    (condition_type_._value == CondType::group) ? 1 : acc.basket_rows.size();
+        }
+    }
+
+    // Compute support and violations
+    bool const is_group = (condition_type_._value == CondType::group);
```
Oddin60F, line 326: Use it in the last cycle too

src/core/algorithms/cind/cind_verifier/cind_verifier.cpp
```diff
@@ -0,0 +1,350 @@
...
+    } else if (cond_vals.size() != conditional.size()) {
+        throw std::runtime_error(
+                "cind_condition_values size must equal number of conditional attributes");
+    }
+    for (auto& s : cond_vals) {
+        s = StripQuotes(s);
+    }
+
+    bool const all_wildcards = std::all_of(cond_vals.begin(), cond_vals.end(),
+                                           [](std::string const& s) { return IsWildcard(s); });
```
Oddin60F:
```suggestion
    bool const all_wildcards = std::ranges::all_of(cond_vals,
                                           [](std::string const& s) { return IsWildcard(s); });
```

src/core/model/table/encoded_column_data.h
```diff
@@ -0,0 +1,77 @@
...
+        return column_->GetIndex();
+    }
+
+    std::set<std::string> const& GetUniqueValues() const noexcept {
+        return unique_values_;
+    }
+
+    std::shared_ptr<ValueDictionary> const& GetValueDict() const noexcept {
+        return value_dictionary_;
+    }
```
Oddin60F: The pointer is constant, but the dictionary itself is not. A constant reference will be enough for the validator.
```suggestion
    ValueDictionary const& GetValueDict() const noexcept {
        return *value_dictionary_;
    }
```

src/core/algorithms/cind/cind_verifier/cind_verifier.cpp
```diff
@@ -0,0 +1,350 @@
...
+        }
+
+        auto const* col = conditional[i];
+        if (!col->GetUniqueValues().contains(want)) {
+            real_validity_ = -1.0;
+            real_completeness_ = 0.0;
+            return;
+        }
+
+        expected_codes.push_back(col->GetValueDict()->ToInt(want));
```
Oddin60F: Now, using the non-constant `ToInt()` method, we can change the dictionary, since `GetValueDict()` will return a pointer to the non-constant dictionary.
Yes, this is not possible at this point in the code, since we have checked for existence, but it is still worth adding the constant `ToInt()` method to the `ValueDictionary` class.

src/core/algorithms/cind/cind_verifier/cind_verifier.cpp
```diff
@@ -0,0 +1,350 @@
...
+
+    auto row_matches = [&](std::size_t row) -> bool {
+        for (std::size_t i = 0; i < conditional.size(); ++i) {
+            if (!expected_codes[i].has_value()) continue;
+            if (conditional[i]->GetValue(row) != *expected_codes[i]) return false;
+        }
+        return true;
+    };
+
+    using Key = std::vector<int>;
```
Oddin60F, line 284: Use it for the allwildcards case too.

src/core/algorithms/cind/condition.h
```diff
@@ -0,0 +1,113 @@
...
+#include <vector>
+
+#include <boost/container_hash/hash.hpp>
+
+#include "condition_miners/itemset_node.h"
+#include "condition_miners/position_lists_set.h"
+#include "core/model/table/encoded_column_data.h"
+
+namespace algos::cind {
+char const* const kAnyValue = "-";
```
Oddin60F:
```suggestion
constexpr auto  kAnyValue = "-";
```

Oddin60F (changes requested):

src/core/algorithms/cind/cind.h
```diff
@@ -0,0 +1,49 @@
...
+    size_t Hash() const {
+        static auto const kCondHasher = std::hash<Condition>{};
+        size_t result = boost::hash_value(&ind);
+        for (auto const& cond : conditions) {
+            boost::hash_combine(result, kCondHasher(cond));
+        }
+        return result;
+    }
+};
+}  // namespace algos::cind
```
Oddin60F, line 50: Add an empty line to the end of the file here.

Oddin60F (approved):
Add an empty line to the end of the files `pli_cind.h` and `cind.h` and that's it.

PR #676, opened 2026-02-05
Add ParMaxFEM algorithm

m4xxx1m:
Adds infrastructure (parser, bindings, model) for event sequences and ParMaxFEM, a parallel version of the MaxFEM algorithm.

MaxFEM (Maximal Frequent Episode Miner) identifies only the maximal frequent episodes in a complex event sequence.


src/core/algorithms/fem/maxfem/composite_episode_miner.cpp
```diff
@@ -0,0 +1,91 @@
...
+                int current_tasks = ctx.tasks_in_flight.load(std::memory_order_relaxed);
+                if (current_tasks < ctx.max_parallel_tasks) {
+                    CompositeEpisode child = episode;
+                    child.Extend(seed, extended_bl->GetSupport());
+                    ctx.tasks_in_flight++;
+                    boost::asio::post(ctx.pool, [this, child = std::move(child),
+                                                 bl = std::move(*extended_bl), &ctx]() mutable {
+                        RunSearchTask(std::move(child), std::move(bl), ctx);
+                        ctx.tasks_in_flight--;
+                    });
```
VinogradovDmitri, line 73: `std::atomic` counter `tasks_in_flight` is incremented after posting a task to the thread pool, and decremented inside the task. This creates violation of logic in `current_tasks` < `ctx.max_parallel_tasks` counter can go negative

```suggestion
                int current_tasks = ctx.tasks_in_flight.load(std::memory_order_acquire);
                if (current_tasks < ctx.max_parallel_tasks) {
                    CompositeEpisode child = episode;
                    child.Extend(seed, extended_bl->GetSupport());
                    ctx.tasks_in_flight.fetch_add(1, std::memory_order_release);
                    boost::asio::post(ctx.pool, [this, child = std::move(child),
                                                 bl = std::move(*extended_bl), &ctx]() mutable {
                        RunSearchTask(std::move(child), std::move(bl), ctx);
                        ctx.tasks_in_flight.fetch_sub(1, std::memory_order_acq_rel);
                    });
```

src/core/algorithms/fem/maxfem/model/composite_episode.cpp
```diff
@@ -0,0 +1,92 @@
...
+    result.second = support_;
+    return result;
+}
+
+bool CompositeEpisodeComparator::operator()(std::unique_ptr<CompositeEpisode> const& lhs,
+                                            std::unique_ptr<CompositeEpisode> const& rhs) const {
+    if (lhs->GetEventsSum() != rhs->GetEventsSum()) {
+        return lhs->GetEventsSum() < rhs->GetEventsSum();
+    }
+    return lhs < rhs;
```
VinogradovDmitri:
```suggestion
    return lhs->GetRaw() < rhs->GetRaw();
```
or just
```suggestion
    return true;
```

src/core/algorithms/fem/maxfem/model/max_episodes_collection.h
```diff
@@ -0,0 +1,33 @@
...
+#include <set>
+
+#include "core/algorithms/fem/maxfem/model/composite_episode.h"
+
+namespace algos::maxfem {
+
+class MaxEpisodesCollection {
+private:
+    std::vector<std::set<std::unique_ptr<CompositeEpisode>, CompositeEpisodeComparator>>
+            max_episodes_;
```
VinogradovDmitri: Do u acknowledged what `std::unique_ptr` addresses are unique, two equal episodes will be considered different, cause duplicates in the `std::set`
if answer is "No" try use `std::vector<std::set<std::shared_ptr<CompositeEpisode>, CompositeEpisodeComparator>>` and update all insertion/extraction code to use `std::shared_ptr`

src/core/algorithms/fem/maxfem/model/max_episodes_collection.cpp
```diff
@@ -0,0 +1,158 @@
...
+        episodes_set.clear();
+    }
+
+    max_episodes_.clear();
+    return result;
+}
+
+bool MaxEpisodesCollection::CheckForSuperEpisode(CompositeEpisode const& episode,
+                                                 size_t length) const {
+    for (int index = max_episodes_.size() - 1; index >= static_cast<int>(length); --index) {
```
VinogradovDmitri: `max_episodes_.size()` is unsigned number so when it zero it cause UB

```suggestion
    if (max_episodes_.empty()) return false;
    for (int index = max_episodes_.size() - 1; index >= static_cast<int>(length); --index) {
```

src/core/algorithms/fem/maxfem/model/bound_list.h
```diff
@@ -0,0 +1,30 @@
...
+#include "core/algorithms/fem/maxfem/model/parallel_episode.h"
+#include "core/model/sequence/timestamp.h"
+
+namespace algos::maxfem {
+
+class BoundList {
+private:
+    std::vector<model::Timestamp> starts_;
+    std::vector<model::Timestamp> ends_;
+
```
VinogradovDmitri, line 14: idk maybe use 1 `std::vector<pair<int, int>>`, should consume less memory

VinogradovDmitri (changes requested):
if I'm wrong let me know)
And if u fixed something, can u leave a comment something like "Done"

src/core/algorithms/fem/maxfem/composite_episode_miner.cpp
m4xxx1m: I am aware of violation `current_tasks` < `ctx.max_parallel_tasks`. It's just an heuristic, so it's okay.
(Though changing memory order of operations won't fix it)

src/core/algorithms/fem/maxfem/model/composite_episode.cpp
m4xxx1m: First option does the same thing.
Second options isn't correct, as we can't have duplicates (though in fact it's `<=`).

src/core/algorithms/fem/maxfem/model/max_episodes_collection.h
m4xxx1m: All episodes inserted into the sets of `max_episdoes_` have unique addresses.
If they weren't, `std::shared_ptr` wouldn't fix the problem.

VinogradovDmitri: I'm not going to lie, I didn't read the article (no one left links), but let me know if I understood correctly that you want to keep the episodes identical (even in content) and use `std::set` as `std::multiset`

m4xxx1m: There are no identical episodes in this structure, they are all unique. Heuristic is to keep them sorted by summary of all events. But for episodes with the same events summary order is not important.

src/core/algorithms/fem/maxfem/model/bound_list.h
m4xxx1m: Done

src/core/algorithms/fem/maxfem/model/max_episodes_collection.cpp
m4xxx1m: Technically it is not UB. Also C++20 guarantees index will be -1 in this case

src/core/algorithms/fem/maxfem/model/composite_episode.cpp
VinogradovDmitri: pls check it again, isn't it break semantics of `std::set`, set defines unique values through !cmp(a,b) && !cmp(b,a), so then u compare pointers it's all unique, and if u want dublicates then use `std::multiset` or correct comparator

src/core/algorithms/fem/maxfem/model/max_episodes_collection.cpp
```diff
@@ -0,0 +1,158 @@
...
+    for (size_t index = 1; index <= length; ++index) {
+        auto& max_episodes_set = max_episodes_[index];
+        auto iter = max_episodes_set.begin();
+
+        while (iter != max_episodes_set.end()) {
+            auto const& max_episode = *iter;
+
+            if (max_episode->GetEvenEventsSum() >= episode.GetEventsSum()) {
+                break;
+            }
```
VinogradovDmitri: Is it correct field? And u stop then it equal, check it i don't know the algorithm, just suggest via logic
```suggestion
            if (max_episode->GetEventsSum() > episode.GetEventsSum()) {
                break;
            }
```

src/core/algorithms/fem/maxfem/model/bound_list.cpp
```diff
@@ -0,0 +1,43 @@
...
+    auto const& locs = parallel_episode.GetLocationList();
+    bound_list_.reserve(locs.size());
+    for (auto ts : locs) {
+        bound_list_.push_back({ts, ts});
+    }
+}
+
+std::optional<BoundList> BoundList::Extend(std::vector<model::Timestamp> const& loc_list,
+                                           size_t min_support, size_t window_length) const {
+    size_t const max_misses = bound_list_.size() - min_support;
```
VinogradovDmitri, line 15: Can `min_support` >`bound_list_.size()`?

src/core/model/sequence/memory_sequence_stream.h
VinogradovDmitri, line 1: Is it never used?

src/core/model/sequence/isequence_stream.h
```diff
@@ -0,0 +1,36 @@
+#pragma once
+
+#pragma once
```
VinogradovDmitri: )

VinogradovDmitri (changes requested):
maybe I'll look again later

src/core/algorithms/fem/maxfem/model/bound_list.cpp
m4xxx1m: No, such episodes are discarded earlier

src/core/algorithms/fem/maxfem/model/max_episodes_collection.cpp
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    max_episodes_.clear();
+    return result;
+}
+
+bool MaxEpisodesCollection::CheckForSuperEpisode(CompositeEpisode const& episode,
+                                                 size_t length) const {
+    for (size_t index = max_episodes_.size(); index-- > length;) {
+        auto& max_episodes_set = max_episodes_[index];
```
VinogradovDmitri, line 99: previously u had `index >= length` and i think it was right, isn't it? Now u check strictly longer than `length` - never `max_episodes_[length]` itself.
```suggestion
    for (size_t index = max_episodes_.size(); index > length + 1; --index) {
        auto& max_episodes_set = max_episodes_[index - 1]; // bypasses underflow size_t
```

VinogradovDmitri (changes requested):

src/core/algorithms/fem/maxfem/model/max_episodes_collection.cpp
m4xxx1m: No, it is correct. Inside `for (size_t i = a; i-- > b;)` body `i` will be from `a-1` to `b`, which is intented.

VinogradovDmitri (approved):

Vdaleke (approved):
LGTM

PR #686, opened 2026-03-04
Remove progress tracking in algorithms

wildsor:
The functionality is currently not used anywhere but still needs additional effort when implementing new algorithms and is another thing to think about when refactoring. Only 3 algorithms ever had proper progress stages: Depminer ("AgreeSets generation", "Finding CMAXSets", "Finding LHS"), FastFDs ("Agree sets generation", "Finding minimal covers") and Cords ("Calculating values frequencies", "SFD and correlations mining").


.github/workflows/check-codestyle.yml
```diff
@@ -28,15 +28,15 @@ jobs:
...
           wget https://apt.llvm.org/llvm-snapshot.gpg.key
           sudo apt-key add llvm-snapshot.gpg.key
           sudo apt-get update
-          sudo apt-get install -y clang-format-18
+          sudo apt-get install -y clang-format-20
       - name: Check formatting
         run: |
           set +e
-          formatterOutput=$( git diff -U0 origin/$GITHUB_BASE_REF...HEAD | clang-format-diff-18 -p 1)
+          formatterOutput=$( git diff -U0 origin/$GITHUB_BASE_REF...HEAD | clang-format-diff-20 -p 1)
```
perfotoran: If we're  bumping the clang-format version, shouldn't we go with the latest  one (v22, for example)?

wildsor: It wasn't available in the apt repo at the time. Updated it now.
image

perfotoran (approved):

PR #688, opened 2026-03-05
Fix afd

Sneper-Breeze:
## Summary
This PR redesigns the class structure for the Tane algorithm, enabling it to return Approximate Functional Dependencies (AFDs) with thresholds. It separates AFD handling from the existing regular FD class structures, paving the way for future algorithms to adopt the AFD class hierarchy. Additionally, this PR adds new metrics to the AFDMetricCalculator class.

### 1.  `AFD` Class
This new class inherits from the existing `FD` class. It extends the regular FD with a `threshold_` field and corresponding methods, allowing AFD mining algorithms to store and retrieve the threshold for each discovered AFD.

### 2.  `AFDAlgorithm` Class
This class replaces the regular `FDAlgorithm` for AFD mining. It collects and returns `AFD` objects instead of `FD` objects, ensuring that thresholds are preserved for every discovered dependency.

### 3. `PliBasedAFDAlgorithm` Class
This class inherits from `AFDAlgorithm` and serves as the AFD counterpart to `PliBasedFDAlgorithm`. It provides a base for implementing AFD algorithms that rely on position list indices (PLI). Previously, AFD mining algorithms collected AFDs as if they were regular FDs, making it impossible to retrieve the threshold for each AFD.

All AFD‑related classes (`AFD`, `AFDAlgorithm`, `PliBasedAFDAlgorithm`) are located in the same directory as their FD analogues: `core/algorithms/fd/`.

### 4. New Metrics in `AFDMetricCalculator`
The following metrics have been added to `AFDMetricCalculator` (formulas are available in [this paper](https://arxiv.org/pdf/2312.06296)):

- `g1`
- `g3`
- `rho`
- `g1 zero-ary` (a variant specifically required by the Tane algorithm)

### 5.  Tane Refactoring
Tane has been refactored to serve as an example of migrating an AFD mining algorithm from `FDAlgorithm` to `AFDAlgorithm`. All metric calculations in Tane now use the new `AFDMetricCalculator`, and the algorithm mines AFDs using all the metrics listed above.

### 6. Python Bindings
Python bindings have been added for the `AFD` class and for algorithm classes derived from `AFDAlgorithm`, enabling their use from Python.

## Details

**Note on the g1 metric**: The current implementation calculates `g1` in a non‑standard way because Tane originally used this variant and all existing tests were written accordingly. A future update may align it with the standard definition.


src/core/algorithms/fd/tane/afd_measures.h
```diff
@@ -4,21 +4,4 @@
...
-
-config::ErrorType CalculateTauMeasure(model::PLIWS const* x_pli, model::PLIWS const* a_pli,
-                                      model::PLIWS const* xa_pli);
-
-config::ErrorType CalculateMuPlusMeasure(model::PLIWS const* x_pli, model::PLIWS const* a_pli,
-                                         model::PLIWS const* xa_pli);
-
-config::ErrorType CalculateRhoMeasure(model::PLIWS const* x_pli, model::PLIWS const* xa_pli);
-}  // namespace algos
+namespace algos {}  // namespace algos
```
wildsor: !

src/core/algorithms/fd/tane/tane_common.cpp
```diff
@@ -28,8 +27,17 @@ double TaneCommon::CalculateUccError(model::PositionListIndex const* pli,
     return pli->GetNepAsLong() / static_cast<double>(relation_data->GetNumTuplePairs());
 }
 
+// void TaneCommon::RegisterAndCountFd(Vertical lhs, Column const* rhs) {
+//     RegisterFd(std::move(lhs), *rhs, relation_->GetSharedPtrSchema());
```
wildsor: Extra comment

wildsor (changes requested):
okay, I'm not going to look any further, please look over the code yourself one more time

src/core/algorithms/fd/afd_metric/afd_metric_calculator.cpp
```diff
@@ -86,6 +100,48 @@ long double AFDMetricCalculator::CalculateG2(model::PLI const* lhs_pli, model::P
     return num_error_rows / num_rows;
 }
 
+long double AFDMetricCalculator::CalculateG3(model::PLI const* lhs_pli, model::PLI const* rhs_pli,
+                                             size_t num_rows) {
+    if (num_rows <= 0) throw std::invalid_argument("received unpositive number of rows");
```
wildsor:
```suggestion
    if (num_rows <= 0) throw std::invalid_argument("received non-positive number of rows");
```

src/core/algorithms/fd/afd_metric/afd_metric_calculator.cpp
```diff
@@ -210,4 +266,20 @@ long double AFDMetricCalculator::CalculateFI(model::PLIWS const* lhs_pli,
...
+config::ErrorType AFDMetricCalculator::CalculateZeroAryG1(ColumnData const* rhs,
+                                                          unsigned long long num_tuple_pairs) {
+    return 1 - rhs->GetPositionListIndex()->GetNepAsLong() /
+                       static_cast<config::ErrorType>(num_tuple_pairs);
+}
+
+config::ErrorType AFDMetricCalculator::CalculateG1Error(model::PLIWS const* lhs_pli,
+                                                        model::PLIWS const* joint_pli,
+                                                        unsigned long long num_tuple_pairs) {
+    // incorrect calculation, it uses number of pairs without repetitions, while it needed to be
```
wildsor: is this what "g1 zero-ary *(a variant specifically required by the Tane algorithm)*" is about? Give more context here in the comment

examples/expert/dedupe.py
```diff
@@ -4,7 +4,7 @@
 import pandas
 
 # Algorithm that finds approximate FDs and its config
-ALGORITHM_TYPE = desbordante.afd.algorithms.Default
+ALGORITHM_TYPE = desbordante.afd.algorithms.Pyro
```
wildsor, line 7: Change the example to use the AFD interface instead (perhaps you need an AFD->FD conversion method?)

src/tests/unit/test_fd_util.h
```diff
@@ -155,4 +155,12 @@ struct ApproximateDatasets<algos::EulerFD> {
...
 };
 
+inline std::vector<unsigned int> BitsetToIndexVector(boost::dynamic_bitset<> const& bitset) {
+    std::vector<unsigned int> res;
+    for (size_t index = bitset.find_first(); index != boost::dynamic_bitset<>::npos;
+         index = bitset.find_next(index)) {
+        res.push_back(index);
+    }
+    return res;
+}
```
wildsor: already available as util::BitsetToIndices

src/tests/unit/test_fd_algorithm.cpp
```diff
@@ -151,14 +142,13 @@ TYPED_TEST_P(AlgorithmTest, MaxLHSOptionWork) {
...
 }
 
-REGISTER_TYPED_TEST_SUITE_P(AlgorithmTest, ThrowsOnEmpty, ReturnsEmptyOnSingleNonKey,
-                            WorksOnLongDataset, WorksOnWideDataset, LightDatasetsConsistentHash,
-                            HeavyDatasetsConsistentHash, ConsistentRepeatedExecution,
-                            MaxLHSOptionWork);
+REGISTER_TYPED_TEST_SUITE_P(AlgorithmTest, MaxLHSOptionWork, ThrowsOnEmpty,
+                            ReturnsEmptyOnSingleNonKey, WorksOnLongDataset, WorksOnWideDataset,
+                            LightDatasetsConsistentHash, HeavyDatasetsConsistentHash,
+                            ConsistentRepeatedExecution);
```
wildsor: Order changed for no reason

src/tests/unit/test_afd_algorithm.cpp
```diff
@@ -0,0 +1,164 @@
...
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
+#include "core/algorithms/fd/depminer/depminer.h"
+#include "core/algorithms/fd/dfd/dfd.h"
+#include "core/algorithms/fd/fastfds/fastfds.h"
+#include "core/algorithms/fd/fdep/fdep.h"
+#include "core/algorithms/fd/fun/fun.h"
+#include "core/algorithms/fd/hyfd/hyfd.h"
+#include "core/algorithms/fd/pyro/pyro.h"
```
wildsor: Extra includes

src/tests/unit/CMakeLists.txt
```diff
@@ -397,6 +390,44 @@ desbordante_add_test(
...
+    ${DESBORDANTE_PREFIX}::model::table
+    ${DESBORDANTE_PREFIX}::model::types
+    ${DESBORDANTE_PREFIX}::testlib::common
+    ${DESBORDANTE_PREFIX}::fd::depminer
+    ${DESBORDANTE_PREFIX}::fd::dfd
+    ${DESBORDANTE_PREFIX}::fd::fastfds
+    ${DESBORDANTE_PREFIX}::fd::fdep
+    ${DESBORDANTE_PREFIX}::fd::fun
+    ${DESBORDANTE_PREFIX}::fd::hy
+    ${DESBORDANTE_PREFIX}::fd::pyro
```
wildsor: Extra targets

src/tests/unit/CMakeLists.txt
```diff
@@ -397,6 +390,44 @@ desbordante_add_test(
...
+    ${DESBORDANTE_PREFIX}::fd::tane::pfd
+    ${DESBORDANTE_PREFIX}::fd::tane
+    ${DESBORDANTE_PREFIX}::fd
```
wildsor: Possibly extra as well?

src/tests/unit/CMakeLists.txt
```diff
@@ -276,6 +266,7 @@ desbordante_add_test(
     ${DESBORDANTE_PREFIX}::model::types
     ${DESBORDANTE_PREFIX}::testlib::common
     ${DESBORDANTE_PREFIX}::fd
+    ${DESBORDANTE_PREFIX}::afd
```
wildsor: Why are these added?

src/python_bindings/CMakeLists.txt
```diff
@@ -159,6 +159,18 @@ desbordante_add_bind(
...
 )
 
+desbordante_add_bind(
+    afd
+    SRCS
+    fd/bind_afd.cpp
+    LIBS
+    Desbordante::fd::tane
+    Desbordante::fd::tane::pfd
+    Desbordante.bindlib.util
```
wildsor:
```suggestion
    ${DESBORDANTE_PREFIX}::fd::tane
    ${DESBORDANTE_PREFIX}::fd::tane::pfd
    ${DESBORDANTE_PREFIX}::bindlib::util
```

src/python_bindings/fd/bind_fd.cpp
```diff
@@ -73,24 +75,20 @@ void BindFd(py::module_& main_module) {
...
         auto algos_module = main_module.def_submodule(name).def_submodule("algorithms");
         for (auto algo_name : algorithms) {
             algos_module.attr(algo_name) = fd_algos_module.attr(algo_name);
         }
-        algos_module.attr("Default") = algos_module.attr(algorithms.front());
     };
 
-    define_submodule("afd", {kPyroName, kTaneName});
-    define_submodule("pfd", {kPFDTaneName});
+    define_submodule("afd", {kPyroName});
```
wildsor: Take all of this out and bind Pyro in bind_afd.cpp (I think it's a `detail::RegisterAlgorithm` call, but see the `BindPrimitive` code to be sure)

src/core/algorithms/fd/pli_based_afd_algorithm.h
```diff
@@ -0,0 +1,35 @@
...
+    std::shared_ptr<ColumnLayoutRelationData> relation_;
+
+    void LoadDataInternal() final;
+
+    ColumnLayoutRelationData const& GetRelation() const noexcept {
+        // GetRelation should be called after the dataset has been parsed, i.e. after algorithm
+        // execution
+        assert(relation_ != nullptr);
+        return *relation_;
+    }
```
wildsor: I don't think this method is used anymore, you can take it out.

src/core/algorithms/fd/CMakeLists.txt
```diff
@@ -15,13 +14,36 @@ set(SUBDIRS
...
     add_subdirectory(${dir})
 endforeach()
 
+# --- AFD ---
+set(NAME afd)
+desbordante_add_lib(NAME OBJECT)
+target_sources(
+    ${NAME}
+    PRIVATE afd_algorithm.cpp
+    PUBLIC FILE_SET HEADERS BASE_DIRS "${PROJECT_SOURCE_DIR}/src"
```
wildsor: extra line

```suggestion
```

src/core/algorithms/fd/CMakeLists.txt
```diff
@@ -15,13 +14,36 @@ set(SUBDIRS
...
+    ${NAME} PRIVATE ${DESBORDANTE_PREFIX}::model::table ${DESBORDANTE_PREFIX}::algos Boost::headers
+)
+
+# --- PliBasedAFD ---
+set(NAME afd.pli)
+desbordante_add_lib(NAME OBJECT)
+target_sources(
+    ${NAME}
+    PRIVATE pli_based_afd_algorithm.cpp
+    PUBLIC FILE_SET HEADERS BASE_DIRS "${PROJECT_SOURCE_DIR}/src"
```
wildsor: extra line
```suggestion
```

src/core/algorithms/fd/tane/tane_common.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "core/algorithms/fd/pli_based_afd_algorithm.h"
 #include "core/algorithms/fd/pli_based_fd_algorithm.h"
```
wildsor: extra include

src/core/algorithms/fd/tane/tane_common.h
```diff
@@ -26,7 +27,9 @@ class TaneCommon : public PliBasedFDAlgorithm {
                                                model::PLIWS const* joint_pli) = 0;
     static double CalculateUccError(model::PositionListIndex const* pli,
                                     ColumnLayoutRelationData const* relation_data);
+
     void RegisterAndCountFd(Vertical lhs, Column const* rhs);
+    void RegisterAndCountAfd(Vertical const& lhs, Column const* rhs, long double threshold);
```
wildsor: Get rid of these methods

wildsor (changes requested):

src/core/algorithms/fd/afd_metric/afd_metric_calculator.cpp
Sneper-Breeze: fixed

src/tests/unit/test_fd_util.h
Sneper-Breeze: replaced

src/tests/unit/test_fd_algorithm.cpp
Sneper-Breeze: reordered

src/tests/unit/test_afd_algorithm.cpp
Sneper-Breeze: deleted

src/tests/unit/CMakeLists.txt
Sneper-Breeze: deleted

src/tests/unit/CMakeLists.txt
Sneper-Breeze: extra. deleted

src/python_bindings/CMakeLists.txt
Sneper-Breeze: applied

src/python_bindings/fd/bind_fd.cpp
Sneper-Breeze: There is some issue with Pyro, Pyro now is not a `afd_algorithm` class, so it provides only the exact FDs, and maybe it will confuse users. Because of it I returned the old interface with Pyro in afd, BUT now it is not defualt algo in afd module and needs to be called explicit with `afd.Pyro` .

src/core/algorithms/fd/pli_based_afd_algorithm.h
Sneper-Breeze: removed

src/core/algorithms/fd/CMakeLists.txt
Sneper-Breeze: deleted

src/core/algorithms/fd/CMakeLists.txt
Sneper-Breeze: deleted

src/core/algorithms/fd/tane/tane_common.h
Sneper-Breeze: deleted

src/core/algorithms/fd/tane/tane_common.h
Sneper-Breeze: removed.

wildsor:
Instead of replying to every comment, just resolve the obvious ones (delete a line, reorder things, etc.).

wildsor:
Remember to press "re-request review" once you're ready

src/tests/unit/CMakeLists.txt
Sneper-Breeze: ::fd and ::afd are both required

Sneper-Breeze:
add explicit pickle definition for AFD class binding, because bug with it was found

AFD was bound as py::class_<AFD, FD> but without its own pickle.
When pickle.loads ran, it fell through to FD's __setstate__, which
reconstructs a plain FD (wrong dynamic type, threshold_ lost).
pybind11's holder machinery then aborted because the C++ type
didn't match the Python class. The explicit py::pickle here
serialises threshold_ as a 4th tuple element and reconstructs via
AFD(lhs, rhs, threshold, schema), so round-tripping preserves both
the type and the threshold value.

src/tests/unit/test_fd_util.h
```diff
@@ -8,6 +8,7 @@
 #include "core/algorithms/fd/pyrocommon/core/parameters.h"
 #include "core/config/error/type.h"
 #include "core/config/names.h"
+#include "core/util/bitset_utils.h"
```
wildsor: I think you meant to add this to test_fd_algorithm.cpp

src/tests/unit/test_afd_algorithm.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
+#include "core/algorithms/fd/pyro/pyro.h"
+#include "core/algorithms/fd/tane/pfdtane.h"
+#include "core/algorithms/fd/tane/tane.h"
+#include "core/model/table/relational_schema.h"
+#include "tests/unit/test_afd_util.h"
+#include "tests/unit/test_fd_util.h"
```
wildsor:
```suggestion
#include "core/algorithms/fd/tane/pfdtane.h"
#include "core/algorithms/fd/tane/tane.h"
#include "core/model/table/relational_schema.h"
#include "tests/unit/test_afd_util.h"
```

src/tests/unit/CMakeLists.txt
wildsor: Could it be because of extra includes in `test_afd_algorithm.cpp`? Remove the headers that are not required and try again

src/core/algorithms/fd/CMakeLists.txt
```diff
@@ -15,13 +14,28 @@ set(SUBDIRS
     pyro
     pyrocommon
     sfd
+    afd_metric
```
wildsor: line moved for no reason

src/core/algorithms/fd/CMakeLists.txt
```diff
@@ -15,13 +14,28 @@ set(SUBDIRS
...
 foreach(dir IN LISTS SUBDIRS)
     add_subdirectory(${dir})
 endforeach()
 
+# --- AFD ---
+set(NAME afd)
+desbordante_add_lib(NAME OBJECT)
+target_sources(${NAME} PRIVATE afd_algorithm.cpp)
+target_link_libraries(
+    ${NAME} PRIVATE ${DESBORDANTE_PREFIX}::model::table ${DESBORDANTE_PREFIX}::algos Boost::headers
```
wildsor: You have algorithm.h included as a header, so the targets that link to this one need it too, use `PUBLIC`. Same is true for the other two targets here, but it should be PUBLIC in the `${DESBORDANTE_PREFIX}::algos` definition for a similar reason, so whatever.

```suggestion
    ${NAME} PUBLIC ${DESBORDANTE_PREFIX}::algos PRIVATE ${DESBORDANTE_PREFIX}::model::table Boost::headers
```

src/core/algorithms/fd/CMakeLists.txt
```diff
@@ -15,13 +14,28 @@ set(SUBDIRS
...
+)
+
+# --- PliBasedAFD ---
+set(NAME afd.pli)
+desbordante_add_lib(NAME OBJECT)
+target_sources(${NAME} PRIVATE pli_based_afd_algorithm.cpp)
+target_link_libraries(${NAME} PRIVATE ${DESBORDANTE_PREFIX}::afd Boost::headers)
```
wildsor: Same as above

```suggestion
target_link_libraries(${NAME} PUBLIC ${DESBORDANTE_PREFIX}::afd PRIVATE Boost::headers)
```

src/core/algorithms/fd/CMakeLists.txt
```diff
@@ -30,6 +44,7 @@ target_link_libraries(
     ${NAME}
     PUBLIC ${DESBORDANTE_PREFIX}::config
     PRIVATE ${DESBORDANTE_PREFIX}::model::table ${DESBORDANTE_PREFIX}::algos Boost::headers
+            ${DESBORDANTE_PREFIX}::afd
```
wildsor: No, that definitely looks wrong, neither of fd.cpp and fd_algorithm.cpp need to link AFDs
```suggestion
```

src/core/algorithms/fd/afd_algorithm.h
```diff
@@ -0,0 +1,93 @@
...
+        if (result.back() == ',') {
+            result.erase(result.size() - 1);
+        }
+        result += "]}";
+        return result;
+    }
+
+public:
+    constexpr static std::string_view kDefaultPhaseName = "FD mining";
+
```
wildsor: `Progress` has been removed a while ago, this is not needed

```suggestion
```

src/core/algorithms/fd/afd_algorithm.h
```diff
@@ -0,0 +1,93 @@
...
+        return afd_collection_.AsList();
+    }
+
+    std::list<AFD>& AfdList() noexcept {
+        return afd_collection_.AsList();
+    }
+
+    std::list<AFD>& SortedAfdList();
+
+    // считает контрольную сумму Флетчера - нужно для тестирования по хешу
```
wildsor: Remove the comment, this method is getting removed anyway
```suggestion
```

src/core/algorithms/fd/afd_algorithm.h
```diff
@@ -0,0 +1,93 @@
...
+    /* Collection of all discovered AFDs
+     * Every AFD mining algorithm should place discovered dependencies here. Don't add new AFDs by
+     * accessing this field directly, use RegisterAfd methods instead
+     */
+    util::PrimitiveCollection<AFD> afd_collection_;
+
+    /* Registers new AFD.
+     * Should be overridden if custom behavior is needed
+     */
+    virtual void RegisterAfd(Vertical lhs, Column rhs, long double threshold,
```
wildsor: No, let's not complicate things, no overrides unless required

```suggestion
    /* Registers new AFD.
     * Should be overridden if custom behavior is needed
     */
    void RegisterAfd(Vertical lhs, Column rhs, long double threshold,
```

src/core/algorithms/fd/afd_algorithm.h
```diff
@@ -0,0 +1,93 @@
...
+                             std::shared_ptr<RelationalSchema const> const& schema) {
+        if (lhs.GetArity() <= max_lhs_)
+            afd_collection_.Register(std::move(lhs), std::move(rhs), std::move(threshold), schema);
+    }
+
+    virtual void RegisterAfd(AFD afd_to_register) {
```
wildsor:
```suggestion
    void RegisterAfd(AFD afd_to_register) {
```

src/core/algorithms/fd/afd_algorithm.h
```diff
@@ -0,0 +1,93 @@
...
+    util::PrimitiveCollection<AFD> afd_collection_;
+
+    /* Registers new AFD.
+     * Should be overridden if custom behavior is needed
+     */
+    virtual void RegisterAfd(Vertical lhs, Column rhs, long double threshold,
+                             std::shared_ptr<RelationalSchema const> const& schema) {
+        if (lhs.GetArity() <= max_lhs_)
+            afd_collection_.Register(std::move(lhs), std::move(rhs), std::move(threshold), schema);
+    }
```
wildsor: Remove this method, it's literally just saving a few symbols over the other one

src/core/algorithms/fd/afd_algorithm.h
```diff
@@ -0,0 +1,93 @@
...
+
+namespace model {
+class AgreeSetFactory;
+}
+
+namespace algos {
+
+/* It is highly recommended to inherit your Algorithm from this class.
+ * Consider TANE as an example of such an AFDAlgorithm usage.
+ * */
```
wildsor: I think providing a class purely for storage and a few options and naming it *Pattern*Algorithm is a bad idea because it looks like a layer of helpful abstraction, so it should be removed, but you couldn't have known that when making this class, so just remove the comment.

```suggestion
```

src/core/algorithms/fd/afd_algorithm.h
```diff
@@ -0,0 +1,93 @@
...
+    virtual void MakeExecuteOptsAvailableFDInternal() {};
+    void MakeExecuteOptsAvailable() override;
+    virtual void ResetStateFd() = 0;
+
+protected:
+    config::MaxLhsType max_lhs_;
+
+    /* Collection of all discovered AFDs
+     * Every AFD mining algorithm should place discovered dependencies here. Don't add new AFDs by
+     * accessing this field directly, use RegisterAfd methods instead
```
wildsor: Same as above, there is no "should"

```suggestion
     * Don't add new AFDs by accessing this field directly, use RegisterAfd methods instead
```

examples/expert/dedupe.py
wildsor: So?

src/core/algorithms/fd/tane/CMakeLists.txt
```diff
@@ -3,17 +3,22 @@ set(NAME fd.tane.common)
 desbordante_add_lib(NAME OBJECT)
 target_sources(${NAME} PRIVATE model/lattice_level.cpp model/lattice_vertex.cpp tane_common.cpp)
 target_link_libraries(
-    ${NAME} PRIVATE spdlog::spdlog_header_only ${DESBORDANTE_PREFIX}::fd::pli Boost::headers
+    ${NAME}
+    PRIVATE spdlog::spdlog_header_only ${DESBORDANTE_PREFIX}::fd::pli Boost::headers
+            ${DESBORDANTE_PREFIX}::afd ${DESBORDANTE_PREFIX}::afd::metric_calc
+            ${DESBORDANTE_PREFIX}::afd::pli
```
wildsor: No, that doesn't look right. Find out which parts of the library are in the headers, make them public, then try removing the extra targets from here.

src/core/algorithms/fd/afd_algorithm.h
```diff
@@ -0,0 +1,93 @@
...
+
+namespace algos {
+
+/* It is highly recommended to inherit your Algorithm from this class.
+ * Consider TANE as an example of such an AFDAlgorithm usage.
+ * */
+class AFDAlgorithm : public Algorithm {
+private:
+    friend model::AgreeSetFactory;
+
```
wildsor: Remove this
```suggestion
```

src/core/algorithms/fd/afd_algorithm.h
```diff
@@ -0,0 +1,93 @@
...
+#include "core/algorithms/algorithm.h"
+#include "core/algorithms/fd/afd.h"
+#include "core/config/max_lhs/type.h"
+#include "core/util/primitive_collection.h"
+
+namespace model {
+class AgreeSetFactory;
+}
+
```
wildsor: Remove this
```suggestion

```

src/core/algorithms/fd/pli_based_afd_algorithm.cpp
```diff
@@ -0,0 +1,26 @@
+#include "core/algorithms/fd/pli_based_afd_algorithm.h"
+
+#include "core/config/equal_nulls/option.h"
```
wildsor:
```suggestion
```

src/core/algorithms/fd/pli_based_afd_algorithm.cpp
```diff
@@ -0,0 +1,26 @@
...
+#include "core/config/tabular_data/input_table/option.h"
+
+namespace algos {
+
+PliBasedAFDAlgorithm::PliBasedAFDAlgorithm() : AFDAlgorithm() {
+    RegisterOptions();
+    MakeOptionsAvailable({config::kTableOpt.GetName(), config::kEqualNullsOpt.GetName()});
```
wildsor:
```suggestion
    MakeOptionsAvailable({config::kTableOpt.GetName()});
```

src/core/algorithms/fd/pli_based_afd_algorithm.cpp
```diff
@@ -0,0 +1,26 @@
...
+}
+
+void PliBasedAFDAlgorithm::RegisterOptions() {
+    RegisterOption(config::kTableOpt(&input_table_));
+    RegisterOption(config::kEqualNullsOpt(&is_null_equal_null_));
```
wildsor:
```suggestion
```

src/core/algorithms/fd/pli_based_afd_algorithm.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+
+#include "core/algorithms/fd/afd_algorithm.h"
+#include "core/config/equal_nulls/type.h"
```
wildsor:
```suggestion
```

src/core/algorithms/fd/pli_based_afd_algorithm.h
```diff
@@ -0,0 +1,26 @@
...
+#include "core/config/tabular_data/input_table_type.h"
+#include "core/model/table/column_layout_relation_data.h"
+
+namespace algos {
+
+class PliBasedAFDAlgorithm : public AFDAlgorithm {
+private:
+    config::InputTable input_table_;
+    config::EqNullsType is_null_equal_null_;
```
wildsor:
```suggestion
```

src/core/algorithms/fd/pli_based_afd_algorithm.h
```diff
@@ -0,0 +1,26 @@
...
+
+    void RegisterOptions();
+
+protected:
+    std::shared_ptr<ColumnLayoutRelationData> relation_;
+
+    void LoadDataInternal() final;
```
wildsor: Should be private

wildsor (changes requested):

examples/expert/dedupe.py
Sneper-Breeze: I described problem with it in the comment below. 
Reapiting
`
There is some issue with Pyro, Pyro now is not a afd_algorithm class, so it provides only the exact FDs, and maybe it will confuse users. Because of it I returned the old interface with Pyro in afd, BUT now it is not defualt algo in afd module and needs to be called explicit with afd.algorithms.Pyro .
`

wildsor: Mhm, my idea was that Pyro would have the AFD interface eventually, so it makes sense to remake the example for that, but it's not going to

src/tests/unit/CMakeLists.txt
wildsor: > Could it be because of extra includes in `test_afd_algorithm.cpp`? Remove the headers that are not required and try again

Did it fail?

src/core/algorithms/fd/CMakeLists.txt
```diff
@@ -16,12 +15,33 @@ set(SUBDIRS
...
     add_subdirectory(${dir})
 endforeach()
 
+# --- AFD ---
+set(NAME afd)
+desbordante_add_lib(NAME OBJECT)
+target_sources(${NAME} PRIVATE afd_algorithm.cpp)
+target_link_libraries(
+    ${NAME}
+    PUBLIC ${DESBORDANTE_PREFIX}::algos
```
wildsor: Given that FD is a parent of AFD, maybe you can try
```suggestion
    PUBLIC ${DESBORDANTE_PREFIX}::algos ${DESBORDANTE_PREFIX}::fd
```

wildsor (approved):
I think this is mostly okay, but look closer at the FD target being linked where it shouldn't be

wildsor:
Oh, and maybe get the commits so that they show what happened here, not how this was developed, squash the fixes.

PR #689, opened 2026-03-05
Remove RV policy argument in BindPrimitive

wildsor:
That argument was added to solve a historical problem where pattern objects depended on algorithm lifetime. It is getting erroneously set to an incorrect value by people copying code but is irrelevant for new patterns.


ol-imorozko (approved):
lgtm

PR #693, opened 2026-03-08
Set hidden visibility, add build target for afd_metric and link it correctly

wildsor:
Hidden visibility enables some optimizations and reduces binary size. In
exchange, we have to manually specify what symbols should be made
public. See https://gcc.gnu.org/wiki/Visibility for details.

For our case, what matters is that we need to share the `type_info`
objects across shared object boundaries: the Option object compares
`type_info`s to alert the user with a proper error message that contains
the option name in case the value type is incorrect. Otherwise, the
comparison would have been done by `boost::any`'s cast method that would
fail with a less comprehensible error message.

Hidden visibility was already set on all pybind11 libraries, for the
reasoning see the -Wattributes section in
https://pybind11.readthedocs.io/en/stable/faq.html.

Previously when compiling bindings we only had to deal with a single
bindings library and the `type_info` issues did not cause trouble as
the corresponding symbols were merged together. Currently, the bindings
consist of several libraries, which appears to have changed linker
behavior on MacOS.
See https://github.com/Desbordante/desbordante-core/pull/652#issuecomment-3944962633 for
details. Now, if we do not specify default visibility for the
non-standard types that are used as algorithm options, the conversions
from Python will fail in the library built on MacOS as the `type_info`s
from different libraries would not compare equal.

Basically, now we have to explicitly specify default visibility for the
types used as options even though the library is not built with hidden
visibility. So we get the troubles, but not the benefits. Thus, we
might as well set hidden visibility for the whole project to get the
benefits too.

The visibility issues are also relevant for exception types, see the
"Problems with C++ exceptions (please read!)" section in the GCC wiki
above. Thus, this commit adds the visibility specifier for the
configuration exception.

### DIFF
--- a/src/core/algorithms/fd/afd_metric/CMakeLists.txt
+++ b/src/core/algorithms/fd/afd_metric/CMakeLists.txt
@@ -1,0 +1,13 @@
+# --- AFD metrics ---
+set(NAME fd.afd_metric)
+desbordante_add_lib(NAME OBJECT)
+target_sources(
+    ${NAME}
+    PRIVATE afd_metric_calculator.cpp
+    PUBLIC FILE_SET HEADERS BASE_DIRS "${PROJECT_SOURCE_DIR}/src"
+)
+target_link_libraries(
+    ${NAME}
+    PUBLIC better-enums
+    PRIVATE Boost::headers
+)
--- a/src/python_bindings/CMakeLists.txt
+++ b/src/python_bindings/CMakeLists.txt
@@ -71,6 +71,11 @@ desbordante_add_bind(
 )
 
 desbordante_add_bind(
+    afd_metric SRCS afd_metric/bind_afd_metric_calculation.cpp LIBS Desbordante::fd::afd_metric
+    Boost::headers
+)
+
+desbordante_add_bind(
     ar
     SRCS
     ar/bind_ar.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -108,6 +108,10 @@ set(CMAKE_COLOR_DIAGNOSTICS ON)
 set(CMAKE_OPTIMIZE_DEPENDENCIES ON)
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/target")
 
+# Set hidden visibility for object symbols for smaller and faster loading binaries
+set(CMAKE_CXX_VISIBILITY_PRESET hidden)
+set(CMAKE_VISIBILITY_INLINES_HIDDEN TRUE)
+
 if(DESBORDANTE_BUILD_NATIVE)
     string(JOIN ";" BUILD_OPTS "${BUILD_OPTS}" "-march=native")
 endif()
--- a/src/core/algorithms/fd/CMakeLists.txt
+++ b/src/core/algorithms/fd/CMakeLists.txt
@@ -1,4 +1,5 @@
 set(SUBDIRS
+    afd_metric
     aidfd
     depminer
     dfd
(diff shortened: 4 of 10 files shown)


src/core/algorithms/fd/afd_metric/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+# --- AFD metrics ---
+set(NAME fd.afd_metric)
+desbordante_add_lib(NAME OBJECT)
+target_sources(
+    ${NAME}
+    PRIVATE afd_metric_calculator.cpp
+    PUBLIC FILE_SET HEADERS BASE_DIRS "${PROJECT_SOURCE_DIR}/src"
+)
+target_link_libraries(${NAME} PRIVATE better-enums Boost::headers)
```
p-senichenkov: Do we need boost headers here?

wildsor: Yeah, they are transitively included with `algorithm.h`. Should probably be linked public or something. It fails on apple-clang without this.

wildsor: algorithm.h uses `boost::any`

p-senichenkov: So... This means that **every** target must link not only to its own dependencies, but also to dependencies of all its dependencies, and all their dependencies, and so on. Do we really want our build system to be so dumb?
Why do we need separate `create_algo` (not linked to boost, and without sources) and `algos` (with boost) targets?

p-senichenkov: Of course, this comment isn't related to this PR. This PR looks good to me (I don't fully understand the new build system, so I'll wait for someone else to approve it)

wildsor: Yeah, that's an issue, we'll need to have it fixed. The person doing it couldn't continue, so we merged it as-is.

polyntsov: Better enum is used in this header and thus is part of the interface, so it should be linked using PUBLIC

src/python_bindings/CMakeLists.txt
```diff
@@ -70,6 +70,11 @@ desbordante_add_bind(
     Boost::headers
 )
 
+desbordante_add_bind(
+    afd_metric SRCS afd_metric/bind_afd_metric_calculation.cpp LIBS Desbordante::fd::afd_metric
+    Boost::headers
+)
```
p-senichenkov, line 76: Other bindings are styled a bit differently. Does `cmake-format` give such result?

I think, `desbordante_add_bind` should be added to [`cmake-format` custom macro list](https://github.com/cheshirekow/cmake_format?tab=readme-ov-file#custom-commands), along with `desbordante_add_test` and others

p-senichenkov (approved):

src/python_bindings/CMakeLists.txt
wildsor: > Other bindings are styled a bit differently. Does `cmake-format` give such result?

Yes

p-senichenkov: Ok, this means that we need to add annotations (in `.cmake-format.yaml`) for all custom macros, so that they'll be formatted like built-in ones:
```CMake
macro(
    positional_arg1 positional_arg2
    ARG_NAME arg1 arg2 ...
    ARG_NAME2 ...
)
```
It shouldn't be very hard, but I don't have much time now.
Could you please do this? If no, let's create an issue

PR #695, opened 2026-03-11
Implement parallel FastADC

ol-imorozko:


src/core/algorithms/dc/FastADC/model/column_operand.h
```diff
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <boost/container_hash/hash.hpp>
```
polyntsov, line 3: should be in cpp file?

src/core/algorithms/dc/FastADC/model/evidence.h
```diff
@@ -1,27 +1,27 @@
...
 
 namespace algos::fastadc {
 
 struct Evidence {
     int64_t count;
     PredicateBitset evidence;
 
-    Evidence(Clue satisfied, int64_t count, PredicateBitset const& cardinalityMask,
+    Evidence(PredicateBitset const& satisfied, int64_t count,
+             PredicateBitset const& cardinalityMask,
```
polyntsov: Shouldn't local vars be snake case?

src/core/algorithms/dc/FastADC/model/operator.cpp
```diff
@@ -26,4 +26,8 @@ bool Operator::Eval(std::byte const* v1, std::byte const* v2, model::Type const&
     }
 }
 
+size_t hash_value(Operator const& k) noexcept {
+    return static_cast<size_t>(k.GetType());
+}
```
polyntsov, line 31: Is it used somewhere?

src/core/algorithms/dc/FastADC/util/denial_constraint_set.h
```diff
@@ -123,7 +141,7 @@ class DenialConstraintSet {
     }
 
     void Clear() {
-        constraints_.clear();
+        constraints_ = decltype(constraints_)(0, DCHash(nullptr), DCEqual(nullptr));
```
polyntsov: nit: `using` for the constraints type would be cleaner

src/core/algorithms/dc/FastADC/util/clue_set_builder.h
```diff
@@ -3,9 +3,13 @@
...
 #include "core/algorithms/dc/FastADC/util/common_clue_set_builder.h"
 #include "core/algorithms/dc/FastADC/util/evidence_aux_structures_builder.h"
+#include "core/util/worker_thread_pool.h"
 
 namespace algos::fastadc {
 
 ClueSet BuildClueSet(std::vector<PliShard> const& pliShards, PredicatePacks const& packs);
 
+ClueSet BuildClueSetParallel(std::vector<PliShard> const& pliShards, PredicatePacks const& packs,
+                             util::WorkerThreadPool* thread_pool);
```
polyntsov: naming conventions :)

src/core/algorithms/dc/FastADC/util/evidence_set_builder.h
```diff
@@ -13,8 +14,10 @@ class EvidenceSetBuilder {
 public:
     EvidenceSet evidence_set;
 
-    EvidenceSetBuilder(std::vector<PliShard> const& pli_shards, PredicatePacks const& packs) {
-        clue_set_ = BuildClueSet(pli_shards, packs);
+    EvidenceSetBuilder(std::vector<PliShard> const& pli_shards, PredicatePacks const& packs,
+                       util::WorkerThreadPool* thread_pool = nullptr) {
+        clue_set_ = thread_pool ? BuildClueSetParallel(pli_shards, packs, thread_pool)
```
polyntsov: Why expecting thread_pool as a pointer in `BuildClueSetParallel`, if here it's always non-null?

src/core/algorithms/dc/FastADC/util/clue_set_builder.cpp
```diff
@@ -45,4 +54,101 @@ ClueSet BuildClueSet(std::vector<PliShard> const& pliShards, PredicatePacks cons
...
+                    throw std::runtime_error("TaskIdToIndices produced out-of-bounds shard index");
+                }
+
+                buffers.task_clue_set.clear();
+
+                if (i == j) {
+                    SingleClueSetBuilder{pliShards[i]}.BuildClueSet(packs, buffers.forward_clues,
+                                                                    buffers.task_clue_set);
+                } else {
+                    CrossClueSetBuilder{pliShards[i], pliShards[j]}.BuildClueSet(
```
polyntsov: there is no false sharing in pli shards or packs, right?

src/core/algorithms/dc/FastADC/util/clue_set_builder.h
```diff
@@ -9,7 +11,25 @@ namespace algos::fastadc {
 
 ClueSet BuildClueSet(std::vector<PliShard> const& pliShards, PredicatePacks const& packs);
 
+ClueSetT<uint8_t> BuildClueSet8(std::vector<PliShard> const& pliShards, PredicatePacks const& packs);
+ClueSetT<uint16_t> BuildClueSet16(std::vector<PliShard> const& pliShards, PredicatePacks const& packs);
+ClueSetT<uint32_t> BuildClueSet32(std::vector<PliShard> const& pliShards, PredicatePacks const& packs);
+ClueSetT<uint64_t> BuildClueSet64(std::vector<PliShard> const& pliShards, PredicatePacks const& packs);
```
polyntsov: hm, why do we use int values and use all these overloads and branching instead of std::bitset parametrized by the number of bits? Is uint64_t indeed a lot more efficient than std::bitset<64>?

polyntsov: Also what is the performance improvement introduced by this commit?

ol-imorozko: > Also what is the performance improvement introduced by this commit?

A lot, like x2 time, but not because of uint64_t faster than std::bitset<64> -- it's not. It's because previously I was just selecting std::bitset<128> as Clue type and clues were unable to fit one cache line etc etc..
This commit checks how many predicates we have on runtime and determines best fit type as u8/u16 etc, but yeah of course that should be just std::bitset with correct template argument

src/core/algorithms/dc/FastADC/model/column_operand.h
ol-imorozko: don't think so, `std::hash<ColumnOperand>` defined in header and it calls `boost::hash_combine`, so this include is needed

src/core/algorithms/dc/FastADC/model/operator.cpp
ol-imorozko: Yes, `boost::hash_combine(seed, k.GetOperator())` in `predicate.h` uses it

src/core/algorithms/dc/FastADC/util/clue_set_builder.cpp
ol-imorozko: no, they are only read in parallel, no writes

polyntsov (approved):

PR #699, opened 2026-03-14
Add SD validation

kamenkremen:
This pull request introduces the Sequential Dependencies (SD) verification functionality. It includes new algorithms and examples to demonstrate SD validation. The most important changes include adding the SD verifier algorithm, Python bindings, and an example script.

### New SD Verification Functionality:

*   [**`src/core/algorithms/sd/sd_verifier/sd_verifier.h`**](src/core/algorithms/sd/sd_verifier/sd_verifier.h): Introduced `SDVerifier` class to handle SD verification.
*   [**`src/core/algorithms/sd/sd_verifier/util/data_structures.h`**](src/core/algorithms/sd/sd_verifier/util/data_structures.h): Added custom `Fenwick` and `SegmentTree` structures to optimize confidence calculations.

### Python Bindings:

*   [**`src/python_bindings/sd/bind_sd_verification.cpp`**](src/python_bindings/sd/bind_sd_verification.cpp): Added bindings for the SD verifier to the Python module.

### Example:

*   [**`examples/basic/verifying_sd.py`**](examples/basic/verifying_sd.py): Added a script to demonstrate how to verify SDs using the Desbordante library, covering data loading, constraint definition, algorithm execution, and result interpretation.


src/core/algorithms/sd/sd_verifier/sd_verifier.cpp
```diff
@@ -0,0 +1,544 @@
...
+    return ops[n - 1];
+}
+
+unsigned long long SDVerifier::ExecuteInternal() {
+    auto start_time = std::chrono::steady_clock::now();
+
+    if (lhs_indices_.size() != 1 || rhs_indices_.size() != 1) {
+        throw std::runtime_error(
+                "SDVerifier currently supports exactly one lhs and one rhs column.");
+    }
```
polyntsov: Shouldn't this be option validation error?

src/core/algorithms/sd/sd_verifier/sd_verifier.cpp
```diff
@@ -0,0 +1,544 @@
...
+    if (g1_ < 0) {
+        throw std::runtime_error("g1 must be non-negative.");
+    }
+    if (g2_ >= 0 && g2_ < g1_) {
+        throw std::runtime_error("g2 must be >= g1 or negative.");
+    }
+    if (g2_ >= 0 && g2_ < 1e-9 && g1_ < 1e-9) {
+        throw std::runtime_error("g1 and g2 cannot both be zero.");
+    }
```
polyntsov: These are also checked during option registration?

polyntsov: I guess `g2_ < g1_` check is needed here because cross-parameter validation can't be done during option set time?

Although is `check_g2` not enough?

src/core/algorithms/sd/sd_verifier/sd_verifier.cpp
```diff
@@ -0,0 +1,544 @@
...
+        throw std::runtime_error(
+                "SDVerifier currently supports exactly one lhs and one rhs column.");
+    }
+    if (g1_ < 0) {
+        throw std::runtime_error("g1 must be non-negative.");
+    }
+    if (g2_ >= 0 && g2_ < g1_) {
+        throw std::runtime_error("g2 must be >= g1 or negative.");
+    }
+    if (g2_ >= 0 && g2_ < 1e-9 && g1_ < 1e-9) {
```
polyntsov: why `1e-9`?

src/core/algorithms/sd/sd_verifier/sd_verifier.cpp
```diff
@@ -0,0 +1,544 @@
...
+        }
+        remainders[i] = {remainder, i};
+    }
+    std::sort(remainders.begin(), remainders.end());
+
+    // Assign class IDs to elements based on their sorted order by remainder.
+    int cid = 0;
+    class_id[remainders[0].second] = cid;
+    for (size_t i = 1; i < n; ++i) {
+        if (remainders[i].first - remainders[i - 1].first > 1e-9) {
```
polyntsov: Which values do we expect here to compare? How big can be `g1_` (since remainders are in `[0; g1_)`)? Asking because comparing difference between doubles to predefined hard-coded epsilon doesn't make much sense in many cases (detailed explanation: https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/)

And the same question about `1e-9` hardcoded into comparison in the `if` just below and in `is_exact_gap`

src/core/algorithms/sd/sd_verifier/sd_verifier.cpp
```diff
@@ -0,0 +1,544 @@
...
+}
+
+// Validates predecessor range [l_rank, r_rank) from segment tree.
+// If valid path exists, updates min2_cost with the best predecessor.
+void SDVerifier::UpdateMin2Interval(double current_val, size_t i, size_t l_rank, size_t r_rank,
+                                    std::vector<double> const& unique_vals,
+                                    ds::SegmentTree const& tree, long& min2_cost,
+                                    long& min2_j) const {
+    auto res = tree.Query(l_rank, r_rank);
+    if (res.rank == static_cast<size_t>(-1)) {
```
polyntsov: Not the best interface for sentinel value. Something like `std::optional` is more readable

examples/basic/verifying_sd.py
polyntsov, line 1: I wonder why we don't change test snapshots after new test is added? Shouldn't example tests fail in ci? Or maybe it's just a warning

src/core/algorithms/sd/sd_verifier/sd_verifier.cpp
```diff
@@ -0,0 +1,544 @@
...
+    auto get_cols_num = [this]() {
+        return input_table_ ? static_cast<config::IndexType>(input_table_->GetNumberOfColumns())
+                            : 0;
+    };
+
+    RegisterOption(config::kLhsIndicesOpt(&lhs_indices_, get_cols_num));
+    RegisterOption(config::kRhsIndicesOpt(&rhs_indices_, get_cols_num));
+    RegisterOption(
+            Option<config::IndicesType>{&indices_, kSdIndices, kDSdIndices}.SetIsRequiredFunc(
+                    []() { return false; }));
```
polyntsov: Weird thing to do

I think it should be option with `calculate_default` set?

polyntsov (changes requested):

src/core/algorithms/sd/sd_verifier/sd_verifier.cpp
kamenkremen: i think `check_g2` is not enough because it only validates `g2_` against the current value of `g1_` at the moment `g2_` is set. If `g1_` is changed afterwards, the pair can become invalid

src/core/algorithms/sd/sd_verifier/sd_verifier.cpp
kamenkremen: Implemented the `AlmostEqual` function from the article to compare doubles

src/core/algorithms/sd/sd_verifier/sd_verifier.cpp
polyntsov: The interface for options is not the most obvious one, but I think it should be possible to confine this option validation logic to the options themselves where it logically belongs

As far as I understand, it's possible to set for example `g2` as conditional option for `g1` via `g1.SetConditionalOpts()` and this defines the order of validation checks for the options -- `g2` will be check strictly after `g1` is set.

Could you please ask @BUYT-1 about this?

wildsor: See MetricVerifier for an example

kamenkremen: Moved `g1`/`g2` validation logic to the option registration. Also moved indices validation there

src/core/algorithms/sd/sd_verifier/sd_verifier.cpp
kamenkremen: I removed it and now use AlmostEqual for these comparisons instead

polyntsov (approved):
Thanks!

PR #701, opened 2026-03-17
Add FDHits mining algorithm

Godrik0:
This pull request introduces a new functional dependency discovery algorithm, `fdhits`. The main changes include adding the `fdhits` module, implementing its core algorithm and supporting components, and integrating it into the build system.

### DIFF
--- a/src/core/algorithms/fd/fdhits/CMakeLists.txt
+++ b/src/core/algorithms/fd/fdhits/CMakeLists.txt
@@ -1,0 +1,19 @@
+set(NAME fd.fdhits)
+desbordante_add_lib(NAME)
+target_sources(
+    ${NAME}
+    PRIVATE fdhits.cpp
+            pli_table.cpp
+            treesearch/dynamic_edgemark_store.cpp
+            treesearch/dynamic_edgemark_store_multi.cpp
+            treesearch/treesearch_de.cpp
+            treesearch/treesearch_multi.cpp
+    PUBLIC FILE_SET HEADERS BASE_DIRS "${PROJECT_SOURCE_DIR}/src"
+)
+target_link_libraries(
+    ${NAME}
+    PRIVATE ${DESBORDANTE_PREFIX}::fd::hy::common
+            ${DESBORDANTE_PREFIX}::fd::pli
+            spdlog::spdlog_header_only
+            Boost::headers
+)
--- a/src/core/algorithms/fd/fdhits/edgecollectors/edge_hash_collector.h
+++ b/src/core/algorithms/fd/fdhits/edgecollectors/edge_hash_collector.h
@@ -1,0 +1,53 @@
+#pragma once
+
+#include <algorithm>
+#include <vector>
+
+#include <boost/unordered_set.hpp>
+
+#include "core/algorithms/fd/fdhits/edges/default_edge.h"
+#include "core/algorithms/fd/fdhits/edges/multi_fd.h"
+#include "core/algorithms/fd/fdhits/treesearch/hypergraph.h"
+
+namespace algos::fd::fdhits {
+
+template <typename E>
+class EdgeHashCollector {
+private:
+    boost::unordered_set<E, typename E::Hash> edges_;
+
+public:
+    void Consume(E const& edge) {
+        edges_.insert(edge);
+    }
+
+    auto const& GetEdgeSet() const {
+        return edges_;
+    }
+
+    bool IsEmpty() const {
+        return edges_.empty();
+    }
+
+    Hypergraph<E> ToGraph(E const& s) {
+        Hypergraph<E> graph(s.Size());
+
+        std::vector<E> sorted_edges;
+        sorted_edges.reserve(edges_.size());
+        for (auto const& edge : edges_) {
+            sorted_edges.push_back(edge);
+        }
+        edges_.clear();
+
+        std::sort(sorted_edges.begin(), sorted_edges.end());
+
+        for (auto const& edge : sorted_edges) {
+            if (!graph.Implies(edge)) {
+                graph.AddEdge(edge);
+            }
+        }
+        return graph;
+    }
+};
+
+}  // namespace algos::fd::fdhits
--- a/src/core/algorithms/fd/fdhits/edgecollectors/vec_edge_collector.h
+++ b/src/core/algorithms/fd/fdhits/edgecollectors/vec_edge_collector.h
@@ -1,0 +1,28 @@
+#pragma once
+#include <vector>
+
+namespace algos::fd::fdhits {
+
+template <typename E>
+class VecEdgeCollector {
+public:
+    std::vector<E> results;
+
+    void Consume(E const& edge) {
+        results.push_back(edge);
+    }
+
+    size_t Size() const {
+        return results.size();
+    }
+
+    auto begin() const {
+        return results.begin();
+    }
+
+    auto end() const {
+        return results.end();
+    }
+};
+
+}  // namespace algos::fd::fdhits
--- a/src/core/algorithms/fd/fdhits/edges/default_edge.h
+++ b/src/core/algorithms/fd/fdhits/edges/default_edge.h
@@ -1,0 +1,191 @@
+#pragma once
+
+#include <cstddef>
+#include <sstream>
+#include <string>
+#include <vector>
+
+#include <boost/dynamic_bitset.hpp>
+#include <boost/functional/hash.hpp>
+
+namespace algos::fd::fdhits::edges {
+class DefaultEdge {
+private:
+    boost::dynamic_bitset<> bits_;
+
+    explicit DefaultEdge(boost::dynamic_bitset<> bits) : bits_(std::move(bits)) {}
+
+public:
+    static constexpr auto kNpos = boost::dynamic_bitset<>::npos;
+
+    explicit DefaultEdge(size_t size) : bits_(size) {}
+
+    static DefaultEdge Empty(size_t size) {
+        return DefaultEdge(size);
+    }
+
+    static DefaultEdge Filled(size_t size) {
+        DefaultEdge edge(size);
+        edge.bits_.set();
+        return edge;
+    }
+
+    void Add(size_t node) {
+        bits_.set(node);
+    }
+
+    void Remove(size_t node) {
+        bits_.reset(node);
+    }
+
+    bool Contains(size_t node) const {
+        return bits_.test(node);
+    }
+
+    size_t Count() const {
+        return bits_.count();
+    }
+
+    size_t Size() const {
+        return bits_.size();
+    }
+
+    bool IsEmpty() const {
+        return bits_.none();
+    }
+
+    size_t Overlap(DefaultEdge const& other) const {
+        return (bits_ & other.bits_).count();
+    }
+
+    bool Covers(DefaultEdge const& other) const {
+        return bits_.intersects(other.bits_);
+    }
+
+    bool Implies(DefaultEdge const& other) const {
+        return bits_.is_subset_of(other.bits_);
+    }
+
+    bool IsImpliedBy(std::vector<DefaultEdge> const& others) const {
+        for (auto const& e : others) {
+            if (e.Implies(*this)) return true;
+        }
+        return false;
+    }
+
+    void RemoveAll(DefaultEdge const& other) {
+        bits_ -= other.bits_;
+    }
+
+    void AddAll(DefaultEdge const& other) {
+        bits_ |= other.bits_;
+    }
+
+    DefaultEdge Intersect(DefaultEdge const& other) const {
+        return DefaultEdge(bits_ & other.bits_);
+    }
+
+    void IntersectWith(DefaultEdge const& other) {
+        bits_ &= other.bits_;
+    }
+
+    void SetTo(DefaultEdge const& other) {
+        bits_ = other.bits_;
+    }
+
+    void Clear() {
+        bits_.reset();
+    }
+
+    std::vector<size_t> GetNodes() const {
+        std::vector<size_t> nodes;
+        nodes.reserve(Count());
+        for (size_t i = bits_.find_first(); i != kNpos; i = bits_.find_next(i)) {
+            nodes.push_back(i);
+        }
+        return nodes;
+    }
+
+    class NodesIterator {
+    private:
+        boost::dynamic_bitset<> const* bits_;
+        size_t current_;
+
+    public:
+        NodesIterator(boost::dynamic_bitset<> const* bits, size_t pos)
+            : bits_(bits), current_(pos) {}
+
+        size_t operator*() const {
+            return current_;
+        }
+
+        NodesIterator& operator++() {
+            current_ = bits_->find_next(current_);
+            return *this;
+        }
+
+        bool operator!=(NodesIterator const& other) const {
+            return current_ != other.current_;
+        }
+    };
+
+    NodesIterator NodesBegin() const {
+        return NodesIterator(&bits_, bits_.find_first());
+    }
+
+    NodesIterator NodesEnd() const {
+        return NodesIterator(&bits_, kNpos);
+    }
+
+    std::string ToString() const {
+        std::ostringstream oss;
+        bool first = true;
+        for (size_t i = bits_.find_first(); i != kNpos; i = bits_.find_next(i)) {
+            if (!first) oss << ",";
+            oss << i;
+            first = false;
+        }
+        return oss.str();
+    }
+
+    bool operator==(DefaultEdge const& other) const {
+        return bits_ == other.bits_;
+    }
+
+    bool operator!=(DefaultEdge const& other) const {
+        return bits_ != other.bits_;
+    }
+
+    bool operator<(DefaultEdge const& other) const {
+        if (bits_.size() != other.bits_.size()) {
+            return bits_.size() < other.bits_.size();
+        }
+        for (size_t i = 0; i < bits_.size(); ++i) {
+            if (bits_[i] != other.bits_[i]) {
+                return bits_[i] < other.bits_[i];
+            }
+        }
+        return false;
+    }
+
+    boost::dynamic_bitset<>& GetBits() {
+        return bits_;
+    }
+
+    boost::dynamic_bitset<> const& GetBits() const {
+        return bits_;
+    }
+
+    struct Hash {
+        std::size_t operator()(DefaultEdge const& edge) const {
+            std::size_t hash = 0;
+            boost::hash_combine(hash, edge.bits_.count());
+            for (size_t i = edge.bits_.find_first(); i != kNpos; i = edge.bits_.find_next(i)) {
+                boost::hash_combine(hash, i);
+            }
+            return hash;
+        }
+    };
+};
+
+}  // namespace algos::fd::fdhits::edges
--- a/src/core/algorithms/fd/fdhits/edges/multi_fd.h
+++ b/src/core/algorithms/fd/fdhits/edges/multi_fd.h
@@ -1,0 +1,193 @@
+#pragma once
+
+#include <algorithm>
+#include <cstddef>
+#include <string>
+#include <vector>
+
+#include <boost/functional/hash.hpp>
+
+#include "core/algorithms/fd/fdhits/edges/default_edge.h"
+
+namespace algos::fd::fdhits::edges {
+class MultiFD {
+private:
+    DefaultEdge lhs_;
+    DefaultEdge rhs_;
+
+public:
+    MultiFD(size_t num_columns) : lhs_(num_columns / 2), rhs_(num_columns / 2) {}
+
+    MultiFD(DefaultEdge lhs, DefaultEdge rhs) : lhs_(std::move(lhs)), rhs_(std::move(rhs)) {}
+
+    static MultiFD Empty(size_t num_columns) {
+        return MultiFD(num_columns);
+    }
+
+    static MultiFD Filled(size_t num_columns) {
+        return MultiFD(DefaultEdge::Filled(num_columns / 2), DefaultEdge::Filled(num_columns / 2));
+    }
+
+    void Add(size_t node) {
+        if (node < lhs_.Size()) {
+            lhs_.Add(node);
+        } else {
+            rhs_.Add(node - lhs_.Size());
+        }
+    }
+
+    void Remove(size_t node) {
+        if (node < lhs_.Size()) {
+            lhs_.Remove(node);
+        } else {
+            rhs_.Remove(node - lhs_.Size());
+        }
+    }
+
+    void Clear() {
+        lhs_.Clear();
+        rhs_.Clear();
+    }
+
+    bool Contains(size_t node) const {
+        if (node < lhs_.Size()) {
+            return lhs_.Contains(node);
+        } else {
+            return rhs_.Contains(node - lhs_.Size());
+        }
+    }
+
+    size_t Count() const {
+        return lhs_.Count() + rhs_.Count();
+    }
+
+    size_t Size() const {
+        return lhs_.Size() + rhs_.Size();
+    }
+
+    size_t Overlap(MultiFD const& other) const {
+        return lhs_.Overlap(other.lhs_) + rhs_.Overlap(other.rhs_);
+    }
+
+    bool Covers(MultiFD const& other) const {
+        return lhs_.Covers(other.lhs_) || rhs_.Implies(other.rhs_);
+    }
+
+    bool Implies(MultiFD const& other) const {
+        return lhs_.Implies(other.lhs_) && rhs_.Implies(other.rhs_);
+    }
+
+    bool IsImpliedBy(std::vector<MultiFD> const& others) const {
+        return std::any_of(others.begin(), others.end(),
+                           [this](MultiFD const& o) { return o.Implies(*this); });
+    }
+
+    void RemoveAll(MultiFD const& other) {
+        lhs_.RemoveAll(other.lhs_);
+        rhs_.RemoveAll(other.rhs_);
+    }
+
+    void AddAll(MultiFD const& other) {
+        lhs_.AddAll(other.lhs_);
+        rhs_.AddAll(other.rhs_);
+    }
+
+    MultiFD Intersect(MultiFD const& other) const {
+        return MultiFD(lhs_.Intersect(other.lhs_), rhs_.Intersect(other.rhs_));
+    }
+
+    void RemoveAllInvalid(MultiFD const& other) {
+        for (size_t rhs_node : other.rhs_.GetNodes()) {
+            lhs_.Remove(rhs_node);
+        }
+        if (other.rhs_.Count() > 0) {
+            rhs_.Clear();
+        } else {
+            for (size_t lhs_node : other.lhs_.GetNodes()) {
+                rhs_.Remove(lhs_node);
+            }
+        }
+    }
+
+    void SetTo(MultiFD const& other) {
+        lhs_.SetTo(other.lhs_);
+        rhs_.SetTo(other.rhs_);
+    }
+
+    std::vector<size_t> GetNodes() const {
+        std::vector<size_t> nodes;
+        nodes.reserve(Count());
+        for (size_t node : lhs_.GetNodes()) {
+            nodes.push_back(node);
+        }
+        size_t offset = lhs_.Size();
+        for (size_t node : rhs_.GetNodes()) {
+            nodes.push_back(node + offset);
+        }
+        return nodes;
+    }
+
+    std::string ToString() const {
+        return lhs_.ToString() + " -> " + rhs_.ToString();
+    }
+
+    DefaultEdge const& GetLhs() const {
+        return lhs_;
+    }
+
+    DefaultEdge const& GetRhs() const {
+        return rhs_;
+    }
+
+    std::vector<size_t> GetRhsNodes() const {
+        return rhs_.GetNodes();
+    }
+
+    void RemoveRhs(size_t node) {
+        rhs_.Remove(node);
+    }
+
+    void RemoveLhs(size_t node) {
+        lhs_.Remove(node);
+    }
+
+    void AddLhs(size_t node) {
+        lhs_.Add(node);
+    }
+
+    void SetRhs(DefaultEdge const& rhs) {
+        rhs_.SetTo(rhs);
+    }
+
+    void IntersectRhs(DefaultEdge const& other) {
+        rhs_.IntersectWith(other);
+    }
+
+    size_t RhsCount() const {
+        return rhs_.Count();
+    }
+
+    bool operator==(MultiFD const& other) const {
+        return lhs_ == other.lhs_ && rhs_ == other.rhs_;
+    }
+
+    bool operator!=(MultiFD const& other) const {
+        return !(*this == other);
+    }
+
+    bool operator<(MultiFD const& other) const {
+        if (lhs_ != other.lhs_) return lhs_ < other.lhs_;
+        return rhs_ < other.rhs_;
+    }
+
+    struct Hash {
+        std::size_t operator()(MultiFD const& fd) const {
+            std::size_t hash = 0;
+            boost::hash_combine(hash, DefaultEdge::Hash{}(fd.lhs_));
+            boost::hash_combine(hash, DefaultEdge::Hash{}(fd.rhs_));
+            return hash;
+        }
+    };
+};
+
+}  // namespace algos::fd::fdhits::edges
--- a/src/core/algorithms/fd/fdhits/problems/fd_problem.h
+++ b/src/core/algorithms/fd/fdhits/problems/fd_problem.h
@@ -1,0 +1,44 @@
+#pragma once
+
+#include "core/algorithms/fd/fdhits/edges/default_edge.h"
+#include "core/algorithms/fd/fdhits/pli_table.h"
+#include "core/algorithms/fd/fdhits/types.h"
+
+namespace algos::fd::fdhits::problems {
+class FDProblem {
+private:
+    size_t column_id_;
+    PLITable const* table_;
+
+public:
+    using EdgeType = edges::DefaultEdge;
+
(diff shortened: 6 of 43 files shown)


src/core/algorithms/fd/fdhits/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+desbordante_add_lib(NAME)
+target_sources(
+    ${NAME}
+    PRIVATE fdhits.cpp
+            pli_table.cpp
+            treesearch/dynamic_edgemark_store.cpp
+            treesearch/dynamic_edgemark_store_multi.cpp
+            treesearch/treesearch_de.cpp
+            treesearch/treesearch_multi.cpp
+    PUBLIC FILE_SET HEADERS BASE_DIRS "${PROJECT_SOURCE_DIR}/src"
```
wildsor, line 11: These `PUBLIC FILE_SET HEADERS` lines are not needed now if you don't want to actually specify headers

src/core/algorithms/fd/fdhits/edgecollectors/edge_hash_collector.h
```diff
@@ -0,0 +1,53 @@
...
+
+template <typename E>
+class EdgeHashCollector {
+private:
+    boost::unordered_set<E, typename E::Hash> edges_;
+
+public:
+    void Consume(E const& edge) {
+        edges_.insert(edge);
+    }
```
ol-imorozko, line 22: If I understood correctly, `EdgeHashCollector` takes ownership of edges. It that case, I would suggest to write Consume like so:
```cpp
  void Consume(E const& edge) {
      edges_.insert(edge);
  }

  void Consume(E&& edge) {
      edges_.insert(std::move(edge));
  }
```
Otherwise you will be copying object of type E each time you call `Consume`

src/core/algorithms/fd/fdhits/edgecollectors/edge_hash_collector.h
```diff
@@ -0,0 +1,53 @@
+#pragma once
+
+#include <algorithm>
+#include <vector>
+
+#include <boost/unordered_set.hpp>
+
+#include "core/algorithms/fd/fdhits/edges/default_edge.h"
+#include "core/algorithms/fd/fdhits/edges/multi_fd.h"
```
ol-imorozko, line 9: unused includes

src/core/algorithms/fd/fdhits/edgecollectors/edge_hash_collector.h
```diff
@@ -0,0 +1,53 @@
...
+
+    Hypergraph<E> ToGraph(E const& s) {
+        Hypergraph<E> graph(s.Size());
+
+        std::vector<E> sorted_edges;
+        sorted_edges.reserve(edges_.size());
+        for (auto const& edge : edges_) {
+            sorted_edges.push_back(edge);
+        }
+        edges_.clear();
```
ol-imorozko, line 40: Each `push_back` is a copy and you even immediately call `clear()` to free the original. 
Just move it:
```cpp
while (!edges_.empty()) {
    sorted_edges.push_back(
        std::move(edges_.extract(edges_.begin()).value())
    );
}
```

src/core/algorithms/fd/fdhits/edgecollectors/edge_hash_collector.h
```diff
@@ -0,0 +1,53 @@
...
+    auto const& GetEdgeSet() const {
+        return edges_;
+    }
+
+    bool IsEmpty() const {
+        return edges_.empty();
+    }
+
+    Hypergraph<E> ToGraph(E const& s) {
+        Hypergraph<E> graph(s.Size());
```
ol-imorozko, line 33: `ToGraph(E const& s)` only reads `s.Size()`. Maybe take `size_t vertex_count` instead? This would save few allocations since we won't have to construct `E` object

src/core/algorithms/fd/fdhits/edgecollectors/edge_hash_collector.h
```diff
@@ -0,0 +1,53 @@
...
+
+        std::vector<E> sorted_edges;
+        sorted_edges.reserve(edges_.size());
+        for (auto const& edge : edges_) {
+            sorted_edges.push_back(edge);
+        }
+        edges_.clear();
+
+        std::sort(sorted_edges.begin(), sorted_edges.end());
```
ol-imorozko, line 42: nit: `std::ranges::sort(sorted_edges)`

src/core/algorithms/fd/fdhits/edgecollectors/edge_hash_collector.h
```diff
@@ -0,0 +1,53 @@
...
+    boost::unordered_set<E, typename E::Hash> edges_;
+
+public:
+    void Consume(E const& edge) {
+        edges_.insert(edge);
+    }
+
+    auto const& GetEdgeSet() const {
+        return edges_;
+    }
```
ol-imorozko, line 26: The only usage is in pli_validator.h: it calls it twice in one expression just to get `begin()/end()`. I would suggest to provide a range accessor instead of giving the whole underlying container

src/core/algorithms/fd/fdhits/edgecollectors/edge_hash_collector.h
```diff
@@ -0,0 +1,53 @@
...
+        for (auto const& edge : edges_) {
+            sorted_edges.push_back(edge);
+        }
+        edges_.clear();
+
+        std::sort(sorted_edges.begin(), sorted_edges.end());
+
+        for (auto const& edge : sorted_edges) {
+            if (!graph.Implies(edge)) {
+                graph.AddEdge(edge);
```
ol-imorozko, line 46:
```suggestion
        for (auto && edge : sorted_edges) {
            if (!graph.Implies(edge)) {
                graph.AddEdge(std::move(edge));
```

src/core/algorithms/fd/fdhits/edgecollectors/edge_hash_collector.h
ol-imorozko, line 42: Also claude told me this:

> Hypergraph::Implies(edge) is ∃e ∈ graph : e ⊆ edge, so this loop can only prune an edge that is a superset of one already inserted — which requires minimal-cardinality-first ordering. But DefaultEdge::operator< (edges/default_edge.h:159) compares bits_.size() — the bitset length, identical for every edge produced by a given Problem — and then  
>   falls through to a lexicographic scan from bit 0 up. So the ordering is by bit pattern, and {1,2,3} sorts before {0}. Cardinality never enters into it, and the minimization mostly doesn't fire.                                                                                                                                                         
>                                                                                                                                                                                                                                                                                                                                                             
>   Cheapest fix is in operator<, not here: compare Count() first, then keep the existing bit-pattern scan as the tiebreak. That restores subset-before-superset and preserves a total order — worth keeping, since boost::unordered_set iteration order is unspecified and the total order is currently the only thing making the minimized graph            
>   deterministic across runs. Relevant given ConsistentRepeatedExecution in the typed suite.

Might want to look into that, I did not researched that deeper though so I can't verify whether this is an issue or not

src/core/algorithms/fd/fdhits/edgecollectors/vec_edge_collector.h
```diff
@@ -0,0 +1,28 @@
...
+namespace algos::fd::fdhits {
+
+template <typename E>
+class VecEdgeCollector {
+public:
+    std::vector<E> results;
+
+    void Consume(E const& edge) {
+        results.push_back(edge);
+    }
```
ol-imorozko, line 13: Same with the rvalue and move

src/core/algorithms/fd/fdhits/edgecollectors/vec_edge_collector.h
```diff
@@ -0,0 +1,28 @@
...
+        return results.size();
+    }
+
+    auto begin() const {
+        return results.begin();
+    }
+
+    auto end() const {
+        return results.end();
+    }
```
ol-imorozko, line 25: What's the point if the `results` is public?

src/core/algorithms/fd/fdhits/edgecollectors/vec_edge_collector.h
```diff
@@ -0,0 +1,28 @@
+#pragma once
+#include <vector>
```
ol-imorozko, line 2: separate with a blank line

src/core/algorithms/fd/fdhits/edges/default_edge.h
```diff
@@ -0,0 +1,191 @@
...
+    explicit DefaultEdge(boost::dynamic_bitset<> bits) : bits_(std::move(bits)) {}
+
+public:
+    static constexpr auto kNpos = boost::dynamic_bitset<>::npos;
+
+    explicit DefaultEdge(size_t size) : bits_(size) {}
+
+    static DefaultEdge Empty(size_t size) {
+        return DefaultEdge(size);
+    }
```
ol-imorozko, line 25: I would name it `EmptyEdge` or something. It gets a bit confusing since we also have `IsEmpty()` here

src/core/algorithms/fd/fdhits/edges/default_edge.h
```diff
@@ -0,0 +1,191 @@
...
+        return bits_.size();
+    }
+
+    bool IsEmpty() const {
+        return bits_.none();
+    }
+
+    size_t Overlap(DefaultEdge const& other) const {
+        return (bits_ & other.bits_).count();
+    }
```
ol-imorozko, line 59: Oh. This materialises a whole temporary `dynamic_bitset` only to call `count()` therefore it's memory alloc and free each call.
I would suggest to do smth like
```cpp
  size_t Overlap(DefaultEdge const& other) const {
      size_t count = 0;
      for (size_t node : Nodes()) {
          if (other.Contains(node)) ++count;
      }
      return count;
  }
```

src/core/algorithms/fd/fdhits/edges/default_edge.h
```diff
@@ -0,0 +1,191 @@
...
+        if (bits_.size() != other.bits_.size()) {
+            return bits_.size() < other.bits_.size();
+        }
+        for (size_t i = 0; i < bits_.size(); ++i) {
+            if (bits_[i] != other.bits_[i]) {
+                return bits_[i] < other.bits_[i];
+            }
+        }
+        return false;
+    }
```
ol-imorozko, line 169: Read review in src/core/algorithms/fd/fdhits/edgecollectors/edge_hash_collector.h. Claude suggests to do this:
```cpp
  bool operator<(DefaultEdge const& other) const {
      if (Count() != other.Count()) return Count() < other.Count();
      return bits_ < other.bits_;
  }
```

src/core/algorithms/fd/fdhits/edges/default_edge.h
```diff
@@ -0,0 +1,191 @@
...
+    bool Implies(DefaultEdge const& other) const {
+        return bits_.is_subset_of(other.bits_);
+    }
+
+    bool IsImpliedBy(std::vector<DefaultEdge> const& others) const {
+        for (auto const& e : others) {
+            if (e.Implies(*this)) return true;
+        }
+        return false;
+    }
```
ol-imorozko, line 74: nit: could be replaces with some algo like `std::any_of`. Fine as is though

src/core/algorithms/fd/fdhits/edges/default_edge.h
```diff
@@ -0,0 +1,191 @@
...
+
+    std::vector<size_t> GetNodes() const {
+        std::vector<size_t> nodes;
+        nodes.reserve(Count());
+        for (size_t i = bits_.find_first(); i != kNpos; i = bits_.find_next(i)) {
+            nodes.push_back(i);
+        }
+        return nodes;
+    }
```
ol-imorozko, line 107: For something like `for (size_t x : edge.GetNodes())` you would have to build a new vector each time. 
Why not just use `NodesIterator` below?
This method simply becomes
```cpp
  auto Nodes() const {
      return std::ranges::subrange(NodesBegin(), NodesEnd());
  }
```

src/core/algorithms/fd/fdhits/edges/default_edge.h
```diff
@@ -0,0 +1,191 @@
...
+
+    NodesIterator NodesEnd() const {
+        return NodesIterator(&bits_, kNpos);
+    }
+
+    std::string ToString() const {
+        std::ostringstream oss;
+        bool first = true;
+        for (size_t i = bits_.find_first(); i != kNpos; i = bits_.find_next(i)) {
```
ol-imorozko, line 143: `for (size_t node : Nodes())`, cause you've implemented the NodesIterator. Use it

src/core/algorithms/fd/fdhits/edges/default_edge.h
```diff
@@ -0,0 +1,191 @@
...
+
+        NodesIterator& operator++() {
+            current_ = bits_->find_next(current_);
+            return *this;
+        }
+
+        bool operator!=(NodesIterator const& other) const {
+            return current_ != other.current_;
+        }
+    };
```
ol-imorozko, line 130: nit:
It's fine, but I want to note that good quality custom iterator also should have `static_assert(std::forward_iterator<DefaultEdge::NodesIterator>);` satisfied so you can use it in standard algorithms. I think even here we have some usage for `std::any_of` and right now we won't be able to do this.
You can add `static_assert(std::forward_iterator<DefaultEdge::NodesIterator>);` right here and see what's missing.

Mainly it's `operator==` (you can fix it by just replacing operator != with `friend bool operator==(NodesIterator const&, NodesIterator const&) = default;`, C++20 can infer != from that)
and some types like
```cpp
    using value_type = size_t;
    using difference_type = std::ptrdiff_t;
```

src/core/algorithms/fd/fdhits/treesearch/dynamic_edgemark_store_multi.cpp
```diff
@@ -0,0 +1,244 @@
...
+    for (size_t rhs : rhs_nodes) {
+        if (!CanAddRhs(rhs)) {
+            GetCurrentCandMut().RemoveRhs(rhs);
+        }
+    }
+}
+
+void DynamicEdgemarkStoreMulti::RemoveLhsForSingularRhs() {
+    if (GetCurrentCand().RhsCount() == 1) {
+        size_t rhs = GetCurrentCand().GetRhsNodes()[0];
```
ol-imorozko, line 88: you can add a method `FirstNode()` returning `bits_.find_first()` instead of creating a vector here only to take it's first element

src/core/algorithms/fd/fdhits/edges/default_edge.h
```diff
@@ -0,0 +1,191 @@
...
+        return oss.str();
+    }
+
+    bool operator==(DefaultEdge const& other) const {
+        return bits_ == other.bits_;
+    }
+
+    bool operator!=(DefaultEdge const& other) const {
+        return bits_ != other.bits_;
+    }
```
ol-imorozko, line 157: `bool operator==(DefaultEdge const& other) const = default;` is enough

src/core/algorithms/fd/fdhits/edges/default_edge.h
```diff
@@ -0,0 +1,191 @@
...
+        return bits_.none();
+    }
+
+    size_t Overlap(DefaultEdge const& other) const {
+        return (bits_ & other.bits_).count();
+    }
+
+    bool Covers(DefaultEdge const& other) const {
+        return bits_.intersects(other.bits_);
+    }
```
ol-imorozko, line 63: `Covers` is `intersects`? That's strange. Usually "covers" means superset

src/core/algorithms/fd/fdhits/edges/default_edge.h
```diff
@@ -0,0 +1,191 @@
...
+        return DefaultEdge(bits_ & other.bits_);
+    }
+
+    void IntersectWith(DefaultEdge const& other) {
+        bits_ &= other.bits_;
+    }
+
+    void SetTo(DefaultEdge const& other) {
+        bits_ = other.bits_;
+    }
```
ol-imorozko, line 94: Isn't that is just copy assignment? It's better to use `DefaultEdge x = y` in my opinion

src/core/algorithms/fd/fdhits/edges/default_edge.h
```diff
@@ -0,0 +1,191 @@
...
+
+    boost::dynamic_bitset<> const& GetBits() const {
+        return bits_;
+    }
+
+    struct Hash {
+        std::size_t operator()(DefaultEdge const& edge) const {
+            std::size_t hash = 0;
+            boost::hash_combine(hash, edge.bits_.count());
+            for (size_t i = edge.bits_.find_first(); i != kNpos; i = edge.bits_.find_next(i)) {
```
ol-imorozko, line 183: `for (size_t node : Nodes())`

src/core/algorithms/fd/fdhits/edges/default_edge.h
```diff
@@ -0,0 +1,191 @@
...
+class DefaultEdge {
+private:
+    boost::dynamic_bitset<> bits_;
+
+    explicit DefaultEdge(boost::dynamic_bitset<> bits) : bits_(std::move(bits)) {}
+
+public:
+    static constexpr auto kNpos = boost::dynamic_bitset<>::npos;
+
+    explicit DefaultEdge(size_t size) : bits_(size) {}
```
ol-imorozko, line 21: maybe move those to private section

src/core/algorithms/fd/fdhits/edges/multi_fd.h
```diff
@@ -0,0 +1,193 @@
...
+#include "core/algorithms/fd/fdhits/edges/default_edge.h"
+
+namespace algos::fd::fdhits::edges {
+class MultiFD {
+private:
+    DefaultEdge lhs_;
+    DefaultEdge rhs_;
+
+public:
+    MultiFD(size_t num_columns) : lhs_(num_columns / 2), rhs_(num_columns / 2) {}
```
ol-imorozko, line 19: DefaultEdge has `explicit`. Add it here also, please

src/core/algorithms/fd/fdhits/edges/multi_fd.h
```diff
@@ -0,0 +1,193 @@
...
+
+    MultiFD(DefaultEdge lhs, DefaultEdge rhs) : lhs_(std::move(lhs)), rhs_(std::move(rhs)) {}
+
+    static MultiFD Empty(size_t num_columns) {
+        return MultiFD(num_columns);
+    }
+
+    static MultiFD Filled(size_t num_columns) {
+        return MultiFD(DefaultEdge::Filled(num_columns / 2), DefaultEdge::Filled(num_columns / 2));
+    }
```
ol-imorozko, line 29: should we add assert here to verify that `num_columns` is even?

src/core/algorithms/fd/fdhits/edges/multi_fd.h
```diff
@@ -0,0 +1,193 @@
...
+
+    bool operator!=(MultiFD const& other) const {
+        return !(*this == other);
+    }
+
+    bool operator<(MultiFD const& other) const {
+        if (lhs_ != other.lhs_) return lhs_ < other.lhs_;
+        return rhs_ < other.rhs_;
+    }
```
ol-imorozko, line 181:
```suggestion
    bool operator<(MultiFD const& other) const {
        if (Count() != other.Count()) return Count() < other.Count();
        if (lhs_ != other.lhs_) return lhs_ < other.lhs_;
        return rhs_ < other.rhs_;
    }
```

src/core/algorithms/fd/fdhits/edges/multi_fd.h
```diff
@@ -0,0 +1,193 @@
...
+        return rhs_.Count();
+    }
+
+    bool operator==(MultiFD const& other) const {
+        return lhs_ == other.lhs_ && rhs_ == other.rhs_;
+    }
+
+    bool operator!=(MultiFD const& other) const {
+        return !(*this == other);
+    }
```
ol-imorozko, line 176: just `bool operator==(MultiFD const& other) const = default;` is enough

src/core/algorithms/fd/fdhits/edges/multi_fd.h
```diff
@@ -0,0 +1,193 @@
...
+        nodes.reserve(Count());
+        for (size_t node : lhs_.GetNodes()) {
+            nodes.push_back(node);
+        }
+        size_t offset = lhs_.Size();
+        for (size_t node : rhs_.GetNodes()) {
+            nodes.push_back(node + offset);
+        }
+        return nodes;
+    }
```
ol-imorozko, line 128: nit:
IDK wherther we would need an actual vector in the future, but if we use this only to iterate over, then implementing an iterator would be a better alternative

src/core/algorithms/fd/fdhits/treesearch/dynamic_edgemark_store_multi.cpp
```diff
@@ -0,0 +1,244 @@
...
+    for (size_t i = 1; i < rhs_nodes.size(); ++i) {
+        size_t col = rhs_nodes[i];
+        current_vertex_hitting_.IntersectWithInverse(vertex_hittings_[col]);
+    }
+}
+
+void DynamicEdgemarkStoreMulti::CheckCriticalityForActiveRhs() {
+    auto rhs_nodes = GetCurrentCand().GetRhsNodes();
+
+    for (size_t rhs : rhs_nodes) {
```
ol-imorozko, line 79: That allocates a vector only to iterate over it. Use iterators (read reveiew comment in default_edge.h)

src/core/algorithms/fd/fdhits/edges/default_edge.h
ol-imorozko: Hm. Why not just `return boost::hash_value(edge.bits_);` and that's it? `dynamic_bitset` has boost's hash

src/core/algorithms/fd/fdhits/problems/fd_problem.h
```diff
@@ -0,0 +1,44 @@
...
+
+    template <typename C>
+    void GetEdge(RowIndex r1, RowIndex r2, C* consumer) const {
+        EdgeType edge = EdgeType::Empty(table_->GetColumnCount());
+        for (size_t col = 0; col < table_->GetColumnCount(); ++col) {
+            if (table_->IsDifferentOnColumn(r1, r2, col)) {
+                edge.Add(col);
+            }
+        }
+        if (edge.Contains(column_id_)) {
```
ol-imorozko, line 26: Isn't `edge.Contains(column_id_)` <=> `table_->IsDifferentOnColumn(r1, r2, column_id_)`? If so, we can just put guard at the very begging and avoid extra work:
```cpp
  template <typename C>
  void GetEdge(RowIndex r1, RowIndex r2, C* consumer) const {
      if (!table_->IsDifferentOnColumn(r1, r2, column_id_)) return;

      size_t const num_columns = table_->GetColumnCount();
      EdgeType edge = EdgeType::Empty(num_columns);
      for (size_t col = 0; col < num_columns; ++col) {
          if (table_->IsDifferentOnColumn(r1, r2, col)) {
              edge.Add(col);
          }
      }
      consumer->Consume(std::move(edge));
  }
```

src/core/algorithms/fd/fdhits/problems/fd_problem.h
```diff
@@ -0,0 +1,44 @@
...
+            consumer->Consume(edge);
+        }
+    }
+
+    bool IsValidResult(EdgeType const& result) const {
+        return !result.Contains(column_id_);
+    }
```
ol-imorozko, line 33: claude noted that, might want to look into that:

> Every edge this class produces contains column_id_ (that's the L26 guard). The initial candidate set is EdgeType::Filled(pli_table.GetColumnCount()) (fdhits.cpp:99), which includes the target. So the target column is present in every edge, which makes it maximally attractive to the min-overlap heuristic at treesearch_de.cpp:26-31 — and         
>   selecting it covers the whole hypergraph in one step, driving straight to IsFullyCovered() → Check → IsValidResult returning false at pli_validator.h:51.                                                                                                                                                                                                 
>                                                                                                                                                                                                                                                                                                                                                             
>   If that reading is right, excluding column_id_ from the initial cand would prune that branch at every node instead of rediscovering and rejecting it, and IsValidResult (L31) becomes unnecessary for this problem. Haven't traced every path, so flagging as a question rather than a claim — but if it is what's happening it's a per-node waste, not a 
>   one-off.

src/core/algorithms/fd/fdhits/problems/fd_problem.h
```diff
@@ -0,0 +1,44 @@
...
+
+#include "core/algorithms/fd/fdhits/edges/default_edge.h"
+#include "core/algorithms/fd/fdhits/pli_table.h"
+#include "core/algorithms/fd/fdhits/types.h"
+
+namespace algos::fd::fdhits::problems {
+class FDProblem {
+private:
+    size_t column_id_;
+    PLITable const* table_;
```
ol-imorozko, line 11: nit:
You use T* in some places and T& in others, sometimes for the same parameter: `ClusterFilter` is `ClusterFilter*` in pli_cache.h:21,29,38,55, `ClusterFilter&` in pli_table.h:57, and `ClusterFilter const&` in refinement_helper.h:32.
If I understood correctly, none of those pointers are null nor we use type erasure or copying - so I would suggest to pick one approach (I suggest references)

src/core/algorithms/fd/fdhits/problems/fd_problem.h
```diff
@@ -0,0 +1,44 @@
...
+
+public:
+    using EdgeType = edges::DefaultEdge;
+
+    FDProblem(size_t column_id, PLITable const* table) : column_id_(column_id), table_(table) {}
+
+    template <typename C>
+    void GetEdge(RowIndex r1, RowIndex r2, C* consumer) const {
```
ol-imorozko, line 19: nit: Here C* is a reference since it's an out-parameter. It's correct. But if you decide to make all out-params a pointers, then `Intersect`'s `Pli& intersection` needs to follow it too. Or make them all references

src/core/algorithms/fd/fdhits/problems/multi_fd_problem.h
```diff
@@ -0,0 +1,68 @@
...
+    bool IsValidResult(EdgeType const& result) const {
+        return result.RhsCount() > 0;
+    }
+
+    bool IsViolatingPair(RowIndex r1, RowIndex r2, EdgeType const& candidate) const {
+        for (size_t node : candidate.GetRhs().GetNodes()) {
+            if (table_->IsDifferentOnColumn(r1, r2, node)) return true;
+        }
+        return false;
+    }
```
ol-imorozko, line 65: Use that Nodes() range from the default_edge.h discussion

src/core/algorithms/fd/fdhits/problems/multi_fd_problem.h
```diff
@@ -0,0 +1,68 @@
...
+                    if (lhs_col == rhs_col) continue;
+                    if (!table_->IsDifferentOnColumn(r1, r2, rhs_col)) {
+                        edges::DefaultEdge lhs = edges::DefaultEdge::Empty(num_columns);
+                        lhs.Add(lhs_col);
+                        edges::DefaultEdge rhs = edges::DefaultEdge::Empty(num_columns);
+                        rhs.Add(rhs_col);
+                        consumer->Consume(EdgeType(std::move(lhs), std::move(rhs)));
+                    }
+                }
+            }
```
ol-imorozko, line 37: Is this branch needed? fdhits.cpp:111 is the only construction and it pass `separate_edges_` as false

src/core/algorithms/fd/fdhits/sampling/fd_sampler.h
```diff
@@ -0,0 +1,118 @@
...
+    mutable std::mt19937 rng_;
+
+public:
+    using EdgeType = typename P::EdgeType;
+
+    Sampler(P const* problem, PLITable const* table, double sampling_factor)
+        : problem_(problem),
+          table_(table),
+          sampling_factor_(sampling_factor),
+          rng_(std::random_device{}()) {}
```
ol-imorozko, line 32: by claude:

`test_fd_algorithm.cpp` runs ConsistentRepeatedExecution and HeavyDatasetsConsistentHash against FDHits, so the suite is currently green only because the search happens to converge on every seed it's seen. First seed that doesn't, and it's an unreproducible CI failure.

The repo already has the pattern: EulerFD — the other sampling-based FD algorithm here — plumbs `config::kCustomRandomFlagOpt` (config/custom_random_seed/option.h) into util/CustomRandom, with the comment at eulerfd.cpp:380 reading "Choose random strategy (it is necessary for stable unit tests)". Please follow that.

src/core/algorithms/fd/fdhits/sampling/fd_sampler.h
```diff
@@ -0,0 +1,118 @@
...
+            bool operator==(SampleKey const& other) const {
+                return cluster_idx == other.cluster_idx && r1 == other.r1 && r2 == other.r2;
+            }
+        };
+
+        struct SampleKeyHash {
+            size_t operator()(SampleKey const& k) const {
+                return k.cluster_idx ^ (k.r1 << 16) ^ (k.r2 << 8);
+            }
+        };
```
ol-imorozko, line 66: Didn't get that part. Why can't we call `GetEdge` directly in the loop? What do we need SampleKey, SampleKeyHash and samples for?

ol-imorozko: I mean, we pass the consumer which already handles duplicates, no?

src/core/algorithms/fd/fdhits/sampling/fd_sampler.h
```diff
@@ -0,0 +1,118 @@
...
+#include "core/algorithms/fd/fdhits/pli_table.h"
+#include "core/algorithms/fd/fdhits/types.h"
+
+namespace algos::fd::fdhits {
+
+template <typename P>
+class Sampler {
+private:
+    P const* problem_;
+    PLITable const* table_;
```
ol-imorozko, line 21: I think that's an unused field

src/core/algorithms/fd/fdhits/sampling/fd_sampler.h
```diff
@@ -0,0 +1,118 @@
...
+        : problem_(problem),
+          table_(table),
+          sampling_factor_(sampling_factor),
+          rng_(std::random_device{}()) {}
+
+    template <typename C>
+    void SamplePLI(Pli const& pli, C* consumer) const {
+        if (pli.empty()) return;
+
+        std::vector<uint64_t> weights;
```
ol-imorozko, line 38: nit: add `reserve`

src/core/algorithms/fd/fdhits/treesearch/bit_util.h
```diff
@@ -0,0 +1,57 @@
...
+}
+
+inline size_t GetIntersectionSize(boost::dynamic_bitset<> const& one,
+                                  boost::dynamic_bitset<> const& other) {
+    boost::dynamic_bitset<> tmp = one;
+    tmp.resize(other.size());
+    return (tmp & other).count();
+}
+
+}  // namespace algos::fd::fdhits::bit_util
```
ol-imorozko, line 57: I don't like that each function here could allocate via resize. It seems that it is trying to fix some problem that should be fixed in another place. 

If we were to fix that, those functions become simple
```cpp
  inline void SetTo(boost::dynamic_bitset<>& one, boost::dynamic_bitset<> const& other) {
      assert(one.size() == other.size());
      one = other;
  }

  inline void SetToNot(boost::dynamic_bitset<>& one, boost::dynamic_bitset<> const& other) {
      assert(one.size() == other.size());
      one = other;
      one.flip();
  }

  inline void IntersectWithInverse(boost::dynamic_bitset<>& one,
                                   boost::dynamic_bitset<> const& other) {
      assert(one.size() == other.size());
      one -= other;
  }

  inline bool IsSubsetInverse(boost::dynamic_bitset<> const& one,
                              boost::dynamic_bitset<> const& other) {
      assert(one.size() == other.size());
      return !one.intersects(other);
  }
```

ol-imorozko: I asked agent on how those functions are used (I haven't reached that code yet) at it told me:

> Every Edgemark is supposed to be sized to the current edge count and grown in lockstep by Extend. Except one gets missed:    
                                                                                                                                          
```cpp
  // dynamic_edgemark_store_multi.cpp:186-197 — Extend grows three of the four
  for (auto& vh : vertex_hittings_) vh.Grow(graph_->GetEdgeCount());
  uncov_.Grow(graph_->GetEdgeCount());
  for (auto& crits : critical_) for (auto& crit : crits) crit.Grow(graph_->GetEdgeCount());
  // current_vertex_hitting_ (declared ..._multi.h:21, sized at ..._multi.cpp:10) — not grown
```
> It only re-syncs by accident: the next CalcRhsHitting calls SetToNot, whose one = ~other assignment silently resizes it.

Look into that pls, it may be wrong

src/core/algorithms/fd/fdhits/treesearch/bit_util.h
```diff
@@ -0,0 +1,57 @@
+#pragma once
+
+#include <algorithm>
```
ol-imorozko, line 3: unused

src/core/algorithms/fd/fdhits/treesearch/bit_util.h
ol-imorozko: If that's the case, then we can implement those helpers in Edgemark file (since they will become very small) to avoid `friend bool IsSubsetInverse` there

ol-imorozko (changes requested):

src/core/algorithms/fd/fdhits/edges/default_edge.h
ol-imorozko: Or `std::ranges::count_if`

PR #709, opened 2026-03-29
Simplify logging initialization

wildsor:


src/core/util/logger.h
```diff
@@ -7,39 +7,30 @@
...
-        logger = spdlog::get(logger_name);
-        assert(logger);
-    }
-    return logger;
-}
+        return logger;
+    }();
 
-inline std::shared_ptr<spdlog::logger>& GetCachedLogger() {
-    static thread_local std::shared_ptr<spdlog::logger> s_logger = ::util::logging::GetLogger();
```
wildsor, line 41: @Tydik42 Do you remember why you put `thread_local` in here?

src/core/util/logger.h
```diff
@@ -7,39 +7,30 @@
...
 
-    if (sinks.empty()) {
-        sinks.push_back(std::make_shared<spdlog::sinks::stderr_color_sink_mt>());
-    }
+inline DESBORDANTE_EXPORT std::shared_ptr<spdlog::logger>& Initialize(
+        std::vector<spdlog::sink_ptr> sinks = {}) {
+    static std::shared_ptr<spdlog::logger> core_logger = [&]() {
+        if (sinks.empty()) {
+            sinks.push_back(std::make_shared<spdlog::sinks::stderr_color_sink_mt>());
+        }
```
p-senichenkov, line 19: As for me, such usage of lambda looks a bit strange. I'd prefer giving it a name:
```suggestion
inline DESBORDANTE_EXPORT std::shared_ptr<spdlog::logger>& Initialize(
        std::vector<spdlog::sink_ptr> sinks = {}) {
    auto make_logger = [&sinks]() { ... };
    
    static std::shared_ptr<spdlog::logger> core_logger = make_logger();
```

_You can ignore this comment, if you'd like_

src/core/util/logger.h
```diff
@@ -7,39 +7,30 @@
...
+        // we are free to assume there will be no logger with this name.
+        auto logger = std::make_shared<spdlog::logger>("desbordante", sinks.begin(), sinks.end());
+        logger->set_pattern("[%Y-%m-%d %H:%M:%S.%e] [%n] [%^%l%$] %v");
 
-    logger->set_level(spdlog::level::trace);
+        logger->set_level(spdlog::level::trace);
 
-    spdlog::register_logger(logger);
-}
+        spdlog::register_logger(logger);
```
p-senichenkov: Actually, it is not necessary to register logger, since we don't use `spdlog::get`

p-senichenkov (approved):

src/core/util/logger.h
wildsor: Looks standard to me ([example](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#res-lambda-init)).

src/core/util/logger.h
```diff
@@ -7,39 +7,27 @@
...
 
-    if (sinks.empty()) {
-        sinks.push_back(std::make_shared<spdlog::sinks::stderr_color_sink_mt>());
-    }
+inline DESBORDANTE_EXPORT std::shared_ptr<spdlog::logger>& Initialize(
+        std::vector<spdlog::sink_ptr> sinks = {}) {
+    static std::shared_ptr<spdlog::logger> core_logger = [&sinks]() {
+        if (sinks.empty()) {
+            sinks.push_back(std::make_shared<spdlog::sinks::stderr_color_sink_mt>());
+        }
```
Tydik42, line 19: nit:
`Initialize()` now permanently captures the sinks from the first call. If any `LOG_*` happens before the Python bridge calls `Initialize({python_sink})`, the logger is created with the default stderr sink and the Python sink is never attached. This is probably fine in the current import path, but it is still order dependent. I would make `Initialize()` idempotent for sink attachment as well: create the logger on first call, and on later calls merge in any new sinks instead of ignoring them.

Tydik42 (approved):
Thanks for catching this nasty bug! The change looks good overall; I just have one small follow up remark

src/core/util/logger.h
wildsor: I don't think this PR changes behavior in any way, unless I'm missing something. Before, if `LOG_*` was invoked before `EnsureInitialized`, it would proceed to initialize the logger with the stderr sink, and later `EnsureInitialized` calls would be ignored in the same way.

PR #713, opened 2026-04-01
Remove null handling in PLIs

wildsor:
1. I don't think nulls should be handled in them
2. The handling was incorrect ("" instead of "NULL")
3. The case with nulls was not fully implemented

### DIFF
--- a/src/core/model/table/position_list_index.cpp
+++ b/src/core/model/table/position_list_index.cpp
@@ -25,8 +25,7 @@ int const PositionListIndex::kSingletonValueId = 0;
 unsigned long long PositionListIndex::micros_ = 0;
 int PositionListIndex::intersection_count_ = 0;
 
-PositionListIndex::PositionListIndex(std::deque<std::vector<int>> index,
-                                     std::vector<int> null_cluster, unsigned int size,
+PositionListIndex::PositionListIndex(std::deque<std::vector<int>> index, unsigned int size,
                                      double entropy, unsigned long long nep,
                                      unsigned int relation_size,
                                      unsigned int original_relation_size, double inverted_entropy,
@@ -34,7 +33,6 @@ PositionListIndex::PositionListIndex(std::deque<std::vector<int>> index,
     : index_(std::move(index)),
       relation_size_(relation_size),
       size_(size),
-      null_cluster_(std::move(null_cluster)),
       entropy_(entropy),
       inverted_entropy_(inverted_entropy),
       gini_impurity_(gini_impurity),
@@ -42,20 +40,11 @@ PositionListIndex::PositionListIndex(std::deque<std::vector<int>> index,
       original_relation_size_(original_relation_size),
       probing_table_cache_() {}
 
-std::unique_ptr<PositionListIndex> PositionListIndex::CreateFor(std::vector<int>& data,
-                                                                bool is_null_eq_null) {
+std::unique_ptr<PositionListIndex> PositionListIndex::CreateFor(std::vector<int>& data) {
     std::unordered_map<int, std::vector<int>> index;
     for (unsigned long position = 0; position < data.size(); ++position) {
         int value_id = data[position];
         index[value_id].push_back(position);
-    }
-
-    std::vector<int> null_cluster;
-    if (index.count(ColumnLayoutRelationData::kNullValueId) != 0) {
-        null_cluster = index[ColumnLayoutRelationData::kNullValueId];
-    }
-    if (!is_null_eq_null) {
-        index.erase(ColumnLayoutRelationData::kNullValueId);  // move?
     }
 
     double key_gap = 0.0;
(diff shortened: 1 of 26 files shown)


src/core/model/table/position_list_index.cpp
```diff
@@ -25,39 +25,28 @@ int const PositionListIndex::kSingletonValueId = 0;
...
         index[value_id].push_back(position);
     }
 
-    std::vector<int> null_cluster;
-    if (index.count(ColumnLayoutRelationData::kNullValueId) != 0) {
-        null_cluster = index[ColumnLayoutRelationData::kNullValueId];
-    }
-    if (!is_null_eq_null) {
-        index.erase(ColumnLayoutRelationData::kNullValueId);  // move?
-    }
```
polyntsov, line 59: While I agree that the way we currently handle nulls is outright bad and probably buggy, but I am not sure if nulls should be removed from plis.

>I don't think nulls should be handled in them

Given that PLI effectively is an equivalence set, we have to compare values for equality when building them, and so we have to handle NULLs somehow. A solution might be to make a decision that all NULLs are always handled in a specific way (i.e. always considered equal to each other), the question is how do we make such decision and what are the consequences

polyntsov (changes requested):

src/core/model/table/position_list_index.cpp
polyntsov: or we can probably just remove any notion of NULL from the codebase, we are a python library to work with csv data, not an sql database

The fact that currently string `"NULL"` is treated in a special way is probably confusing

wildsor: Okay, this ended up horribly jumbled, but I don't want to spend more time organizing it, it should be clear enough.

I've thought about it a bit more. NULLs are not clearly defined entities and require special treatment in every case. We don't have the resources for that. But if the person (people) making an algorithm decides to (incompletely) support them, they should be able to, and if they don't want to deal with them, they likewise should be able to avoid that.

`PositionListIndex` is one of those structures that is going to be looked at often. It is also one that people are going to copy to adapt to their needs. Thus, I believe it would be better if it was as understandable as possible. The simpler, the better. And the simpler case would be not having entities as annoying to deal with as NULLs at all in it.

### Context

To be more specific, what exactly is a NULL? How do we treat it if we encounter it? Well, it makes sense to treat each NULL as a distinct value (`is_null_equal_null = false`) in a column that identifies a person (like personnummer, SSN). It also makes sense to treat it as "no value" (`is_null_equal_null = true`) where there is always a known value or no value (for example, discount code for an order). In addition to that, there is also a situation where the value is unknown but present in reality (first name). Or we could have either no value or unknown value (patronymic).

### PLIs

Okay, why not just leave it be and have two possible semantics? Well, I don't think we can have that just by having a switch in one data structure. I think, some people saw the flag you can pass to the `ColumnLayoutRelationData::CreateFrom` method and cargo culted the EqualNulls option onto their algorithm. And this is going to keep happening in the future. It might break something eventually.

It also makes PLIs harder to understand, as it's another thing to keep track of.

Besides, Desbordante doesn't even have proper NULLs, as in a value distinct from every other possible value. We still read data as if it is a text file (a sequence of `std::vector<std::string>`), even with DataFrames, so there is no NULL, "NULL" and "" are just strings. We can't have an actual, project-level NULL like this.

Given all that, I suggest we axe the NULL support in the PLI for now and leave it to the algorithms themselves. Anyone willing to implement (partial) support for NULLs in a manner that works with a switch like this in their algorithm could copy the PLI and adapt it.

### NULL handling in general

On another note, what do we say about a FD holding if we encounter a NULL as one of the relevant values? Well, after a bit of searching, I found lots of articles from the 80's discussing exactly that. It looks like quite the rabbit hole. Some even suggest using 3-valued logic, which definitely can't be dealt with with just a data structure switch.

Now imagine that, but for every single table pattern type. I'm of the opinion that we can't have project-wide NULL value handling and should probably get rid of any traces of it in other places too.

### Response

> so we have to handle NULLs somehow

We don't have to handle them if we just say they don't exist. There is no NULL `std::string`. As for future usage, see the part about NULL handling in general. Whoever wants to handle NULLs implements their own thing, if it ends up reusable in other places, the other places can go and reuse it.

> what are the consequences

We kill valid use cases and complicate their reintroduction in the future. The project's facilities are not mature enough to make that call.

> not sure if nulls should be removed from plis

They should be because handling NULLs is likely to be more complicated than having a single switch in a data structure and people using PLIs might not realize that. They might not even realize what the flag does at all.

### Conclusion

Let's isolate NULL handling to the algorithms that actually do it, starting from the PLIs.

polyntsov: >Let's isolate NULL handling to the algorithms that actually do it, starting from the PLIs.

I'd just remove notion of NULLs completely, it was a bad design decision made with the goal to have feature parity with metanome. There is no notion of NULL in csv, so I don't think we should come up with our own notion for it. Instead we should let the user decide if they want to handle some values of their input data in a special way and preprocess the data accordingly

polyntsov (approved):
>I'd just remove notion of NULLs completely, it was a bad design decision made with the goal to have feature parity with metanome. There is no notion of NULL in csv, so I don't think we should come up with our own notion for it. Instead we should let the user decide if they want to handle some values of their input data in a special way and preprocess the data accordingly

Regardless of the decision made on this comment, the change is good

PR #714, opened 2026-04-03
Check LHS size before inserting into FDTree

wildsor:
Merging #798 first is preferred

This helps cut runtime and memory if the max LHS option is specified. Before, FDs with LHSs that are too large were only removed at the point of result construction.

The result is constructed by traversing an internal trie (FDTree). The trie has an invariant that no nodes corresponding to an FD that generalizes some other FD that is in the tree ever exist in it. This is achieved by checking for that condition before adding a new node.

A FD with a larger LHS by definition never generalizes a FD with a smaller LHS. The only way FDs are added to the trie is by adding one new attribute to the LHS of a FD that was already in the trie.

Thus, if we stop all FDs with LHSs beyond a certain size from being added, then the trie's invariant would still hold, and the FDs with LHSs that are too large would not end up in the trie (and in the answer).

We will not miss any FDs with a smaller size either, because the checks for the FDs with smaller LHSs are not affected in any manner by the FDs with larger LHSs.

The situation is exactly the same as in HyMD, and it's being done there already. I suspect it's true for all the Hy* algorithms too.


src/core/algorithms/fd/hyfd/validator.cpp
```diff
@@ -296,9 +298,10 @@ algos::hy::IdPairs Validator::ValidateAndExtendCandidates() {
         comparison_suggestions.insert(comparison_suggestions.end(),
                                       result.ComparisonSuggestions().begin(),
                                       result.ComparisonSuggestions().end());
-        if (current_level_number_ >= fds_->GetNumAttributes()) {
+        if (current_level_number_ == max_lhs) {
```
wildsor, line 301: The invalidated FDs get removed during validation above

polyntsov (approved):

PR #715, opened 2026-04-03
Avoid bitset copies in HyFD Validator

wildsor:


src/core/algorithms/fd/hyfd/validator.cpp
```diff
@@ -217,8 +217,8 @@ Validator::FDValidations Validator::ProcessHigherLevel(LhsPair const& lhs_pair)
                                                       *compressed_records_, lhs, rhs, first_attr);
     lhs.set(first_attr);
 
-    rhs &= ~valid_rhss;
-    vertex->SetFds(valid_rhss);
+    rhs -= valid_rhss;
```
wildsor, line 220: https://www.boost.org/doc/libs/latest/libs/dynamic_bitset/doc/html/dynamic_bitset/reference/boost/dynamic_bitset/operator_minus_eq.html

ol-imorozko (approved):

PR #721, opened 2026-04-11
Move compile DB generation to clang-tidy action

wildsor:
Generating compile_commands.json in a separate step can lead to mismatches between Python versions on the runner and in the action's container.


BUYT-1:
Seems to work in https://github.com/Desbordante/desbordante-core/actions/runs/24289309095/job/70923897100?pr=681

Tydik42 (approved):

PR #722, opened 2026-04-11
Actions job for cmake-format

Dmitry909:
Add actions job for cmake-format linter for cmake files. It uses existed config, I didn't change the config. The jobs searches all CMakeLists.txt and *cmake files (expecting one), and check their style.

Firstly, the job failed, then I linted files, and the job passed. So, the job works.


.github/workflows/check-codestyle.yml
```diff
@@ -94,6 +94,43 @@ jobs:
...
+      - name: Install cmake-format
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y cmake-format
+
+      - name: Check CMake formatting
+        run: |
+          set +e
+
+          files=$(find . -name "CMakeLists.txt" -o -name "*.cmake" | grep -v -e "cmake/CPM.cmake")
```
p-senichenkov: Maybe it's better to keep a separate list, like this:
```
ingore_list='cmake/CPM.cmake'
files=$(find ... | grep -v -e $ignore_list)
```
?
In this case the comment won't be needed

p-senichenkov: It's recommended to use single quotes when you don't need substitutions, and wrap variables in double quotes:

```suggestion
          files="$(find . -name 'CMakeLists.txt' -o -name '*.cmake' | grep -v -e 'cmake/CPM.cmake')"
```

.github/workflows/check-codestyle.yml
```diff
@@ -94,6 +94,43 @@ jobs:
...
+          cmake-format --check $files
+
+          if [ $? -ne 0 ]; then
+            echo ":x: CMake files are not properly formatted" >> $GITHUB_STEP_SUMMARY
+            echo "Run 'cmake-format -i' on the following files:" >> $GITHUB_STEP_SUMMARY
+            echo "\`\`\`" >> $GITHUB_STEP_SUMMARY
+            echo "$files" >> $GITHUB_STEP_SUMMARY
+            echo "\`\`\`" >> $GITHUB_STEP_SUMMARY
+            exit 1
+          fi
```
p-senichenkov: If you want to print file list to step summary, let's use `for` loop so that only files with bad formatting appear in the list:
```
BAD_FILES=''
for file in "$files"; do
    cmake-format --check "$file"
    if [ $? -ne 0 ]; then
        BAD_FILES="$BAD_FILES\n$file"
    fi
done
```

Another option is to print the whole formatter output, like in `clang-format` action. It's also acceptable. But I'm not sure if `cmake-format` has "print only diff" mode

p-senichenkov (changes requested):

.github/workflows/check-codestyle.yml
Dmitry909: done, thanks

Dmitry909: Thanks, replaced. Also I replaced in existing job `clang-format-check`, where there is no subsitution (including echos)

.github/workflows/check-codestyle.yml
Dmitry909: I did the option with the same style of output, but only bad files. Good idea, thanks

.github/workflows/check-codestyle.yml
```diff
@@ -94,6 +94,57 @@ jobs:
...
+          set +e
+
+          ingore_list=(
+            'cmake/CPM.cmake'
+          )
+
+          grep_args='-v'
+          for f in "${ingore_list[@]}"; do
+            grep_args="$grep_args -e $f"
+          done
```
p-senichenkov: I didn't notice that `grep` doesn't support passing a list directly.

Probably, a separate file will be bit more readable:
```
cat<<EOF>ignore_list
cmake/CMP.cmake
EOF

files="$(... | grep -f ignore_list)"
```
but I'm not sure. It's up to you

Dmitry909: I actually didn't understand this syntax for cat at first, so it seems to me, that the way with a variable is more readable. If would prefer to use my latest variant. So, okay, let's leave it

p-senichenkov: FYI: it's [heredoc](https://en.wikipedia.org/wiki/Here_document)

.github/workflows/check-codestyle.yml
```diff
@@ -94,6 +94,57 @@ jobs:
...
+
+          files="$(find . -name 'CMakeLists.txt' -o -name '*.cmake' | grep -v -e 'cmake/CPM.cmake')"
```
p-senichenkov: Looks like you've forgot to use `grep_args` here

BUYT-1: Why don't we actually format it too, though?
```suggestion
          ignore_list=(
            'cmake/CPM.cmake' # This file is copied verbatim from the CPM.cmake docs
          )

          find \( -name 'CMakeLists.txt' -o -name '*.cmake' \) $(printf ' -not -path %s' "${ignore_list[@]}")
```

.github/workflows/check-codestyle.yml
```diff
@@ -94,6 +94,57 @@ jobs:
...
+
+          bad_files=''
+          for file in "$files"; do
+              cmake-format --check "$file"
+              if [ $? -ne 0 ]; then
+                  bad_files="$bad_files\n$file"
+              fi
+          done
+
+          if [ $? -ne 0 ]; then
```
BUYT-1: I think you're supposed to check `bad_files` here

.github/workflows/check-codestyle.yml
```diff
@@ -94,6 +94,57 @@ jobs:
...
+          files="$(find . -name 'CMakeLists.txt' -o -name '*.cmake' | grep -v -e 'cmake/CPM.cmake')"
+
+          if [ -z "$files" ]; then
+            echo '### No CMake files to check :heavy_check_mark:' >> $GITHUB_STEP_SUMMARY
+            exit 0
+          fi
+
+          bad_files=''
+          for file in "$files"; do
+              cmake-format --check "$file"
```
BUYT-1, line 134: Use `cmake-format "$file" | diff "$file" -` to output the changes

BUYT-1 (changes requested):

.github/workflows/check-codestyle.yml
p-senichenkov: I'm not sure that it's a good idea to replace a little overly verbose, but clear loop with a Perl-style expression, which takes several minutes to understand😅

.github/workflows/check-codestyle.yml
p-senichenkov: Does someone actually look at these diffs? If a file has bad formatting, you'll need to run `cmake-format` locally anyway. I don't think that someone fixes formatting manually. So, do we need to print diffs here (except for action debugging purposes)?

.github/workflows/check-codestyle.yml
Dmitry909: I agree with @p-senichenkov

.github/workflows/check-codestyle.yml
Dmitry909: yes, thanks

.github/workflows/check-codestyle.yml
Dmitry909: thanks, I added diffs below the list of files. Please, see the last failed and last successfull cmake-format runs in Actions tab, the output is pretty cute.

.github/workflows/check-codestyle.yml
Dmitry909: yes, thanks

BUYT-1: Why don't we format CPM's CML too?

.github/workflows/check-codestyle.yml
BUYT-1: > Does someone actually look at these diffs? If a file has bad formatting, you'll need to run `cmake-format` locally anyway. I don't think that someone fixes formatting manually. So, do we need to print diffs here (except for action debugging purposes)?

I think having a quick look at what needs to be formatted is nice to have. Sometimes it is faster to do it manually. Also, you can spot if it suggests something weird right on the action's page.

.github/workflows/check-codestyle.yml
BUYT-1: > I'm not sure that it's a good idea to replace a little overly verbose, but clear loop with a Perl-style expression, which takes several minutes to understand😅

Okay, I took it out of an LLM and didn't check it properly. It inserted some details that don't really matter. Fixed it.

All it does is filter the files with `find` itself by sticking explicit `-not -path $path` to the end of the expression.

p-senichenkov: I thought about using `-not`, but I didn't realize that it doesn't require `-and`s. This way it's definitely better than `grep`.
To add a comment about each file being ignored is also a good idea

p-senichenkov: > Why don't we format CPM's CML too?

Actually, both options have their pros and cons. Personally I would prefer to keep the file as is, since we don't modify it, except for formatting

.github/workflows/check-codestyle.yml
p-senichenkov: > thanks, I added diffs below the list of files. Please, see the last failed and last successfull cmake-format runs in Actions tab, the output is pretty cute.

Okay, the output really looks great. The only thing I'd suggest is to remove first file list from the output, since they are printed with diffs anyway.
Note that the most common case is a couple of CMake files in a single PR, each of them is smaller than 10 lines, so it's unlikely that there will be a lot of `diff` output.
If one changes a lot of CMake files, most probably, they've ran `cmake-format` locally (inside the IDE)

.github/workflows/check-codestyle.yml
BUYT-1: What are the cons?

p-senichenkov: I've ran `cmake-format` on `cmake/CPM.cmake`, and it's already formatted almost properly. So, I don't mind removing it from ignore list.
In that case, the whole "ignore list" logic should be removed. It won't be hard to implement one if needed (but, most probably, we won't need it)

Dmitry909: I run `cmake-format --check cmake/CPM.cmake` and it reports that the file is not properly formatted. So, we need to keep the ignore list.

p-senichenkov: Yeah, but the only thing it complains are too thin tabs (let me know if I'm wrong).
I mean, we should format it, and *then* remove all the ignore logic. I think, @BUYT-1 suggests the same

.github/workflows/check-codestyle.yml
Dmitry909: Deleted list of bad files, thanks! Checked with and without non-formatted files, works correctly

.github/workflows/check-codestyle.yml
Dmitry909: Ah, ok, understood. Formatted this file and removed the ignore, thanks.

wildsor (changes requested):
LGTM, just squash the commits

p-senichenkov (approved):

wildsor (approved):
Would be perfect if the quote change in .github/workflows/check-codestyle.yml was in a separate commit, but good enough for now.

.github/workflows/check-codestyle.yml
```diff
@@ -79,6 +94,50 @@ jobs:
...
+                  if [ -n "$diff_output" ]; then
+                      bad_diffs="$bad_diffs\n\n:x: $file\n\`\`\`diff\n$diff_output\n\`\`\`"
+                  fi
+              fi
+          done <<< "$files"
+
+          if [ -n "$bad_diffs" ]; then
+            echo ':x: CMake files are not properly formatted' >> $GITHUB_STEP_SUMMARY
+            echo 'Run "cmake-format -i" on the following files:' >> $GITHUB_STEP_SUMMARY
+            echo -e "$bad_diffs" >> $GITHUB_STEP_SUMMARY
```
wildsor: It just shows diffs with filenames now, the phrasing isn't quite right.
```suggestion
            echo ':x: CMake files are not properly formatted, run "cmake-format -i" on the files specified:' >> $GITHUB_STEP_SUMMARY
            echo -e "$bad_diffs" >> $GITHUB_STEP_SUMMARY
```

Dmitry909: I didn't really understand why phrasing is so important, they look familiar to me, but done.

PR #728, opened 2026-04-20
Implemented class ADDVerifier

unfadingDawn:
Implemented class ADDVerifier.
Implemented tests and Python bindings for this class.
Implemented code of the ADD validation example on Python.


src/core/algorithms/dd/add_verifier/add_verifier.cpp
```diff
@@ -0,0 +1,91 @@
+#include "core/algorithms/dd/add_verifier/add_verifier.h"
+
+#include "core/algorithms/dd/dd_verifier/dd_verifier.h"
```
VinogradovDmitri, line 3: Its does not suppose to be here because u have it in header. I write it only because i want u to fix `CMakeLists.txt` (read about PUBLIC in [wiki](https://github.com/Desbordante/desbordante-core/wiki/Development:-target%E2%80%90based-CMake-quick-start-guide#core))

src/core/algorithms/dd/dd_verifier/dd_verifier.h
```diff
@@ -38,13 +37,13 @@ class DDVerifier : public Algorithm {
     double CalculateDistance(model::ColumnIndex column_index,
                              std::pair<std::size_t, std::size_t> const &tuple_pair) const;
 
-    void CheckDFOnRhs(std::vector<std::pair<std::size_t, std::size_t>> const &lhs);
+    virtual void CheckDFOnRhs(std::vector<std::pair<std::size_t, std::size_t>> const& lhs);
 
     void VerifyDD();
```
VinogradovDmitri, line 45: why didn't this `virtual`?

src/core/algorithms/dd/add_verifier/add_verifier.cpp
```diff
@@ -0,0 +1,91 @@
...
+            throw config::ConfigurationError("Satisfaction threshold out of range");
+    };
+
+    RegisterOption(
+            Option{&satisfaction_threshold_, kSatisfactionThreshold, kDSatisfactionThreshold, 0.}
+                    .SetValueCheck(check_threshold));
+}
+
+void ADDVerifier::CheckDFOnRhs(std::vector<std::pair<std::size_t, std::size_t>> const& lhs) {
+    double min_dist = -1.;
```
VinogradovDmitri:
```suggestion
void ADDVerifier::CheckDFOnRhs(std::vector<std::pair<std::size_t, std::size_t>> const& lhs) {
    if (lhs.empty()) {
       error_ = 0.0; // or 1.0 idk 
       return;
    }
    double min_dist = -1.;
```

src/core/algorithms/dd/add_verifier/add_verifier.cpp
```diff
@@ -0,0 +1,91 @@
...
+        }
+    };
+
+    std::ranges::for_each(dd_.left, check_constraint);
+    std::ranges::for_each(dd_.right, check_constraint);
+    if (dd_.right.size() > 1) {
+        throw std::invalid_argument(
+                "The RHS of an ADD must have a differential function on a single attribute.");
+    }
+}
```
VinogradovDmitri, line 92: i suggest do all exceptions in begins of methods so it will not waste time

src/core/algorithms/dd/add_verifier/add_verifier.cpp
```diff
@@ -0,0 +1,91 @@
...
+        for (auto const column_index : rhs_column_indices_) {
+            double const dif = CalculateDistance(column_index, pair);
+            if (!curr_constraint->constraint.Contains(dif)) {
+                highlights_.emplace_back(column_index, pair, dif);
+                ++num_error_rhs_;
+            }
+            min_dist = min_dist == -1 ? dif : std::min(min_dist, dif);
+            ++curr_constraint;
+        }
+    }
```
VinogradovDmitri, line 46: i believe -O3 optimize this but probably u should change 

```suggestion
void ADDVerifier::CheckDFOnRhs(std::vector<std::pair<std::size_t, std::size_t>> const& lhs) {
    double min_dist = std::numeric_limits<double>::max();
    for (auto const& pair : lhs) {
        auto curr_constraint = dd_.right.cbegin();
        for (auto const column_index : rhs_column_indices_) {
            double const dif = CalculateDistance(column_index, pair);
            if (!curr_constraint->constraint.Contains(dif)) {
                highlights_.emplace_back(column_index, pair, dif);
                ++num_error_rhs_;
            }
            min_dist = std::min(min_dist, dif);
            ++curr_constraint;
        }
    }
```

src/core/algorithms/dd/add_verifier/add_verifier.cpp
VinogradovDmitri: division by zero

src/core/algorithms/dd/add_verifier/add_verifier.cpp
```diff
@@ -0,0 +1,91 @@
...
+    std::size_t num_pairs_with_min_dist = 0;
+    for (auto const& pair : lhs) {
+        auto curr_constraint = dd_.right.cbegin();
+        for (auto const column_index : rhs_column_indices_) {
+            double const dif = CalculateDistance(column_index, pair);
+            if (dif == min_dist) {
```
VinogradovDmitri: i don't know about that, we should compare `double` like (`dif` - `min_dist` < `epsilon`)

VinogradovDmitri (changes requested):
That's enough for now, maybe I'll look again later

src/core/algorithms/dd/dd_verifier/dd_verifier.h
unfadingDawn: Looks like a merge/sync issue during the squash. I've double-checked, and the function is marked as virtual in the final commit.

src/core/algorithms/dd/add_verifier/add_verifier.cpp
```diff
@@ -0,0 +1,94 @@
...
+    }
+}
+
+bool ADDVerifier::DDHolds() const {
+    return 1 - error_ >= satisfaction_threshold_;
+}
+
+void ADDVerifier::MakeExecuteOptsAvailable() {
+    using namespace config::names;
+    MakeOptionsAvailable({kDDString, kSatisfactionThreshold});
```
VinogradovDmitri, line 71: i don't know how we code `verifier`, but i think we should reuse base class opts

```suggestion
    using namespace config::names;
    DDVerifier::MakeExecuteOptsAvailable();
    MakeOptionsAvailable({config::names::kSatisfactionThreshold});
```

src/core/algorithms/dd/add_verifier/add_verifier.cpp
```diff
@@ -0,0 +1,94 @@
...
+        }
+        if (constraint.constraint.lower_bound < 0 || constraint.constraint.upper_bound < 0) {
+            throw std::invalid_argument("Constraint bounds cannot be negative for column: " +
+                                        constraint.column_name);
+        }
+    };
+
+    std::ranges::for_each(dd_.left, check_constraint);
+    std::ranges::for_each(dd_.right, check_constraint);
+}
```
VinogradovDmitri, line 92: Same, reuse base methods

```suggestion
void ADDVerifier::CheckCorrectnessDd() const {
    DDVerifier::CheckCorrectnessDd();
    
    if (dd_.right.size() != 1) {
        throw std::invalid_argument(
                "The RHS of an ADD must have a differential function on a single attribute.");
    }
}
```

VinogradovDmitri (changes requested):

PR #729, opened 2026-04-21
Add CIND Cure Algorithm with Example and Tests

ALanovaya:
## Summary

This PR adds the **Cure** condition mining algorithm to the CIND framework, complementing the existing `CINDERELLA` and `PLI-CIND` miners.

## Background

A **Conditional Inclusion Dependency (CIND)** is an IND that holds only on the subset of rows matching a pattern condition on non-inclusion attributes (see #664 for the formal definition). Desbordante already provides `CINDERELLA` (itemset-based) and `PLI-CIND` (PLI-based) condition miners, both controlled by validity/completeness thresholds. Cure offers a different trade-off: it is controlled by a single support threshold (minimum number of joined tuples) and produces compact patterns with disjunctive RHS values.

## The Cure algorithm

O. Curé (2012) discovers conditional patterns in two phases:

1. **Discovery** — for each approximate IND, enumerate pairs of (LHS conditional attribute, RHS conditional attribute) and hash-join rows on the inclusion key. For each joined LHS/RHS value pair, count co-occurrences and keep those with count ≥ `support`.

2. **Minimal cover** — merge patterns sharing the same `(LHS conditional attribute, LHS value)` key into a single tableau row. When multiple RHS values occur for the same LHS key, they are folded into a comma-separated disjunction in the corresponding RHS slot.

Per-pattern validity and completeness are derived from the pattern support and the total number of joined tuples.

### Reference

- O. Curé. *"Improving the Data Quality of Drug Databases using Conditional Dependencies and Ontologies"*. ACM JDIQ 4(1):20, 2012.

## Changes

- **`CureCind`** (`core/algorithms/cind/condition_miners/cure_cind.{h,cpp}`) — new condition miner implementing the two-phase Cure algorithm.
- **`AlgoType::cure_cind`** — added alongside existing `cinderella` and `pli_cind` in `core/algorithms/cind/types.h`.
- **`CindAlgorithm`** dispatch — instantiates `CureCind` when `algo_type = cure_cind` (`core/algorithms/cind/cind_algorithm.cpp`).
- **`min_support_`** member added to `CindMiner` base class (used only by Cure).
- **`support` config option** (`core/config/conditions/support/option.{h,cpp}`) — new minimum-tuples threshold, default `2`.
- **Tests** — three parameterized cases (`support` in {1, 2, 3}) added to `src/tests/unit/test_cind_algorithms.cpp`; a `cind.algos` test target is also wired up in the unit test CMakeLists.
- **Example** — `examples/basic/mining_cind_cure.py` runs CINDERELLA and Cure on the same DBpedia German/English persons dataset and prints a side-by-side comparison.


src/core/algorithms/cind/condition_miners/cind_miner.cpp
```diff
@@ -0,0 +1,73 @@
+#include "cind_miner.h"
```
wildsor, line 1: Use the full path for all headers, check other files too
```suggestion
#include "core/algorithms/cind/condition_miners/cind_miner.h"
```

wildsor (changes requested):

src/core/config/CMakeLists.txt
```diff
@@ -1,8 +1,11 @@
+file(GLOB_RECURSE CONDITIONS_SRCS conditions/*.cpp)
```
wildsor: Do not glob for files in CMLs, it makes it more difficult to rebuild the project when you need to add more files because CMake doesn't know we need to compile them.

src/core/model/table/encoded_tables.h
```diff
@@ -0,0 +1,40 @@
...
+        : value_dictionary_(std::make_shared<ValueDictionary>()) {
+        for (size_t table_id = 0; table_id < input_tables.size(); ++table_id) {
+            input_tables[table_id]->Reset();
+            tables_.push_back(ColumnEncodedRelationData::CreateFrom(input_tables[table_id],
+                                                                    table_id, value_dictionary_));
+        }
+    }
+
+    ColumnEncodedRelationData& GetTable(size_t index) {
+        assert(index < tables_.size());
```
wildsor, line 23: These asserts are already present in std::vector. Please remove them.

wildsor (changes requested):

src/core/algorithms/cind/condition_miners/cure_cind.cpp
```diff
@@ -0,0 +1,208 @@
...
+};
+
+std::vector<int> MakeKey(std::size_t row, AttrsType const& cols) {
+    std::vector<int> key;
+    key.reserve(cols.size());
+    for (auto const* c : cols) {
+        key.push_back(c->GetValue(row));
+    }
+    return key;
+}
```
Oddin60F: I think `VecIntHash` and `MakeKey()` should be moved to a separate namespace `cind::utils`. This will avoid code duplication in the verifier and all mining algorithms. Also, the `MakeKey()` logic is often implicitly used in the `pli_cind` and `cinderella` algorithms and can be replaced by an explicit call to this function.

src/core/algorithms/cind/condition_miners/cure_cind.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+        if (it == cover.end()) {
+            CoverEntry entry;
+            entry.values.resize(total_attrs, kAnyValue);
+            entry.values[p.lhs_attr_idx] =
+                    attrs.lhs_conditional[p.lhs_attr_idx]->DecodeValue(p.lhs_value);
+            entry.values[lhs_cond_size + p.rhs_attr_idx] =
+                    attrs.rhs_conditional[p.rhs_attr_idx]->DecodeValue(p.rhs_value);
+            entry.support = p.support;
+            cover.emplace(key, std::move(entry));
```
Oddin60F:
```suggestion
            cover.emplace(std::move(key), std::move(entry));
```

src/core/algorithms/cind/condition_miners/cure_cind.cpp
```diff
@@ -0,0 +1,208 @@
...
+                entry.values[rhs_pos] += ", " + rhs_decoded;
+            }
+            entry.support += p.support;
+        }
+    }
+
+    std::vector<Condition> conditions;
+    conditions.reserve(cover.size());
+
+    for (auto& [key, entry] : cover) {
```
Oddin60F:
```suggestion
    for (auto& [_, entry] : cover) {
```

src/core/algorithms/cind/condition_miners/cure_cind.cpp
```diff
@@ -0,0 +1,208 @@
...
+CIND CureCind::ExecuteSingle(model::IND const& aind) {
+    auto attrs = ClassifyCureAttributes(aind);
+
+    auto patterns = DiscoverPatterns(attrs);
+
+    // Total joined tuples used as denominator for validity/completeness.
+    std::size_t total_joined = 0;
+    for (auto const& p : patterns) {
+        total_joined += p.support;
+    }
```
Oddin60F: IMHO, it is better to transfer the calculation to the `MinimalCover()` method since the result is used only there and is not used further.

src/core/algorithms/cind/condition_miners/cure_cind.h
```diff
@@ -0,0 +1,34 @@
+#pragma once
+
+#include "core/algorithms/cind/condition_miners/cind_miner.h"
+
+namespace algos::cind {
+
+class CureCind final : public CindMiner {
+public:
+    explicit CureCind(config::InputTables& input_tables);
+
```
Oddin60F: Our order of filling in fields in classes is `private`/`protected`/`public`.

src/core/config/names.h
```diff
@@ -142,6 +142,7 @@ constexpr auto kValidity = "validity";
 constexpr auto kCompleteness = "completeness";
 constexpr auto kConditionType = "condition_type";
 constexpr auto kAlgoType = "algo_type";
+constexpr auto kSupport = "support";
```
Oddin60F: IMHO, it would be better to designate the parameter as `kCindMinSupport`, because support is a very common indicator for many primitives, and they should be separated.

src/python_bindings/cind/bind_cind.cpp
```diff
@@ -76,7 +76,8 @@ void BindCind(py::module_& main_module) {
                      return result;
                  })
             .def("get_condition_attributes",
-                 [](CIND const& cind) { return VectorToTuple(cind.conditional_attributes); });
+                 [](CIND const& cind) { return VectorToTuple(cind.conditional_attributes); })
+            .def("get_ind_string", [](CIND const& cind) { return cind.ind.ToLongString(); });
```
Oddin60F, line 70: From the point of view of binding requirements _(you can look at Wiki/Development‐Features)_, it would be more correct if the user could get the full IND object embedded in the CIND. You can simply add the `get_ind` method so that the user can do whatever he wants with the embedded IND.

Oddin60F: Please note that due to the fact that the embedded IND is stored by reference in the CIND class, accessing it after destroying the algorithm class will result in an error. It seems worth rewriting the code so that IND is copied or moved to the CIND object.

src/python_bindings/cind/bind_cind.cpp
Oddin60F: A couple of days ago, a common function appeared in `py_util.h` for such code. Use it.

src/tests/unit/test_cind_algorithms.cpp
```diff
@@ -106,4 +106,46 @@ INSTANTIATE_TEST_SUITE_P(
...
+    auto mp = MakeCureParams(p.support);
+    auto cind_algo = algos::CreateAndLoadAlgorithm<algos::cind::CindAlgorithm>(mp);
+    cind_algo->Execute();
+    ASSERT_FALSE(cind_algo->CINDList().empty());
+    size_t total_conditions = 0;
+    for (auto const& cind : cind_algo->CINDList()) {
+        total_conditions += cind.ConditionsNumber();
+    }
+    ASSERT_EQ(total_conditions, p.expected_total_conditions);
+}
```
Oddin60F, line 161: It's my mistake that I missed this last time, but it's worth rewriting the tests so that the conditions themselves are checked, too, and not just their number. You can see how it's done, for example, in `test_apriori.cpp` or `test_cfd_algos.cpp`

src/core/algorithms/cind/condition_miners/cure_cind.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+            for (std::size_t lhs_row = 0; lhs_row < lhs_rows; ++lhs_row) {
+                auto key = MakeKey(lhs_row, attrs.lhs_inclusion);
+                auto it = rhs_index.find(key);
+                if (it == rhs_index.end()) continue;
+
+                int const lv = lhs_attr->GetValue(lhs_row);
+                for (std::size_t rhs_row : it->second) {
+                    int const rv = rhs_attr->GetValue(rhs_row);
+                    pair_counts[{lv, rv}]++;
```
Oddin60F:
```suggestion
                    ++pair_counts[{lv, rv}];
```

src/core/algorithms/cind/condition_miners/cure_cind.cpp
```diff
@@ -0,0 +1,208 @@
...
+        auto const* lhs_attr = attrs.lhs_conditional[li];
+
+        for (std::size_t ri = 0; ri < attrs.rhs_conditional.size(); ++ri) {
+            auto const* rhs_attr = attrs.rhs_conditional[ri];
+
+            std::unordered_map<std::pair<int, int>, std::size_t, PairIntHash> pair_counts;
+
+            for (std::size_t lhs_row = 0; lhs_row < lhs_rows; ++lhs_row) {
+                auto key = MakeKey(lhs_row, attrs.lhs_inclusion);
+                auto it = rhs_index.find(key);
```
Oddin60F: IMHO, it's better this way because we don't need the key anymore.
```suggestion
                auto it = rhs_index.find(MakeKey(lhs_row, attrs.lhs_inclusion));
```

Oddin60F (changes requested):

src/core/algorithms/cind/condition_miners/CMakeLists.txt
```diff
@@ -1,4 +1,6 @@
 set(NAME cind.miners)
```
wildsor: This thing wasn't correct, we have at least a target for each algorithm. Make yours a separate one and do some renaming for this one.

src/core/algorithms/cind/condition_miners/cure_cind.cpp
```diff
@@ -0,0 +1,190 @@
...
+    }
+};
+}  // namespace
+
+CureCind::CureCind(config::InputTables& input_tables) : CindMiner(input_tables) {}
+
+CureCind::CureAttributes CureCind::ClassifyCureAttributes(model::IND const& aind) const {
+    CureAttributes result;
+
+    auto const& lhs = aind.GetLhs();
```
wildsor: Do not use `auto` to save some typing, only use it when the type is obvious, see https://google.github.io/styleguide/cppguide.html#Type_deduction (that's for all code)

src/core/algorithms/cind/cind_algorithm.cpp
```diff
@@ -58,11 +62,13 @@ void CindAlgorithm::RegisterCindMinerOptions() {
     RegisterOption(config::kValidityOpt(&cind_miner_->min_validity_));
     RegisterOption(config::kCompletenessOpt(&cind_miner_->min_completeness_));
     RegisterOption(config::kConditionTypeOpt(&cind_miner_->condition_type_));
+    RegisterOption(config::kCindMinSupportOpt(&cind_miner_->min_support_));
```
wildsor: Is this relevant for the other mining algorithm? If no, don't add the option here, add it in your algorithm.

src/core/algorithms/cind/cind_algorithm.cpp
```diff
@@ -48,6 +50,8 @@ void CindAlgorithm::LoadDataInternal() {
 void CindAlgorithm::CreateCindMinerAlgo() {
     if (algo_type_._value == AlgoType::cinderella) {
         cind_miner_ = std::make_unique<Cinderella>(spider_algo_->input_tables_);
+    } else if (algo_type_._value == AlgoType::cure_cind) {
```
wildsor, line 55: Why is this here?

src/core/config/conditions/support/option.cpp
```diff
@@ -0,0 +1,14 @@
+#include "core/config/conditions/support/option.h"
+
+#include "core/config/exceptions.h"
+#include "core/config/names_and_descriptions.h"
+
+namespace config {
+using names::kCindMinSupport, descriptions::kDCindMinSupport;
+extern CommonOption<unsigned int> const kCindMinSupportOpt{
```
wildsor: If this option is not relevant for other CIND algorithms, create it directly instead of making another CommonOption

src/tests/unit/test_cind_algorithms.cpp
```diff
@@ -106,4 +106,83 @@ INSTANTIATE_TEST_SUITE_P(
...
+
+// clang-format off
```
wildsor: Why are you turning off clang-format here?

wildsor (changes requested):

src/core/algorithms/cind/cind_algorithm.cpp
ALanovaya: Cure uses the same pipeline as Cinderella and Pli cind. Whats wrong?

src/tests/unit/test_cind_algorithms.cpp
ALanovaya: legacy from another author. i ll fix it

src/core/algorithms/cind/condition_miners/cure_cind.cpp
ALanovaya: do you mean all code in this pr?

wildsor: yes

src/core/algorithms/cind/cind_algorithm.cpp
wildsor: Okay, I see how it is now. I'm not a fan of how this works now, but that's not something to fix in this PR.

src/core/algorithms/cind/cind_algorithm.cpp
```diff
@@ -58,11 +64,29 @@ void CindAlgorithm::RegisterCindMinerOptions() {
...
+        config::Option<unsigned int> support_opt{&cure->min_support_,
+                                                 config::names::kCindMinSupport,
+                                                 config::descriptions::kDCindMinSupport, 2u};
+        support_opt.SetValueCheck([](unsigned int support) {
+            if (support < 1) {
+                throw config::ConfigurationError("ERROR: support must be >= 1.");
+            }
+        });
+        RegisterOption(std::move(support_opt));
+    }
```
wildsor, line 79: oh no

This is not how options are supposed to be used at all. But you'd have to remake everything here to fix it. Let's leave it be, I guess

src/core/algorithms/cind/cind_algorithm.cpp
```diff
@@ -58,11 +64,29 @@ void CindAlgorithm::RegisterCindMinerOptions() {
...
+        auto* cure = static_cast<CureCind*>(cind_miner_.get());
+        config::Option<unsigned int> support_opt{&cure->min_support_,
+                                                 config::names::kCindMinSupport,
+                                                 config::descriptions::kDCindMinSupport, 2u};
+        support_opt.SetValueCheck([](unsigned int support) {
+            if (support < 1) {
+                throw config::ConfigurationError("ERROR: support must be >= 1.");
+            }
+        });
+        RegisterOption(std::move(support_opt));
```
Oddin60F, line 78: For the consistency of the code, it is better to write like this.
```suggestion
        auto* cure = static_cast<CureCind*>(cind_miner_.get());                                         
        auto support_check = [](usnigned int support) {
            if(support < 1){
                throw config::ConfigurationError("Support must be >= 1.");
            }
        };
        RegisterOption(config::Option<unsigned int>{&cure->min_support_, config::names::kCindMinSupport, config::descriptions::kDCindMinSupport, 2u}.SetValueCheck(std::move(support_check)));
```

src/core/algorithms/cind/condition_miners/cure_cind.h
```diff
@@ -0,0 +1,40 @@
...
+#include "core/algorithms/cind/condition_miners/cind_miner.h"
+
+namespace algos::cind {
+
+class CureCind final : public CindMiner {
+private:
+    friend class CindAlgorithm;
+
+    uint min_support_{2};
```
Oddin60F, line 13: `uint` from `sys/types.h` is not a standard C++ type and, in theory, its use may lead to compilation errors for other platforms. Just replace it with a standard `unsigned int`.

src/core/algorithms/cind/condition_miners/cure_cind.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+            if (entry.values[rhs_pos] == kAnyValue) {
+                entry.values[rhs_pos] = rhs_decoded;
+            } else if (entry.values[rhs_pos].find(rhs_decoded) == std::string::npos) {
+                // Disjunction: append with comma
+                entry.values[rhs_pos] += ", " + rhs_decoded;
+            }
+            entry.support += p.support;
+        }
+    }
```
Oddin60F, line 158: It seems like it's more understandable what's going on.
```suggestion
    std::size_t total_joined = 0;
    for (PatternPair const& p : patterns) {
        total_joined += p.support;
    }
    
    struct CoverEntry {
        std::vector<std::string> values;
        std::size_t support{0};
        explicit CoverEntry(std::size_t total_attrs): values(total_attrs, kAnyValue) {}
    };

    using CoverKey = std::pair<std::size_t, int>;
    std::unordered_map<CoverKey, CoverEntry, PairIntHash> cover;

    for (PatternPair const& p : patterns) {
        CoverKey key{p.lhs_attr_idx, p.lhs_value};
        auto [it, inserted] = cover.try_emplace(std::move(key), total_attrs);
        CoverEntry& entry = it->second;
        if (inserted) {
             entry.values[p.lhs_attr_idx] =
                    attrs.lhs_conditional[p.lhs_attr_idx]->DecodeValue(p.lhs_value);
        }
        std::size_t const rhs_pos = lhs_cond_size + p.rhs_attr_idx;
        std::string const rhs_decoded =
                    attrs.rhs_conditional[p.rhs_attr_idx]->DecodeValue(p.rhs_value);
        if (entry.values[rhs_pos] == kAnyValue) {
              entry.values[rhs_pos] = rhs_decoded;
         } else if (entry.values[rhs_pos].find(rhs_decoded) == std::string::npos) {
               // Disjunction: append with comma
               entry.values[rhs_pos] += ", " + rhs_decoded;
         }
        
        entry.support += p.support;
    }
```

Oddin60F (changes requested):

src/core/algorithms/cind/condition_miners/cure_cind.cpp
```diff
@@ -0,0 +1,205 @@
...
+namespace {
+struct PairIntHash {
+    std::size_t operator()(std::pair<int, int> const& p) const noexcept {
+        std::size_t seed = 0;
+        boost::hash_combine(seed, p.first);
+        boost::hash_combine(seed, p.second);
+        return seed;
+    }
+};
+}  // namespace
```
Oddin60F: Delete and use your `utils.cpp`.

src/core/algorithms/cind/condition_miners/cure_cind.cpp
```diff
@@ -0,0 +1,205 @@
...
+        std::vector<std::string> values;
+        std::vector<std::unordered_set<std::string>> rhs_seen;
+        std::size_t support{0};
+
+        CoverEntry(std::size_t total_attrs, std::size_t rhs_cond_size)
+            : values(total_attrs, kAnyValue), rhs_seen(rhs_cond_size) {}
+    };
+
+    using CoverKey = std::pair<std::size_t, int>;
+    std::unordered_map<CoverKey, CoverEntry, PairIntHash> cover;
```
Oddin60F, line 151: I think it’s worth moving the structure definition and the alias to a separate file. You shouldn’t define classes and structures in `.cpp`.

src/python_bindings/cind/bind_cind.cpp
Oddin60F, line 1: Our Python classes lack the `__repr__` method for printing from collections and point 6) [from here](https://github.com/Desbordante/desbordante-core/wiki/Development‐Features#что-значит-разработать-байндинги).

src/core/algorithms/cind/condition_miners/cure_cind.cpp
```diff
@@ -0,0 +1,205 @@
...
+struct PairIntHash {
+    std::size_t operator()(std::pair<int, int> const& p) const noexcept {
+        std::size_t seed = 0;
+        boost::hash_combine(seed, p.first);
+        boost::hash_combine(seed, p.second);
+        return seed;
+    }
+};
+}  // namespace
+
```
Oddin60F, line 27: Delete and use your utils.cpp.

Oddin60F (changes requested):

src/core/algorithms/cind/condition_miners/cure_cind.cpp
wildsor: You are going to search the map three times when hitting the `else if` case, use `.find`

src/core/algorithms/cind/condition_miners/cure_cind.cpp
wildsor: > You shouldn’t define classes and structures in `.cpp`.

That's too broad. I'd say classes and structs in anonymous namespace are okay.

I'm not totally against having a struct in a function like this, but it does make the function larger, so perhaps moving to `namespace {` may make sense still.

examples/basic/mining_cind_2.py
```diff
@@ -0,0 +1,344 @@
...
+    print(f"{label}{lines[0]}")
+    cont_pad = " " * len(label)
+    for line in lines[1:]:
+        print(f"{cont_pad}{line}")
+    print(f"{cont_pad}validity = {cond.validity():.3f}, "
+          f"completeness = {cond.completeness():.3f}")
+
+
+def fmt_ind(ind_string):
+    import re
```
wildsor, line 70: Move the import to the top of the file

PR #611, opened 2025-10-15
Implement Domain PAC validation

p-senichenkov:
Implement base classes for Probabilistic Approximate Constratins validation.
Probabilistic Approximate Constraint is a primitve intoduced in "Checks and Balances: Monitoring Data Quality Problems in Network Traffic Databases" by Flip Korn et al. PAC with parameters epsilon and delta means that *Probability that approximate constraint with error=epsilon holds is greater than delta*.

Implement Domain PAC verifier.
Domain PAC on domain D means that *Pr(x in D+-epsilon) > delta*.

Add Domain PAC verifier bindings and examples.

PR #734, opened 2026-04-25
Bindings for user-defined metrics

p-senichenkov:
Add classes for user-defined metrics that allow transparent usage from Python (i. e. user's function gets values directly, not strings or something), without introducing extra overhead in C++.

There are two types of metrics:
- A single-column metric (and a separate option for a set of single-column metrics)
- A multi-column metric, that treats its arguments as vectors

There are three classes provided by default:
- `Py*Metric` -- allows the user to pass a Python function, which will be called on real values
- `Static*Metric` -- C++ metric that converts all values to some fixed type (very convenient in tests!)
- `Dynamic*Metric` -- C++ metric that passes real column type and raw value to the user function, allowing it to do all conversions
Other types of metrics can be introduced by implementing the `IMetric` interface, if some non-standard behavior is required in some algorithm.

Also, there are default metrics for metrizable types (vector metric uses the euclidean distance)

### DIFF
--- a/src/core/config/custom_metric/custom_metric/option.h
+++ b/src/core/config/custom_metric/custom_metric/option.h
@@ -1,0 +1,14 @@
+#pragma once
+
+#include <string_view>
+
+#include "core/config/custom_metric/custom_metric/type.h"
+#include "core/config/descriptions.h"
+#include "core/config/names.h"
+#include "core/config/option.h"
+
+namespace config {
+Option<CustomMetricType> MetricOption(CustomMetricType* value_ptr,
+                                      std::string_view name = names::kCustomMetric,
+                                      std::string_view description = descriptions::kDCustomMetric);
+}  // namespace config
--- a/src/core/config/custom_metric/custom_metrics/option.h
+++ b/src/core/config/custom_metric/custom_metrics/option.h
@@ -1,0 +1,53 @@
+#pragma once
+
+#include <algorithm>
+#include <cstddef>
+#include <functional>
+#include <memory>
+#include <sstream>
+#include <string_view>
+
+#include "core/config/column_index/type.h"
+#include "core/config/custom_metric/custom_metrics/type.h"
+#include "core/config/descriptions.h"
+#include "core/config/exceptions.h"
+#include "core/config/names.h"
+#include "core/config/option.h"
+
+namespace config {
+/// @brief Option for a collection of user-defined metrics
+class MetricsOption {
+private:
+    std::string_view name_;
+    std::string_view description_;
+
+    static CustomMetricsType MakeDefaultMetrics(std::size_t indices_count) {
+        return CustomMetricsType(indices_count, nullptr);
+    }
+
+    static void CheckMetrics(CustomMetricsType const& value, std::size_t indices_count) {
+        if (value.size() != indices_count) {
+            std::ostringstream msg;
+            msg << "Expected " << indices_count << " user-defined metrics, got " << value.size();
+            throw ConfigurationError(msg.str());
+        }
+    }
+
+    /// User can pass @c nullptr to use the default metric explicitly
+    static void NormalizeMetrics(CustomMetricsType& value) {
+        auto default_metric = std::make_shared<util::DefaultCustomMetric>();
+
+        std::ranges::replace_if(value, std::logical_not{}, default_metric);
+    }
+
+public:
+    MetricsOption(std::string_view name = names::kCustomMetrics,
+                  std::string_view description = descriptions::kDCustomMetrics)
+        : name_(name), description_(description) {}
+
+    // NOTE: This option should depend on indices option (see @c SetConditionalOpts)
+    // to properly get column count
+    [[nodiscard]] Option<CustomMetricsType> operator()(
+            CustomMetricsType* value_ptr, std::function<IndexType()> get_col_count) const;
+};
+}  // namespace config
--- a/src/core/util/custom_metric/custom_metric.h
+++ b/src/core/util/custom_metric/custom_metric.h
@@ -1,0 +1,96 @@
+#pragma once
+
+#include <cassert>
+#include <cstddef>
+#include <functional>
+#include <optional>
+#include <sstream>
+#include <utility>
+
+#include "core/config/exceptions.h"
+#include "core/model/types/imetrizable_type.h"
+#include "core/model/types/type.h"
+#include "core/util/export.h"
+
+namespace util {
+/// @brief User-defined metric on a single typed column
+/// Together with @c PyCustomMetric, @c DynamicCustomMetric and @c StaticCustomMetric provides
+/// a convenient user interface (especially, in Python) without extra overhead on conversions
+// NOTE: these objects are wrapped in `shared_ptr`, and this encourages user to use the same object
+// for several columns. Keep it in mind if you are planning to implement some complex internal state
+/// WARN: NULL value is represented by nullptr
+class DESBORDANTE_EXPORT ICustomMetric {
+public:
+    virtual ~ICustomMetric() = default;
+
+    virtual double Dist(model::Type const* type, std::byte const* first,
+                        std::byte const* second) const = 0;
+};
+
+/// @brief Provides a convenient way to define custom metric, when column type is known in advance
+/// WARN: Ignores real column type. Passing incorrect type leads to undefined behaviour
+template <typename ArgType>
+class StaticCustomMetric : public ICustomMetric {
+private:
+    using OptArg = std::optional<ArgType>;
+    using Metric = std::function<double(OptArg const&, OptArg const&)>;
+
+    Metric metric_;
+
+    static OptArg GetValue(std::byte const* value) {
+        if (!value) {
+            return std::nullopt;
+        }
+        return model::Type::GetValue<ArgType>(value);
+    }
+
+public:
+    explicit StaticCustomMetric(Metric metric) : metric_(std::move(metric)) {}
+
+    double Dist(model::Type const*, std::byte const* first,
+                std::byte const* second) const override {
+        return metric_(GetValue(first), GetValue(second));
+    }
+};
+
+/// @brief A custom metric, which uses real column type
+class DynamicCustomMetric : public ICustomMetric {
+private:
+    using Metric = std::function<double(model::Type const*, std::byte const*, std::byte const*)>;
+
+    Metric metric_;
+
+public:
+    explicit DynamicCustomMetric(Metric metric) : metric_(std::move(metric)) {}
+
+    double Dist(model::Type const* type, std::byte const* first,
+                std::byte const* second) const override {
+        return metric_(type, first, second);
+    }
+};
+
+/// @brief A default value for custom metric option
+/// Uses default metric for the type. Works only with metrizable types
+class DefaultCustomMetric : public ICustomMetric {
+private:
+    static model::IMetrizableType const* ConvertType(model::Type const* type) {
+        auto const* metr_type = dynamic_cast<model::IMetrizableType const*>(type);
+        if (!metr_type) {
+            std::ostringstream msg;
+            msg << "Cannot use default metric, because column type " << type->ToString()
+                << " is not metrizable. Consider defining custom metric";
+            throw config::ConfigurationError(msg.str());
+        }
+        return metr_type;
+    }
+
+public:
+    double Dist(model::Type const* type, std::byte const* first,
+                std::byte const* second) const override {
+        if (!first || !second) {
+            return 0;
+        }
+        return ConvertType(type)->Dist(first, second);
+    }
+};
+}  // namespace util
--- a/src/core/util/custom_metric/custom_vector_metric.h
+++ b/src/core/util/custom_metric/custom_vector_metric.h
@@ -1,0 +1,78 @@
+#pragma once
+
+#include <cassert>
+#include <cmath>
+#include <cstddef>
+#include <functional>
+#include <utility>
+#include <vector>
+
+#include "core/config/exceptions.h"
+#include "core/model/types/imetrizable_type.h"
+#include "core/model/types/type.h"
+#include "core/util/export.h"
+
+namespace util {
+/// @brief User-defined metric on a set of columns, that treats values as tuples (vectors)
+/// Together with @c PyCustomVectorMetric and @c CustomVectorMetric provides
+/// a convenient user interface (especially, in Python) without extra overhead on conversions
+// NOTE: these objects are wrapped in `shared_ptr`, and this encourages user to use the same object
+// for several columns. Keep it in mind if you are planning to implement some complex internal state
+/// WARN: NULL value is represented by nullptr
+class DESBORDANTE_EXPORT ICustomVectorMetric {
+protected:
+    using Types = std::vector<model::Type const*>;
+    using Values = std::vector<std::byte const*>;
+
+public:
+    virtual ~ICustomVectorMetric() = default;
+
+    virtual double Dist(Types const& types, Values const& first, Values const& second) const = 0;
+};
+
+/// @brief A custom vector metric, which uses real column types
+class CustomVectorMetric : public ICustomVectorMetric {
+private:
+    using Metric = std::function<double(Types const&, Values const&, Values const&)>;
+
+    Metric metric_;
+
+public:
+    explicit CustomVectorMetric(Metric metric) : metric_(std::move(metric)) {}
+
+    double Dist(Types const& types, Values const& first, Values const& second) const override {
+        return metric_(types, first, second);
+    }
+};
+
+/// @brief A default value for custom vector metric option
+/// Uses euclidean metric and default metrics for individual coordinates.
+/// Works only with metrizable types
+class DefaultCustomVectorMetric : public ICustomVectorMetric {
+private:
+    static model::IMetrizableType const* ConvertType(model::Type const* type) {
+        auto const* metr_type = dynamic_cast<model::IMetrizableType const*>(type);
+        if (!metr_type) {
+            std::ostringstream msg;
+            msg << "Cannot use default metric, because column type " << type->ToString()
+                << " is not metrizable. Consider defining custom metric";
+            throw config::ConfigurationError(msg.str());
+        }
+        return metr_type;
+    }
+
+public:
+    double Dist(Types const& types, Values const& first, Values const& second) const override {
+        assert(types.size() == first.size() && types.size() == second.size());
+
+        double result = 0;
+        for (std::size_t i = 0; i < types.size(); ++i) {
+            if (first[i] && second[i]) {
+                double dist = ConvertType(types[i])->Dist(first[i], second[i]);
+                result += dist * dist;
+            }
+        }
+        return std::sqrt(result);
+    }
+};
+}  // namespace util
--- a/src/python_bindings/CMakeLists.txt
+++ b/src/python_bindings/CMakeLists.txt
@@ -47,6 +47,7 @@ target_sources(
             py_util/iterable_sequence_stream.cpp
             py_util/logging.cpp
             py_util/opt_to_py.cpp
+            py_util/py_custom_metrics.cpp
             py_util/py_to_any.cpp
             py_util/table_serialization.cpp
             py_util/value_to_py.cpp
--- a/src/python_bindings/py_util/py_custom_metrics.cpp
+++ b/src/python_bindings/py_util/py_custom_metrics.cpp
@@ -1,0 +1,22 @@
+#include "python_bindings/py_util/py_custom_metrics.h"
+
+#include <pybind11/pybind11.h>
+
+#include <memory>
+
+#include "core/util/custom_metric/custom_metric.h"
+#include "core/util/custom_metric/custom_vector_metric.h"
+
+namespace py = pybind11;
+
+namespace python_bindings {
+void BindCustomMetrics(pybind11::module_& main_module) {
+    auto metrics_module = main_module.def_submodule("metrics");
+
+    // These classes are used only in get_opts
+    py::class_<util::ICustomMetric, std::shared_ptr<util::ICustomMetric>>(metrics_module,
+                                                                          "CustomMetric");
+    py::class_<util::ICustomVectorMetric, std::shared_ptr<util::ICustomVectorMetric>>(
+            metrics_module, "CustomVectorMetrics");
+}
+}  // namespace python_bindings
--- a/src/python_bindings/py_util/py_custom_metrics.h
+++ b/src/python_bindings/py_util/py_custom_metrics.h
@@ -1,0 +1,42 @@
+#pragma once
+
+#include <pybind11/pybind11.h>
+
+#include <cstddef>
+#include <utility>
+
+#include <pybind11/pytypes.h>
+
+#include "core/model/types/type.h"
+#include "core/util/custom_metric/custom_metric.h"
+#include "core/util/custom_metric/custom_vector_metric.h"
+#include "python_bindings/py_util/value_to_py.h"
+
+namespace python_bindings {
+class PyCustomMetric : public util::ICustomMetric {
+private:
+    pybind11::object metric_;
+
+public:
+    explicit PyCustomMetric(pybind11::object&& metric) : metric_(std::move(metric)) {}
+
+    double Dist(model::Type const* type, std::byte const* first,
+                std::byte const* second) const override {
+        return pybind11::cast<double>(metric_(ValueToPy(type, first), ValueToPy(type, second)));
+    }
+};
+
+class PyCustomVectorMetric : public util::ICustomVectorMetric {
+private:
+    pybind11::object metric_;
+
+public:
+    explicit PyCustomVectorMetric(pybind11::object&& metric) : metric_(std::move(metric)) {}
+
+    double Dist(Types const& types, Values const& first, Values const& second) const override {
+        return pybind11::cast<double>(metric_(ValuesToPy(types, first), ValuesToPy(types, second)));
+    }
+};
+
+void BindCustomMetrics(pybind11::module_& main_module);
+}  // namespace python_bindings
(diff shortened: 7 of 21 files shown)


src/python_bindings/data/py_custom_metrics.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include "core/model/types/create_type.h"
+#include "core/model/types/mixed_type.h"
+#include "core/model/types/type.h"
+
+using namespace model;
+namespace py = pybind11;
+
+namespace {
+py::object DateToPyObject(Date const& date) {
+    static auto py_date = py::module_::import("datetime").attr("date");
```
wildsor: This will not work with subinterpreters. Do not store Python objects in non-local storage.

wildsor (changes requested):
> `PyMetric`

Could you please update the description? That doesn't seem to exist anymore.

> (very convenient in tests!)

If it's only needed in tests, then it belongs in tests.

> some non-standard behavior

Overabstraction has had a pretty bad effect on the code of the project. Please do not implement things unless you're certain (and can explain) that they're needed.

p-senichenkov:
> Could you please update the description? That doesn't seem to exist anymore.

The description is up-to-date. I meant `Py{Custom,CustomVector}Metric`

> If it's only needed in tests, then it belongs in tests.

No, it's needed **not** only in tests

> Overabstraction has had a pretty bad effect on the code of the project. Please do not implement things unless you're certain (and can explain) that they're needed.

How am I supposed to implement this thing without any interfaces?😐

wildsor:
> How am I supposed to implement this thing without any interfaces?😐

nevermind, I misread, all ok

src/python_bindings/py_util/py_to_any.cpp
```diff
@@ -119,6 +124,30 @@ boost::any InputTablesToAny(std::string_view option_name, py::handle obj) {
...
+            new python_bindings::PyCustomMetric(py::reinterpret_borrow<py::object>(obj)));
+}
+
+boost::any CustomMetricsToAny(std::string_view option_name, py::handle obj) {
+    auto metric_handles = CastAndReplaceCastError<std::vector<py::object>>(option_name, obj);
+    std::vector<std::shared_ptr<config::ICustomMetric>> result(metric_handles.size());
+    std::ranges::transform(
+            metric_handles, result.begin(),
+            [](py::object obj) -> std::shared_ptr<config::ICustomMetric> {
+                if (obj.is_none()) {
```
wildsor: What does this check do? Isn't it going to cause segfaults where a Python error would have been raised otherwise?

p-senichenkov: No, it isn't going to. Actually, we'd get segfaults in low-level python code if we removed this check, since `null` is not checked there

`nullptr` is **not** an error. It says "use the default metric for this column", just like None does in most of Python standard library functions. It's being replaced with `DefaultCustomMetric` later in option's normalization function. Passing `py::none` there would cause a segfault

src/core/util/custom_metric/custom_metric.h
```diff
@@ -0,0 +1,94 @@
+#pragma once
+
+#include <cassert>
+#include <cstddef>
+#include <functional>
+#include <optional>
+#include <sstream>
```
MichaelS239, line 7:
```suggestion
#include <sstream>
#include <utility>
```

src/core/config/custom_metric/custom_metric_option.h
```diff
@@ -0,0 +1,32 @@
...
+#include <string_view>
+
+#include "core/config/custom_metric/custom_metric.h"
+#include "core/config/descriptions.h"
+#include "core/config/names.h"
+#include "core/config/option.h"
+
+namespace config {
+/// @brief An option that represents a custom metric on a single column in the table
+class MetricOption : public Option<std::shared_ptr<ICustomMetric>> {
```
MichaelS239: We could use `config::CommonOption` here instead of creating a new class.

src/core/config/custom_metric/custom_metrics_option.h
```diff
@@ -0,0 +1,43 @@
...
+
+#include "core/config/custom_metric/custom_metric.h"
+#include "core/config/descriptions.h"
+#include "core/config/exceptions.h"
+#include "core/config/names.h"
+#include "core/config/option.h"
+
+namespace config {
+/// @brief Option for a collection of user-defined metrics
+class MetricsOption : public Option<std::vector<std::shared_ptr<ICustomMetric>>> {
```
MichaelS239: As far as I understand, the normalization function must be called manually from user's code, and the only correct option for this function is `NormalizeMetrics`. Also this function does the job of both `NormalizeFunc` and `ValueCheckFunc` (it throws an exception if the check is failed). I think that a better interface can be made.

We can do something like in `config::IndicesOption`:
- Use `CommonOption<ValueType>` object as a field
- Construct it with name, description and default value in constructor
- Add `operator()` that takes `value_ptr` and `indices_count` as parameters, creates `Option` object from `CommonOption`, calls `SetNormalizeFunc` and `SetValueCheck` with specified lambdas (with captured `indices_count`), and returns the resulting `Option` object
- Add global variables as constructed `MetricsOption` objects

The usage would be much simpler:
`RegisterOption(kLhsMetricsOpt(&lhs_metrics_, lhs_indices_.size()));`.

src/core/util/custom_metric/custom_vector_metric.h
```diff
@@ -0,0 +1,75 @@
+#pragma once
+
+#include <cassert>
+#include <cmath>
+#include <cstddef>
+#include <functional>
+#include <vector>
```
MichaelS239, line 8:
```suggestion
#include <vector>
#include <utility>
```

src/core/config/custom_metric/custom_vector_metric_option.h
```diff
@@ -0,0 +1,35 @@
...
+
+#include "core/config/custom_metric/custom_vector_metric.h"
+#include "core/config/descriptions.h"
+#include "core/config/names.h"
+#include "core/config/option.h"
+
+namespace config {
+/// @brief An option that represents a custom metric on a set of columns, that treats values as a
+/// vector
+class VectorMetricOption : public Option<std::shared_ptr<ICustomVectorMetric>> {
```
MichaelS239: Here we can also use `CommonOption` instead of creating a class.

src/core/util/custom_metric/custom_metric.h
MichaelS239, line 1: Maybe `util` is a better place for `custom_metric.h` and `custom_vector_metric.h` because `config` is responsible only for setting options.

Also other options in `config` have a common structure: `option.cpp`, `option.h` and `type.h`. Maybe we should follow this structure for consistency.

src/python_bindings/py_util/py_to_any.cpp
```diff
@@ -119,6 +124,30 @@ boost::any InputTablesToAny(std::string_view option_name, py::handle obj) {
...
+                    return nullptr;
+                }
+                return std::make_shared<python_bindings::PyCustomMetric>(std::move(obj));
```
MichaelS239: Why didn't we use `reinterpret_borrow` here like in other functions?

src/core/config/custom_metric/custom_metric.h
```diff
@@ -0,0 +1,94 @@
...
+#include "core/model/types/type.h"
+
+namespace config {
+/// @brief User-defined metric on a single typed column
+/// Together with @c PyCustomMetric, @c DynamicCustomMetric and @c StaticCustomMetric provides
+/// a convenient user interface (especially, in Python) without extra overhead on conversions
+// NOTE: these objects are wrapped in `shared_ptr`, and this encourages user to use the same object
+// for several columns. Keep it in mind if you are planning to implement some complex internal state
+/// WARN: NULL value is represented by nullptr
+class ICustomMetric {
```
MichaelS239:
```suggestion
class DESBORDANTE_EXPORT ICustomMetric {
```
I'm not sure, but it seems to be the case when this macro should be used.

I'm a little confused though, because there are option types like `model::Gdd`, which don't use this macro and still don't break CI.

src/python_bindings/data/py_custom_metrics.h
```diff
@@ -0,0 +1,51 @@
...
+#include "core/config/custom_metric/custom_metric.h"
+#include "core/config/custom_metric/custom_vector_metric.h"
+#include "core/model/types/type.h"
+
+namespace python_bindings {
+namespace detail {
+pybind11::object GetPyObjectFromMixed(model::Type const* type, std::byte const* value);
+pybind11::object GetPyObject(model::Type const* type, std::byte const* value);
+std::vector<pybind11::object> GetPyObjects(std::vector<model::Type const*> const& types,
+                                           std::vector<std::byte const*> const& values);
```
MichaelS239: Currently, all methods for type conversions are defined in separate files, like `py_to_any`, `get_py_type` and `opt_to_py`. Also these conversions can be used not only for metrics: several primitives (including Domain PACs) are already using Python functions and more may use them in the future. So, here I suggest moving these functions to a separate file in `py_util`.

MichaelS239: As far as I understand, these conversion functions are the only part of this PR that is used by Domain PACs. Maybe we could implement the conversions in #611 and leave the rest here in order to merge Domain PACs sooner.

src/python_bindings/data/py_custom_metrics.cpp
```diff
@@ -0,0 +1,95 @@
...
+    auto real_type = CreateType(mixed_type->RetrieveTypeId(value), mixed_type->IsNullEqNull());
+    return GetPyObject(real_type.get(), mixed_type->RetrieveValue(value));
+}
+
+py::object GetPyObject(Type const* type, std::byte const* value) {
+    if (!value) {
+        return py::none{};
+    }
+
+    switch (type->GetTypeId()) {
```
MichaelS239, line 46: Bool type has been recently added, it would be nice to support this type as well.

src/python_bindings/py_util/py_to_any.cpp
p-senichenkov: Because `obj` is already `py::object` here, there is no need in casting it again

src/python_bindings/data/py_custom_metrics.h
p-senichenkov: I'm not sure that Domain PACs are easier to merge than this PR. Let's just move converters into a separate PR and rebase both PRs onto it

p-senichenkov: https://github.com/Desbordante/desbordante-core/pull/776

src/core/config/custom_metric/custom_metric_option.h
p-senichenkov: `CommonOption` doesn't provide ability to set `name` and `description`, and I really dislike the idea of creating a global object for each possible option name, that will be used only once (like "metric", "lhs_metric", "rhs_metric", "some_other_metric")

src/core/util/custom_metric/custom_metric.h
p-senichenkov: > Maybe `util` is a better place for `custom_metric.h` and `custom_vector_metric.h` because `config` is responsible only for setting options.

I'm not sure. Custom metric types are related to configuration, like `config::IndicesType`

> Also other options in `config` have a common structure: `option.cpp`, `option.h` and `type.h`. Maybe we should follow this structure for consistency.

Here we have several types and options related to a single thing. Maybe something like
```
custom_metric
|- metric
|  |- type.h
|  `- option.h
|- metrics
|  `- option.h
`- vector_metric.h
   |- type.h
    `- option.h
```

wildsor: All the logic in src/config is used only in the configuration process itself, your class's code is used by the actual algorithms. Please do move this stuff out of here for consistency, if only because it'll be easier to refactor the stuff later. Just have an alias in `type.h` for `std::shared_ptr<ICustomMetric>`

src/core/config/custom_metric/custom_metric_option.h
wildsor: You can just use a function that returns an option

src/core/config/custom_metric/custom_vector_metric.h
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+public:
+    double Dist(Types const& types, Values const& first, Values const& second) const override {
+        assert(types.size() == first.size() && types.size() == second.size());
+
+        double result = 0;
+        for (std::size_t i = 0; i < types.size(); ++i) {
+            if (first[i] && second[i]) {
+                result += std::pow(ConvertType(types[i])->Dist(first[i], second[i]), 2);
```
wildsor: gcc isn't smart enough for this, it actually calls `pow`.

```suggestion
                double dist = ConvertType(types[i])->Dist(first[i], second[i])
                result += dist * dist;
```

src/core/config/custom_metric/custom_metrics/option.h
```diff
@@ -0,0 +1,50 @@
...
+    CommonOption<MetricsType> const common_option_;
+
+    static void CheckMetrics(MetricsType const& value, std::size_t indices_count) {
+        if (value.size() > indices_count) {
+            throw ConfigurationError("Too many user-defined metrics");
+        }
+    }
+
+    static void NormalizeMetrics(MetricsType& value, std::size_t indices_count) {
+        // A single static instance could be used, but Python would try to destruct it
```
wildsor: Wouldn't Python just decrement the pointer? It looks like it's fine, except google's style guide doesn't want static objects with non-trivial destructors, but it's not like we actually follow that.

src/python_bindings/CMakeLists.txt
```diff
@@ -49,10 +49,11 @@ target_sources(
             py_util/opt_to_py.cpp
             py_util/py_to_any.cpp
             py_util/table_serialization.cpp
+            py_util/value_to_py.cpp
 )
 target_link_libraries(
-    ${NAME} PRIVATE magic_enum::magic_enum spdlog::spdlog_header_only Boost::headers
-                    pybind11::pybind11
+    ${NAME} PRIVATE ${DESBORDANTE_PREFIX}::bindlib::data magic_enum::magic_enum
```
wildsor: May want to add bindlib::data as public, since it's used in the header

src/core/util/custom_metric/custom_vector_metric.h
wildsor, line 1: same as for custom_metric.h, move this out of core/config. util could work

src/core/config/custom_metric/custom_vector_metric_option.h
```diff
@@ -0,0 +1,35 @@
...
+        }
+    }
+
+public:
+    VectorMetricOption(std::shared_ptr<ICustomVectorMetric>* value_ptr,
+                       std::string_view name = names::kCustomMetric,
+                       std::string_view description = descriptions::kDCustomMetric)
+        : Option(value_ptr, name, description, ValueType{nullptr}) {
+        // User can pass `nullptr` to explicitly say that they want to use default metric
+        SetNormalizeFunc(&Normalize);
```
wildsor: Just make this a function

src/core/config/custom_metric/custom_vector_metric_option.h
```diff
@@ -0,0 +1,35 @@
...
+#include "core/config/option.h"
+
+namespace config {
+/// @brief An option that represents a custom metric on a set of columns, that treats values as a
+/// vector
+class VectorMetricOption : public Option<std::shared_ptr<ICustomVectorMetric>> {
+private:
+    using ValueType = std::shared_ptr<ICustomVectorMetric>;
+
+    static ValueType const kDefaultCustomMetric;
```
wildsor: Extra variable?

src/python_bindings/data/bind_data_types.cpp
```diff
@@ -44,5 +45,7 @@ void BindDataTypes(py::module_& main_module) {
                         auto col_indices = t[1].cast<std::vector<model::ColumnIndex>>();
                         return ColumnCombination(table_index, col_indices);
                     }));
+
+    BindCustomMetrics(data_module);
```
wildsor: Perhaps have a separate `desbordante.metrics` module?

wildsor (changes requested):

src/core/config/custom_metric/custom_metrics/option.h
p-senichenkov: I caught segfaults in low-level Python object destruction logic, when I tried to use a single static object here, so I suppose Python don't just decrement the pointer. But I haven't engaged a thorough study of Python's GC behavior on static C++ objects

src/python_bindings/data/bind_data_types.cpp
p-senichenkov: As for me, `desbordante.metrics` sounds like we provide some non-trivial metrics that the user can call in their code there, it's not about "bindings for user-defined metric utility classes", that user won't use directly. But I agree that putting all things into `data` is not a good idea too

src/core/config/custom_metric/custom_metrics/option.h
wildsor: I want to figure out what's happening here, do you have test code?

p-senichenkov: No, I've simply ran FD PAC verifier examples

src/core/config/custom_metric/custom_metric/option.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+#include <string_view>
+
+#include "core/config/custom_metric/custom_metric/type.h"
+#include "core/config/descriptions.h"
+#include "core/config/names.h"
```
wildsor, line 7: One of the reasons I did the extern variable stuff is to avoid having these in the headers. Do you really need defaults here and in the other option?

p-senichenkov: No really, but it seems convenient. If having these includes in headers is really an issue, default could be defined in source file, but it's a little unnatural

Actually, for me it seems like these headers are included everywhere, and there are some other places where they are included in headers, not in source files

src/core/util/custom_metric/custom_metric.h
```diff
@@ -0,0 +1,96 @@
...
+/// @brief A default value for custom metric option
+/// Uses default metric for the type. Works only with metrizable types
+class DefaultCustomMetric : public ICustomMetric {
+private:
+    static model::IMetrizableType const* ConvertType(model::Type const* type) {
+        auto const* metr_type = dynamic_cast<model::IMetrizableType const*>(type);
+        if (!metr_type) {
+            std::ostringstream msg;
+            msg << "Cannot use default metric, because column type " << type->ToString()
+                << " is not metrizable. Consider defining custom metric";
```
wildsor, line 81:
```suggestion
            msg << "Cannot use the default metric, because column type " << type->ToString()
                << " is not metrizable. Consider defining a custom metric";
```

wildsor (changes requested):

src/core/config/custom_metric/custom_metric/option.h
wildsor: Oh, if that's the case then whatever

src/python_bindings/data/bind_data_types.cpp
wildsor: `desbordante.data` was created solely to have types for different (input) data, like table, graph, etc. (look at the docstring for the module above), so let's not put anything there. `ColumnCombination` shouldn't be there either, it looks like it's an IND-specific thing, idk how I missed that during review.

Also, can you really be sure that anyone other than you is going to use these classes? Maybe you should put them into a `desbordante.pac` module or something like that, and we'll move them out if they happen to become useful for something else?

src/core/config/custom_metric/custom_metrics/option.h
wildsor: I'm not getting any issues when I use `static auto` in your `FD-PAC-validation` branch when running `examples/basic/verifying_pac/verifying_fd_pac.py`

p-senichenkov: Ok, I'll rebase that branch (just to be sure) and check it later

src/python_bindings/data/bind_data_types.cpp
p-senichenkov: Ok, `desbordante.metrics` is still better than `desbordante.data`, let's use this option

I cannot force the team to use these classes whenever they want to allow user to provide custom metrics. But I believe that it's the most convenient option for the user (am I wrong?), and I've tried to make them more or less convenient for the developer. Therefore, I think that they should be used in all new algorithms that need user-defined metrics
Putting these classes into `desbordante.pac` will make them seem like they shouldn't be used by other algorithms, so I don't think we should do so

wildsor: > will make them seem like they shouldn't be used by other algorithms, so I don't think we should do so

Okay, that's fair.

src/python_bindings/py_util/py_to_any.cpp
MichaelS239: I mean, all of the functions originally accept `py::handle` and in other functions we use `py::reinterpret_borrow` to cast it to `py::object`. But here we cast `py::handle` to `std::vector<py::object>` using `py::cast` in `CastAndReplaceCastError`. I'm not sure what happens to reference counting during such conversion.

Maybe we could cast `py::handle` to `std::vector<py::handle>` first and then use `py::reinterpret_borrow` on each `py::handle`.

p-senichenkov: As far as I understand, `cast` is a more smart function than `reinterpret_borrow`/`steal`. It performs type checking (which we do not need when casting `handle` to `object`) and then
- moves an object if its ref counter is 1
- else copies it

`reinterpret_borrow`/`steal` always do a copy/move respectively.
I've replaced `std::move` with `reinterpret_borrow` here for consistency

src/core/config/custom_metric/custom_metrics/option.h
p-senichenkov: Okay, I don't have any issues with running the FD PAC example, and even with the following script:
```Python
import desbordante

import gc

MARINE_URCHINS = "examples/datasets/verifying_pac/marine_urchins.csv"


def execute():
    algo = desbordante.pac_verification.algorithms.FDPACVerifier()
    algo.load_data(
        table=(MARINE_URCHINS, ",", True), lhs_indices=[0], rhs_indices=[1], lhs_deltas=[10]
    )
    algo.execute()
    print(f"Result: {algo.get_pac()}")


def execute2():
    algo = desbordante.pac_verification.algorithms.FDPACVerifier()
    algo.load_data(
        table=(MARINE_URCHINS, ",", True), lhs_indices=[1], rhs_indices=[2], lhs_deltas=[10], lhs_metrics=[None],
    )
    algo.execute()
    print(f"Result: {algo.get_pac()}")


execute()
gc.collect()
execute2()
gc.collect()

print("Execution finished")
```

Perhaps there were some other problems that caused segfaults, which are corrected now.

I still don't see any advantages in using a single static object. The only disadvantage of current implementation is that we use a few extra memory (typically there are no more than 5 custom metrics in a single dependency, and each object is rather compact -- it has no fields). So I'll just remove the comment if you are okay with that

src/python_bindings/data/bind_data_types.cpp
```diff
@@ -44,5 +45,7 @@ void BindDataTypes(py::module_& main_module) {
                         auto col_indices = t[1].cast<std::vector<model::ColumnIndex>>();
                         return ColumnCombination(table_index, col_indices);
                     }));
+
+    BindCustomMetrics(main_module);
```
wildsor: Call this in bindings.cpp (and remove the header from here)

src/python_bindings/py_util/py_to_any.cpp
```diff
@@ -1,13 +1,16 @@
...
 #include <functional>
+#include <memory>
 #include <sstream>
 #include <stdexcept>
 #include <string>
 #include <unordered_map>
 
 #include <boost/any.hpp>
 #include <boost/core/demangle.hpp>
+#include <pybind11/pytypes.h>
```
wildsor: Are you using anything from this header?

src/python_bindings/py_util/py_custom_metrics.h
wildsor, line 1: Old directory, consider moving to py_util/

wildsor (changes requested):

src/python_bindings/py_util/py_custom_metrics.cpp
```diff
@@ -0,0 +1,25 @@
+#include "python_bindings/py_util/py_custom_metrics.h"
+
+#include <pybind11/pybind11.h>
+
+#include <cassert>
+#include <memory>
+
+#include <pybind11/pytypes.h>
```
wildsor:
```suggestion
#include <memory>
```

src/python_bindings/py_util/py_to_any.cpp
```diff
@@ -194,6 +200,30 @@ boost::any StringVectorToAny(std::string_view option_name, py::handle obj) {
...
+                               }
+                               return std::make_shared<python_bindings::PyCustomMetric>(
+                                       py::reinterpret_borrow<py::object>(handle));
+                           });
+    return result;
+}
+
+boost::any CustomVectorMetricToAny(std::string_view, py::handle obj) {
+    return config::CustomVectorMetricType{
+            new python_bindings::PyCustomVectorMetric(py::reinterpret_borrow<py::object>(obj))};
```
wildsor, line 217: I guess it works, but it is unexpected to encounter a `new`, so consider changing to `std::make_shared` here and above.

wildsor (changes requested):
Looks mostly okay

src/python_bindings/py_util/py_to_any.cpp
p-senichenkov: If I do `make_shared` here, I'll get a `boost::any` with `std::shared_ptr<PyCustomMetric>` inside it, and "bad any cast" later

wildsor: You could use `std::dynamic_pointer_cast` but it's not critical

wildsor (approved):

PR #742, opened 2026-04-28
Refactor: make type checker order deterministic

snookumsff:
Replaces unordered_map iteration with deterministic ordering for type checking

### DIFF
--- a/src/core/model/table/typed_column_data.cpp
+++ b/src/core/model/table/typed_column_data.cpp
@@ -25,11 +25,15 @@ TypeId TypedColumnDataFactory::DeduceColumnType() const {
     bool is_undefined = true;
     std::bitset<5> candidate_types_bitset("11111");
     TypeId first_type_id = TypeId::kUndefined;
+    auto matcher_map_iter = kTypeIdToChecker.begin();
     for (std::size_t i = 0; i != unparsed_.size(); ++i) {
         if (!kNullCheck(unparsed_[i]) && !kEmptyCheck(unparsed_[i])) {
             is_undefined = false;
             if (first_type_id != TypeId::kUndefined) {
-                auto& type_check = kTypeIdToChecker.at(first_type_id);
+                auto it =
+                        std::find_if(matcher_map_iter, kTypeIdToChecker.end(),
+                                     [&](auto const& pair) { return pair.first == first_type_id; });
+                auto& type_check = it->second;
                 if (type_check(unparsed_[i])) {
                     // undelimited and delimited dates have different bitsets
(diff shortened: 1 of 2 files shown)


wildsor (changes requested):
The code looks fine

Please make this all a single commit and explain in the body that you needed to reorder the type detector map elements because the previous tests relied on a particular order. See the wiki, particularly the [linked article](https://cbea.ms/git-commit/)

src/core/model/table/typed_column_data.cpp
```diff
@@ -29,7 +29,10 @@ TypeId TypedColumnDataFactory::DeduceColumnType() const {
         if (!kNullCheck(unparsed_[i]) && !kEmptyCheck(unparsed_[i])) {
             is_undefined = false;
             if (first_type_id != TypeId::kUndefined) {
-                auto& type_check = kTypeIdToChecker.at(first_type_id);
+                auto it =
+                        std::find_if(kTypeIdToChecker.begin(), kTypeIdToChecker.end(),
```
wildsor: I'm a little bit concerned that this will do the iteration for every single value, but it's probably not too big a deal, given the vector is only 4 elements long? Still, it'd be better to do something like
```suggestion
                        std::find_if(checker_map_iter, kTypeIdToChecker.end(),
```
and store `auto checker_map_iter = kTypeIdToChecker.begin()` outside the loop.

wildsor (changes requested):
There is some bad naming in this class, this thing is not doing any "checking", it's trying to determine what type to convert the strings that have been read to.

> This caused TestTypeParsing tests to fail intermittently, as numbers were sometimes detected as kDouble instead of kInt depending on the map's internal ordering.

From what I'm seeing, the tests were fine before the addition of kBool, so "intermittently" is not correct. It's more so the addition exposed that flaw. I think it'd be a good idea to include that detail so that people looking through the commit history can get some lore.

Maybe something like
```
Fix reliance on unordered_map iteration order in type detection

The type checker previously iterated through unordered_map, which does
not have order guarantees. This caused the TestTypeParsing tests to fail
when a new type was added because numbers that were detected as ints
before started being detected as doubles.

This replaces unordered_map with vector and reorders entries to ensure
kDate -> kInt -> kBigInt -> kDouble -> kString order.
```

wildsor (approved):
Oops, I said the "checker" naming was bad and then went and suggested the commit description and code that still uses "checker". If you're low on time, we can let this through. But the more precise name for the function in that map is "matcher" (matches a string to its possible type), so `matcher_map_iter`. And "The type checker previously iterated" should be something like "The type *detection* previously iterated" or "inference".

wildsor (approved):

PR #744, opened 2026-04-30
Prune time tracking

wildsor:
Currently there are four areas where time is tracked in the project:
- ExecuteInternal return value
- In some logging statements
- Public methods in CIND/IND providing timing info
- Algorithm timeouts

The `ExecuteInternal` value has long gone unused but people implement its calculation to comply with the interface. The CIND/IND public methods are not used anywhere, looks like they were part of some preliminary testing.
I don't think logging random timings provides any benefit to the user or to the developer. It looks like these logs were either used for some preliminary testing or cargo-culted from other algorithms. I decided to remove them.
Algorithm timeouts are technically part of the implementation and need a bit more thinking to get right, I decided to leave them as is for now.

### DIFF
--- a/src/core/algorithms/fd/tane/tane_common.cpp
+++ b/src/core/algorithms/fd/tane/tane_common.cpp
@@ -229,7 +229,6 @@ void TaneCommon::ExecuteInternal() {
         // TODO: printProfilingData
     }
 
-    LOG_DEBUG("Total intersections: {}", model::PositionListIndex::intersection_count_);
     LOG_DEBUG("Total FD count: {}", fd_collection_.Size());
     LOG_DEBUG("HASH: {}", Fletcher16());
 }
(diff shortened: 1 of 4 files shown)


polyntsov (approved):
I agree that current state is a bit of a mess and needs a cleanup. However, for the algorithm internal logs which are useless for the end-user, it seems like this should be controlled via log levels

While Desbordante is not a benchmarking suite, benchmarking is still important, I am not sure how algorithm authors do it currently. By measuring `Execute` elapsed time?

wildsor:
> While Desbordante is not a benchmarking suite, benchmarking is still important, I am not sure how algorithm authors do it currently. By measuring `Execute` elapsed time?

It depends, I guess. I don't really see an issue with measuring by inserting extra statements and removing them once in the actual library. Or, indeed, just measuring that, if it's enough.

wildsor:
> However, for the algorithm internal logs which are useless for the end-user, it seems like this should be controlled via log levels

I've only removed timing logs, I just don't think those are meaningfully useful for anything, no matter the log level

wildsor:
Merged in #768

PR #748, opened 2026-05-02
Add optional PLIWS

Sneper-Breeze:
### Changes:

- In AfdMetricCalc (src/core/algorithms/fd/afd_metric):
  - Added methods to support computations with both PLI and PLIWS.
  - Rewrote ExecuteInternal logic using `std::variant` to select between PLI and PLIWS based on the `use_pliws` flag (default: false).

- In TaneCommon:
  - Rewrote ComputeDependencies logic with `std::variant` to work with both PLI and PLIWS.

### Behavior:

- The public interface remains unchanged for existing callers. An optional `use_pliws` parameter (default false) can now be passed at algorithm creation.
- Old code continues to work as before.

### Testing:

- Added new tests for AfdMetricCalc with and without PLIWS to verify correctness.
- All existing tests still pass, confirming no regression.

### DIFF
--- a/src/core/algorithms/fd/afd_metric/afd_metric_calculator.cpp
+++ b/src/core/algorithms/fd/afd_metric/afd_metric_calculator.cpp
@@ -30,13 +30,14 @@ void AFDMetricCalculator::RegisterOptions() {
     RegisterOption(config::kLhsIndicesOpt(&lhs_indices_, get_schema_cols));
     RegisterOption(config::kRhsIndicesOpt(&rhs_indices_, get_schema_cols));
     RegisterOption(Option{&metric_, kMetric, kDAFDMetric});
+    RegisterOption(config::kUsePliwsOpt(&use_pliws_));
 }
 
 void AFDMetricCalculator::MakeExecuteOptsAvailable() {
     using namespace config::names;
 
-    MakeOptionsAvailable(
-            {kMetric, config::kLhsIndicesOpt.GetName(), config::kRhsIndicesOpt.GetName()});
+    MakeOptionsAvailable({kMetric, config::kLhsIndicesOpt.GetName(),
+                          config::kRhsIndicesOpt.GetName(), config::kUsePliwsOpt.GetName()});
 }
 
 void AFDMetricCalculator::LoadDataInternal() {
@@ -54,26 +55,46 @@ unsigned long long AFDMetricCalculator::ExecuteInternal() {
     return elapsed_milliseconds;
 }
 
+namespace {
+
+template <typename PLIT>
+auto CalculateMetric(PLIT&& lhs_ptr, PLIT&& rhs_ptr, AFDMetric metric, size_t num_rows) {
+    auto* lhs_raw = lhs_ptr.get();
+    auto* rhs_raw = rhs_ptr.get();
+
+    switch (metric) {
+        case AFDMetric::kG2:
+            return AFDMetricCalculator::CalculateG2(lhs_raw, rhs_raw, num_rows);
+        case AFDMetric::kTau: {
+            auto joint = lhs_raw->Intersect(rhs_raw);
+            return AFDMetricCalculator::CalculateTau(lhs_raw, rhs_raw, joint.get());
+        }
+        case AFDMetric::kMuPlus: {
+            auto joint = lhs_raw->Intersect(rhs_raw);
+            return AFDMetricCalculator::CalculateMuPlus(lhs_raw, rhs_raw, joint.get());
+        }
+        case AFDMetric::kFi:
+            return AFDMetricCalculator::CalculateFI(lhs_raw, rhs_raw, num_rows);
+    }
+
+    assert(false);
+    __builtin_unreachable();
+}
+}  // namespace
+
 void AFDMetricCalculator::CalculateMetric() {
     auto num_rows = relation_->GetNumRows();
-    auto lhs_pli = relation_->CalculatePLIWS(lhs_indices_);
-    auto rhs_pli = relation_->CalculatePLIWS(rhs_indices_);
 
-    switch (metric_) {
-        case AFDMetric::kG2:
-            result_ = CalculateG2(lhs_pli.get(), rhs_pli.get(), num_rows);
-            break;
-        case AFDMetric::kTau:
-            result_ = CalculateTau(lhs_pli.get(), rhs_pli.get(),
-                                   lhs_pli->Intersect(rhs_pli.get()).get());
-            break;
-        case AFDMetric::kMuPlus:
-            result_ = CalculateMuPlus(lhs_pli.get(), rhs_pli.get(),
-                                      lhs_pli->Intersect(rhs_pli.get()).get());
-            break;
-        case AFDMetric::kFi:
-            result_ = CalculateFI(lhs_pli.get(), rhs_pli.get(), num_rows);
-            break;
+    if (use_pliws_) {
+        auto lhs_pli = relation_->CalculatePLIWS(lhs_indices_);
+        auto rhs_pli = relation_->CalculatePLIWS(rhs_indices_);
+        result_ = ::algos::afd_metric_calculator::CalculateMetric(lhs_pli, rhs_pli, metric_,
+                                                                  num_rows);
+    } else {
+        auto lhs_pli = relation_->CalculatePLI(lhs_indices_);
+        auto rhs_pli = relation_->CalculatePLI(rhs_indices_);
+        result_ = ::algos::afd_metric_calculator::CalculateMetric(lhs_pli, rhs_pli, metric_,
+                                                                  num_rows);
     }
 }
 
@@ -95,7 +116,7 @@ long double AFDMetricCalculator::CalculateG2(model::PLI const* lhs_pli, model::P
     return num_error_rows / num_rows;
 }
 
-long double AFDMetricCalculator::CalculatePdepSelf(model::PLIWithSingletons const* x_pli) {
+long double AFDMetricCalculator::CalculatePdepSelf(model::PLI const* x_pli) {
     size_t n = x_pli->GetRelationSize();
     config::ErrorType sum = 0;
     std::size_t cluster_rows_count = 0;
@@ -150,9 +171,58 @@ long double AFDMetricCalculator::CalculatePdepMeasure(model::PLIWithSingletons c
     return (sum / static_cast<config::ErrorType>(n));
 }
 
+long double AFDMetricCalculator::CalculatePdepMeasure(model::PositionListIndex const* x_pli,
+                                                      model::PositionListIndex const* xa_pli) {
+    std::deque<Cluster> xa_index = xa_pli->GetIndex();
+    std::deque<Cluster> x_index = x_pli->GetIndex();
+    size_t n = x_pli->GetRelationSize();
+
+    config::ErrorType sum = 0;
+
+    std::unordered_map<int, size_t> x_frequencies;
+
+    int x_value_id = 1;
+    for (Cluster const& x_cluster : x_index) {
+        x_frequencies[x_value_id++] = x_cluster.size();
+    }
+
+    x_frequencies[model::PositionListIndex::kSingletonValueId] = 1;
+
+    auto x_prob = x_pli->CalculateAndGetProbingTable();
+
+    auto get_x_freq_by_tuple_ind{[&x_prob, &x_frequencies](int tuple_ind) {
+        int value_id = x_prob->at(tuple_ind);
+        return static_cast<config::ErrorType>(x_frequencies[value_id]);
+    }};
+
+    for (Cluster const& xa_cluster : xa_index) {
+        config::ErrorType num = xa_cluster.size() * xa_cluster.size();
+        config::ErrorType denum = get_x_freq_by_tuple_ind(xa_cluster.front());
+        sum += num / denum;
+    }
+
+    auto xa_prob = xa_pli->CalculateAndGetProbingTable();
+    for (size_t i = 0; i < xa_prob->size(); i++) {
+        if (xa_prob->at(i) == 0) {
+            sum += 1 / get_x_freq_by_tuple_ind(i);
+        }
+    }
+    return (sum / static_cast<config::ErrorType>(n));
+}
+
 long double AFDMetricCalculator::CalculateTau(model::PLIWS const* lhs_pli,
                                               model::PLIWS const* rhs_pli,
                                               model::PLIWS const* joint_pli) {
+    auto p1 = CalculatePdepSelf(rhs_pli);
+    if (p1 == 1) return 1;
+
+    auto p2 = CalculatePdepMeasure(lhs_pli, joint_pli);
+
+    return (p2 - p1) / (1 - p1);
+}
+
+long double AFDMetricCalculator::CalculateTau(model::PLI const* lhs_pli, model::PLI const* rhs_pli,
+                                              model::PLI const* joint_pli) {
     auto p1 = CalculatePdepSelf(rhs_pli);
     if (p1 == 1) return 1;
 
@@ -186,6 +256,33 @@ long double AFDMetricCalculator::CalculateMuPlus(model::PLIWS const* lhs_pli,
     return std::max(mu, 0.L);
 }
 
+long double AFDMetricCalculator::CalculateMuPlus(model::PositionListIndex const* x_pli,
+                                                 model::PositionListIndex const* a_pli,
+                                                 model::PositionListIndex const* xa_pli) {
+    config::ErrorType pdep_y = CalculatePdepSelf(a_pli);
+    if (pdep_y == 1) return 1;
+
+    config::ErrorType pdep_xy = CalculatePdepMeasure(x_pli, xa_pli);
+
+    size_t n = x_pli->GetRelationSize();
+    std::size_t cluster_rows_count = 0;
+    std::deque<Cluster> const& x_index = x_pli->GetIndex();
+    size_t k = x_index.size();
+
+    for (Cluster const& x_cluster : x_index) {
+        cluster_rows_count += x_cluster.size();
+    }
+
+    std::size_t unique_rows = x_pli->GetRelationSize() - cluster_rows_count;
+    k += unique_rows;
+
+    if (k == n) return 1;
+
+    config::ErrorType mu = 1 - (1 - pdep_xy) / (1 - pdep_y) * (n - 1) / (n - k);
+    config::ErrorType mu_plus = std::max(0., mu);
+    return mu_plus;
+}
+
 long double AFDMetricCalculator::CalculateFI(model::PLIWS const* lhs_pli,
                                              model::PLIWS const* rhs_pli, size_t num_rows) {
     if (num_rows <= 0) throw std::invalid_argument("received unpositive number of rows");
@@ -213,6 +310,44 @@ long double AFDMetricCalculator::CalculateFI(model::PLIWS const* lhs_pli,
             if (size == 0.L) continue;
             conditional_entropy -= size * (std::log(size) - log_x);
         }
+    }
+    conditional_entropy /= num_rows;
+    auto mutual_information = entropy - conditional_entropy;
+    return mutual_information / entropy;
+}
+
+long double AFDMetricCalculator::CalculateFI(model::PLI const* lhs_pli, model::PLI const* rhs_pli,
+                                             size_t num_rows) {
+    if (num_rows <= 0) throw std::invalid_argument("received unpositive number of rows");
+
+    if (rhs_pli->GetNumCluster() < 2) {
+        return 0.L;
+    }
+
+    auto entropy = rhs_pli->GetEntropy();
+
+    std::deque<Cluster> rhs_clusters{rhs_pli->GetIndex()};
+    for (auto& y : rhs_clusters) {
+        std::sort(y.begin(), y.end());
+    }
+
+    auto conditional_entropy = 0.L;
+    for (auto& x : std::deque<Cluster>{lhs_pli->GetIndex()}) {
+        std::sort(x.begin(), x.end());
+        auto log_x = std::log(x.size());
+
+        size_t xy_calc_cnt = 0;
+        for (auto const& y : rhs_clusters) {
+            model::PositionListIndex::Cluster xy;
+            std::set_intersection(x.begin(), x.end(), y.begin(), y.end(), std::back_inserter(xy));
+
+            auto size = (long double)xy.size();
+            if (size == 0.L) continue;
+            xy_calc_cnt += size;
+            conditional_entropy -= size * (std::log(size) - log_x);
+        }
+
+        conditional_entropy -= (long double)(x.size() - xy_calc_cnt) * (-log_x);
     }
     conditional_entropy /= num_rows;
     auto mutual_information = entropy - conditional_entropy;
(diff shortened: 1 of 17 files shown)


src/core/algorithms/fd/tane/tane_common.cpp
```diff
@@ -21,6 +22,12 @@ namespace tane {
 
 TaneCommon::TaneCommon() : PliBasedFDAlgorithm() {
     RegisterOption(config::kErrorOpt(&max_ucc_error_));
+    RegisterOption(
+            config::Option{&use_pliws_, "use_pliws",
```
wildsor: Isn't this exactly the same as the other use_pliws option? You can put it into src/config

Sneper-Breeze: Yes, It is. Sure, reasonable.

src/core/algorithms/fd/afd_metric/afd_metric_calculator.cpp
```diff
@@ -56,25 +58,52 @@ unsigned long long AFDMetricCalculator::ExecuteInternal() {
...
+                case AFDMetric::mu_plus: {
+                    auto joint = lhs_raw->Intersect(rhs_raw);
+                    result_ = CalculateMuPlus(lhs_raw, rhs_raw, joint.get());
+                    break;
+                }
+                case AFDMetric::fi:
+                    result_ = CalculateFI(lhs_raw, rhs_raw, num_rows);
+                    break;
+            }
+        } else {
```
mitya-radivonchik: Should we consider inverting this constexpr if statement to call std::terminate when the types differ, and then have the actual logic outside the if statement? That would strip an identation level

src/core/algorithms/fd/afd_metric/afd_metric_calculator.cpp
```diff
@@ -56,25 +58,52 @@ unsigned long long AFDMetricCalculator::ExecuteInternal() {
...
+            std::terminate();
```
wildsor: assert(false); __builtin_unreachable();
so the compiler knows it's not possible.

wildsor: Actually, nevermind, the gcc and clang are not smart enough to figure out an optimization here. And it's not really worth it either.
But do insert an `assert(false)` here

wildsor: Actually, just get rid of this `if` for clarity. Use std::visit with one parameter, then do `*std::get_if<LhsType>(&rhs_pli)` on the other. Same for others like this, if any.

src/tests/unit/test_afd_metric_calculator.cpp
```diff
@@ -41,13 +42,21 @@ INSTANTIATE_TEST_SUITE_P(
         AFDMetricCalculatorTestSuite, TestAFDMetrics,
         ::testing::Values(
             AFDMetricCalculatorParams({4}, {3}, AFDMetric::tau, 78.L/90),
+            AFDMetricCalculatorParams({4}, {3}, AFDMetric::tau, 78.L/90, true),
```
mitya-radivonchik: Do we want to do the same in `src/tests/unit/test_pfdtane.cpp`?

src/core/algorithms/fd/afd_metric/afd_metric_calculator.cpp
```diff
@@ -30,13 +31,14 @@ void AFDMetricCalculator::RegisterOptions() {
...
+    RegisterOption(config::kUsePliwsOpt(&use_pliws_));
 }
 
 void AFDMetricCalculator::MakeExecuteOptsAvailable() {
     using namespace config::names;
 
-    MakeOptionsAvailable(
-            {kMetric, config::kLhsIndicesOpt.GetName(), config::kRhsIndicesOpt.GetName()});
+    MakeOptionsAvailable({kMetric, config::kLhsIndicesOpt.GetName(),
+                          config::kRhsIndicesOpt.GetName(), "use_pliws"});
```
mitya-radivonchik: Let's use kUsePliws here instead of a hardcoded string

src/tests/unit/test_afd_metric_calculator.cpp
```diff
@@ -15,13 +15,14 @@ struct AFDMetricCalculatorParams {
...
-                              AFDMetric metric, long double expected,
+                              AFDMetric metric, long double expected, bool use_pliws = false,
                               CSVConfig const& csv_config = kTestFD)
         : params({{kCsvConfig, csv_config},
                   {kLhsIndices, std::move(lhs_indices)},
                   {kRhsIndices, std::move(rhs_indices)},
                   {kEqualNulls, true},
-                  {kMetric, metric}}),
+                  {kMetric, metric},
+                  {"use_pliws", use_pliws}}),
```
mitya-radivonchik: Let's use kUsePliws here instead of a hardcoded string

src/core/algorithms/fd/afd_metric/afd_metric_calculator.cpp
mitya-radivonchik: I was going to propose a different approach:

1. define this in an anonymous namespace:
```cpp
namespace {

template<typename PLIT>
auto CalculateMetric(PLIT&& lhs_ptr, PLIT&& rhs_ptr, AFDMetric metric, size_t num_rows) {
    auto* lhs_raw = lhs_ptr.get();
    auto* rhs_raw = rhs_ptr.get();

    switch (metric) {
        case AFDMetric::g2:
            return AFDMetricCalculator::CalculateG2(lhs_raw, rhs_raw, num_rows);
        case AFDMetric::tau: {
            auto joint = lhs_raw->Intersect(rhs_raw);
            return AFDMetricCalculator::CalculateTau(lhs_raw, rhs_raw, joint.get());
        }
        case AFDMetric::mu_plus: {
            auto joint = lhs_raw->Intersect(rhs_raw);
            return AFDMetricCalculator::CalculateMuPlus(lhs_raw, rhs_raw, joint.get());
        }
        case AFDMetric::fi:
            return AFDMetricCalculator::CalculateFI(lhs_raw, rhs_raw, num_rows);
    }

    assert(false);
    __builtin_unreachable();
}

}
```

2. Let ADFMetricCalculator::CalculateMetric be:
```cpp
void AFDMetricCalculator::CalculateMetric() {
    auto num_rows = relation_->GetNumRows();

    if (use_pliws_) {
        auto lhs_pli = relation_->CalculatePLIWS(lhs_indices_);
        auto rhs_pli = relation_->CalculatePLIWS(rhs_indices_);
        result_ = ::algos::afd_metric_calculator::CalculateMetric(lhs_pli, rhs_pli, metric_, num_rows);
    } else {
        auto lhs_pli = relation_->CalculatePLI(lhs_indices_);
        auto rhs_pli = relation_->CalculatePLI(rhs_indices_);
        result_ = ::algos::afd_metric_calculator::CalculateMetric(lhs_pli, rhs_pli, metric_, num_rows);
    }
}
```

This allows for inlining here (as the template function is statically compiled) and moves all of the type matching machinery to happen in the template substitution

src/core/algorithms/fd/tane/tane.cpp
```diff
@@ -24,24 +24,36 @@ config::ErrorType Tane::CalculateZeroAryFdError(ColumnData const* rhs) {
...
-                                                                                joint_pli);
-        case +AfdErrorMeasure::mu_plus:
-            return 1 - afd_metric_calculator::AFDMetricCalculator::CalculateMuPlus(lhs_pli, rhs_pli,
-                                                                                   joint_pli);
-        case +AfdErrorMeasure::rho:
-            return 1 - CalculateRhoMeasure(lhs_pli, joint_pli);
-        default:
-            return CalculateG1Error(lhs_pli, joint_pli, relation_.get()->GetNumTuplePairs());
-    }
+config::ErrorType Tane::CalculateFdError(tane::PLIVariantPtr lhs_pli, tane::PLIVariantPtr rhs_pli,
```
mitya-radivonchik: If @BUYT-1 agrees with my proposal here https://github.com/Desbordante/desbordante-core/pull/748/changes#r3261625795 I'd suggest to apply that approach universally to replace std::variant/std::visit. Here I would suggest introducing two overloads for `Tane::CalculateFdError` -- one using normal PLI, one using PLIWS, and implement those overloads with a template function with single template argument for the pli type.

src/core/algorithms/fd/tane/tane.h
```diff
@@ -13,9 +13,8 @@ class Tane : public tane::TaneCommon {
     AfdErrorMeasure afd_error_measure_ = +AfdErrorMeasure::g1;
     void MakeExecuteOptsAvailableFDInternal() override final;
     config::ErrorType CalculateZeroAryFdError(ColumnData const* rhs) override;
-    config::ErrorType CalculateFdError(model::PLIWithSingletons const* lhs_pli,
-                                       model::PLIWithSingletons const* rhs_pli,
-                                       model::PLIWithSingletons const* joint_pli) override;
+    config::ErrorType CalculateFdError(tane::PLIVariantPtr lhs_pli, tane::PLIVariantPtr rhs_pli,
```
mitya-radivonchik: Should we decide to follow https://github.com/Desbordante/desbordante-core/pull/748/changes#r3261654275, this signature using variants would be replaced with two signatures -- one per pli type.

src/core/algorithms/fd/tane/tane_common.cpp
```diff
@@ -110,8 +126,14 @@ void TaneCommon::ComputeDependencies(model::LatticeLevel* level) {
...
+            std::variant<model::PLI const*, model::PLIWS const*> x_pli, a_pli;
+            if (use_pliws_) {
+                x_pli = x_vertex->GetPositionListIndexWithSingletons();
+                a_pli = relation_->GetColumnData(a_index).GetPLWSIndex();
+            } else {
+                x_pli = x_vertex->GetPositionListIndex();
+                a_pli = relation_->GetColumnData(a_index).GetPositionListIndex();
+            }
             // Check X -> A
             config::ErrorType error = CalculateFdError(x_pli, a_pli, xa_pli);
```
mitya-radivonchik: With https://github.com/Desbordante/desbordante-core/pull/748/changes#r3261654275 the CalculateFdError call would need to be moved into both branches of the if statement above

src/tests/unit/test_afd_metric_calculator.cpp
```diff
@@ -15,13 +15,14 @@ struct AFDMetricCalculatorParams {
     long double const expected = 0.L;
 
     AFDMetricCalculatorParams(config::IndicesType lhs_indices, config::IndicesType rhs_indices,
-                              AFDMetric metric, long double expected,
+                              AFDMetric metric, long double expected, bool use_pliws = false,
```
mitya-radivonchik, line 19: I'd  say we should probably move `bool use_pliws = false` to come after csv_config

src/core/algorithms/fd/tane/tane_common.h
```diff
@@ -8,11 +8,13 @@
 #include "core/model/table/position_list_index.h"
 
 namespace algos::tane {
+using PLIVariantPtr = std::variant<model::PLI const*, model::PLIWS const*>;
```
mitya-radivonchik: In case we end up not removing std::variant, please make sure to add `#include <variant>` at the top -- currently it's included transitively which is brittle.

src/core/config/use_pliws/option.cpp
```diff
@@ -0,0 +1,10 @@
+#include "core/config/use_pliws/option.h"
+
+#include <limits>
```
mitya-radivonchik: This include seems unused in this file

src/core/algorithms/fd/tane/tane_common.cpp
```diff
@@ -93,14 +96,27 @@ void TaneCommon::ComputeDependencies(model::LatticeLevel* level) {
...
+                auto parent_pli_1 = xa_vertex->GetParents()[0]->GetPositionListIndex();
+                auto parent_pli_2 = xa_vertex->GetParents()[1]->GetPositionListIndex();
+                xa_vertex->AcquirePositionListIndex(parent_pli_1->Intersect(parent_pli_2));
+            }
         }
 
         dynamic_bitset<> xa_indices = xa.GetColumnIndices();
         dynamic_bitset<> a_candidates = xa_vertex->GetRhsCandidates();
-        auto xa_pli = xa_vertex->GetPositionListIndexWithSingletons();
+        std::variant<model::PLI const*, model::PLIWS const*> xa_pli;
```
mitya-radivonchik, line 115: I guess we would have to leave the std::variant here and use std::get_if when it's accessed in the if below

src/core/algorithms/fd/afd_metric/afd_metric_calculator.cpp
```diff
@@ -150,6 +179,45 @@ long double AFDMetricCalculator::CalculatePdepMeasure(model::PLIWithSingletons c
     return (sum / static_cast<config::ErrorType>(n));
 }
 
+long double AFDMetricCalculator::CalculatePdepMeasure(model::PositionListIndex const* x_pli,
```
mitya-radivonchik, line 174: This function seems to be an exact copy paste of `CalculatePdepMeasure` in `src/core/algorithms/fd/tane/afd_measures.cpp`. Should we just call the function from `src/core/algorithms/fd/tane/afd_measures.cpp` then? If we do, then it seems to me that the other types of measures are also implemented there, and we should look into reusing them, having confirmed that they are indeed the same.

mitya-radivonchik (changes requested):

src/core/algorithms/fd/tane/tane.cpp
wildsor: Yeah, you've done a proper review, I've just looked through stuff, your idea seems better

src/tests/unit/test_afd_metric_calculator.cpp
Sneper-Breeze: Yea, probably should

src/core/algorithms/fd/afd_metric/afd_metric_calculator.cpp
wildsor:
```suggestion
                          config::kRhsIndicesOpt.GetName(), config::kUsePliwsOpt.GetName()});
```

src/core/algorithms/fd/afd_metric/afd_metric_calculator.cpp
Sneper-Breeze: Yes, this function is a copy of `src/core/algorithms/fd/tane/afd_measures.cpp `. But call `tane/afd_measures.cpp ` from it seems like a not good idea, cause in the future we want to delete `tane/afd_measures.cpp ` and make the afd_metric_calc calss is the only one thing to calculating mesaures for fd/afd.

src/core/algorithms/fd/tane/tane_common.cpp
Sneper-Breeze: I think it is bad to having nullptr from std::get_if for troubleshooting and correctness of the algorithm, so maybe using std::get is more correct? Also std::get_if will return a pointer, so we will need extra pointer dereference, so error will  be thrown anyway.

src/tests/unit/test_afd_metric_calculator.cpp
Sneper-Breeze: So then we will can`t do the `AFDMetricCalculatorParams({4}, {3}, AFDMetric::tau, 78.L/90, true)`, and will be needed to add kTestFD parameters also. Maybe just move initialization `{kUsePliws, use_pliws}` upper, above the `{kMetric, metric}`?

src/core/algorithms/fd/tane/tane_common.cpp
mitya-radivonchik: That is fair, the reason I suggested std::get_if is that it would return a null-pointer instead of throwing, which is marginally better in terms of performance even on the happy path. But I'm pretty sure it's negligible here, so let's leave it as is. Thanks

src/core/algorithms/fd/afd_metric/afd_metric_calculator.cpp
mitya-radivonchik: Sounds good

src/core/algorithms/fd/afd_metric/afd_metric_calculator.cpp
```diff
@@ -3,6 +3,7 @@
 #include <chrono>
 #include <cmath>
 #include <iterator>
+#include <variant>
```
mitya-radivonchik: Seems like you don't need the `#include <variant>` anymore

src/tests/unit/test_afd_metric_calculator.cpp
mitya-radivonchik: I'm fine to leave as is

mitya-radivonchik (approved):

src/core/algorithms/fd/tane/tane.cpp
```diff
@@ -24,9 +24,27 @@ config::ErrorType Tane::CalculateZeroAryFdError(ColumnData const* rhs) {
...
+                                                                                        joint_pli);
+        case AfdErrorMeasure::kTau:
+            return 1 - afd_metric_calculator::AFDMetricCalculator::CalculateTau(lhs_pli, rhs_pli,
+                                                                                joint_pli);
+        case AfdErrorMeasure::kMuPlus:
+            return 1 - afd_metric_calculator::AFDMetricCalculator::CalculateMuPlus(lhs_pli, rhs_pli,
+                                                                                   joint_pli);
+        case AfdErrorMeasure::kRho:
+            return 1 - CalculateRhoMeasure(lhs_pli, joint_pli);
+        default:
```
wildsor, line 41: Specify the case explicitly so you can get a build failure by means of compiler warning if you forget one
```suggestion
        case AfdErrorMeasure::kG1:
```

src/core/algorithms/fd/afd_metric/afd_metric_calculator.cpp
```diff
@@ -186,6 +256,33 @@ long double AFDMetricCalculator::CalculateMuPlus(model::PLIWS const* lhs_pli,
...
+    config::ErrorType pdep_xy = CalculatePdepMeasure(x_pli, xa_pli);
+
+    size_t n = x_pli->GetRelationSize();
+    std::size_t cluster_rows_count = 0;
+    std::deque<Cluster> const& x_index = x_pli->GetIndex();
+    size_t k = x_index.size();
+
+    for (Cluster const& x_cluster : x_index) {
+        cluster_rows_count += x_cluster.size();
+    }
```
wildsor, line 274: This value is `PositionListIndex::GetSize`

src/core/algorithms/fd/afd_metric/afd_metric_calculator.cpp
```diff
@@ -186,6 +256,33 @@ long double AFDMetricCalculator::CalculateMuPlus(model::PLIWS const* lhs_pli,
...
+
+    std::size_t unique_rows = x_pli->GetRelationSize() - cluster_rows_count;
```
wildsor, line 276:
```suggestion
    std::size_t unique_rows = x_pli->GetRelationSize() - x_pli->GetSize();
```

src/core/algorithms/fd/afd_metric/afd_metric_calculator.cpp
```diff
@@ -219,4 +316,42 @@ long double AFDMetricCalculator::CalculateFI(model::PLIWS const* lhs_pli,
     return mutual_information / entropy;
 }
 
+long double AFDMetricCalculator::CalculateFI(model::PLI const* lhs_pli, model::PLI const* rhs_pli,
+                                             size_t num_rows) {
+    if (num_rows <= 0) throw std::invalid_argument("received unpositive number of rows");
```
wildsor, line 321:
```suggestion
    if (num_rows <= 0) throw std::invalid_argument("received non-positive number of rows");
```

src/core/algorithms/fd/tane/pfdtane.cpp
```diff
@@ -33,6 +33,12 @@ config::ErrorType PFDTane::CalculateFdError(model::PLIWS const* lhs_pli,
     return CalculatePFDError(lhs_pli, joint_pli, pfd_error_measure_);
 }
 
+config::ErrorType PFDTane::CalculateFdError(model::PLI const* lhs_pli,
+                                            [[maybe_unused]] model::PLI const* rhs_pli,
```
wildsor, line 37:
```suggestion
                                            model::PLI const*,
```

src/core/algorithms/fd/tane/pfdtane.h
```diff
@@ -15,7 +15,11 @@ class PFDTane : public tane::TaneCommon {
     void RegisterOptions();
     void MakeExecuteOptsAvailableFDInternal() final;
     config::ErrorType CalculateZeroAryFdError(ColumnData const* rhs) override;
-    config::ErrorType CalculateFdError(model::PLIWS const* lhs_pli, model::PLIWS const* rhs_pli,
+    config::ErrorType CalculateFdError(model::PLI const* lhs_pli,
+                                       [[maybe_unused]] model::PLI const* rhs_pli,
+                                       model::PLI const* joint_pli) override;
+    config::ErrorType CalculateFdError(model::PLIWS const* lhs_pli,
+                                       [[maybe_unused]] model::PLIWS const* rhs_pli,
```
wildsor, line 22:
```suggestion
                                       model::PLI const*,
                                       model::PLI const* joint_pli) override;
    config::ErrorType CalculateFdError(model::PLIWS const* lhs_pli,
                                       model::PLIWS const*,
```

src/core/algorithms/fd/tane/tane_common.h
```diff
@@ -21,6 +22,9 @@ class TaneCommon : public PliBasedFDAlgorithm {
     void ComputeDependencies(model::LatticeLevel* level);
     unsigned long long ExecuteInternal() final;
     virtual config::ErrorType CalculateZeroAryFdError(ColumnData const* rhs) = 0;
+    virtual config::ErrorType CalculateFdError(model::PLI const* lhs_pli,
+                                               [[maybe_unused]] model::PLI const* rhs_pli,
+                                               model::PLI const* joint_pli) = 0;
```
wildsor, line 27: Extra `[[maybe_unused]]`
```suggestion
    virtual config::ErrorType CalculateFdError(model::PLI const* lhs_pli,
                                               model::PLI const* rhs_pli,
                                               model::PLI const* joint_pli) = 0;
```

src/core/algorithms/fd/afd_metric/afd_metric_calculator.cpp
wildsor: And now that I see this, I'm skeptical if we need PLI with singletons at all. Please tell me where it helps.

wildsor (changes requested):
`[[maybe_unused]]` is supposed to be used when a parameter might be used, but it depends on a macro. If you have to have the parameter due to interface restrictions, just leave the name empty.

chernishev:
Closed due to merging [828](https://github.com/Desbordante/desbordante-core/pull/828) and [821](https://github.com/Desbordante/desbordante-core/pull/821). As it turned out, singleton PLIs are no longer needed, so we are removing them from the codebase and examples.

PR #753, opened 2026-05-06
Fix ThreadPool race condition that fails only on MacOS

ol-imorozko:
The pool's std::mutex and std::condition_variable are declared
after worker_threads_ in the class body, so they are destroyed
before the worker JThreads are joined as members destruct in
reverse declaration order.

If a worker is still inside its unique_lock dtor when
~WorkerThreadPool returns and member destruction begins,
the worker touches a destroyed mutex. pthread_mutex_unlock on a
destroyed mutex returns EINVAL on macOS, which libc++'s
std::mutex translates into a system_error and aborts:

  libc++abi: terminating due to uncaught exception of type
  std::__1::system_error: mutex lock failed: Invalid argument

Explicitly clear worker_threads_ at the end of the dtor body.
Reordering the member declarations would also work but IMHO
current fix is better

### DIFF
--- a/src/tests/unit/test_worker_thread_pool.cpp
+++ b/src/tests/unit/test_worker_thread_pool.cpp
@@ -1,0 +1,17 @@
+#include <gtest/gtest.h>
+
+#include "core/util/worker_thread_pool.h"
+
+namespace tests {
+
+TEST(WorkerThreadPool, ConstructAndDestroyWithoutWork) {
+    util::WorkerThreadPool pool{4};
+}
+
+TEST(WorkerThreadPool, RepeatedConstructAndDestroyWithoutWork) {
+    for (int i = 0; i < 100; ++i) {
+        util::WorkerThreadPool pool{4};
+    }
+}
+
+}  // namespace tests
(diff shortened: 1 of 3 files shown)


src/tests/unit/test_worker_thread_pool.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+namespace tests {
+
+TEST(WorkerThreadPool, ConstructAndDestroyWithoutWork) {
+    util::WorkerThreadPool pool{4};
+}
+
+TEST(WorkerThreadPool, RepeatedConstructAndDestroyWithoutWork) {
+    constexpr std::size_t iterations = 100;
+    for (std::size_t i = 0; i < iterations; ++i) {
```
wildsor: nit: I'm assuming you did this because the compiler was complaining about comparison of signed and unsigned integers.
```suggestion
    for (int i = 0; i < 100; ++i) {
```
```suggestion
    for (std::size_t i = 0; i < 100u; ++i) {
```

wildsor (approved):
Thanks.
I wonder if it's possible to create a test situation that triggers this reliably or semi-reliably.

PR #756, opened 2026-05-07
Enable safe vertical hashing by default

wildsor:
There are many datasets with more than 64 columns. It's unexpected that they would fail to be processed by default, especially in the pip package. This did make sense when Desbordante was a benchmarking tool where any issues like that would be handled by recompiling but not when it tries to pose as a proper library.


chernishev (approved):
LGTM

PR #758, opened 2026-05-11
Remove dataset parsing in gSpan ResetState method

perfotoran:
Previously, `ResetState` parsed the graph database for every execution.  Since loading dataset should only be done in LoadData method, `graph_database_` field was split into two fields: `raw_dataset_` for original dataset and `pruned_graphs` for modified one.

### DIFF
--- a/src/core/algorithms/fsm/gspan/gspan.h
+++ b/src/core/algorithms/fsm/gspan/gspan.h
@@ -32,7 +32,8 @@ protected:
 
     std::filesystem::path graph_database_path_;
     std::filesystem::path output_path_;
-    std::vector<gspan::graph_t> graph_database_;
+    std::vector<gspan::graph_t> raw_dataset_;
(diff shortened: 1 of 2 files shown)


src/core/algorithms/fsm/gspan/gspan.h
```diff
@@ -65,7 +66,7 @@ class GSpan : public Algorithm {
     }
 
     std::vector<gspan::graph_t> const& GetGraphDatabase() const {
-        return graph_database_;
+        return raw_dataset_;
```
VinogradovDmitri: I'm wondering if the correct graph is being returned, and moreover, I haven't found any use of the `GetGraphDatabase` function in your primitive.

perfotoran: I guess I wrote it on autopilot by analogy with the GetFrequentSubgraphs. Will be removed.

VinogradovDmitri (approved):

PR #759, opened 2026-05-12
Make DataFrame reader take table name from .attrs

wildsor:
DataFrames do not have a dedicated attribute to specify their name, so there is nothing to pass to Desbordante that pattern objects can use. Before, every DataFrame was using "Pandas dataframe" as its name internally, which made it difficult to disambigulate tables in patterns that are defined on several tables, like INDs. For an example where it matters see https://groups.google.com/g/desbordante/c/BqpE5dK0EXo.

The attrs attribute on DataFrames is specifically designated for custom metadata like that. Creating a separate class just for this as in https://pandas.pydata.org/pandas-docs/stable/development/extending.html#define-original-properties would be excessive.

See also
https://github.com/pandas-dev/pandas/issues/447#issuecomment-868822049
https://github.com/pandas-dev/pandas/issues/2485
https://github.com/pandas-dev/pandas/issues/52166

### DIFF
--- a/examples/basic/mining_ind.py
+++ b/examples/basic/mining_ind.py
@@ -37,3 +37,5 @@ print_table('examples/datasets/ind_datasets/course.csv')
 print()
 print('department.csv:\n')
 print_table('examples/datasets/ind_datasets/department.csv')
+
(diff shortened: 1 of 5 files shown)


vs9h (approved):
Looks good, I have no comments

PR #771, opened 2026-05-27
fix: FdFirst options and refactor CFDRelationData class

### LINKED ISSUE #278: FDFirstAlgorithm: fail on incorrect option values early
The `Option` class has facilities to reject incorrect values immediately, yet `FDFirstAlgorithm` fails only during its execution stage. Its `RegisterOptions` method should be modified to make use of `Option`.

Oddin60F:
**CFDRelationData:**
1. Remove the parameters for the number of rows and columns to load part of the table into the algorithm.
2. Remove unused methods.

**FdFirst options:**
Add parameter verification during loading. Previously, an incorrect parameter caused an error when running the algorithm, not when loading the parameters.


src/core/algorithms/cfd/cfd_discovery.cpp
```diff
@@ -36,13 +36,32 @@ void CFDDiscovery::ResetState() {
 void CFDDiscovery::RegisterOptions() {
     DESBORDANTE_OPTION_USING;
 
+    auto check_num_cols = [this](unsigned int val) {
+        if (OptionIsSet(kCfdTuplesNumber) && (tuples_number_ != 0 && val == 0)) {
```
wildsor: Use SetConditionalOpts with empty condition, like in `MetricVerifier` (kMetric option)

src/core/config/option.h
```diff
@@ -79,7 +79,6 @@ class Option : public IOption {
     // predicate is equivalent to an always-true predicate, and thus must
     // always be last.
     Option& SetConditionalOpts(OptCondVector opt_cond) {
-        assert(opt_cond_.empty());
```
wildsor, line 82: Shouldn't be removed

wildsor (changes requested):

wildsor:
The assert says that `SetConditionalOpts` must only be called once (notice the underscore at the end of the name).

wildsor:
Could you be more specific on what was done to `CFDRelationData`? It looks like you did one thing, and then it pulled in a bunch of others? Also, the first commit contains stuff that gets removed in the next one. I think it'd make it easier to review if you split the commits in a way that reflects what changes happened.

src/core/algorithms/cfd/fd_first_algorithm.cpp
```diff
@@ -25,11 +25,33 @@ FDFirstAlgorithm::FDFirstAlgorithm() : CFDDiscovery() {
...
+        }
+    };
+    auto check_lhs = [](unsigned int val) {
+        if (val == 0) {
+            throw config::ConfigurationError("Maximum LHS size must be more than 0.");
+        }
+    };
+
+    RegisterOption(Option{&min_supp_, kCfdMinimumSupport, kDCfdMinimumSupport, 0u}.SetValueCheck(
+            std::move(check_supp)));
```
wildsor: Is this not going to error out if we don't set the value? Since the default is 0 and 0 errors out. I think 1 makes more sense.

src/core/algorithms/cfd/fd_first_algorithm.cpp
```diff
@@ -25,11 +25,33 @@ FDFirstAlgorithm::FDFirstAlgorithm() : CFDDiscovery() {
 void FDFirstAlgorithm::RegisterOptions() {
     DESBORDANTE_OPTION_USING;
 
-    Substrategy default_val = Substrategy::kDfs;
-    RegisterOption(Option{&min_supp_, kCfdMinimumSupport, kDCfdMinimumSupport, 0u});
-    RegisterOption(Option{&min_conf_, kCfdMinimumConfidence, kDCfdMinimumConfidence, 0.0});
-    RegisterOption(Option{&max_lhs_, kCfdMaximumLhs, kDCfdMaximumLhs, 0u});
-    RegisterOption(Option{&substrategy_, kCfdSubstrategy, kDCfdSubstrategy, default_val});
+    auto check_conf = [](double val) {
+        if (val <= 0 || val > 1) {
```
wildsor: Why did you change the interval?

src/core/algorithms/cfd/fd_first_algorithm.cpp
```diff
@@ -25,11 +25,33 @@ FDFirstAlgorithm::FDFirstAlgorithm() : CFDDiscovery() {
...
+            throw config::ConfigurationError("Minimum confidence must be a value between (0,1].");
+        }
+    };
+    auto check_supp = [this](unsigned int val) {
+        if (val == 0) {
+            throw config::ConfigurationError("Minimum support must be more than 0.");
+        } else if (val > relation_->GetNumRows()) {
+            throw config::ConfigurationError(
+                    "Minimum support must be less than or equal to the number of tuples.");
```
wildsor, line 38:
```suggestion
                    "Minimum support must be less than or equal to the number of tuples (" + std::to_string(relation_->GetNumRows()) + ").");
```

src/core/algorithms/cfd/fd_first_algorithm.cpp
```diff
@@ -25,11 +25,33 @@ FDFirstAlgorithm::FDFirstAlgorithm() : CFDDiscovery() {
...
-    RegisterOption(Option{&max_lhs_, kCfdMaximumLhs, kDCfdMaximumLhs, 0u});
-    RegisterOption(Option{&substrategy_, kCfdSubstrategy, kDCfdSubstrategy, default_val});
+    auto check_conf = [](double val) {
+        if (val <= 0 || val > 1) {
+            throw config::ConfigurationError("Minimum confidence must be a value between (0,1].");
+        }
+    };
+    auto check_supp = [this](unsigned int val) {
+        if (val == 0) {
+            throw config::ConfigurationError("Minimum support must be more than 0.");
```
wildsor:
```suggestion
            throw config::ConfigurationError("Minimum support must be greater than 0.");
```

src/core/algorithms/cfd/model/cfd_relation_data.cpp
```diff
@@ -204,25 +111,10 @@ int CFDRelationData::GetAttr(std::string const& s) const {
...
-    }
-}
-
-void CFDRelationData::Sort() {
-    std::sort(data_rows_.begin(), data_rows_.end(),
-              [](std::vector<int> const& a, std::vector<int> const& b) {
-                  return std::lexicographical_compare(a.begin(), b.begin(), a.end(), b.end());
-              });
-}
+    auto const& item_dict = column_data_.at(attr).GetValueDict();
```
wildsor: I think we don't need error checking here since the input is always going to be correct? I'm going off the assumption that this is always going to stay an algorithm-internal operation.
```suggestion
    auto const& item_dict = column_data_[attr].GetValueDict();
```

src/tests/unit/test_cfd_algos.cpp
```diff
@@ -87,13 +74,13 @@ TEST_F(CFDAlgorithmTest, FullTennisDataset) {
                                            "(windy, humidity, outlook) => play"};
     CheckCfdSetsEquality(actual_cfds, expected_cfds);
 
-    algorithm = CreateAlgorithmInstance(kTennis, 8, 0.85, "kBfs", 3);
+    algorithm = CreateAlgorithmInstance(kTennis, 8, 0.85, algos::cfd::Substrategy::kBfs, 3);
     algorithm->Execute();
     CheckCfdSetsEquality(actual_cfds, expected_cfds);
 }
 
 TEST_F(CFDAlgorithmTest, PartialMushroomDataset) {
```
wildsor, line 82: Idk if the "mushroom" dataset is from some article or not. If it is, leaving "partial" in makes sense, but then the file should be renamed. If it's not, then this test should be renamed. May want to look at other test names to see if they reference the functionality of cutting an existing dataset you've removed here and change names there too.

wildsor (changes requested):
First off, thanks. But I do want to be more specific on what I asked for [here](https://github.com/Desbordante/desbordante-core/pull/771#issuecomment-4704587594)

This is [the "what" and "why"](https://github.com/Desbordante/desbordante-core/wiki/Development%E2%80%90General#%D0%BE%D1%84%D0%BE%D1%80%D0%BC%D0%BB%D0%B5%D0%BD%D0%B8%D0%B5-%D0%BA%D0%BE%D0%BC%D0%BC%D0%B8%D1%82%D0%BE%D0%B2) (and a bit too much "how" for a commit body) I see in this PR.
Change 1:
Picking rows and columns from the data was removed because the user can do that easily themselves. So you removed the extra `CreateFrom`s and now have just one standard method. The very specifically named AddNewItemsIn{Full,Partial}Table were removed because they were only used for the overengineered `CreateFrom`s. You also removed testing relating to that functionality. And you also removed the options for this from option names and descriptions. And also shrunk the mushroom.csv dataset (might want to rename either the dataset or the test here) because it only made sense to have in full so that `CFDRelationData` could cut it.

Change 2:
`CreateFrom` now actually detects and logs if the number of columns in the data row from IDatasetStream is incorrect, instead of silently causing UB when it gets fewer and just ignoring the extra values when it gets more.

Change 3:
String to item ID mapping was moved to individual CFDColumnData instead of having a global (column number, value) mapping in CFDRelationData (why: slightly faster access, easier reasoning because of smaller scope). The `ItemDictionary` type alias was changed because it is in the context of a single column now.

Change 4:
`CFDRelationData::Size` was removed because it duplicates `AbstractRelationData::GetNumRows`.

Change 5:
Moved option checks in `FDFirstAlgorithm` into the options themselves, along with simplifying them.

Change 6:
Changed the confidence possible values to (0, 1] for some reason along the way? Because it doesn't actually make sense to search for dependencies with confidence 0?

Change 7:
Removed `CFDRelationData::ToFront`, `CFDRelationData::Sort`, `CFDDiscovery::NrCfds`. Because they were unused, I assume?

Change 8:
Simplified and corrected variable names in `FDFirstAlgorithm::ExecuteInternal` time calculations.

Next time, please take time to split your commits and explain what happened in each one of them in the body. If you don't, the reviewer has to reestablish what happened from the code and guess why it was done to gain enough understanding for a good review. But you already have all that context, so you could guide them instead.

The change grouping I've given is pretty loose, I didn't think that much about it. I guess for a compromise between reviewer convenience and reorganization time the commits could look like this: changes 1+2+3, change 4, changes 5+6, change 7, change 8. Look into `git add -p` and `git reset -p`. I'm not asking you to reorganize them here, not really a point except for having a good commit history, since I've already laid out what is happening in the PR here, but you can try just for practice. Oh, it's also generally a good idea to have each commit work correctly so that it's easier to figure out what broke what (see `git bisect`).

Edit: or maybe the method removals were all because of Change 1 and should be in the commit with it? I may have missed that at first

wildsor:
Is this ready to review? Press "re-request review" in the reviewers panel on the right if so.

Oddin60F:
> First off, thanks. But I do want to be more specific on what I asked for [here](https://github.com/Desbordante/desbordante-core/pull/771#issuecomment-4704587594)
> 
> This is [the "what" and "why"](https://github.com/Desbordante/desbordante-core/wiki/Development%E2%80%90General#%D0%BE%D1%84%D0%BE%D1%80%D0%BC%D0%BB%D0%B5%D0%BD%D0%B8%D0%B5-%D0%BA%D0%BE%D0%BC%D0%BC%D0%B8%D1%82%D0%BE%D0%B2) (and a bit too much "how" for a commit body) I see in this PR. Change 1: Picking rows and columns from the data was removed because the user can do that easily themselves. So you removed the extra `CreateFrom`s and now have just one standard method. The very specifically named AddNewItemsIn{Full,Partial}Table were removed because they were only used for the overengineered `CreateFrom`s. You also removed testing relating to that functionality. And you also removed the options for this from option names and descriptions. And also shrunk the mushroom.csv dataset (might want to rename either the dataset or the test here) because it only made sense to have in full so that `CFDRelationData` could cut it.
> 
> Change 2: `CreateFrom` now actually detects and logs if the number of columns in the data row from IDatasetStream is incorrect, instead of silently causing UB when it gets fewer and just ignoring the extra values when it gets more.
> 
> Change 3: String to item ID mapping was moved to individual CFDColumnData instead of having a global (column number, value) mapping in CFDRelationData (why: slightly faster access, easier reasoning because of smaller scope). The `ItemDictionary` type alias was changed because it is in the context of a single column now.
> 
> Change 4: `CFDRelationData::Size` was removed because it duplicates `AbstractRelationData::GetNumRows`.
> 
> Change 5: Moved option checks in `FDFirstAlgorithm` into the options themselves, along with simplifying them.
> 
> Change 6: Changed the confidence possible values to (0, 1] for some reason along the way? Because it doesn't actually make sense to search for dependencies with confidence 0?
> 
> Change 7: Removed `CFDRelationData::ToFront`, `CFDRelationData::Sort`, `CFDDiscovery::NrCfds`. Because they were unused, I assume?
> 
> Change 8: Simplified and corrected variable names in `FDFirstAlgorithm::ExecuteInternal` time calculations.
> 
> Next time, please take time to split your commits and explain what happened in each one of them in the body. If you don't, the reviewer has to reestablish what happened from the code and guess why it was done to gain enough understanding for a good review. But you already have all that context, so you could guide them instead.
> 
> The change grouping I've given is pretty loose, I didn't think that much about it. I guess for a compromise between reviewer convenience and reorganization time the commits could look like this: changes 1+2+3, change 4, changes 5+6, change 7, change 8. Look into `git add -p` and `git reset -p`. I'm not asking you to reorganize them here, not really a point except for having a good commit history, since I've already laid out what is happening in the PR here, but you can try just for practice. Oh, it's also generally a good idea to have each commit work correctly so that it's easier to figure out what broke what (see `git bisect`).
> 
> Edit: or maybe the method removals were all because of Change 1 and should be in the commit with it? I may have missed that at first

Thank you so much for the detailed review. I'll try to write more detailed comments next time.

src/core/algorithms/cfd/fd_first_algorithm.cpp
Oddin60F: My mistake. Returned the old check.

src/core/algorithms/cfd/fd_first_algorithm.cpp
```diff
@@ -42,48 +64,13 @@ void FDFirstAlgorithm::ResetStateCFD() {
 }
 
 void FDFirstAlgorithm::ExecuteInternal() {
+    LOG_DEBUG("Start");
+
```
wildsor: Extra code

wildsor (changes requested):
Looks good, except we don't do fixup commits in the same PR, put the changes into the other two commits. We don't use Conventional Commits either, but that's not critical

wildsor (approved):
LGTM

PR #776, opened 2026-06-13
Implement ValueToPy

p-senichenkov:
Implement functions to convert value in table to Python value

### DIFF
--- a/src/python_bindings/py_util/value_to_py.cpp
+++ b/src/python_bindings/py_util/value_to_py.cpp
@@ -1,0 +1,84 @@
+#include "python_bindings/py_util/value_to_py.h"
+
+#include <pybind11/pybind11.h>
+
+#include <algorithm>
+#include <cassert>
+#include <iterator>
+#include <stdexcept>
+#include <string>
+#include <vector>
+
+#include <pybind11/pytypes.h>
+
+#include "core/model/types/builtin.h"
+#include "core/model/types/create_type.h"
+#include "core/model/types/mixed_type.h"
+#include "core/model/types/type.h"
+
+using namespace model;
+namespace py = pybind11;
+
+namespace {
+py::object DateToPyObject(Date const& date) {
+    auto py_date = py::module_::import("datetime").attr("date");
+
+    auto ymd = date.year_month_day();
+    // Unlike the Python API, pybind cannot cast types here
+    return py_date(static_cast<unsigned>(ymd.year), static_cast<unsigned>(ymd.month),
+                   static_cast<unsigned>(ymd.day));
+}
+
+py::object MixedToPy(model::Type const* type, std::byte const* value) {
+    auto const* mixed_type = dynamic_cast<model::MixedType const*>(type);
+    assert(mixed_type);
+    auto real_type = CreateType(mixed_type->RetrieveTypeId(value), mixed_type->IsNullEqNull());
+    return python_bindings::ValueToPy(real_type.get(), mixed_type->RetrieveValue(value));
+}
+}  // namespace
+
+namespace python_bindings {
+py::object ValueToPy(Type const* type, std::byte const* value) {
+    if (!value) {
+        return py::none{};
+    }
+
+    switch (type->GetTypeId()) {
+        case TypeId::kBool:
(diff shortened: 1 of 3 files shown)


src/python_bindings/py_util/value_to_py.cpp
```diff
@@ -0,0 +1,82 @@
...
+        case TypeId::kDate:
+            return DateToPyObject(Type::GetValue<Date>(value));
+        case TypeId::kNull:
+        case TypeId::kEmpty:
+        case TypeId::kUndefined:
+            return py::none{};
+        case TypeId::kMixed:
+            return detail::MixedToPy(type, value);
+    }
+    throw std::runtime_error("Unknown type: " + type->ToString());
```
MichaelS239, line 72:
```suggestion
        case TypeId::kMixed:
            return detail::MixedToPy(type, value);
        default:
            throw std::runtime_error("Unknown type: " + type->ToString());
    }
```

src/python_bindings/py_util/value_to_py.cpp
```diff
@@ -0,0 +1,82 @@
+#include "python_bindings/py_util/value_to_py.h"
+
+#include <pybind11/pybind11.h>
+
+#include <algorithm>
+#include <vector>
```
MichaelS239, line 10:
```suggestion
#include <vector>
#include <iterator>
#include <stdexcept>
#include <cassert>
```

src/python_bindings/py_util/value_to_py.cpp
```diff
@@ -0,0 +1,82 @@
...
+            return py::int_(Type::GetValue<Int>(value));
+        case TypeId::kDouble:
+            return py::float_(Type::GetValue<Double>(value));
+        case TypeId::kBigInt:
+            // Parameters:
+            // 1. C string
+            // 2. First character after integer representation (nullptr -- take the entire string)
+            // 3. Base (0 -- auto-detect)
+            return py::reinterpret_steal<py::object>(
+                    PyLong_FromString(Type::GetValue<BigInt>(value).t.c_str(), nullptr, 0));
```
MichaelS239:
```suggestion
                    PyLong_FromString(static_cast<std::string const&>(Type::GetValue<BigInt>(value)).c_str(), nullptr, 0));
```
I guess this is more general way that doesn't depend on internal implementation.

src/python_bindings/py_util/value_to_py.cpp
wildsor: The benefit of the existing version is the compiler will warn when a new type id is added without an update here, I think it's fine.

src/python_bindings/py_util/value_to_py.h
```diff
@@ -0,0 +1,18 @@
...
+#include <pybind11/pybind11.h>
+
+#include <cstddef>
+#include <vector>
+
+#include "core/model/types/type.h"
+
+namespace python_bindings {
+namespace detail {
+pybind11::object MixedToPy(model::Type const* type, std::byte const* value);
```
wildsor: If this isn't going to be used by anything other than `ValueToPy`, then it should be moved to the cpp file. If it is, then it likely shouldn't be in `namespace detail`

wildsor (changes requested):

src/python_bindings/py_util/value_to_py.cpp
p-senichenkov: Yeah, that's exactly what I wanted

src/python_bindings/py_util/value_to_py.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+    auto ymd = date.year_month_day();
+    // Unlike the Python API, pybind cannot cast types here
+    return py_date(static_cast<unsigned>(ymd.year), static_cast<unsigned>(ymd.month),
+                   static_cast<unsigned>(ymd.day));
+}
+}  // namespace
+
+namespace python_bindings {
+py::object MixedToPy(model::Type const* type, std::byte const* value) {
```
wildsor, line 32: Move to unnamed namespace

wildsor (approved):

PR #777, opened 2026-06-13
Optimize gSpan algorithm 

perfotoran:
This PR brings major performance and memory improvements to the gSpan algorithm through data structure optimizations and multi-threading.


### Multi-threading Support:
Introduced parallel execution using a custom thread pool. The core mining logic was isolated into a dedicated `SubgraphMiner` class, allowing worker threads to traverse independent branches of the DFS tree safely.       
                                                                                                                                                                                                  
### Memory Optimizations:
Significantly reduced memory fragmentation and dynamic allocation overhead across the algorithm. The graph representation was migrated to a contiguous memory format (`Compressed Sparse Row)`. Additionally, nested arrays were flattened, standard maps were replaced with open-addressing hash maps (via Boost), and internal buffers are now reused during recursion.

### Algorithmic Improvements:

- Projection Caching: Optimized the projection mechanism (inverted index). The algorithm now effectively tracks graph embeddings and caches edge sources, avoiding redundant subgraph isomorphism checks from scratch.

- History Tree: Added a history structure to track the current DFS path, providing O(1) lookups to check if vertices/edges are already occupied by the current pattern.

- Rightmost Path: Refined the rightmost path calculation to reduce redundant checks and early-prune invalid DFS codes.

### DIFF
--- a/src/core/algorithms/fsm/gspan/gspan.cpp
+++ b/src/core/algorithms/fsm/gspan/gspan.cpp
@@ -480,7 +480,7 @@ bool GSpan::IsProjectionMin(DFSCode const& code) {
         } else {
             // edge is forward, so ensure no backward edges exist,
             // then ensure the edge is minimal.
-            if (ExistsBackwards(code, projection_start_index) ||
+            if (ExistsBackwards(projection_start_index) ||
                 !IsForwardMin(code, ee, projection_start_index)) {
                 return false;
             }
@@ -620,7 +620,7 @@ bool GSpan::IsForwardMin(gspan::DFSCode const& code, ExtendedEdge const& ee,
     return true;
 }
 
-bool GSpan::ExistsBackwards(DFSCode const& code, size_t projection_start_index) {
+bool GSpan::ExistsBackwards(size_t projection_start_index) {
     size_t projection_end_index = min_projection_.size();
 
     for (auto j = projection_start_index; j < projection_end_index; j++) {
--- a/src/core/algorithms/fsm/gspan/gspan.cpp
+++ b/src/core/algorithms/fsm/gspan/gspan.cpp
@@ -372,7 +372,8 @@ void GSpan::MineSubgraph(Projection const& projection, DFSCode& code) {
     for (auto const& [ee, proj] : backward_pmap) {
         MineChild(proj, ee, code);
     }
-    for (auto const& [ee, proj] : forward_pmap) {
+    for (auto it = forward_pmap.rbegin(); it != forward_pmap.rend(); it++) {
+        auto [ee, proj] = *it;
         MineChild(proj, ee, code);
     }
 }
--- a/src/core/algorithms/fsm/gspan/gspan.cpp
+++ b/src/core/algorithms/fsm/gspan/gspan.cpp
@@ -22,171 +22,6 @@ std::vector<vertex_t> const& FindAllWithLabel(int target_label, graph_t const& g
     return graph[boost::graph_bundle].label_to_vertices.at(target_label);
 }
 
-struct FlatIsoms {
-    std::vector<vertex_t> data;
-    std::vector<size_t> offsets;
-
-    void clear() {
-        data.clear();
-        offsets.clear();
-    }
-
-    size_t size() const {
-        return offsets.size();
-    }
-
-    void swap(FlatIsoms& other) {
-        data.swap(other.data);
-        offsets.swap(other.offsets);
-    }
-
-    struct Proxy {
-        vertex_t const* ptr;
-        size_t sz;
-
-        size_t size() const {
-            return sz;
-        }
-
-        vertex_t operator[](size_t i) const {
-            return ptr[i];
-        }
-
-        vertex_t const* begin() const {
-            return ptr;
-        }
-
-        vertex_t const* end() const {
-            return ptr + sz;
-        }
-    };
-
-    void push(Proxy const& iso) {
-        offsets.push_back(data.size());
-        data.insert(data.end(), iso.begin(), iso.end());
-    }
-
-    void push_with_extra(Proxy const& iso, vertex_t extra) {
-        offsets.push_back(data.size());
-        data.insert(data.end(), iso.begin(), iso.end());
-        data.push_back(extra);
-    }
-
-    Proxy operator[](size_t i) const {
-        size_t start = offsets[i];
-        size_t end = (i + 1 < offsets.size()) ? offsets[i + 1] : data.size();
-        return {data.data() + start, end - start};
-    }
-
-    struct Iterator {
-        FlatIsoms const* list;
-        size_t idx;
-
-        Proxy operator*() const {
-            return (*list)[idx];
-        }
-
-        Iterator& operator++() {
-            ++idx;
-            return *this;
-        }
-
-        bool operator!=(Iterator const& other) const {
-            return idx != other.idx;
-        }
-    };
-
-    Iterator begin() const {
-        return {this, 0};
-    }
-
-    Iterator end() const {
-        return {this, offsets.size()};
-    }
-};
-
-// Find all isomorphisms between graph described by DFSCode and boost::graph, each
-// isomorphism is represented by a map
-FlatIsoms const& SubgraphIsomorphisms(DFSCode const& code, graph_t const& graph) {
-    LOG_TRACE("Finding subgraph isomorphisms: pattern size={}, target vertices={}, edges={}",
-              code.Size(), boost::num_vertices(graph), boost::num_edges(graph));
-    thread_local FlatIsoms isoms;
-    thread_local FlatIsoms update_isoms;
-    thread_local std::vector<int> inv;
-
-    isoms.clear();
-    size_t n = boost::num_vertices(graph);
-
-    if (inv.size() < n) {
-        inv.resize(n, -1);
-    }
-
-    // Initial isomorphisms by finding all vertices with same label as vertex 0 in code
-    int start_label = code.GetExtendedEdges()[0].vertex1.label;
-    auto const vertices = FindAllWithLabel(start_label, graph);
-    isoms.offsets.reserve(vertices.size());
-    isoms.data.reserve(vertices.size());
-
-    for (vertex_t vertex : vertices) {
-        isoms.offsets.push_back(isoms.data.size());
-        isoms.data.push_back(vertex);
-    }
-    LOG_TRACE("Initial candidate mappings: {}", isoms.size());
-
-    // Each extended edge will update partial isomorphisms.
-    // For forward edge, each isomorphism will be either extended or discarded.
-    // For backward edge, each isomorphism will be either unchanged or discarded.
-
-    for (ExtendedEdge const& ee : code.GetExtendedEdges()) {
-        int v1 = ee.vertex1.id;
-        int v2 = ee.vertex2.id;
-        int v2_label = ee.vertex2.label;
-        int edge_label = ee.label;
-
-        update_isoms.clear();
-        update_isoms.offsets.reserve(isoms.size());
-        update_isoms.data.reserve(isoms.data.size());
-        for (auto iso : isoms) {
-            auto mapped_v1 = iso[v1];
-
-            // If it is a forward edge extension
-            if (v1 < v2) {
-                for (int dfs_id = 0; dfs_id < iso.size(); ++dfs_id) {
-                    inv[iso[dfs_id]] = dfs_id;
-                }
-                // For each neighbor of the vertex corresponding to v1
-                for (auto edge : boost::make_iterator_range(boost::out_edges(mapped_v1, graph))) {
-                    vertex_t mapped_v2 = (boost::source(edge, graph) == mapped_v1)
-                                                 ? boost::target(edge, graph)
-                                                 : boost::source(edge, graph);
-                    auto& mapped_edge = graph[edge];
-
-                    if (v2_label == graph[mapped_v2].label && (inv[mapped_v2] == -1) &&
-                        (edge_label == mapped_edge.label)) {
-                        update_isoms.push_with_extra(iso, mapped_v2);
-                    }
-                }
-            } else {
-                // If it is a backward edge extension.
-                // V2 has been visited, only require mappedV1 and mappedV2 are connected in graph.
-                auto mapped_v2 = iso[v2];
-                auto [mapped_edge, is_neighbors] = boost::edge(mapped_v1, mapped_v2, graph);
-                if (is_neighbors && edge_label == graph[mapped_edge].label) {
-                    update_isoms.push(iso);
-                }
-            }
-
-            for (size_t dfs_id = 0; dfs_id < iso.size(); ++dfs_id) {
-                inv[iso[dfs_id]] = -1;
-            }
-        }
-        isoms.swap(update_isoms);
-    }
-
-    LOG_TRACE("Found {} valid isomorphisms", isoms.size());
-    return isoms;
-}
-
 // Precalculate the list of vertices having each label
 void PrecalculateLabelsToVertices(graph_t& graph) {
     auto& label_to_verts = graph[boost::graph_bundle].label_to_vertices;
@@ -325,22 +160,22 @@ void GSpan::MineSubgraphs() {
 
     // Set with all the graph ids
     LOG_DEBUG("Building active graph set");
-    boost::unordered_flat_set<int> graph_ids;
+    Projection embeddings;
     for (size_t i = 0; i < pruned_graphs_.size(); i++) {
         graph_t& graph = pruned_graphs_[i];
         if (boost::num_vertices(graph) != 0) {
-            graph_ids.insert(i);
+            embeddings.emplace_back(i, FlatIsoms{});
             PrecalculateLabelsToVertices(graph);
         } else {
             empty_graphs_removed_++;
         }
     }
-    LOG_DEBUG("Active graphs: {}, empty graphs removed: {}", graph_ids.size(),
+    LOG_DEBUG("Active graphs: {}, empty graphs removed: {}", embeddings.size(),
               empty_graphs_removed_);
 
     if (frequent_vertex_labels_.size() != 0) {
-        LOG_DEBUG("Starting DFS search with {} graphs", graph_ids.size());
-        GSpanDFS(DFSCode(), std::move(graph_ids));
+        LOG_DEBUG("Starting DFS search with {} graphs", embeddings.size());
+        GSpanDFS(DFSCode(), std::move(embeddings));
     }
 
     LOG_INFO("GSpan complete: {} frequent subgraphs found", frequent_subgraphs_.size());
@@ -436,9 +271,9 @@ void GSpan::RemoveInfrequentVertexPairs() {
     }
 }
 
-void GSpan::GSpanDFS(gspan::DFSCode const& code, boost::unordered_flat_set<int> graph_ids) {
+void GSpan::GSpanDFS(gspan::DFSCode const& code, Projection embeddings) {
     LOG_TRACE("DFS step: pattern size={}, candidate graphs={}, patterns found={}", code.Size(),
-              graph_ids.size(), frequent_subgraphs_.size());
+              embeddings.size(), frequent_subgraphs_.size());
 
     // If we have reached the maximum size, we do not need to extend this graph
     if (code.Size() == static_cast<size_t>(max_number_of_edges_)) {
@@ -449,11 +284,11 @@ void GSpan::GSpanDFS(gspan::DFSCode const& code, boost::unordered_flat_set<int> 
     // Find all the extensions of this graph, with their support values.
     // They are stored in a map where the key is an extended edge, and the value
     // is the list of graph ids where this edge extends the current subgraph.
-    auto extensions = RightMostPathExtensions(code, std::move(graph_ids));
+    auto extensions = RightMostPathExtensions(code, std::move(embeddings));
     LOG_TRACE("Found {} candidate extensions", extensions.size());
 
-    for (auto& [extension, new_graph_ids] : extensions) {
-        int sup = new_graph_ids.size();
+    for (auto& [extension, projection] : extensions) {
+        int sup = projection.size();
 
         // If the support is enough
         if (sup >= min_sup_) {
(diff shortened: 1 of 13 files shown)


src/core/algorithms/fsm/gspan/miner/subgraph_miner.cpp
```diff
@@ -0,0 +1,442 @@
...
+
+#include <boost/unordered/unordered_flat_set.hpp>
+
+#include "core/util/logger.h"
+
+namespace gspan {
+
+namespace {
+
+int CountSupport(Projection const& projection) {
```
pppppptttttt: `support` is not negative, `int` -> `size_t`? this change also removes `static_cast` below

src/core/algorithms/fsm/gspan/miner/subgraph_miner.cpp
```diff
@@ -0,0 +1,442 @@
...
+    }
+
+    return support;
+}
+
+}  // namespace
+
+void SubgraphMiner::MineChild(Projection const& projection, ExtendedEdge const& new_edge,
+                              DFSCode code) {
+    int support = CountSupport(projection);
```
pppppptttttt:
```suggestion
    int const support = CountSupport(projection);
```

src/core/algorithms/fsm/gspan/miner/subgraph_miner.cpp
```diff
@@ -0,0 +1,442 @@
...
+        frequent_subgraphs_.emplace_back(frequent_subgraphs_.size(), code,
+                                         std::move(original_graph_ids), support);
+        MineSubgraph(projection, code);
+    }
+
+    code.Pop();
+}
+
+void SubgraphMiner::MineSubgraph(Projection const& projection, DFSCode const& code) {
+    if (code.Size() == static_cast<size_t>(max_number_of_edges_)) {
```
pppppptttttt: `max_number_of_edges_` is already declared as `size_t`

src/core/algorithms/fsm/gspan/miner/subgraph_miner.cpp
```diff
@@ -0,0 +1,442 @@
...
+            auto p_ptr = std::make_shared<Projection>(std::move(proj));
+            thread_pool_->Spawn(
+                    [this, p_ptr = std::move(p_ptr), ee, code](int t_id) {
+                        (*miners_)[t_id]->MineChild(std::move(*p_ptr), ee, code);
+                    },
+                    pending);
+        } else {
+            MineChild(std::move(proj), ee, code);
+        }
+    }
```
pppppptttttt, line 63: move cycle body to lambda/member function to avoid code duplication?

```suggestion
  auto dispatch = [this, code, &pending](Projection&& proj, ExtendedEdge const& ee) {
      if (thread_pool_) {
          auto p_ptr = std::make_shared<Projection>(std::move(proj));
          thread_pool_->Spawn(
                  [this, p_ptr = std::move(p_ptr), ee, code](int t_id) {
                      (*miners_)[t_id]->MineChild(std::move(*p_ptr), ee, code);
                  },
                  pending);
      } else {
          MineChild(std::move(proj), ee, code);
      }
  };
  
  for (auto& [ee, proj] : backward_pmap) {
      dispatch(std::move(proj), ee);
  }
  for (auto it = forward_pmap.rbegin(); it != forward_pmap.rend(); ++it) {
      auto& [ee, proj] = *it;
      dispatch(std::move(proj), ee);
  }
```

src/core/algorithms/fsm/gspan/types/extended_edge.h
```diff
@@ -41,4 +41,33 @@ struct ExtendedEdge {
     std::string ToString() const;
 };
 
+struct ExtendedEdgeProjectCompare {
+    bool operator()(ExtendedEdge const& first, ExtendedEdge const& second) const {
```
pppppptttttt: comparators may be marked as `noexcept`

src/core/algorithms/fsm/gspan/gspan.cpp
```diff
@@ -224,30 +97,113 @@ void GSpan::MineSubgraphs() {
...
+    ProjectionMap embeddings = GetInitialEdges();
+
+    ThreadPool pool(threads_num_);
+    std::vector<std::unique_ptr<SubgraphMiner>> miners;
+    for (size_t i = 0; i < threads_num_ + 1; ++i) {
+        auto miner =
+                std::make_unique<SubgraphMiner>(pruned_csr_graphs_, min_sup_, max_number_of_edges_);
+        miner->SetParallelContext(&pool, &miners, i);
+        miners.push_back(std::move(miner));
+    }
```
pppppptttttt: for each `miner` `thread_pool != nullptr`, and we get `if (thread_pool_) {...} else {/*dead code*/}` in SubgraphMiner methods. maybe just add assertion for thread pool nullability and remove all else branches in corresponding checks?

pppppptttttt: also, if `thread_num_ == 1,` user might expect one thread, but he will get one main thread and one worker insied thread pool; if it is a correct behaviour, I suggest documenting it for the user, otherwise create thread pool if and only if `thread_nums_ > 1` with `thread_nums_ - 1` workers

src/core/algorithms/fsm/gspan/utils/thread_pool.h
```diff
@@ -0,0 +1,81 @@
...
+        for (auto& t : threads_) t.join();
+    }
+
+    template <typename F>
+    void Spawn(F&& f, std::atomic<int>& pending_counter) {
+        pending_counter.fetch_add(1, std::memory_order_relaxed);
+        {
+            std::unique_lock<std::mutex> lock(mutex_);
+            tasks_.emplace_back([f = std::forward<F>(f), &pending_counter](int thread_id) {
+                f(thread_id);
```
pppppptttttt: exception from `f` (for example `std::bad_alloc`) will cause `std::terminate` in thread, and `pending` won't be decremented; `util::WorkerThreadPool` solves it this way:

```cpp
try {
    pool.work_();
} catch (...) {
    pool.barrier_.ArriveAndWait();
    throw;
}
pool.barrier_.ArriveAndWait();
```

src/core/algorithms/fsm/gspan/miner/subgraph_miner.cpp
```diff
@@ -0,0 +1,442 @@
...
+            original_graph_ids.insert(
+                    graph_database_[proj_entry.graph_id][boost::graph_bundle].original_id);
+        }
+
+        frequent_subgraphs_.emplace_back(frequent_subgraphs_.size(), code,
+                                         std::move(original_graph_ids), support);
+        MineSubgraph(projection, code);
+    }
+
+    code.Pop();
```
pppppptttttt: code is passed by value, it will be destructed in the next line

src/core/algorithms/fsm/gspan/miner/subgraph_miner.cpp
```diff
@@ -0,0 +1,442 @@
...
+}
+
+}  // namespace
+
+void SubgraphMiner::MineChild(Projection const& projection, ExtendedEdge const& new_edge,
+                              DFSCode code) {
+    int support = CountSupport(projection);
+    if (static_cast<size_t>(support) < min_sup_) {
+        return;
+    }
```
pppppptttttt: task spawns for every extension, and many of them will do early return

I would check this before creating task in thread pool, it reduces `shared_pointer` and `std::function` allocations, two mutex locks (once in `Spawn` itself, and once in `WorkerLoop` to acquire task), dfscode copies, etc.

imo, it should be in separate tasks if a) `CountSupport` is slow or b) it filters out the minority

however, benchmarcking should clarify it

src/core/algorithms/fsm/gspan/miner/subgraph_miner.cpp
```diff
@@ -0,0 +1,442 @@
...
+        }
+    }
+    for (auto it = forward_pmap.rbegin(); it != forward_pmap.rend(); it++) {
+        auto& [ee, proj] = *it;
+        if (thread_pool_) {
+            auto p_ptr = std::make_shared<Projection>(std::move(proj));
+            thread_pool_->Spawn(
+                    [this, p_ptr = std::move(p_ptr), ee, code](int t_id) {
+                        (*miners_)[t_id]->MineChild(std::move(*p_ptr), ee, code);
+                    },
```
pppppptttttt: `code` is being copied twice here; if lambda will become mutable, `code` may be moved
```suggestion
                    [this, p_ptr = std::move(p_ptr), ee, code](int t_id) mutable {
                        (*miners_)[t_id]->MineChild(std::move(*p_ptr), ee, std::move(code));
                    },
```

src/core/algorithms/fsm/gspan/gspan.h
```diff
@@ -57,10 +58,10 @@ class GSpan : public Algorithm {
     void RegisterOptions();
 
 public:
-    void MineSubgraphs();
-
     GSpan();
 
+    void Launch();
```
pppppptttttt: imo, the old method name was more informative than new one

src/core/algorithms/fsm/gspan/miner/subgraph_miner.h
```diff
@@ -0,0 +1,81 @@
...
+
+    SubgraphMiner(std::vector<csr_graph_t> const& graph_database, int min_sup,
+                  int max_number_of_edges)
+        : graph_database_(graph_database),
+          min_sup_(min_sup),
+          max_number_of_edges_(max_number_of_edges) {
+        int max_edges = 0;
+        int max_vertices = 0;
+        for (size_t i = 0; i < graph_database_.size(); i++) {
+            auto& graph = graph_database_[i];
```
pppppptttttt:
```suggestion
        for (auto const& graph : graph_database_) {
```

src/core/algorithms/fsm/gspan/utils/utils.h
```diff
@@ -0,0 +1,52 @@
...
+        std::vector<GraphType> const& graph_db) {
+    boost::unordered_flat_set<int> original_ids;
+    original_ids.reserve(internal_ids.size());
+    for (int id : internal_ids) {
+        original_ids.insert(graph_db[id][boost::graph_bundle].original_id);
+    }
+    return original_ids;
+}
+
+inline csr_graph_t ConvertToCSR(graph_t const& src_graph) {
```
pppppptttttt: this function does not look trivial, maybe move it to utils.cpp or graph.cpp?

src/core/algorithms/fsm/gspan/utils/utils.h
```diff
@@ -0,0 +1,52 @@
...
+#include <boost/unordered/unordered_flat_set.hpp>
+
+#include "types/graph.h"
+
+namespace gspan {
+
+template <typename GraphType>
+boost::unordered_flat_set<int> TranslateToOriginalIds(
+        boost::unordered_flat_set<int> const& internal_ids,
+        std::vector<GraphType> const& graph_db) {
```
pppppptttttt: this function is called only with `GSpan::pruned_graphs_`, so it is instanciated only by one type; consider removing temlpate and moving it's definition to utils.cpp or graph.cpp

src/core/algorithms/fsm/gspan/miner/subgraph_miner.cpp
```diff
@@ -0,0 +1,442 @@
...
+
+bool SubgraphMiner::IsCanonical(DFSCode const& code) {
+    LOG_TRACE("Checking canonicity: pattern size={}", code.Size());
+    min_graph_.BuildFromDFSCode(code);
+    rightmost_path_.clear();
+    rightmost_path_.push_back(0);
+
+    if (code.Size() == 1) {
+        return true;
+    }
```
pppppptttttt: do we need to build min grah before early return?

src/tests/unit/test_gspan.cpp
pppppptttttt, line 1: consider adding tests that validate correctness with different number of threads

src/core/algorithms/fsm/gspan/miner/subgraph_miner.cpp
pppppptttttt: my main issue with this is that this function does not have any conditions and upper bounds for task creation. each costs `Projection` allocation, mutex lock (which is global to thread pool), allocation for std::function, etc.

on lower levels of DFS useful work may take less time than task creation and it's execution inside the thread pool.

possible solution - parallel upper levels of tree, and compute lower levels sequentially. consider adding bounds for depth and amount of work (projection.size() for example), and refactoring to something like that:

```cpp
static constexpr size_t kMaxParallelDepth = ...;
static constexpr size_t kMinProjectionSizeForTask = ...;

bool ShouldSpawn(size_t code_size, size_t projection_size) const noexcept {
    return thread_pool_ != nullptr && code_size < kMaxParallelDepth &&
           projection_size >= kMinProjectionSizeForTask;
}

void Dispatch(ExtendedEdge const& ee, Projection&& proj, DFSCode const& code, std::atomic<int>& pending) {
    if (...) { return; } // early exits check
    if (!ShouldSpawn(code.size(), proj.size())) { MineChild(...); return; }
    // spawn task in thread pool otherwise
}

void MineSubgraph() {
    // ...
    for (auto& [ee, proj] : backward_pmap) {
        Dispatch(ee, std::move(proj), code, pending);
    }
    for (auto it = forward_pmap.rbegin(); it != forward_pmap.rend(); ++it) {
        auto& [ee, proj] = *it;
        Dispatch(ee, std::move(proj), code, pending);
    }
    // ...
}
```

and as in early return for MineChild - I may be wrong and benchmarking should clarify this situation also

pppppptttttt (changes requested):
Good job overall. I can see why a custom ThreadPool was needed, but there are issues with it and its usage. Please run benchmarks before and after making changes related to it.

src/core/algorithms/fsm/gspan/miner/subgraph_miner.cpp
perfotoran: It helped, thanks. I also developed this idea further, counting support "on the fly" while building Projection

src/core/algorithms/fsm/gspan/miner/subgraph_miner.cpp
perfotoran: Thanks for the suggestion. I benchmarked and found its time-memory tradeoff (saving memory costs proportional performance), to me its not worth it

src/core/algorithms/fsm/gspan/miner/subgraph_miner.h
```diff
@@ -0,0 +1,89 @@
...
+    bool IsCanonical(DFSCode const& code);
+    bool IsProjectionMin(DFSCode const& code);
+    bool IsBackwardMin(DFSCode const& code, ExtendedEdge const& ee, size_t projection_start_index);
+    bool IsForwardMin(DFSCode const& code, ExtendedEdge const& ee, size_t projection_start_index);
+    bool ExistsBackwards(size_t projection_start_index);
+
+    void UpdateRightmostPath(DFSCode const& code, size_t size);
+
+public:
+    void SetParallelContext(ThreadPool* pool, std::vector<std::unique_ptr<SubgraphMiner>>* miners,
```
pppppptttttt: it's better to put these asserts in `MineSubgraph`, because if we will forget to call this method, we will dereference nullptr (and it will be better to fail with assertion or exception there). also, if this setter is required for `SubgraphMiner` to work, why not pass these parameters in constructor?

src/core/algorithms/fsm/gspan/types/min_graph.h
```diff
@@ -0,0 +1,75 @@
...
+            vertices_[ee.vertex1.id].edges.push_back(
+                    {ee.vertex1.id, ee.label, ee.vertex2.id, edge_id});
+            vertices_[ee.vertex2.id].edges.push_back(
+                    {ee.vertex2.id, ee.label, ee.vertex1.id, edge_id});
+            edge_id++;
+        }
+        num_edges_ = edge_id;
+    }
+
+    MinVertex const& operator[](size_t i) const {
```
pppppptttttt, line 54: please mark small methods and functions that can't throw as `noexcept`. it helps a tiny bit with the resulting binary size (compiler doesn't have to worry about stack unwinding), but mostly with the code readability - reader don't have to make assumptions about your code

pppppptttttt: the same goes with `const` modifier on variables - it is easier to read block of code while understanding which variables are actually immutable

src/core/algorithms/fsm/gspan/utils/utils.cpp
```diff
@@ -0,0 +1,44 @@
...
+    edge_props.reserve(boost::num_edges(src_graph));
+
+    auto edge_iters = boost::edges(src_graph);
+    for (auto it = edge_iters.first; it != edge_iters.second; ++it) {
+        int u = boost::source(*it, src_graph);
+        int v = boost::target(*it, src_graph);
+
+        edges.emplace_back(u, v);
+        edge_props.push_back(src_graph[*it]);
+    }
```
pppppptttttt:
```suggestion
    for (auto edge : boost::make_iterator_range(boost::edges(src_graph))) {
        edges.emplace_back(boost::source(edge, src_graph), boost::target(edge, src_graph));
        edge_props.push_back(src_graph[edge]);
    }
```

src/core/algorithms/fsm/gspan/utils/utils.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+    csr_graph_t csr(boost::edges_are_unsorted_multi_pass, edges.begin(), edges.end(),
+                    edge_props.begin(), boost::num_vertices(src_graph),
+                    src_graph[boost::graph_bundle]);
+
+    auto vertex_iters = boost::vertices(src_graph);
+    for (auto it = vertex_iters.first; it != vertex_iters.second; ++it) {
+        csr[*it] = src_graph[*it];
+    }
+
```
pppppptttttt: same as for edges

src/core/algorithms/fsm/gspan/miner/subgraph_miner.cpp
```diff
@@ -0,0 +1,412 @@
...
+    auto last_edge = history_.GetEdge(rightmost_path_[0]);
+    auto last_node = boost::target(last_edge, graph);
+
+    for (auto ln_edge : boost::make_iterator_range(boost::out_edges(last_node, graph))) {
+        if (history_.HasEdge(graph[ln_edge].id)) {
+            continue;
+        }
+
+        auto ln_edge_to = boost::target(ln_edge, graph);
+        for (size_t i = rightmost_path_.size() - 1; i > 0; i--) {
```
pppppptttttt, line 97: add `assert(!rightmost_path_.empty())` (or respective exception) above; if someday it will be empty, this will be painful to debug

src/core/algorithms/fsm/gspan/miner/subgraph_miner.cpp
```diff
@@ -0,0 +1,412 @@
...
+// index of the edge 'discovering' the rightmost vertex, the second is the index
+// of the edge discovering the 'from' vertex of the first edge, and so on.
+// DFSCode is treated as if it is truncated to the given size.
+void SubgraphMiner::UpdateRightmostPath(gspan::DFSCode const& code, size_t size) {
+    rightmost_path_.clear();
+    int prev_id = -1;
+
+    // Go in reverse, since we need to first look for the edge that discovered
+    // the rightmost vertex
+    for (auto i = size; i > 0; --i) {
```
pppppptttttt: `auto` -> `size_t`?

src/core/algorithms/fsm/gspan/miner/subgraph_miner.h
```diff
@@ -0,0 +1,89 @@
...
+    SubgraphMiner(std::vector<csr_graph_t> const& graph_database, int min_sup,
+                  int max_number_of_edges)
+        : graph_database_(graph_database),
+          min_sup_(min_sup),
+          max_number_of_edges_(max_number_of_edges) {
+        int max_edges = 0;
+        int max_vertices = 0;
+        for (auto const& graph : graph_database_) {
+            max_edges = std::max(max_edges, static_cast<int>(boost::num_edges(graph)));
+            max_vertices = std::max(max_vertices, static_cast<int>(boost::num_vertices(graph)));
```
pppppptttttt: they can be declared as `size_t`; in `History::Reset` they are used exactly as `size_t`

src/core/algorithms/fsm/gspan/miner/subgraph_miner.h
```diff
@@ -0,0 +1,89 @@
...
+        assert(pool != nullptr);
+        assert(miners != nullptr);
+
+        thread_pool_ = pool;
+        miners_ = miners;
+        thread_id_ = thread_id;
+    }
+
+    SubgraphMiner(std::vector<csr_graph_t> const& graph_database, int min_sup,
+                  int max_number_of_edges)
```
pppppptttttt: corresponding fields (`min_sup_`, `max_number_of_edges_`) are declared as `size_t`

src/core/algorithms/fsm/gspan/utils/utils.cpp
```diff
@@ -0,0 +1,44 @@
...
+    csr_graph_t csr(boost::edges_are_unsorted_multi_pass, edges.begin(), edges.end(),
+                    edge_props.begin(), boost::num_vertices(src_graph),
+                    src_graph[boost::graph_bundle]);
+
+    auto vertex_iters = boost::vertices(src_graph);
+    for (auto it = vertex_iters.first; it != vertex_iters.second; ++it) {
+        csr[*it] = src_graph[*it];
+    }
```
pppppptttttt: same as for edges

src/core/algorithms/fsm/gspan/gspan.cpp
```diff
@@ -7,155 +7,27 @@
...
-}
-
-}  // namespace
-
 GSpan::GSpan() : Algorithm() {
     RegisterOptions();
     MakeOptionsAvailable({config::names::kGraphDatabase, config::names::kGSpanMinimumSupport,
                           config::names::kOutputSingleVertices, config::names::kMaxNumberOfEdges,
-                          config::names::kGSpanOutputPath});
+                          config::names::kGSpanOutputPath, config::names::kThreads});
```
pppppptttttt: `MakeOptionsAvailable` is called on `config::names::kThreads` twice (the second time inside `MakeExecuteOptsAvailable`); seems unnecessary

src/core/algorithms/fsm/gspan/types/graph.h
```diff
@@ -0,0 +1,32 @@
...
+struct GraphProps {
+    int original_id = 0;
+};
+
+using graph_t =
+        boost::adjacency_list<boost::vecS, boost::vecS, boost::directedS, Vertex, Edge, GraphProps>;
+using vertex_t = boost::graph_traits<graph_t>::vertex_descriptor;
+using edge_t = boost::graph_traits<graph_t>::edge_descriptor;
+
+using csr_graph_t = boost::compressed_sparse_row_graph<boost::directedS, Vertex, Edge, GraphProps>;
```
pppppptttttt: setting Vertex and EdgeIndex (two parameters after `GraphProps`) template parameters in graph definition as `uint32_t` seems interesting - they are `size_t` by default as far as i know. These parameters are defining underlying integer types for vertices and edges which `boost::compressed_sparse_row_graph` operates(https://www.boost.org/doc/libs/latest/libs/graph/doc/html/graph/graph_classes/overview.html#_compressed_sparse_row_graph), so sizeof(vertex_t) and sizeof(edge_t) will also reduce. Besides the memory optimization, it can make graph more cache friendly

however, I can’t estimate how long this refactoring will take and whether it will be worth it; so I suggest to try it later on different PR

pppppptttttt (commented):
A few minor improvements remains from my point of view. I would like to see more `const` and `noexcept` modifiers, IMO they are significantly improve code readability.

src/core/algorithms/fsm/gspan/miner/subgraph_miner.h
pppppptttttt: if it is the case maybe declare them as references rather than pointers

src/core/algorithms/fsm/gspan/types/graph.h
perfotoran: Updating `csr_graph_t` turned out to be a 3 line change that doesn't break the build. I will leave this optimization  here as it noticeably benefits both memory usage and performance. A separate follow-up PR will address aligning the internal `int` IDs to `uint32_t` for  type consistency

pppppptttttt: Good to hear it, I was concerned that it could break something on the way and lead to the massive diff

src/core/algorithms/fsm/gspan/miner/subgraph_miner.cpp
```diff
@@ -0,0 +1,414 @@
...
+    auto const last_node = boost::target(last_edge, graph);
+
+    for (auto const ln_edge : boost::make_iterator_range(boost::out_edges(last_node, graph))) {
+        if (history_.HasEdge(graph[ln_edge].id)) {
+            continue;
+        }
+
+        auto const ln_edge_to = boost::target(ln_edge, graph);
+        assert(!rightmost_path_.empty());
+        for (size_t i = rightmost_path_.size() - 1; i > 0; i--) {
```
pppppptttttt: there are actually `rightmost_path_[0]` reads above (on line 84) I didn't notice previous time; please move assertion above

also, there are similar loops in IsBackwardMin and ExistsBackwards (lines 262 and 274)

I may be wrong and it is never empty before these loops, and if I am, please document why

pppppptttttt (commented):
LGTM. Rewrite the commits history and I'll approve

src/core/algorithms/fsm/gspan/miner/subgraph_miner.cpp
perfotoran: `rightmost_path_` guaranteed non-empty since it initialized with 0 in `IsCanonical`, and `UpdateRightmostPath` always restores at least the root forward edge

pppppptttttt (approved):

PR #778, opened 2026-06-14
Improve kBool type

MichaelS239:
This PR adds various fixes and improvements for kBool type:
- Type deduction for kBool was fixed for cases when one column contains both integers and bools
- kBool was made metrizable
- Info about kBool was added to `builtin.h`
- Tests for kBool type deduction were added

PR #779, opened 2026-06-16
Optimize TypedColumnData

MichaelS239:
This PR adds a few optimizations for type deduction process, achieving 2x speedup on `iowa1kk.csv` and `EpicMeds.csv`.

Depends on #778


src/core/model/types/bool_type.h
```diff
@@ -49,6 +50,10 @@ class BoolType final : public Type {
         std::memcpy(&v, value, sizeof(v));
         return std::hash<Bool>{}(v);
     }
+
+    double Dist(std::byte const* l, std::byte const* r) const override {
+        return (GetValue<Bool>(l) == GetValue<Bool>(r)) ? 0 : 1;
```
wildsor:
```suggestion
        return GetValue<Bool>(l) != GetValue<Bool>(r);
```

src/core/model/table/typed_column_data.cpp
```diff
@@ -36,15 +37,20 @@ TypeId TypedColumnDataFactory::DeduceColumnType() const {
...
-                    // undelimited and delimited dates have different bitsets
-                    // t/f and other representations of bools also have different bitsets
-                    if ((first_type_id == TypeId::kDate && kDelimitedDateCheck(unparsed_[i])) ||
-                        (first_type_id == TypeId::kBool && !kStringBoolCheck(unparsed_[i]))) {
-                        candidate_types_bitset &= kTypeIdToBitset.at(first_type_id);
-                    }
-                    // integer and non-integer representations of bools also have different bitsets
-                    if (first_type_id == TypeId::kInt && !kIntegerBoolCheck(unparsed_[i])) {
-                        candidate_types_bitset &= kTypeIdToBitset.at(TypeId::kInt);
+                    if (!reduced_candidates) {
```
wildsor: Is the more complicated code worth it? Also, please think of a more specific name.

MichaelS239: This optimization accounts for 10% performance gain.

PR #780, opened 2026-06-17
Fix SDVerifier tests

MichaelS239:
Currently, all SDVerifier tests share one generated file with the same filename, which may have caused unexpected errors during multithreaded execution of tests in CI ([1](https://github.com/Desbordante/desbordante-core/actions/runs/27509531257/job/81306955798), [2](https://github.com/Desbordante/desbordante-core/actions/runs/27509531257/job/81306955768), [3](https://github.com/Desbordante/desbordante-core/actions/runs/27552576243/job/81442655920), [4](https://github.com/Desbordante/desbordante-core/actions/runs/27552576243/job/81487681046)).

### DIFF
--- a/src/tests/unit/test_sd_verifier.cpp
+++ b/src/tests/unit/test_sd_verifier.cpp
@@ -28,16 +28,16 @@ struct SDVerifierParams {
     std::optional<std::string> custom_csv_content;
     bool expect_exception;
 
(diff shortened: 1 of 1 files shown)


wildsor:
Could you describe the actual issue (or what you think it is, no need to investigate if it doesn't look necessary) in the commit description?

wildsor (approved):

PR #782, opened 2026-06-20
Reduce debug info in sanitizer CI builds

Vdaleke:
Add a dedicated CMake option for reduced CI debug info and enable it for ASan and UBSan core test jobs.

This switches sanitizer builds from full -g debug info to -g1 while leaving local and non-sanitized builds unchanged.

### DIFF
--- a/.github/workflows/core-tests.yml
+++ b/.github/workflows/core-tests.yml
@@ -71,9 +71,9 @@ jobs:
 
         cfg:
           - { BUILD_TYPE: Release }
-          - { BUILD_TYPE: Debug }
-          - { BUILD_TYPE: Debug, SANITIZER : ADDRESS, CMAKE_OPTS : '-DCMAKE_CXX_FLAGS_DEBUG="-O1"' }
-          - { BUILD_TYPE: Debug, SANITIZER : UB, CMAKE_OPTS : '-DCMAKE_CXX_FLAGS_DEBUG="-O1"' }
+          - { BUILD_TYPE: Debug, CMAKE_OPTS : '-DDESBORDANTE_CI_REDUCED_DEBUG_INFO=ON' }
+          - { BUILD_TYPE: Debug, SANITIZER : ADDRESS, CMAKE_OPTS : '-DCMAKE_CXX_FLAGS_DEBUG="-O1" -DDESBORDANTE_CI_REDUCED_DEBUG_INFO=ON' }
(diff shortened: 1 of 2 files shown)


.github/workflows/core-tests.yml
```diff
@@ -72,8 +72,8 @@ jobs:
         cfg:
           - { BUILD_TYPE: Release }
           - { BUILD_TYPE: Debug }
-          - { BUILD_TYPE: Debug, SANITIZER : ADDRESS, CMAKE_OPTS : '-DCMAKE_CXX_FLAGS_DEBUG="-O1"' }
-          - { BUILD_TYPE: Debug, SANITIZER : UB, CMAKE_OPTS : '-DCMAKE_CXX_FLAGS_DEBUG="-O1"' }
+          - { BUILD_TYPE: Debug, SANITIZER : ADDRESS, CMAKE_OPTS : '-DCMAKE_CXX_FLAGS_DEBUG="-O1" -DDESBORDANTE_CI_REDUCED_DEBUG_INFO=ON' }
+          - { BUILD_TYPE: Debug, SANITIZER : UB, CMAKE_OPTS : '-DCMAKE_CXX_FLAGS_DEBUG="-O1" -DDESBORDANTE_CI_REDUCED_DEBUG_INFO=ON' }
```
wildsor, line 76: I don't think CI specific stuff warrants a separate build system option

```suggestion
          - { BUILD_TYPE: Debug, SANITIZER : ADDRESS, CMAKE_OPTS : '-DCMAKE_CXX_FLAGS_DEBUG="-O1 -g1"' }
          - { BUILD_TYPE: Debug, SANITIZER : UB, CMAKE_OPTS : '-DCMAKE_CXX_FLAGS_DEBUG="-O1 -g1"' }
```

wildsor (changes requested):

.github/workflows/core-tests.yml
Vdaleke: The -g flag is set by default in debug mode; there seems to be no way to do this without a specific option.

wildsor: Nevermind, that doesn't work, the name is misleading. What about [CMAKE_CXX_FLAGS_DEBUG](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_FLAGS_CONFIG.html#variable:CMAKE_%3CLANG%3E_FLAGS_%3CCONFIG%3E)?

wildsor (approved):
okay, I don't know if that's the best way to do it, but I couldn't find anything better

PR #784, opened 2026-06-22
Remove extra clang-format off around enums

wildsor:
These used to be defined using the BETTER_ENUM macro, for which the default formatting was bad, now they're just plain enums, no special treatment needed.

### DIFF
--- a/src/core/algorithms/algorithm_types.h
+++ b/src/core/algorithms/algorithm_types.h
@@ -14,16 +14,15 @@ using AlgorithmTypes =
                    Fastod, GfdValidator, EGfdValidator, NaiveGfdValidator, order::Order, dd::Split,
                    Cords, hymd::HyMD, PFDVerifier, cfd_verifier::CFDVerifier,
                    ar_verifier::ARVerifier, GSpan, sd_verifier::SDVerifier, maxfem::MaxFEM>;
-// clang-format off
+
 /* Enumeration of all supported non-pipeline algorithms. If you implement a new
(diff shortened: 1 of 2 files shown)


p-senichenkov (approved):

src/core/algorithms/algorithm_types.h
```diff
@@ -14,16 +14,15 @@ using AlgorithmTypes =
...
-// clang-format off
+
 /* Enumeration of all supported non-pipeline algorithms. If you implement a new
  * algorithm please add its corresponding value to this enum and to the type
  * tuple above.
  * NOTE: algorithm string name representation is taken from the value in this
  * enum, so name it appropriately (lowercase and without additional symbols).
  */
-
 enum class AlgorithmType : char {
```
p-senichenkov, line 24: A little offtop: could you please remind me, why do we need this enum? As I can see, some algorithms are missing here, and it seems like all the things are still working fine

wildsor: I believe we do not, except for Mind using CreateAlgorithmInstance, which can be removed easily (but later)

p-senichenkov: Do we have this task in the task list? If no, let's add it there, otherwise it'll get lost right after this PR gets merged.
Because for now this enum looks like some junk, that is not used, but still needs to be maintained

wildsor: done

PR #787, opened 2026-06-25
Fix option visibility errors

wildsor:
Found linking failures on MacOS


polyntsov (approved):
I guess we need a CI check that verifies that any class/enum class defined under `src/core` and referenced from `src/python_bindings` is marked as exported

PR #788, opened 2026-06-26
Remove custom bitset hashing

wildsor:
Tested Pyro on iowa1kk, everything hashing related took less than 0.01% of time, and the custom hashing is a point of confusion, given that there is a standard way.

docs/papers/Pyro - Maxim Strutovsky - 2020 spring.pdf does not refer to the hashing in any way. The custom hashing was added with Pyro in 7571906f1e5d366915a75ff3477fc5a7. I doubt it is going to affect performance negatively because, first off, it doesn't make any sense to have the algorithm error out when the dataset is too wide in a public library, so we should have SAFE_VERTICAL_HASHING set, and second, the safe method allocates memory for a new bitset and cuts off the last bits, which decreases randomness.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,14 +59,6 @@ option(DESBORDANTE_BUILD_TESTS "Build tests" ON)
 option(DESBORDANTE_FETCH_DATASETS "Fetch datasets for tests or benchmarks" ON)
 option(DESBORDANTE_GDB_SYMBOLS "Include debug information for use by GDB" OFF)
 option(DESBORDANTE_CI_REDUCED_DEBUG_INFO "Use reduced debug info intended for CI builds" OFF)
-# ---- Vertical Hashing Optimization ----
-# Performance/functionality trade-off:
-# - Disabled by default for maximum performance
-# - Enabled for wide datasets (>32 columns) with potential performance penalty
-option(DESBORDANTE_SAFE_VERTICAL_HASHING
-       "Enable safe vertical hashing mode. Allows processing wide datasets (>32 columns) \
-       at the cost of reduced performance. Recommended for exploratory data analysis." ON
-)
 option(DESBORDANTE_UNPACK_DATASETS "Unpack datasets" ON)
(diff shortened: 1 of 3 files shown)


vs9h:
Firstly, I want to fully understand the original motivation for adding this custom hashing method. I'd like to be sure we won't change existing behavior along the way.

So, I'd like to 1) know the original motivation and 2) prove that the new solution doesn't change the existing logic (or how changes and why it is correct).

Regarding the second point, I see a macro `DESBORDANTE_SAFE_VERTICAL_HASHING` in the code and different behaviour depending on the macro. Why does it exist, is it used? How does hashing behave now?

And apparently, this change affects all the algorithms that store `Vertical` in some kind of hash map? Which algorithms are affected? I'd guess it probably should't affect performance in any way

wildsor:
Updated the description with some stuff.

> 1) know the original motivation

Looks like premature optimization along with not knowing about dynamic_bitset hashing support to me, but you'd have to ask @Mstrutov to be sure.

> 2) prove that the new solution doesn't change the existing logic (or how changes and why it is correct)

It just hashes bitsets differently, with the standard boost hasher, I doubt it'll change any logic.

> Regarding the second point, I see a macro `DESBORDANTE_SAFE_VERTICAL_HASHING` in the code and different behaviour depending on the macro. Why does it exist, is it used? How does hashing behave now?

Because originally Pyro was hashing bitsets by turning them into `unsigned long`, but that doesn't work when they are too large, so an alternative method was inserted so that it doesn't fail. It doesn't make sense to fail when the dataset is too large in a public library, so there is no way we're compiling the "unsafe" way in. The current "safe" method just cuts off the bits by copying the original bitset, cutting off the extra bits, and turning what's left to unsigned long.

> Which algorithms are affected?

I'd have to look, but not HyFD, should be just the older FD mining ones, maybe UCC mining too.

wildsor:
> Which algorithms are affected?

DFD, Pyro, Depminer

vs9h (approved):
Nice catch overall, also thanks for the detailed description. This helps to quickly understand the context of the changes. 

I have no more questions.

PR #792, opened 2026-06-30
Remove most of non-const non-automatic variable usage

wildsor:
Split off from #637, depends on #791

Removed most non-const non-automatic variable usage. These are generally dangerous to use when we're dealing with sub-interpreters and multiple threads. The three cases that remain are the logging variables in core and Python bindings and the hash counter for the Null type variant the values of which do not compare equal. All are safe.

The core library shouldn't account for Python in particular, so it makes sense to forbid usage of variables like that entirely except in special cases. This can be done using clang-tidy. The check cppcoreguidelines-avoid-non-const-global-variables covers the global variable case. The custom matcher below should catch most of the bad stuff.
```
match varDecl(
    hasGlobalStorage()
    isExpansionInMainFile(),
    unless(isConstexpr()),
    unless(hasType(isConstQualified())),
    unless(
        anyOf(
            isExpandedFromMacro("TEST_P"),
            isExpandedFromMacro("INSTANTIATE_TEST_SUITE_P"),
            isExpandedFromMacro("TYPED_TEST_SUITE_P"),
            isExpandedFromMacro("TYPED_TEST_P"),
            isExpandedFromMacro("INSTANTIATE_TYPED_TEST_SUITE_P"),
            isExpandedFromMacro("TYPED_TEST"),
            isExpandedFromMacro("PYBIND11_MODULE")
        )
    )
)
```

### DIFF
--- a/src/core/algorithms/fd/pyro/pyro.cpp
+++ b/src/core/algorithms/fd/pyro/pyro.cpp
@@ -13,8 +13,6 @@
 
 namespace algos {
 
-std::mutex search_spaces_mutex;
-
 Pyro::Pyro() : PliBasedFDAlgorithm() {
     RegisterOptions();
     fd_consumer_ = [this](auto const& fd) {
@@ -72,13 +70,13 @@ void Pyro::ExecuteInternal() {
 
     unsigned int total_error_calc_count = 0;
 
-    auto const work_on_search_space = [](std::list<std::unique_ptr<SearchSpace>>& search_spaces,
-                                         ProfilingContext* profiling_context,
-                                         [[maybe_unused]] int id) {
+    auto const work_on_search_space = [this](std::list<std::unique_ptr<SearchSpace>>& search_spaces,
+                                             ProfilingContext* profiling_context,
+                                             [[maybe_unused]] int id) {
         while (true) {
             std::unique_ptr<SearchSpace> polled_space;
             {
-                std::scoped_lock<std::mutex> lock(search_spaces_mutex);
+                std::scoped_lock<std::mutex> lock(search_spaces_mutex_);
                 if (search_spaces.empty()) {
                     break;
                 }
--- a/src/core/algorithms/fd/pyrocommon/model/agree_set_sample.cpp
+++ b/src/core/algorithms/fd/pyrocommon/model/agree_set_sample.cpp
@@ -11,8 +11,6 @@ namespace model {
 
 using namespace std;
 
-double AgreeSetSample::std_dev_smoothing_ = 1;
-
 AgreeSetSample::AgreeSetSample(ColumnLayoutRelationData const* relation_data, Vertical focus,
                                unsigned int sample_size, unsigned long long population_size)
     : relation_data_(relation_data),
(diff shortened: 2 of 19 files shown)


src/core/algorithms/fd/pyro/pyro.cpp
```diff
@@ -72,13 +70,13 @@ void Pyro::ExecuteInternal() {
...
-                                         ProfilingContext* profiling_context,
-                                         [[maybe_unused]] int id) {
+    auto const work_on_search_space = [this](std::list<std::unique_ptr<SearchSpace>>& search_spaces,
+                                             ProfilingContext* profiling_context,
+                                             [[maybe_unused]] int id) {
         while (true) {
             std::unique_ptr<SearchSpace> polled_space;
             {
-                std::scoped_lock<std::mutex> lock(search_spaces_mutex);
+                std::scoped_lock<std::mutex> lock(search_spaces_mutex_);
```
p-senichenkov, line 79: Not related to this PR, but why not `lock_guard`?

src/core/algorithms/fd/pyrocommon/model/agree_set_sample.cpp
```diff
@@ -106,29 +104,31 @@ ConfidenceInterval AgreeSetSample::EstimateGivenNumHits(unsigned long long num_h
...
-    static double b[4] = {-8.47351093090, 23.08336743743, -21.06224101826, 3.13082909833};
+    static constexpr double kB[4] = {-8.47351093090, 23.08336743743, -21.06224101826,
+                                     3.13082909833};
 
-    static double c[9] = {0.3374754822726147, 0.9761690190917186, 0.1607979714918209,
-                          0.0276438810333863, 0.0038405729373609, 0.0003951896511919,
-                          0.0000321767881768, 0.0000002888167364, 0.0000003960315187};
+    static constexpr double kC[9] = {0.3374754822726147, 0.9761690190917186, 0.1607979714918209,
+                                     0.0276438810333863, 0.0038405729373609, 0.0003951896511919,
+                                     0.0000321767881768, 0.0000002888167364, 0.0000003960315187};
```
p-senichenkov, line 115: 🤯️

p-senichenkov (approved):
This PR looks good to me, but there are some issues unrelated to your changes -- we may want to fix them in separate PRs.

Also, I haven't checked how many non-const not-auto-duration variables are left -- it's up to you. Would you kindly run `clang-tidy` with the check from description and post the result?

p-senichenkov:
Oh, and rename the PR, please. It's hard to notice that the word `auto` is related not to types, but to the storage duration

wildsor:
Reproducibility info in https://github.com/Desbordante/desbordante-core/pull/637#issuecomment-4537412132

wildsor:
> Would you kindly run `clang-tidy` with the check from description and post the result?

I've not looked into how to run matchers with clang-tidy, but here is the result with `clang-query` (using [find_globals.sh](https://github.com/user-attachments/files/28235073/find_globals.sh)): [globals.txt](https://github.com/user-attachments/files/29549079/globals.txt). Only logging variables and a counter for Null type hash match.

wildsor:
I also inserted some missing includes that mess up the output with that script

src/core/algorithms/fd/pyro/pyro.cpp
wildsor: Idk, someone forgot, I guess, added to task list

p-senichenkov (approved):

PR #793, opened 2026-07-07
Drop Python 3.8 and 3.9 support

wildsor:
End-of-life Python versions are unsupported by other software, so can cause unexpected problems during CI wheel builds. Also makes the builds faster since we have to build fewer versions.


wildsor:
See #637 for pybind11 bump, the older versions probably do support these Python versions

toadharvard (approved):

PR #795, opened 2026-07-10
Remove unused code in FDep

wildsor:
Removes printing dependencies to file (not needed) and an extra column_names_ field.


polyntsov (approved):

src/core/algorithms/fd/fdep/fd_tree_element.h
polyntsov: I guess fstream is no longer needed

PR #796, opened 2026-07-10
Simplify RelationalSchema creation

wildsor:
Depends on #795

Creating RelationalSchema involves the same loop every time, this puts it into the class itself.

### DIFF
--- a/src/core/model/table/relational_schema.cpp
+++ b/src/core/model/table/relational_schema.cpp
@@ -6,7 +6,31 @@
 #include "core/model/table/vertical.h"
 #include "core/model/table/vertical_map.h"
 
-RelationalSchema::RelationalSchema(std::string name) : columns_(), name_(std::move(name)) {}
+namespace {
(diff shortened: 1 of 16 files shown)


wildsor:
Waiting on #771, CFDRelationData has column cutting functions that need to be cut out

src/core/model/table/relational_schema.cpp
```diff
@@ -6,7 +6,31 @@
 #include "core/model/table/vertical.h"
 #include "core/model/table/vertical_map.h"
 
-RelationalSchema::RelationalSchema(std::string name) : columns_(), name_(std::move(name)) {}
+namespace {
+std::vector<std::unique_ptr<Column>> MakeColumns(RelationalSchema* schema,
```
polyntsov: nit: apparently can be `RelationalSchema const* schema`

polyntsov (approved):

PR #798, opened 2026-07-11
Remove ColumnLayoutRelationData use in HPIValid, HyUCC, HyFD

wildsor:
Depends on #796 

I'm looking to deprecate the class because it obscures what is happening. It also stores RelationalSchema, which I'm also looking to deprecate due to its recursive members, and that will be easier if it's separated from the PLIs instead of in one structure.

### DIFF
--- a/src/core/algorithms/fd/hyfd/hyfd.cpp
+++ b/src/core/algorithms/fd/hyfd/hyfd.cpp
@@ -14,33 +14,41 @@
 #include "core/algorithms/fd/hyfd/sampler.h"
 #include "core/algorithms/fd/hyfd/validator.h"
 #include "core/config/names.h"
+#include "core/config/tabular_data/input_table/option.h"
 #include "core/config/thread_number/option.h"
 #include "core/util/logger.h"
 
 namespace algos::hyfd {
 
-HyFD::HyFD() : PliBasedFDAlgorithm() {
+HyFD::HyFD() : FDAlgorithm() {
+    RegisterOption(config::kTableOpt(&input_table_));
     RegisterOption(config::kThreadNumberOpt(&threads_num_));
+
+    MakeOptionsAvailable({config::names::kTable});
 }
 
 void HyFD::MakeExecuteOptsAvailableFDInternal() {
     MakeOptionsAvailable({config::names::kThreads});
 }
 
+void HyFD::LoadDataInternal() {
+    schema_ = RelationalSchema::CreateFrom(*input_table_);
+
+    std::tie(plis_, pli_records_, og_mapping_) = hy::Preprocess(*input_table_);
+    if (plis_->empty()) {
+        throw std::runtime_error("Got an empty dataset: FD mining is meaningless.");
+    }
+}
+
 void HyFD::ExecuteInternal() {
     using namespace hy;
     LOG_TRACE("Executing");
 
-    auto [plis, pli_records, og_mapping] = Preprocess(relation_.get());
-    auto const plis_shared = std::make_shared<PLIs>(std::move(plis));
-    auto const pli_records_shared = std::make_shared<Rows>(std::move(pli_records));
+    Sampler sampler(plis_, pli_records_, threads_num_);
 
-    Sampler sampler(plis_shared, pli_records_shared, threads_num_);
-
-    auto const positive_cover_tree =
-            std::make_shared<fd_tree::FDTree>(GetRelation().GetNumColumns());
+    auto const positive_cover_tree = std::make_shared<fd_tree::FDTree>(schema_->GetNumColumns());
     Inductor inductor(positive_cover_tree);
-    Validator validator(positive_cover_tree, plis_shared, pli_records_shared, threads_num_);
+    Validator validator(positive_cover_tree, plis_, pli_records_, threads_num_);
 
     IdPairs comparison_suggestions;
 
@@ -59,20 +67,19 @@ void HyFD::ExecuteInternal() {
     }
 
     auto fds = positive_cover_tree->FillFDs();
-    RegisterFDs(std::move(fds), og_mapping);
+    RegisterFDs(std::move(fds), og_mapping_);
 }
 
 void HyFD::RegisterFDs(std::vector<RawFD>&& fds, std::vector<hy::ClusterId> const& og_mapping) {
-    auto const* const schema = GetRelation().GetSchema();
     for (auto&& [lhs, rhs] : fds) {
         boost::dynamic_bitset<> mapped_lhs =
-                hy::RestoreAgreeSet(lhs, og_mapping, schema->GetNumColumns());
-        Vertical lhs_v(schema, std::move(mapped_lhs));
+                hy::RestoreAgreeSet(lhs, og_mapping, schema_->GetNumColumns());
+        Vertical lhs_v(schema_.get(), std::move(mapped_lhs));
 
         auto const mapped_rhs = og_mapping[rhs];
-        Column rhs_c(schema, schema->GetColumn(mapped_rhs)->GetName(), mapped_rhs);
+        Column rhs_c(schema_.get(), schema_->GetColumn(mapped_rhs)->GetName(), mapped_rhs);
 
-        RegisterFd(std::move(lhs_v), std::move(rhs_c), relation_->GetSharedPtrSchema());
+        RegisterFd(std::move(lhs_v), std::move(rhs_c), schema_);
     }
 }
 
--- a/src/core/algorithms/ucc/hpivalid/hpivalid.cpp
+++ b/src/core/algorithms/ucc/hpivalid/hpivalid.cpp
@@ -9,6 +9,8 @@
 #include "core/algorithms/ucc/hpivalid/config.h"
 #include "core/algorithms/ucc/hpivalid/result_collector.h"
 #include "core/algorithms/ucc/hpivalid/tree_search.h"
+#include "core/model/table/create_stripped_partitions.h"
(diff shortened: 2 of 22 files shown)


src/core/model/table/to_value_id_mapped_columns.cpp
```diff
@@ -0,0 +1,38 @@
...
+#include <cstddef>
+#include <string>
+#include <unordered_map>
+#include <utility>
+
+#include "core/model/index.h"
+#include "core/util/logger.h"
+
+namespace model {
+std::vector<std::vector<int>> ToValueIdMappedColumns(model::IDatasetStream& data_stream) {
```
polyntsov: nit: this might benefit from having an alias for the internal `std::vector<int>` and maybe for the `int` itself. I have some blurry memories about introducing `ValueId` or something like that, but at the moment can't find it in the codebase 🤔 

Also the name of the function is not the most obvious one

src/core/model/table/create_stripped_partitions.cpp
```diff
@@ -0,0 +1,21 @@
...
+
+#include "core/model/table/to_value_id_mapped_columns.h"
+
+namespace model {
+std::vector<model::PositionListIndex> CreateStrippedPartitions(
+        std::vector<std::vector<int>> const& value_id_mapped_table) {
+    std::vector<model::PositionListIndex> stripped_partitions;
+    for (std::vector<int> const& mapped_column : value_id_mapped_table) {
+        stripped_partitions.push_back(
+                std::move(*model::PositionListIndex::CreateFor(mapped_column)));
```
polyntsov: A bit weird to see unique ptr being created only to be dereferenced and discarded immediately. Although I understand that it's a restriction of the underlying interface

src/core/algorithms/fd/hyfd/hyfd.cpp
```diff
@@ -14,33 +14,41 @@
...
 void HyFD::MakeExecuteOptsAvailableFDInternal() {
     MakeOptionsAvailable({config::names::kThreads});
 }
 
+void HyFD::LoadDataInternal() {
+    schema_ = RelationalSchema::CreateFrom(*input_table_);
+
+    std::tie(plis_, pli_records_, og_mapping_) = hy::Preprocess(*input_table_);
+    if (plis_->empty()) {
+        throw std::runtime_error("Got an empty dataset: UCC mining is meaningless.");
```
polyntsov:
```suggestion
        throw std::runtime_error("Got an empty dataset: FD mining is meaningless.");
```

src/core/model/table/create_stripped_partitions.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include <vector>
+
+#include "core/model/table/idataset_stream.h"
+#include "core/model/table/position_list_index.h"
+
+namespace model {
+std::vector<model::PositionListIndex> CreateStrippedPartitions(
```
polyntsov: nit: `model::` is redundant everywhere here and in cpp file

src/core/algorithms/ucc/hyucc/hyucc.h
```diff
@@ -6,16 +6,20 @@
...
-#include "core/model/table/column_layout_relation_data.h"
 #include "core/model/table/idataset_stream.h"
+#include "core/model/table/relational_schema.h"
 
 namespace algos {
 
 class HyUCC : public UCCAlgorithm {
 private:
-    std::unique_ptr<ColumnLayoutRelationData> relation_;
+    std::shared_ptr<RelationalSchema const> schema_;
```
polyntsov, line 16: I wonder why it is `shared_ptr`? Where it's copied and why? I'd expect schema to be owned in a single place

src/core/algorithms/ucc/hyucc/hyucc.cpp
```diff
@@ -10,9 +10,10 @@
 namespace algos {
 
 void HyUCC::LoadDataInternal() {
-    relation_ = ColumnLayoutRelationData::CreateFrom(*input_table_);
+    schema_ = RelationalSchema::CreateFrom(*input_table_);
 
-    if (relation_->GetColumnData().empty()) {
+    std::tie(plis_, pli_records_, og_mapping_) = hy::Preprocess(*input_table_);
```
polyntsov, line 15: I guess conceptually it makes sense to preprocess the table once during data loading step reusing the datastructures across executions, but it affects benchmarking. I believe in most cases we benchmark only execution step when comparing with third-party implementations, so by moving preprocessing to data loading step, we are artificially making the execution step cheaper. Probably we no longer benchmark these old algorithms, so it's fine, but it's still something to keep in mind

src/core/algorithms/ucc/hpivalid/hpivalid.cpp
```diff
@@ -9,45 +9,40 @@
...
-
-    model::ColumnIndex const num_columns = relation_->GetNumColumns();
-    auto plis = hy::util::BuildPLIs(relation_.get());
-    for (model::ColumnIndex column_index = 0; column_index < num_columns; column_index++) {
-        std::deque<model::PLI::Cluster> const& index = plis[column_index]->GetIndex();
-        tab.plis.push_back(index);
+    auto plis = model::CreateStrippedPartitions(value_id_mapped_columns);
+    for (model::PositionListIndex const& pli : plis) {
+        std::deque<model::PLI::Cluster> const& index = pli.GetIndex();
+        tab_.plis.push_back(index);
```
polyntsov, line 33: `tab_` is now preserved across `LoadData` calls and is never reset, so consequent calls on the same object lead to a corrupted state. I wonder why it's not caught by the tests 🤔

src/core/algorithms/ucc/hyucc/hyucc.h
wildsor: It's copied to the FD objects so they can have lifetime separate from the algorithm object. It's needed for converting to string.

src/core/algorithms/ucc/hpivalid/hpivalid.cpp
wildsor: `tab_` is only [passed to `hpiv::TreeSearch`](https://github.com/wildsor/Desbordante/blob/2f2ab60fa04949411e51274ebddf62fb81be3a89/src/core/algorithms/ucc/hpivalid/hpivalid.cpp#L45) during execution and that class [stores it as a const-reference](https://github.com/wildsor/Desbordante/blob/2f2ab60fa04949411e51274ebddf62fb81be3a89/src/core/algorithms/ucc/hpivalid/tree_search.h#L21), so its state is always the same after `LoadData`.

src/core/algorithms/ucc/hyucc/hyucc.cpp
wildsor: The whole `LoadData` -> `Execute` thing was only introduced for superficial resemblance to sklearn, so I just don't think the particulars matter too much, I'd only aim for not making the division too surprising.

src/core/model/table/to_value_id_mapped_columns.cpp
wildsor: I'd have to look more carefully for the `int` alias, so I'm not going to do that for now.

The idea with the name is that we get a stream of rows, map the values in them to their IDs, and organize the output by column. I guess I'll remove "mapped" from the name since it doesn't really matter.

src/core/model/table/create_stripped_partitions.cpp
wildsor: I'm planning to get rid of `PositionListIndex` later on because it's a pain to deal with, not the least bit because it's named after an abstract structure that can be implemented in a number of ways, so I figured I wouldn't bother doing this the right way and just use what's there.

src/core/algorithms/ucc/hpivalid/hpivalid.cpp
polyntsov: but here we do `tab_.plis.push_back(index)`, so next call to `LoadData` will also pushback to a non-empty vector with data from the previous table?

src/core/algorithms/ucc/hyucc/hyucc.cpp
polyntsov: I am not so certain that it's completely useless. I think besides resemblance to sklearn, there are two reasons for it to exist:
- it makes it possible to run the same algorithm with different parameters on the same data with the input data parsed once 
- it makes it possible to benchmark the execution time in isolation from parsing time without changing algorithm internals

So for the first reason this change makes sense -- all internal data structures representing underlying table should be built once in `LoadData`. I am just saying that in general we should be careful with such changes because of the benchmarking, especially for the new yet to be benchmarked algorithms. Nevertheless, this change is fine

wildsor: > without changing algorithm internals

I don't see a point to this

src/core/algorithms/ucc/hpivalid/hpivalid.cpp
wildsor: > so next call to `LoadData`

That's not supposed to happen

polyntsov: This is surprising.. Do we check it somewhere?

polyntsov (approved):

src/core/algorithms/ucc/hpivalid/hpivalid.cpp
wildsor: No, though we're supposed to.

PR #799, opened 2026-07-11
Make internal methods of Algorithm in Python private

p-senichenkov:
Add underscore to the beginning of `set_option`, `get_needed_options` and `get_option_type`, since these methods are not intended to be called by the users, and may give unexpected results if called without deep understanding.
For example `set_option` may fail on complex combinations of conditional options, and `get_option_type` gives types in unconventional format, that is not very useful for end users

Also change their docstrings

### DIFF
--- a/src/python_bindings/bind_main_classes.cpp
+++ b/src/python_bindings/bind_main_classes.cpp
@@ -45,38 +45,13 @@ void BindMainClasses(py::module_& main_module) {
     py::register_exception<config::ConfigurationError>(main_module, "ConfigurationError",
                                                        PyExc_ValueError);
 
-#define CERTAIN_SCRIPTS_ONLY                                                       \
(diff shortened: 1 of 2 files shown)


wildsor (approved):
Spared `get_possible_options` and `get_description`? Ok, whatever, those would be basically zero-cost to support anyway, may as well pretend they are, since they don't have any unexpected behavior.

p-senichenkov:
> Spared `get_possible_options` and `get_description`? Ok, whatever, those would be basically zero-cost to support anyway, may as well pretend they are, since they don't have any unexpected behavior.

I suppose that `get_possible_options` and `get_description` may be useful not only in CLI

wildsor:
> may be useful

This project is plagued by overabstraction, that's, like, *the* words said before disaster here. But in this case, it doesn't matter, so whatever. But if something happens with these methods, I'm shooting them immediately.  It just most likely won't.

PR #800, opened 2026-07-12
Add TKE and AFEM algorithms

m4xxx1m:
Adds TKE (Top-k Episode Miner) and AFEM (All Frequent Episode Miner) algorithms and examples to 3 Frequent Episode Mining algorithms (AFEM, MaxFEM and TKE)


src/core/algorithms/fem/tke/tke.cpp
```diff
@@ -0,0 +1,139 @@
...
+    }
+    LOG_DEBUG("Distinct raw events: {}", supports.size());
+    return supports;
+}
+
+void TKE::RemoveInfrequentEvents(std::map<model::Event, size_t> const& events_supports,
+                                 size_t event_minsup) {
+    model::Event new_events_num = model::kStartEvent;
+    reverse_mapping_.clear();
+    reverse_mapping_.resize(new_events_num);
```
mitya-radivonchik: It appears to me that this resize is redundant, am I missing something?

src/core/algorithms/fem/tke/tke.cpp
```diff
@@ -0,0 +1,139 @@
...
+    LOG_DEBUG("Frequent events: {}", events_num_);
+
+    for (model::TimedEventSet& event_set : *event_sequence_) {
+        event_set.MapEventsAndRemoveInfrequent(mapping);
+    }
+}
+
+std::vector<std::shared_ptr<LocationList>> TKE::BuildEventsLocationLists() const {
+    std::vector<std::shared_ptr<LocationList>> location_lists(events_num_);
+    std::generate(location_lists.begin(), location_lists.end(),
```
mitya-radivonchik, line 114: I think it's worth leaving a comment here that the location lists the we end up building do not exactly follow what the algorithm research paper prescribes, but are functionally equivalent and obviously superior from the performance standpoint due to much easier and more contained `LocationList::Merge` implementation that this deviation allows. Nice spot btw.

src/core/algorithms/fem/tke/parallel_topk_miner.cpp
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    size_t const tasks_threshold = threads_num_ * 4;
+
+    while (true) {
+        ParallelEpisode* ep_ptr = nullptr;
+        while (processed.pop(ep_ptr)) {
+            std::unique_ptr<ParallelEpisode> ep(ep_ptr);
+            if (TryAdd(std::move(*ep), top_k, explore)) {
+                atomic_minsup.store(top_k.top().GetSupport(), std::memory_order_relaxed);
```
mitya-radivonchik: Here we're storing `top_k.top().GetSupport()` into `atomic_minsup` -- don't we need to store a 1 instead in case `top_k.size() != k_`?

src/core/algorithms/fem/tke/parallel_topk_miner.cpp
```diff
@@ -0,0 +1,192 @@
...
+                if (TryAdd(std::move(*ep), top_k, explore)) {
+                    atomic_minsup.store(top_k.top().GetSupport(), std::memory_order_relaxed);
+                }
+            }
+            if (got_any) continue;
+            break;
+        }
+    }
+
+    terminate_flag.store(true, std::memory_order_relaxed);
```
mitya-radivonchik: I think you might want to use `memory_order_release` for storing and `memory_order_acquire` for loading the `terminate_flag`, since `boost::interprocess::interprocess_semaphore` doesn't technically have any memory ordering guarantees. Otherwise, depending on the platform, it's technically possible for one of the worker threads to load `false` from the "terminate_flag" while processing that final semaphore bump.

src/core/algorithms/fem/tke/parallel_topk_miner.cpp
```diff
@@ -0,0 +1,192 @@
...
+                                                    : parent_ep.GetSupport() < minsup_now;
+            if (stale) continue;
+
+            model::Event const start_event = parent_ep.GetLastEvent() + 1;
+            if (start_event >= events_num_) continue;
+
+            auto* parent = new ParallelEpisode(std::move(parent_ep));
+            tasks_in_flight.fetch_add(1, std::memory_order_relaxed);
+            if (!tasks.push({parent, start_event})) {
+                delete parent;
```
mitya-radivonchik, line 126: Here if the push fails we'll be silently dropping entries from `explore`, since we've already popped it but never processed the resulting item. I wouldn't expect this codepath to be practically reachable, given that users are unlikely to specify kThreads of > 10000 / 4 = 2500, but since you're handling it anyway I'd suggest fixing it by e.g. putting the item back into `explore`.

src/core/algorithms/fem/tke/parallel_topk_miner.cpp
```diff
@@ -0,0 +1,192 @@
...
+                atomic_minsup.store(top_k.top().GetSupport(), std::memory_order_relaxed);
+            }
+        }
+
+        while (tasks_in_flight.load(std::memory_order_relaxed) < tasks_threshold &&
+               !explore.empty()) {
+            ParallelEpisode parent_ep = std::move(const_cast<ParallelEpisode&>(explore.top()));
+            explore.pop();
+
+            size_t const minsup_now = atomic_minsup.load(std::memory_order_relaxed);
```
mitya-radivonchik: Since the dispatcher thread is the only one actually writing to `atomic_minsup`, it should be able to know its value without having to load the atomic variable. I think just tracking it in another, non-shared variable would be beneficial performance wise.

src/core/algorithms/fem/tke/parallel_topk_miner.cpp
```diff
@@ -0,0 +1,192 @@
...
+                continue;
+            }
+
+            for (model::Event event = task.start_event; event < events_num_; ++event) {
+                ParallelEpisode child =
+                        task.parent->ParallelExtension(event, *events_loc_lists_[event]);
+
+                if (child.GetSupport() >= atomic_minsup.load(std::memory_order_relaxed)) {
+                    auto* result = new ParallelEpisode(std::move(child));
+                    while (!processed.push(result)) {
```
mitya-radivonchik: It's probably better to add a `std::this_thread::yield()` call in this busy while loop

src/core/algorithms/fem/tke/parallel_topk_miner.cpp
```diff
@@ -0,0 +1,192 @@
...
+        while (true) {
+            tasks_sem.wait();
+            if (terminate_flag.load(std::memory_order_relaxed)) break;
+
+            Task task;
+            tasks.pop(task);
+
+            size_t const task_minsup = atomic_minsup.load(std::memory_order_relaxed);
+            if (task.parent->GetSupport() < task_minsup) {
+                delete task.parent;
```
mitya-radivonchik: I would probably invert the condition in the if-statement to avoid having to duplicate the delete and fetch_sub calls, like so:

```cpp
tasks.pop(task);

size_t const task_minsup = atomic_minsup.load(std::memory_order_relaxed);
if (task.parent->GetSupport >= task_minsup) {
    <do the work>
}

delete task.parent;
tasks_in_flight.fetch_sub(1, std::memory_order_release);
```

src/core/algorithms/fem/tke/parallel_topk_miner.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+        while (tasks_in_flight.load(std::memory_order_relaxed) < tasks_threshold &&
+               !explore.empty()) {
+            ParallelEpisode parent_ep = std::move(const_cast<ParallelEpisode&>(explore.top()));
+            explore.pop();
+
+            size_t const minsup_now = atomic_minsup.load(std::memory_order_relaxed);
+            bool const stale = (top_k.size() == k_) ? parent_ep.GetSupport() <= minsup_now
+                                                    : parent_ep.GetSupport() < minsup_now;
+            if (stale) continue;
```
mitya-radivonchik: When `stale` is true the entire `explore` queue is stale, so you should be free to just empty it the way you do it in the ExploreSequential codepath, right?

src/core/algorithms/fem/tke/parallel_topk_miner.cpp
```diff
@@ -0,0 +1,192 @@
...
+        TryAdd(std::move(seed), top_k, explore);
+    }
+
+    if (threads_num_ > 1) {
+        ExploreParallel(top_k, explore);
+    } else {
+        ExploreSequential(top_k, explore);
+    }
+
+    std::vector<ParallelEpisode> result(top_k.size());
```
mitya-radivonchik, line 187: It seems like you're intentionally making sure that the result vector is sorted in non-support-descending order, and not the other way around, and so you end up having to fiddle with indices. I've checked the callers of this method and it seems like the ordering shouldn't matter, or am I missing something? If it doesn't matter, then I'd suggest something like

```cpp
  std::vector<ParallelEpisode> result;
  result.reserve(top_k.size());
  while (!top_k.empty()) {
      result.emplace_back(std::move(const_cast<ParallelEpisode&>(top_k.top())));
      top_k.pop();
  }
  // std::reverse(result.begin(), result.end());  -- could also just std::reverse here if the ordering really matters
```

mitya-radivonchik (commented):
I'm not done reviewing yet, but I have to pause for now so sending what I've already got

src/core/algorithms/fem/tke/model/composite_episode.cpp
```diff
@@ -0,0 +1,82 @@
...
+    child.sequence_.push_back(ext.GetEventSetPtr());
+    child.CountDataForEventSet(*child.sequence_.back());
+    return child;
+}
+
+void CompositeEpisode::CountDataForEventSet(model::EventSet const& event_set, bool add) {
+    int sign = add ? 1 : -1;
+    for (model::Event event : event_set) {
+        if (event % 2 == 0) {
+            sum_of_even_items_ += sign * (event + 1);
```
mitya-radivonchik: It seems like these even and odd sums are a copy-paste from MaxFEM where they're used to optimize subset/superset checking. However, since TKE doesn't actually perform such checks, maybe we should remove that logic from it?

src/core/algorithms/fem/tke/model/composite_episode.cpp
```diff
@@ -0,0 +1,82 @@
...
+        } else {
+            sum_of_odd_items_ += sign * (event + 1);
+        }
+    }
+    events_count_ += sign * event_set.GetSize();
+}
+
+bool CompositeEpisode::StrictlyContains(CompositeEpisode const& other) const {
```
mitya-radivonchik: Same here -- this seems to be unused, and since it lives under `tke/model` and TKE doesn't actually do it, I'd vote for removing it.

src/core/algorithms/fem/tke/composite_topk_miner.cpp
```diff
@@ -0,0 +1,208 @@
...
+        return top_k.size() == k_;
+    } else if (ep.GetSupport() > top_k.top().GetSupport()) {
+        top_k.pop();
+        top_k.push(std::move(ep));
+        return true;
+    }
+    return false;
+}
+
+void CompositeTopKMiner::ExploreParallel(std::vector<ParallelEpisode> const& parallel_episodes,
```
mitya-radivonchik, line 46: A lot of my comments from the parallel_topk_miner.cpp apply here too, namely:
- we should be able to discard the Explore queue when we find that the top element is stale
- we should avoid the atomic load from the dispatching thread, since it's the only thread actually updating the variable
- we probably want to try to avoid busy loops without yields

And most critically:
- upon failing to dispatch a task we currently just drop it forever and never come back to it, which undermines algorithm correctness

src/core/algorithms/fem/tke/composite_topk_miner.cpp
```diff
@@ -0,0 +1,208 @@
...
+            if (stale) continue;
+
+            size_t valid_n = 0;
+            while (valid_n < n && parallel_episodes[valid_n].GetSupport() >= minsup_now) {
+                ++valid_n;
+            }
+            if (valid_n == 0) continue;
+
+            auto* parent = new CompositeEpisode(std::move(parent_ep));
+            tasks_in_flight.fetch_add(1, std::memory_order_relaxed);
```
mitya-radivonchik: Probably better to just fetch_add only upon successful push to the queue, rather than calling fetch_add unconditionally and then subtracting if it failed

Also if you fetch_add before pushing to the queue, then it's possible for a worker process to wake up, call `boost::lockfree::queue::pop` (which is non-blocking and returns `false` if the queue is empty) and get nothing. First of all, since the worker doesn't check the return value from the pop, it would end up processing a default-initalized Task and probably segfault. Second, it might mean that in the end one of the tasks will not get processed at all, since `tasks_in_flight` would no longer correctly reflect the actual number of tasks in flight.

Also, this fetch_add should use `std::memory_order_release` and not `std::memory_order_relaxed` for the same reasons

Also this is the place where we are dropping the task upon failing to dispatch it, which needs to be fixed

src/core/algorithms/fem/tke/composite_topk_miner.cpp
```diff
@@ -0,0 +1,208 @@
...
+            CompositeEpisode parent_ep = std::move(const_cast<CompositeEpisode&>(explore.top()));
+            explore.pop();
+
+            size_t const minsup_now = atomic_minsup.load(std::memory_order_relaxed);
+            bool const stale = (top_k.size() == k_) ? parent_ep.GetSupport() <= minsup_now
+                                                    : parent_ep.GetSupport() < minsup_now;
+            if (stale) continue;
+
+            size_t valid_n = 0;
+            while (valid_n < n && parallel_episodes[valid_n].GetSupport() >= minsup_now) {
```
mitya-radivonchik, line 128: As far as I understand, this assumes that the `parallel_episodes` is sorted in non-support-ascending order. I think it would be great to document that, because otherwise it's quite hard to intuit what this part does.

src/core/algorithms/fem/tke/composite_topk_miner.cpp
```diff
@@ -0,0 +1,208 @@
...
+                if (TryAdd(std::move(*ep), top_k, explore, initial_minsup)) {
+                    atomic_minsup.store(top_k.top().GetSupport(), std::memory_order_relaxed);
+                }
+            }
+            if (got_any) continue;
+            break;
+        }
+    }
+
+    terminate_flag.store(true, std::memory_order_relaxed);
```
mitya-radivonchik: Similar to the above, it's probably better to use `std::memory_order_release` on the terminate_flag to avoid the task push getting reordered before that load.

src/core/algorithms/fem/afem/afem.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+void AFEM::FindFrequentCompositeEpisodes(
+        std::vector<maxfem::ParallelEpisode> const& parallel_episodes) {
+    CompositeEpisodeMiner miner(min_support_, window_length_, threads_num_, tasks_num_multiplier_);
+    auto raw_results = miner.Mine(parallel_episodes);
+
+    for (auto const& parallel_episode : parallel_episodes) {
+        parallel_episode.GetEventSetPtr()->MapEvents(reverse_mapping_);
+    }
+
```
mitya-radivonchik, line 146: You might need to clear the `frequent_episodes_` here because Execute might get called multiple times for the same object as far as I know.

mitya-radivonchik (changes requested):

src/core/algorithms/fem/tke/parallel_topk_miner.cpp
```diff
@@ -0,0 +1,192 @@
+#include "core/algorithms/fem/tke/parallel_topk_miner.h"
+
+#include <atomic>
+#include <memory>
+#include <thread>
+#include <vector>
+
+#include <boost/interprocess/sync/interprocess_semaphore.hpp>
```
mitya-radivonchik: Since https://github.com/Desbordante/desbordante-core/pull/809 has merged, you might consider going back to `#include <semaphore>`

src/core/algorithms/fem/tke/parallel_topk_miner.cpp
m4xxx1m: TryAdd() returns true only when top_k.size() == k. We do not need to store 1 here

src/core/algorithms/fem/tke/parallel_topk_miner.cpp
m4xxx1m: The ordering does matter. CompositeTopKMiner relies on parallel_episodes being sorted by descending support for its prefix-based pruning, and TKE::ExecuteInternal() uses parallel_episodes.back() as the initial minimum support. Therefore, the descending order must be preserved. I wrote comment here about it.
The result is filled directly from the end to avoid an additional std::reverse() pass.

mitya-radivonchik (approved):
LGTM, thanks!

PR #801, opened 2026-07-14
Use correct option setting logic in Python tests

wildsor:
The previous one did not allow setting default values, requiring to provide all values explicitly, which made it more difficult to test algorithms that had a more complex default value calculation logic, like the domain PAC verifier.

### DIFF
--- a/src/python_bindings/py_util/opt_to_py.cpp
+++ b/src/python_bindings/py_util/opt_to_py.cpp
@@ -10,6 +10,7 @@
 #include <pybind11/stl.h>
 
 #include "core/algorithms/dd/dd.h"
+#include "core/algorithms/fd/tane/enums.h"
 #include "core/algorithms/gdd/gdd.h"
 #include "core/algorithms/md/hymd/enums.h"
 #include "core/algorithms/metric/enums.h"
@@ -57,7 +58,8 @@ std::unordered_map<std::type_index, ConvFunction> const kConverters{
         kEnumConvPair<model::InputFormatType>,
         kEnumConvPair<algos::hymd::LevelDefinition>,
         kEnumConvPair<algos::des::DifferentialStrategy>,
-        kEnumConvPair<algos::od::Ordering>};
+        kEnumConvPair<algos::od::Ordering>,
+        kEnumConvPair<algos::AfdErrorMeasure>};
 }  // namespace
 
 namespace python_bindings {
--- a/src/python_bindings/test_bindings.py
+++ b/src/python_bindings/test_bindings.py
@@ -170,14 +170,22 @@ METRIC_VERIFIER_FAILURE_CASES = [
 
 
 class TestPythonBindings(unittest.TestCase):
+    def _configure_and_verify_no_extras(self, testing_algo, params):
+        while option_names := testing_algo._get_needed_options():
+            for option_name in option_names:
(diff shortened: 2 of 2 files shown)


src/python_bindings/test_bindings.py
```diff
@@ -170,14 +170,25 @@ def check_metric_verifier_failure(dataset, options) -> bool:
 
 
 class TestPythonBindings(unittest.TestCase):
+    def _test_correct_stage_configure(self, testing_algo, params):
+        options_encountered = set()
+        while option_names := testing_algo._get_needed_options():
+            options_encountered |= option_names
+            for option_name in option_names:
+                testing_algo._set_option(option_name, params.get(option_name))
```
p-senichenkov: Maybe doing `del params[option_name]` here and checking that `params` are empty later would be more clear

src/python_bindings/py_util/opt_to_py.cpp
```diff
@@ -57,7 +58,8 @@ std::unordered_map<std::type_index, ConvFunction> const kConverters{
         kEnumConvPair<model::InputFormatType>,
         kEnumConvPair<algos::hymd::LevelDefinition>,
         kEnumConvPair<algos::des::DifferentialStrategy>,
-        kEnumConvPair<algos::od::Ordering>};
+        kEnumConvPair<algos::od::Ordering>,
+        kEnumConvPair<algos::AfdErrorMeasure>};
```
p-senichenkov, line 62: Looks like this change is not related to this PR
Should at least move it to a separate commit

p-senichenkov (commented):
Looks good, just some minor issues

p-senichenkov:
Oh, and write a detailed description please, maybe even with references to comments on #611 

We know what was the issue, but it's very hard to figure out from the changes in code, in particular because you use the new syntax, which may be not obvious to a newbie in Python

src/python_bindings/test_bindings.py
```diff
@@ -170,14 +170,25 @@ def check_metric_verifier_failure(dataset, options) -> bool:
 
 
 class TestPythonBindings(unittest.TestCase):
+    def _test_correct_stage_configure(self, testing_algo, params):
+        options_encountered = set()
+        while option_names := testing_algo._get_needed_options():
```
p-senichenkov, line 174: Oh, and another issue: wouldn't this loop become infinite, if one forgets some of the mandatory options? Looks like we should `assert` that `option_names \cap params.keys()` is not empty on each iteration

p-senichenkov (changes requested):

src/python_bindings/test_bindings.py
wildsor: > if one forgets some of the mandatory options

no, that would just give an error like "no value provided to option without a default value"

src/python_bindings/py_util/opt_to_py.cpp
wildsor: It kind of is, but I'll use a separate commit

wildsor:
> even with references to comments on #611

Ehh, I don't see anything I can reference there

> We know what was the issue, but it's very hard to figure out from the changes in code

Okay, I tried

p-senichenkov (approved):

PR #802, opened 2026-07-18
Move cmake test discovery to test run from post build

polyntsov:
Should fix build failures on newer cmake where it fails to discover the tests during parallel build

### DIFF
--- a/cmake/desbordante_helpers.cmake
+++ b/cmake/desbordante_helpers.cmake
@@ -141,5 +141,7 @@ function(desbordante_add_test name)
         Desbordante.copy_test_input_data
     )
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -230,6 +230,7 @@ add_subdirectory("src/core")
 # Configure tests
 if(DESBORDANTE_BUILD_TESTS)
     enable_testing()
+    set(CMAKE_GTEST_DISCOVER_TESTS_DISCOVERY_MODE PRE_TEST)
```
wildsor: I'm not entirely confident whether this should should be here or in `gtest_discover_tests` calls

polyntsov: Let's add to `gtest_discover_tests` call under `desbordante_add_test`, seems cleaner. I just forgot about the helper.

wildsor (approved):

PR #803, opened 2026-07-18
Refactor CFD directory

Oddin60F:
This PR contains the cfd module edits.
Major changes:

1. The FDFirst algorithm files and classes have been separated from `CFDDiscovery` base class.
2. Now the `CFDDiscovery` base class immediately stores `RawCFD` objects as a result, rather than performing the conversion when getting the list from `Getcfd()`.
3. `PrefixTree` template class has been simplified to a normal class, since FDFirst uses only one implementation.
4. The CFD dataset printing test has been removed.
5. Redundant utility files have been deleted. The functions from them have either been moved to class methods, or replaced with direct calls. Unused functions have been removed.


src/core/algorithms/cfd/fd_first_algorithm.cpp
```diff
@@ -152,16 +152,17 @@ void FDFirstAlgorithm::MineFD(MinerNode<PartitionTIdList> const& inode, Itemset
 
 // Initializing all objects that will be used in algorithm
 void FDFirstAlgorithm::FdsFirstDFS() {
-    all_attrs_ = Range(-static_cast<int>(relation_->GetAttrsNumber()), 0);
+    Itemset all_attrs_(relation_->GetNumColumns());
```
wildsor: I don't know why clang-tidy wasn't tripped on this one

```suggestion
    Itemset all_attrs(relation_->GetNumColumns());
```

src/tests/unit/test_cfd_relation_data.cpp
```diff
@@ -1,47 +0,0 @@
...
-            "rainy mild high false yes\n"
-            "rainy cool normal true no\n"
-            "overcast cool normal true yes\n";
-    std::vector<int> my_tidlist = {1, 3, 5, 6};
-    ASSERT_EQ(relation->GetNumRows(), 14);
-    ASSERT_EQ(relation->GetStringFormat(), tennis_string);
-    ASSERT_EQ(relation->GetStringFormat(my_tidlist), tennis_partial_string);
-}
-
-}  // namespace tests
```
artem-smola, line 47: As I see, the tests for CFD algorithms do not directly check table loading and transformation. Thus, it may be correct to retain this test in some form to avoid decreasing the overall code coverage.

src/core/algorithms/cfd/model/partition_tidlist.cpp
```diff
@@ -65,6 +65,53 @@ PartitionTIdList PartitionTIdList::Intersection(PartitionTIdList const& rhs) con
     return res;
 }
 
+std::vector<PartitionTIdList> PartitionTIdList::Intersection(
+        std::vector<PartitionTIdList const*> const& rhses) const {
+    std::unordered_map<int, int> eq_indices(tids.size());
```
artem-smola: Possibly, extra memory is being allocated for delimiters. It should probably be something like:
```
std::unordered_map<int, int> eq_indices(Support());
```

src/core/algorithms/cfd/cfd_discovery.h
```diff
@@ -26,14 +26,13 @@ class CFDDiscovery : public Algorithm {
...
-    ItemsetCFDList cfd_list_;
+    CFDList cfd_list_;
     std::shared_ptr<CFDRelationData> relation_;
 
 public:
     explicit CFDDiscovery();
     void LoadDataInternal() final;
-    ItemsetCFDList const& GetItemsetCfds() const;
-    CFDList GetCfds() const;
+    CFDList const& GetCfds() const;
```
artem-smola, line 33: If I'm right, the pointer to CFDList is invalidated after ResetState().

src/core/algorithms/cfd/util/prefix_tree.h
```diff
@@ -1,91 +1,32 @@
...
-        Value value;
-        std::map<typename Key::value_type, PrefixNode> sub_trees;
-        PrefixNode* parent;
-        typename Key::value_type key;
-        int depth;
+        Itemset value;
+        std::map<Item, PrefixNode> sub_trees;
+        PrefixNode* parent = nullptr;
+        Item key{};
+        size_t depth = 0;
```
artem-smola: `std::size_t`

src/core/algorithms/cfd/cfd_discovery.h
wildsor: That's expected behavior. Python copies, and it's acceptable for C++.

src/tests/unit/test_cfd_relation_data.cpp
wildsor: I think this test forced extra fields to be present. If so, good riddance. @Oddin60F please give some context, it's been a while since I looked here.

Oddin60F: Yes, the functions for converting the dataset into a string are not used anywhere except in tests and are not exposed in the API. At the same time, the removed test from `test_cfd_algos.cpp` and the assert on line 44 essentially check the operation of the function that saves the specified strings from the dataset, and nothing more.
The test for parsing the dataset could, in theory, be kept or rewritten, but I decided that tests of the algorithm’s results would be sufficient, by analogy with some algorithms (for example, INDs) that have their own data representation. It seems to me that it would be better to add more FD-First tests for coverage, but on different datasets.

artem-smola (approved):

PR #804, opened 2026-07-19
Explicitly specify clang++ for clang-tidy-check

wildsor:
The compiler that gets chosen in CI for compile_commands.json generation is GCC, which means the compiler options are generated for it. In particular, CMake does not specify the C++ standard, because the standard we are using is the default in GCC. But clang uses C++17 by default, which leads to it getting confused by the C++20 syntax.

I also have to install clang here because I've not found a supported way to force CMake to emit commands for a compiler that is not installed.

Clangd wants compile_commands to be generated for clang, otherwise parsing issues can happen https://clangd.llvm.org/design/compile-commands#subtle-parsing-problems

### DIFF
--- a/.github/workflows/check-codestyle.yml
+++ b/.github/workflows/check-codestyle.yml
@@ -66,8 +66,10 @@ jobs:
       - uses: ZedThree/clang-tidy-review@v0.23.1
         id: review
         with:
-          apt_packages: python3-dev,libboost1.88-all-dev
-          cmake_command: cmake . -B build -DCMAKE_BUILD_TYPE=Debug -Dgtest_disable_pthreads=OFF -DDESBORDANTE_BINDINGS=BUILD -DDESBORDANTE_BUILD_BENCHMARKS=ON -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DCPM_SOURCE_CACHE=cmake/cpm-cache
+          # Using Clang for compile commands generation to avoid parsing issues
+          # https://clangd.llvm.org/design/compile-commands#subtle-parsing-problems
+          apt_packages: python3-dev,libboost1.88-all-dev,clang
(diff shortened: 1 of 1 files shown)


.github/workflows/check-codestyle.yml
```diff
@@ -66,8 +66,8 @@ jobs:
       - uses: ZedThree/clang-tidy-review@v0.23.1
         id: review
         with:
-          apt_packages: python3-dev,libboost1.88-all-dev
-          cmake_command: cmake . -B build -DCMAKE_BUILD_TYPE=Debug -Dgtest_disable_pthreads=OFF -DDESBORDANTE_BINDINGS=BUILD -DDESBORDANTE_BUILD_BENCHMARKS=ON -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DCPM_SOURCE_CACHE=cmake/cpm-cache
```
p-senichenkov, line 70: Looks like the following will work as well:
```CMake
apt_packages: python3-dev,libboost1.88-all-dev
          cmake_command: cmake . -B build -DCMAKE_BUILD_TYPE=Debug -Dgtest_disable_pthreads=OFF -DDESBORDANTE_BINDINGS=BUILD -DDESBORDANTE_BUILD_BENCHMARKS=ON -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DCPM_SOURCE_CACHE=cmake/cpm-cache -DCMAKE_CXX_FLAGS='-std=c++20'
```
Both ways are not very clear, and it worth adding a comment, but the second one is a little cheaper

wildsor: Comparing https://github.com/Desbordante/desbordante-core/actions/runs/30073836254/job/89420181351?pr=804 ("150 newly installed") and https://github.com/Desbordante/desbordante-core/actions/runs/30072730244/job/89416796581 ("149 newly installed") apt outputs, only `clang` is installed, and that package weighs less than a megabyte [1](https://packages.ubuntu.com/questing/clang) [2](https://packages.ubuntu.com/questing/clang-20). Looks like it's just a symlink to an already installed clang.

p-senichenkov (approved):

PR #805, opened 2026-07-19
Generate enum option descriptions at compile time

wildsor:
I didn't like the unexpected magic_enum dependency from including names_and_descriptions, so figured I'd finally move the description generation to another file, made it properly compile-time as well.


src/core/config/enum_members_string.h
```diff
@@ -0,0 +1,45 @@
...
+
+    constexpr auto&& enum_values = magic_enum::enum_values<EnumType>(); /* C++23: |
+                                   std::views::transform([](auto&& v) { return EnumToStr(v); });*/
+
+    constexpr std::size_t enum_count = enum_values.size();
+    constexpr std::size_t total_length = []() {
+        constexpr std::size_t separator_count = enum_count - 1;
+        constexpr std::size_t brackets_count = 2;
+        std::size_t names_len = 0;
+        for (auto member : enum_values) names_len += EnumToStr(member).size();
```
p-senichenkov: nit: As far as I know, we never write several statements on a single line

wildsor: > Put any controlled statements inside blocks (i.e., use curly braces).

Oh, I thought clang-tidy would catch [that](https://google.github.io/styleguide/cppguide.html#Formatting_Looping_Branching) but I guess not.

src/core/config/enum_members_string.h
```diff
@@ -0,0 +1,47 @@
...
+#include <magic_enum/magic_enum.hpp>
+
+#include "core/util/enum_to_str.h"
+
+namespace util {
+template <typename EnumType>
+inline constexpr auto kEnumValuesCStrBuffer = []() constexpr {
+    // example: [name1|name2|name3]\0
+
+    constexpr auto&& enum_values = magic_enum::enum_values<EnumType>(); /* C++23: |
```
p-senichenkov: Why is an rvalue needed here?

src/core/config/enum_members_string.h
```diff
@@ -0,0 +1,47 @@
...
+        std::size_t names_len = 0;
+        for (auto member : enum_values) {
+            names_len += EnumToStr(member).size();
+        }
+        return brackets_count + names_len + separator_count;
+    }();
+
+    std::array<char, total_length> arr;
+
+    auto it = arr.begin();
```
p-senichenkov: Actually you can use `std::string` inside of the constant expression, since it doesn't leak out of it. For example, the following script works fine:
```cpp
template <std::size_t n>
constexpr auto constexpr_memcpy(char const* str) {
    std::array<char, n> arr;
    std::ranges::copy(str, str + n, arr.begin());
    return arr;
}

template <typename EnumType>
constexpr std::size_t TotalLength() {
    constexpr auto&& enum_values = magic_enum::enum_values<EnumType>();
    constexpr std::size_t enum_count = enum_values.size();
    constexpr std::size_t total_length = []() {
        constexpr std::size_t separator_count = enum_count - 1;
        constexpr std::size_t brackets_count = 2;
        std::size_t names_len = 0;
        for (auto member : enum_values) {
            names_len += EnumToStr(member).size();
        }
        return brackets_count + names_len + separator_count;
    }();
    return total_length;
}

template <typename EnumType>
constexpr auto Concatenate() {
    constexpr auto&& enum_values = magic_enum::enum_values<EnumType>();
    constexpr auto total_len = TotalLength<EnumType>();

    std::string result{"["};
    result += EnumToStr(enum_values.front());
    for (auto member : enum_values | std::views::drop(1)) {
        result += '|';
        result += EnumToStr(member);
    }
    result += "]";
    return constexpr_memcpy<total_len>(result.data());
}

enum class E1 { kFirst, kSecond, kThird };

int main() {
    constexpr auto size = TotalLength<E1>();
    std::cout << "Total " << size << " characters are needed for enum E1\n";
    constexpr auto conc = Concatenate<E1>();
    std::cout << conc.data() << '\n';
}
```

Can we make use of this idea? Note that not only `std::string`, but any dynamic memory can be allocated this way

p-senichenkov: Maybe you will even find a way to make `strcpy` constexpr, and the `TotalLength` call won't be needed. I cannot come up with something useful, since `strlen` can be made constexpr trivially, but the string itself (`result` in my example) is not constexpr, and `MyStrlen(result)` won't be too

src/core/config/enum_members_string.h
wildsor: That's a universal reference, I was just too lazy to write the actual type

wildsor:
Okay, I figured out how to work with strings in compile-time. Maybe it'd be a good idea to have a function for this, but repeating twice without obvious other uses isn't that big a deal IMO.

src/core/config/enum_members_string.h
p-senichenkov: Didn't know that a forwarding reference can be used not only as function parameter, but also this way...

src/core/config/enum_members_string.h
```diff
@@ -0,0 +1,37 @@
...
+
+    constexpr auto make_string = []() {
+        constexpr auto enum_names =
+                std::views::transform(magic_enum::enum_values<EnumType>(),
+                                      [](EnumType v) { return std::string(util::EnumToStr(v)); });
+        static_assert(!magic_enum::enum_values<EnumType>().empty());
+        return "["s +
+               std::accumulate(
+                       std::next(enum_names.begin()), enum_names.end(), *enum_names.begin(),
+                       [&](std::string acc, std::string name) { return acc + "|"s + name; }) +
```
p-senichenkov, line 31: It's a matter of taste, but as for me, this is more clear:

```suggestion
        return std::accumulate(
                       enum_names.begin(), enum_names.end(), "["s,
                       [&](std::string acc, std::string name) { return acc + "|"s + name; }) +
```

src/core/config/descriptions.cpp
```diff
@@ -0,0 +1,80 @@
...
+// IMO decrease readability. Under the assumption that these strings aren't going to change from
+// "description" + "members", very short names are used here. If it doesn't hold up later, consider
+// changing them.
+
+// Workaround for passing a string literal as a template argument.
+template <std::size_t N>
+struct StringLiteralTemplateWrapper {
+    char contents[N];
+
+    consteval StringLiteralTemplateWrapper(char const (&str)[N]) {
```
p-senichenkov:
```suggestion
    explicit consteval StringLiteralTemplateWrapper(char const (&str)[N]) {
```

p-senichenkov (approved):
Looks much better this way

CI fails because of the ancient compiler, isn't it?

src/core/config/enum_members_string.h
wildsor: That's gonna do "[|name1|name2|name3]", you mean this:
```suggestion
        return std::accumulate(
                       std::next(enum_names.begin()), enum_names.end(), "["s + enum_names.front(),
                       [&](std::string acc, std::string name) { return acc + "|"s + name; }) +
```

IMO it doesn't look that good, given that this string is "name1|name2|name3 surrounded by square brackets".

p-senichenkov: A little offtop: I don't really like how `std::accumulate` expresses such a simple thing as "join strings".
Won't we benefit from adding an utility function for this and using it everywhere? Seems like almost every algorithm needs to print some sequences of values, and using `std::accumulate` or `std::string_stream` and `for` loop with raw iterators (to correctly process first element) is not very clear.
Seems like there should be two versions -- a `constexpr` one for strings (based on `std::accumulate`) and based on `std::ostringstream` for other values

Inventing bicycles for every possible case just to make a single call more convenient is definitely a bad idea, but looks like we join strings really often

PR #806, opened 2026-07-19
Remove arbitrary runtime algorithm creation code

wildsor:
It is no longer used in the project. This code used to be useful when creating an algorithm by its name from the C++ CLI and for pipelines, but now there is no C++ CLI and pipelines are no longer supported.


src/core/algorithms/ind/mind/mind.cpp
```diff
@@ -8,8 +8,8 @@
 #include <algorithm>
 #include <unordered_set>
```
p-senichenkov, line 10: nit:
```
#include <memory>
```

p-senichenkov (approved):

p-senichenkov:
And let's delete all "export" files like `mining_algorithms.h` -- here or in a separate PR

PR #807, opened 2026-07-19
Reduce Magic Enum linking

wildsor:
Depends on #805 and #806

Trying to cut how much Magic Enum is linked. Picked some low-hanging fruit for now: removing obviously extra stuff from the Better Enums migration, removing some CML links that didn't cause a build failure (and where `magic_enum` functions were not present in the files), moved `create_dd.h` to the proper location, did proper linking for the DD target (gets rid of `magic_enum` link in Python).


p-senichenkov (approved):
I can't tell if it's all places where removing `magic_enum` is easy enough, but the changes themselves look good

PR #808, opened 2026-07-19
Move ICU find command to desbordante_deps.cmake

wildsor:


.github/workflows/wheel.yml
```diff
@@ -151,10 +151,12 @@ jobs:
             export BOOST_ROOT=/usr &&
             export CXX=g++-10 
           CIBW_BEFORE_ALL_MACOS: >
+            brew install icu4c &&
```
Vdaleke: https://github.com/Desbordante/desbordante-core/blob/a5971d174452378a9337beec52e253eaae8897f1/.github/composite-actions/install-dependencies/action.yml#L72

Is it necessary to duplicate the line for some reason?

wildsor: I don't know. I've not been able to make wheels work

Vdaleke:
I decided to try this, it seems like that's the problem

.github/workflows/wheel.yml
```diff
@@ -147,16 +147,19 @@ jobs:
...
-            export CXX=clang++ &&
-            export DYLD_LIBRARY_PATH=/usr/local/lib:${DYLD_LIBRARY_PATH} 
+            sudo ./b2 install -j3 --prefix=/usr/local cxxflags="-mmacosx-version-min=${MACOSX_DEPLOYMENT_TARGET} -std=c++20" linkflags="-mmacosx-version-min=${MACOSX_DEPLOYMENT_TARGET}"
+          CIBW_ENVIRONMENT_MACOS: >
+            BOOST_ROOT=/usr/local
+            ICU_ROOT=$(brew --prefix icu4c)
+            DYLD_LIBRARY_PATH=/usr/local/lib:$(brew --prefix icu4c)/lib:${DYLD_LIBRARY_PATH}
+            CXX=clang++
           CIBW_REPAIR_WHEEL_COMMAND_MACOS: >
             DYLD_LIBRARY_PATH=/usr/local/lib:${DYLD_LIBRARY_PATH} delocate-wheel --require-archs {delocate_archs} -w {dest_dir} -v {wheel}
```
polyntsov, line 164: Here we add `/usr/local/lib` twice, one of those can be removed

polyntsov (approved):
I'd squash the commits since the first one alone breaks the build

PR #814, opened 2026-08-04
Rename _Find_xx to FindXx in "self-hosted" bitset

p-senichenkov:
Rename _Find_first and _Find_next functions to FindFirst and FindNext (along with some other functions) in a "self-hosted" copy of GCC bitset as names like `_Xx` are "reserved by implementation for any use" (see [lex.name](https://timsong-cpp.github.io/cppwp/std23/lex.name#3.1))

### DIFF
--- a/src/core/algorithms/dc/FastADC/util/evidence_set_builder.h
+++ b/src/core/algorithms/dc/FastADC/util/evidence_set_builder.h
@@ -7,6 +7,7 @@
 
 #include "core/algorithms/dc/FastADC/model/evidence_set.h"
 #include "core/algorithms/dc/FastADC/util/clue_set_builder.h"
+#include "core/model/types/bitset.h"
 #include "core/util/logger.h"
 #include "core/util/worker_thread_pool.h"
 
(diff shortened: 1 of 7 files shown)


p-senichenkov:
Need to benchmark (and, maybe, `perf`) FastOD first, since it used `_Unchecked_set` for some reason

wildsor:
I think the names are from gcc's implementation of the bitset, STL code is the one that can use these "reserved" names

p-senichenkov:
> I think the names are from gcc's implementation of the bitset, STL code is the one that can use these "reserved" names

Nevertheless, using such names is UB

p-senichenkov:
> Need to benchmark (and, maybe, `perf`) FastOD first, since it used `_Unchecked_set` for some reason

As expected, benchmarks show exactly the same results, and `AttributeSet` don't appear in the flamegraph at all
flamegraph

wildsor (approved):
Mhm, I don't like that `Check` actually does something in release build, but that's not for this PR.

p-senichenkov:
> Mhm, I don't like that `Check` actually does something in release build, but that's not for this PR.

You can send your suggestions to GNU :)

wildsor:
> You can send your suggestions to GNU :)

Standards committee, actually. It just doesn't make sense in our context ever, since I wouldn't expect an exception to be tripped during normal algorithm execution (bar user code calls, I guess). boost::dynamic_bitset doesn't do any checks on a `set` call, and given that it's used everywhere, I think it's confusing that the check *is* suddenly done here.

PR #815, opened 2026-08-05
Get rid of create_algo target

p-senichenkov:
- Delete the `create_algo` Cmake target
- Delete `CreateAlgorithmInstance` and all related functions
- Use `std::make_unique` instead of `CreateAlgorithmInstance` in MInd


p-senichenkov:
Duplicate

PR #817, opened 2026-08-06
Rewrite build.sh into Python

p-senichenkov:
Handling options in bash is tricky and unnatural, leads to subtle bugs, and produces an inconvenient interface (for example, `-L TRACE` won't work)

Currently only `click` dependency of the script is managed by `uv`. In future we should consider managing the entire development environment by `uv`


wildsor:
typo in the commit summary

wildsor:
You have "rewrtie" instead of "rewrite" in the first commit's summary

PR #818, opened 2026-08-07
Remove headers listing all algorithms for a task

wildsor:
The divisions among these files were not clear and some algorithms ended up being included in files where they should not have been. They were only used in the Python bindings and in the runtime algorithm creation function, the latter of which does not exist, so it is better to just include them directly now.


p-senichenkov (commented):
Seems like you missed two files
https://github.com/Desbordante/desbordante-core/blob/main/src/core/algorithms/statistics/algorithms.h
https://github.com/Desbordante/desbordante-core/blob/main/src/core/algorithms/fem/algorithms.h

Everything else looks good

p-senichenkov (approved):

PR #824, opened 2026-08-29
Implement custom metric holder

p-senichenkov:
Implement a holder for custom metric to optimize `IMetrizableType`-based metrics
TODO: plots, flamegraphs, more detailed description

See usage in https://github.com/p-senichenkov/Desbordante/pull/11

### DIFF
--- a/src/core/util/custom_metric/custom_metric_holder.h
+++ b/src/core/util/custom_metric/custom_metric_holder.h
@@ -1,0 +1,83 @@
+#pragma once
+
+#include <cassert>
+#include <cstddef>
+#include <memory>
+#include <variant>
+#include <vector>
+
+#include "core/config/exceptions.h"
+#include "core/config/indices/type.h"
+#include "core/model/table/column_layout_typed_relation_data.h"
+#include "core/model/types/imetrizable_type.h"
+#include "core/model/types/type.h"
+#include "core/util/custom_metric/custom_metric.h"
+
+namespace util {
+/// @brief CustomMetricHolder is aimed to save a lot of `dynamic_cast`s for `IMetrizableType`-based
+/// metrics
+/// You should always prefer CustomMetricHolder over plain `std::shared_ptr<ICustomMetric>` until
+/// you have custom type handling
+class CustomMetricHolder {
+private:
+    friend void InitializeCustomMetricHolders(std::vector<CustomMetricHolder>& holders,
+                                              model::TypedRelationData const& typed_relation_data,
+                                              config::IndicesType const& column_indices);
+
+    std::variant<model::Type const*, model::IMetrizableType const*> type_;
+    std::shared_ptr<ICustomMetric> metric_;
+
+public:
+    // This method is intended to be used only in Option constructor
+    std::shared_ptr<ICustomMetric>* ValuePtr() {
+        return &metric_;
+    }
+
+    void SetType(model::Type const* type) {
+        auto metr_type = dynamic_cast<model::IMetrizableType const*>(type);
+        if (metr_type) {
+            type_ = metr_type;
+        } else {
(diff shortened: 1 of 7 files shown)


src/core/util/custom_metric/custom_metric_holder.h
```diff
@@ -0,0 +1,53 @@
...
+/// You should always prefer CustomMetricHolder over plain `std::shared_ptr<ICustomMetric>` until
+/// you have custom type handling
+class CustomMetricHolder {
+private:
+    std::variant<model::Type const*, model::IMetrizableType const*> type_;
+    config::CustomMetricType metric_;
+
+public:
+    // This method is intended to be used only in Option constructor
+    config::CustomMetricType* ValuePtr() {
```
p-senichenkov, line 25: Returns a pointer intentionally to make undesired calls like `(*holder.ValuePtr())->Dist()` very inconvenient

src/core/util/custom_metric/custom_metric_holder.h
```diff
@@ -0,0 +1,53 @@
...
+                        "Provided metric requires a metrizable type, but type {} is not metrizable",
+                        type->ToString()));
+            }
+            type_ = type;
+        }
+    }
+
+    double Dist(std::byte const* a, std::byte const* b) const {
+        auto valueless = [](auto const* type) { return type == nullptr; };
+        assert(((void)"Custom metric holder must be initialized first",
```
p-senichenkov, line 46: How about
```
#define DESBORDANTE_ASSERT(expr, msg) assert(((void)msg, !!(expr)))
```

src/core/util/custom_metric/custom_metrics_holder.h
```diff
@@ -0,0 +1,65 @@
...
+    std::vector<std::variant<model::Type const*, model::IMetrizableType const*>> types_;
+    config::CustomMetricsType metrics_;
+
+public:
+    // This method is intended to be used only in Option constructor
+    config::CustomMetricsType* ValuePtr() {
+        return &metrics_;
+    }
+
+    void SetTypes(model::TypedRelationData const& typed_relation_data,
```
p-senichenkov, line 31: Takes typed relation data instead of a vector of types only to emit more informative error messages

p-senichenkov: 1. Could add an overload `SetTypes(std::vector<model::Type const*>)` here
2. Could add an overload `SetType(model::TypedRelationData const&, config::IndexType)` in `CustomMetricHolder`

src/core/util/custom_metric/custom_metrics_holder.h
```diff
@@ -0,0 +1,65 @@
...
+        types_.reserve(metrics_.size());
+        for (std::size_t i = 0; i < column_indices.size(); ++i) {
+            auto const& column_data = typed_relation_data.GetColumnData(column_indices[i]);
+            auto const* type = &column_data.GetType();
+            auto const* metrizable_type = dynamic_cast<model::IMetrizableType const*>(type);
+            if (metrizable_type) {
+                types_.push_back(metrizable_type);
+            } else {
+                if (metrics_[i]->RequiresMetrizableType()) {
+                    throw config::ConfigurationError(std::format(
```
p-senichenkov, line 45: Why don't we use `std::format`?)

p-senichenkov:
Another option is to disallow direct calls (not through `CustomMetricHolder`) at all: https://github.com/Desbordante/desbordante-core/pull/825. This may be more effective due to a less number of virtual dispatches

Note: there is no separate "usage example" PR, since usage is completely the same in both approaches

p-senichenkov:
https://github.com/Desbordante/desbordante-core/pull/826

PR #825, opened 2026-08-30
Another approach to custom metric holder

p-senichenkov:


p-senichenkov:
https://github.com/Desbordante/desbordante-core/pull/826

PR #828, opened 2026-09-06
Remove PLI with singletons

Sneper-Breeze:
Use regular position list indexes for AFD metrics and update the related core tests.


wildsor (approved):
Would have preferred unrelated changes like formatting and missing includes in separate commits, but good enough

src/core/algorithms/fd/afd_metric/afd_metric_calculator.cpp
```diff
@@ -155,85 +156,64 @@ long double AFDMetricCalculator::CalculatePdepSelf(model::PLIWithSingletons cons
...
+    config::ErrorType sum = n - x_pli->GetSize();
+    std::shared_ptr<std::vector<int> const> const a_prob = a_pli->CalculateAndGetProbingTable();
+
+    for (Cluster const& x_cluster : x_pli->GetIndex()) {
+        config::ErrorType x_cluster_size = x_cluster.size();
+        std::unordered_map<int, unsigned> frequencies;
+        size_t singleton_records = 0;
+
+        for (int tuple_index : x_cluster) {
+            int const value_id = a_prob->at(tuple_index);
```
wildsor:
```suggestion
            int const value_id = (*a_prob)[tuple_index];
```

wildsor (changes requested):

wildsor (approved):

PR #829, opened 2026-09-07
Fixes for AFD measure calculation

wildsor:
see commits


Sneper-Breeze (approved):
LGTM. I am only confused by the G3 measure calculation, cause all other G metrics are calculating with g1/2Error and only the G3 like an AFDmeasure; even if in the paper g1/2 were defined through the 1 - error, it is seems to be confused. Maybe it will be cool to do all static methods calculating like AFD measure.

PR #830, opened 2026-09-11
Rewrite Tane

wildsor:
Depends on #829 

Not needed for release

NOTES: removed UCC, explicitly documented implementation details


wildsor:
The schema has attributes A B C D E
[A B] is a superkey
AFDs [A] -> B, [C] -> B, [D] -> B, [A C] -> B, [A D] -> B, [C D] -> B do not hold
AFD [A C D] -> B holds
[<=> g3 error of all of those is greater than g3([A C D] -> B), g3(...) = 0 is impossible, though]
[A C D] is not a superkey

Threshold = 1/5 = 0.2 (g3) breaks TANE on this table, [A C D] -> B is not reported because the key [A B] gets cut, so [A B C D] never makes it to the level. Even just marking superkeys and skipping like was attempted here does not work, as the dependency has to be detected during COMPUTE_DEPENDENCIES, where it ends up being skipped.
| tuple | A | B | C | D | E |
|---|---:|---:|---:|---:|---:|
| t1 | 1 | 1 | 1 | 1 | 0 |
| t2 | 1 | 2 | 1 | 1 | 1 |
| t3 | 2 | 3 | 2 | 2 | 0 |
| t4 | 2 | 4 | 2 | 3 | 1 |
| t5 | 3 | 5 | 4 | 4 | 0 |
| t6 | 3 | 6 | 5 | 4 | 1 |
| t7 | 6 | 7 | 6 | 6 | 0 |
| t8 | 7 | 8 | 6 | 6 | 1 |

The current implementation seems to work fine and I've spent way too much time on this already. Closing.

PR #833, opened 2026-09-13
Rewrite TANE

wildsor:
Should have reopened #830 but force-pushed too early, the comments from there apply


wildsor:
I've thought through the fix but didn't want to write the code, so asked DeepSeek to do it. I sent it the paper and the original files with this prompt:
```
TANE as described in the original paper is not entirely correct. The code I've written fixes it for when it searches for exact FDs, but it still fails when max_error is not 0.0 due to key pruning being invalid in that case. Here is the outline of the fix I want to make. It's supposed to disable key pruning when error != 0.0. What exact code do I need to modify and insert?

Meaning of .pli after GenerateNextLevel when error != 0.0:
pli == nullptr if all children are superkeys
pli is a superkey when some children are not superkeys but at least one is
otherwise everything is as before

For error == 0.0:
pli == nullptr if any of the children are superkeys

GenerateNextLevel changes:
If all children are superkeys (pli == nullptr), set PLI == nullptr. If just one is a superkey (with one non-superkey), set a (non-NULL) superkey PLI. For error == 0.0 just set nullptr if any is nullptr (even if one of the others is not a superkey).

Prune changes:
pli == nullptr means skip, direct superkey check => execute the procedure already there (including setting pli to nullptr with MarkSuperkey (may need renaming of the two methods on AttributeSetData))

ComputeDependencies:
nullptr check for error == 0.0.
no checks if error != 0.0, but if LHS PLI is nullptr, skip that (if (...) return), since that was already processed in Prune for the previous level

A superkey PLI can just be constructed when found (PositionListIndex::MakeSuperkeyPLI(pli->GetRelationSize())), since all superkey PLIs look the same.

The tane_common_initial.cpp steps need to avoid the effects of PRUNE, so there can be an AFD from non-key to key (Register and modify the candidates accordingly), and that's all that needs to be added, I think?
```

I then took its code and rewrote it, since it was being a little stupid in a few places, namely
* using `IsSuperkey` in `GenerateNextLevel` when `IsMarkedSuperkey` would work just fine (Prune from the previous level has already set non-null key PLIs to nullptr)
* thinking that `prev_level.find(lhs_attribute_mask)` can return prev_level.end() in `ComputeDependencies` (it can't, the attribute set being present implies all the subsets were in the previous level)
* copying the candidates in the new manual first levels FD (`ComputeDependenciesLevel2KeysNotZeroaryAfdRhs`) checks instead of reusing `superkey_rhs_candidates`

I think it should work, but I've not looked too carefully and it needs a few more comments in these places

wildsor:
I should also add the key pruning counterexample into the comments

wildsor:
Better counterexample (B is not a superkey):
tuple | A | B | C | D | E
-- | -- | -- | -- | -- | --
t1 | 1 | 0 | 1 | 1 | 1
t2 | 1 | 1 | 1 | 1 | 2
t3 | 2 | 0 | 2 | 2 | 3
t4 | 2 | 1 | 2 | 3 | 4
t5 | 3 | 0 | 3 | 2 | 5
t6 | 3 | 1 | 4 | 2 | 6
t7 | 4 | 0 | 5 | 5 | 7
t8 | 5 | 1 | 5 | 5 | 8

PR #835, opened 2026-09-18
Temporarily remove ICU

wildsor:
It was only used in one statistic that counted diacritic characters. The library takes a long time to build from source and the version in the OSX repos on the runners in GitHub Actions is too new, which precludes users of older OSX versions from using our wheel.

The statistic was removed as well.


.github/workflows/wheel.yml
```diff
@@ -142,7 +142,6 @@ jobs:
         env:
           MACOSX_DEPLOYMENT_TARGET: 15.0
```
Vdaleke: Please lower the version from 15 to 11 and change the runners to `macos-latest`. We need to revert the build to the previous versions. You can use blame to find the commit where the version was bumped and partially revert it.

Vdaleke (changes requested):
It looks okay, but the previous macOS versions need to be brought back.

Vdaleke (approved):

